PR #175, opened 2015-01-22
README: Add mailing lists

sils1297:


sils1297:
please check if the links work for you too before acking.

PR #176, opened 2015-01-22
Parsing: change way to set custom settings

smfn3321:
* Settings can now be specified with the parameter --settings (-S)
* Settings can still be of the form section.key=value
* Settings can not be mixed with traditional params anymore


sils1297:
and please rebase to current masteer from time to time.

We only allow fast forward merges so the commits have to be rebased to current master at merge time.

PR #181, opened 2015-01-22
Wip/foreign bears

smfn3321:


smfn3321:
This adds the needed functionality for the time being until the collectors will be reworked

sils1297:
some test seems to be missing: see decreased coverage. Would be cool if you had 100% branch coverage for that file after commit.

sils1297:
and obviously needs a rebase as mentioned in the other PR.

smfn3321:
* Coverage: That's what I meant with my comment above. At this moment we still have a parameter to specify ignored bear directories and bears. We also had a unittest that ensured that we do not import ignored bears. This test failed for this commit since bears imported by the __additional_bears__ statement were imported without question. Since we just recently decided to drop that functionality anyways, I did not bother trying to fix the behavior for __additional_bears__ and removed the test instead. The coverage decreased therefore because the functionality is still there for the old-fashioned way of collecting bears but it is no longer tested without the test.

* commits will be squashed as we finish this discourse because I feel that it will vanish thanks to git.

* rebased to what and where is this mentioned?

* import_bears function: I will keep that in mind for the collector 'facelift'

sils1297:
rebased to current master. In some other comment, isnt important.

sils1297:
I won't ack a commit that decreases code coverage. Either you adjust the test or you strip off the ignore functionality first, shouldnt take long btw.

smfn3321:
How can Coverage be 100% when "Coverage decreased (-0.06%) to 99.94% when pulling 793403f on wip/foreign_bears into 0366bd9 on master."

smfn3321:
Anyway, I rebased on current master, squashed and most importantly added the ignore functionality to satisfy the test_ignored() test.

sils1297:
because the second coverage report refers to 5fe35e4 instead of 793403f. So the first one was for your first version.

smfn3321:
Just notices that, anyway it shouldn't show 100% in that case

smfn3321:
well, p_objects are python objects, hence the name, that are extracted from a module. In the traditional case, these are bear classes and they do not really get "imported" (the modules do) but only appended to a bears list that is returned by the collect method.

Now in the new case p_object is a list containing classes, hence the bears list gets extended with p_object, there is nothing more to it really. The new and admittedly not sufficiently named method merely performs checks for validation and verefication of the bear classes

smfn3321:
so do you mean that you'd rather collect all p_objects into a list first to then check all of them with a validation/verification method? that's my plan for the new collectors but I'd have to do the globs and new input first anyways.

smfn3321:
basically we have something similar with the is_target() method but it's not easily applicable in this case.

sils1297:
ok, please rename _less_confusing_method_to_import_referenced_bears to something less confusing and then lets merge this stuff. Seeing forward to the refactoring.

sils1297:
this is merged (I did another rebase)

PR #189, opened 2015-01-26
Correct shebang

sils1297:
According to several sources it is more platform independent to invoke
python via /usr/bin/env.

Please try it out and report if it works on your distribution!

### DIFF
--- a/coala
+++ b/coala
@@ -1,4 +1,4 @@
-#! /bin/python3
+#!/usr/bin/env python3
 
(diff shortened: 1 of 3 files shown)


sils1297:
rebased

smfn3321:
#! /usr/bin/env python3 works for me (Opensuse 13.2)

sils1297:
FYI: works also on arch and fedora.

PR #201, opened 2015-01-28
tests: Shorten output

sils1297:
One can get verbose output through passing the -v argument. If this
doesn't meet your needs because you need the ordering between
stderr/stdout you are encouraged to execute your test manually for
debugging.

### DIFF
--- a/coalib/tests/TestHelper.py
+++ b/coalib/tests/TestHelper.py
@@ -13,35 +13,48 @@ class TestHelper:
             pass
 
     @staticmethod
-    def execute_python3_file(filename, use_coverage, ignored_files):
+    def print_output(command_array, verbose):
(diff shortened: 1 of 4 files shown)


sils1297:
Isn't that a nice output?
http://pastebin.com/6TgJxgT0

smfn3321:
Slick, but is it as verbose as the current configuration if a test fails?

Also, is there a way to show line numbers for missing branches? Sometimes it's really hard to track them down especially if all statements are covered.

sils1297:
Non-verbose output: http://pastebin.com/DG9L0Ciq
With a failure: http://pastebin.com/2syW4Dny

sils1297:
Its as verbose as before. Only difference is that if you have something from stderr and stdout it will be seperated. E.g. the "test_non_printing (__main__.NullPrinterTestCase) ... ERROR" lines go through stderr, the test output usually goes through stdout.

PR #233, opened 2015-02-06
Discoverage

smfn3321:


sils1297:
We have a few branches too much for these commits, could you tidy up at times? I'd love to reduce them to two, wip/smfn/input and wip/smfn/choosewhateveryouwant while the first one is your wip/input branch.

PR #243, opened 2015-02-07
files: Create empty text files

smfn3321:
Empty text files created by dolphin or nautilus feature a line of no text while empty files created by pycharm or the tough command do not feature a line of text. The latter are therefore more empty.


sils1297:
yes, the latter is a really empty file. gnome programs usually automatically add trailing newline, pycharm does not. (Thats why you get those missing trailing newline comments from me every time)

sils1297:
we don't really need this PR, do we?

PR #253, opened 2015-02-10
dreamlined rebased on misc

sils1297:
holds quite some bugfixes, our new enum type, major refactorings and the switch to globs with major CLI arg simplifications.


sils1297:
cool, this PR actually removes more lines than it adds - @smfn3321 good job!

sils1297:
revised version lies on #258

PR #258, opened 2015-02-11
Wip/sils/deps v2

sils1297:
holds quite some bugfixes, our new enum type, major refactorings, the switch to globs with major CLI arg simplifications and bear dependency support for local bears.


sils1297:
merged third version

PR #259, opened 2015-02-11
Wip/sils/deps v3

sils1297:
third revision


sils1297:
all those commits are implicitly acked due to only minor corrections.

PR #313, opened 2015-02-21
Misc fixes

sils1297:


Makman2:
ack from my side, I have no errors when executing coala, so the imports seem to be unneeded^^

PR #315, opened 2015-02-21
pre release preparations

sils1297:


sils1297:
self-reviewed non-code commits. merging...

Makman2:
sorry, I totally missed your mail :(
I hope that was not so bad (but are small commits)

PR #324, opened 2015-02-23
Diff: Fix multiple line handling.

sils1297:
We do matcher.get_grouped_opcodes(1) and assume that each change group
affects only one line in the origin and target file. However despit it
may affect only one origin line it may affect several target lines thus
we need to remove that assumption.

Fixes: https://github.com/coala-analyzer/coala/issues/316


Makman2:
Didn't see that this was closed... But anyway :)

PR #331, opened 2015-02-24
Interactor: Save backup before writing changes

sils1297:


sils1297:
I'd love to have another review from someone else before pushing.

Issue #334, opened 2015-02-24
"./execute_all_tests.py" command doesnot execute test files 


AhmedKamal1432:
"Run our tests. Just execute `./execute_all_tests.py`" 
in "doc/getting_involved/README.md" and "doc/getting_involved/TESTING.md" files 

when run it this message appear "bash: ./execute_all_tests.py: No such file or directory"

but there is no file with this name (I think it should be ./run_tests.py)

sils1297:
Oh, yeah, thanks! Missed documentation correction.

2015-02-24 12:36 GMT+01:00 AhmedKamal1432 <notifications@github.com>:

PR #341, opened 2015-02-24
Documentation : change  Run all test command

AhmedKamal1432:
change ./execute_all_tests.py command in these files doc/getting_involved/README.md  ,  doc/getting_involved/TESTING.md
to  ./run_tests.md

Fixing : this bug #334

### DIFF
--- a/doc/getting_involved/TESTING.md
+++ b/doc/getting_involved/TESTING.md
@@ -8,7 +8,7 @@ Executing our Tests
 
 coala has a big test suite. It is meant to work on every platform on every PC.
 If you just execute our tests you are doing us a favor. You can execute our
-tests with our `./execute_all_tests.py` script and report any errors you get!
+tests with our `./run_tests.py` script and report any errors you get!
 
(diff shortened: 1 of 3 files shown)


sils1297:
You're right, good spot. You can replace it in index.rst because that's actually handwritten. Then regenerate the documentation with the script in coala repository root as an own commit with the [GENERATED] flag in the commit message to mark that only commit message is to be reviewed. Probably in the same commit you want to delete the execute_all_tests.rst because we don't do that automatically :/

sils1297:
Something's gone wrong with rebasing, please pull from upstream again. Something like `git pull upstream --rebase` may help you, care to keep your commits ordered. (You want both your commit on top of all the ones we did in the meantime and not somewhere in between.)

Thanks for reworking by the way.

sils1297:
Merged to master - thanks for submitting!

sils1297:
urgh, seems not to work on master, something went wrong there. I'm reverting the commits for further investigation.

PR #352, opened 2015-03-02
Reimplemented SectionManager

Udayan12167:


Udayan12167:
So i fixed the minor problems.

sils1297:
the code changes of the last two commits look good. However I'm not sure if the tests suffice to prove that the code is right.

A little testing lesson :) I suppose you know of statement and branch coverage. I suppose you optimized statement coverage to 100%, at least travis confirms that. Thats good but actually only a first step. So besides the coverage it's important to test corner cases. Meaning, you try to find input values for which your implementation does **not** work. So you're actively trying to break your code. This is especially important here because all this merging and defaults and several files is a bit error prone and we had bugs with this before.

In order to do this I'd e.g. *at least* add some other sections to the default_coafile test file and also some others to your .coafile so you get all combinations of None/not None merging and so on.

sils1297:
Basically at coala we aim that if the code passes the tests it should not be possible that the code is wrong. (Makes refactoring very easy :))

Udayan12167:
Test cases added please review :)

sils1297:
I pushed the first two commits on master since they're ready and valid on their own.

sils1297:
Please look at those spacig issues:
https://scrutinizer-ci.com/g/coala-analyzer/coala/inspections/6b4d4c61-a8d8-4ae5-a1df-4b1987e3ba90/issues/files/coalib/tests/settings/SectionManagerTest.py?status=new&orderField=path&order=asc

Udayan12167:
Fixed the spacing issues. Added test cases as required.

sils1297:
please rebase onto current master and fix the spacing issues in an own commit. General rule: when something is published one *never* changes it. (The history I mean.)

Udayan12167:
Rebased off current master and fixed spacing issue and converted that method into static.

sils1297:
merged

PR #353, opened 2015-03-03
Add guide for writing tests

Makman2:


Makman2:
Maybe missing: Some asserts I don't covered, for example one that checks whether a call does **not** throw an error.

Makman2:
Pushed new patch commit^^

sils1297:
how about merging them again?

2015-03-03 21:16 GMT+01:00 Makman2 <notifications@github.com>:

Makman2:
Okay everything settled, I'll squash each commit... stand by :P

sils1297:
and maybe review some commits...?

2015-03-03 21:21 GMT+01:00 Makman2 <notifications@github.com>:

Makman2:
update done, ready for merge

Makman2:
Patch commit pushed^^

sils1297:
could you squash and rebase?

sils1297:
I took the freedom of doing that for you.

sils1297:
still missing a hint about the -c and -H argument, and thats a really important one IMHO.

Makman2:
I tried to install coverage3 but I think I got the wrong package. I don't know how this works^^
Could you maybe write the paragraph/hint/note?
Maybe we could keep it then also as an extra commit, No need to squash everytime, the guide is a good starting point now I think.

sils1297:
I think the package is usually named `python-coverage3`. It works that way that you just install it and then run_tests.py takes care of everything else, no setup from you required, it's all done by me :)

sils1297:
I'm so picky about this being in initially because it's very *very* important and you usually want to check your coverage whenever you write something. (At least for coala.)

Makman2:
I'll try it, I first installed `python-coverage`. Everything seemed to point to it when googling for coverage3^^

Makman2:
no such package in ubuntu. I had kept back packages, but I upgraded them and still not working.
Really can you please write the paragraph? As I said even no need to squash it, can be kept as an own commit since it extends the guide.

sils1297:
ok, then squash your stuff and I'll push it to master.

sils1297:
squashing, also inserting full bug URL instead of number only

Issue #367, opened 2015-03-06
Creating .coafile from coala yields warning


Makman2:
When creating a coafile (so .coafile doesn't exist first) it throws me a warning that it is not used, but the sections are all properly executed.

```
coala -S SECTION.a=a SECTION.b=b TODO.enabled=True --save

[WARNING][18:24:57] The requested coafile '/home/mak/dev/coala-tutorial/.coafile' does not exist. Thus it will not be used.
Executing section Default...
Executing section SECTION...
Executing section TODO...
```

Invoking coala again with this .coafile throws no warning then^^

sils1297:
So straightforward solution would be to not throw this warning if we're about to save anyway. Or probably better to change the warning to something like `Given coafile does not exist, a new one will be created.` or so...

Makman2:
That's really better, a warning confuses the user and lets him think he has done something wrong^^

shivanipoddariiith:
Created a Pull request to Fix this HERE -> https://github.com/coala-analyzer/coala/pull/369

sils1297:
still valid?

Makman2:
yes :(
```
[WARNING][02:46:15] The requested coafile '/home/mak/dev/coala-tutorial/.coafile' does not exist. Thus it will not be used.
Executing section Default...
Executing section SECTION...
Executing section TODO...
```
If the warning system doesn't allow other messages but warnings then we should make a new tag like `NOTE` or `INFO` to display such kind of messages.
(Tested on master at ba295bb3ae741e4ebaaedc4258ef714aaf5e8782)

sils1297:
> If the warning system doesn't allow other messages but warnings then we should make a new tag like `NOTE` or `INFO` to display such kind of messages.

Try executing `coala -L DEBUG` and you'll see we have a DEBUG type of messages.

Makman2:
Then creating a new tag shouldn't be that complicated^^

sils1297:
I'd in this case not log any message at all to be honest... (yes it isn't complicated to add another one especially with our enum class.)

Makman2:
Also my intention, the user knows implicitly that it gets added if it's non-existent^^

sils1297:
@shivanipoddariiith progress?

sils1297:
or shall someon else do this bug?

shivanipoddariiith:
Hey, just saw this one, so the workaround is to add another tag here right ?

sils1297:
No

[...]

PR #369, opened 2015-03-07
Misc message fixes

shivanipoddariiith:
Executing "run_tests.py" without installing gettext, prompts the user with the following message:
screen shot 2015-03-07 at 1 34 05 pm
The message is not clear enough to state that, msgfmt requires "gettext" to be installed in the correct path so as to recover from the error.
The error message needs to be fixed for the same.

### DIFF
--- a/doc/getting_involved/MAC_hints.md
+++ b/doc/getting_involved/MAC_hints.md
@@ -1,0 +1,49 @@
+## Coverage Installation Hints for OSX Users:
+
+##### 1. Make sure you have installed Xcode and Homebrew.
+
+##### 2. Install Python3.
+
+
+    brew search python  # This should display python3
+    brew install python3
+    python3 --version   # To check the installed version
+
+##### 3. Create Virtual environments with pyvenv
+
+    # Create Virtual Env named myenv
+    pyvenv myenv
+
+    # This will create a folder named myenv in the
+    # current directory. To activate this environment just type
+    source myenv/bin/activate
+
+    # You can start Python3 by typing:
+    python
+
+##### 4. Virtualenvwrapper with Python 3:
+
+    # Installation
+    pip install virtualenv
+    pip install virtualenvwrapper
+
+    # Folder to contain Virtual Environments
+    mkdir ~/.virtualenvs
+
+    # Add the following in ~/.bash_profile
+    export WORKON_HOME=~/.virtualenvs
+    source /usr/local/bin/virtualenvwrapper.sh
+
+    # Activate Changes
+    source ~/.bash_profile
+
+    # Get Python3 path (python3_pth)
+    where python3
+
+    # Create a new virtual environment with Python3
+    mkvirtualenv --python=python3_path myenv
+
+##### Finally!
+
+    # Install python-coverage3 by
+    easy_install coverage


sils1297:
Can you retry on current master? Got a fix pushed yesterday as I promised
you.
Am 07.03.2015 09:27 schrieb "Shivani Poddar" <notifications@github.com>:

shivanipoddariiith:
I did update my repo before I tried again today. Wonder why it showed me the same error.

sils1297:
I had an unstable internet connection while pushing. Seems commits didn't
get up. (These trains are too fast these days...) Could you try on
wip/sils/newlines branch?
Am 07.03.2015 19:44 schrieb "Shivani Poddar" <notifications@github.com>:

sils1297:
Sorry this something slightly related. Message could be better indeed. I'll
take a look at this. Thanks.
Am 07.03.2015 21:00 schrieb "Lasse Schuirmann" <lasse.schuirmann@gmail.com>:

shivanipoddariiith:
Ah, hard luck, I fixed another issue around, so could you check if those 2 are in sync now. I could move on to other things then :)

sils1297:
Go, move on. I'll try to tidy up as much as possible on Monday.
Am 07.03.2015 21:06 schrieb "Shivani Poddar" <notifications@github.com>:

sils1297:
See also https://github.com/coala-analyzer/coala/commit/d3f890938225636b87c03794851d1bc787f619b1 by the way. Is already acknowledged so it will probably be pushed this evening or so.

sils1297:
Please rephrase your commit messages. We're using the same rules as GNOME, written down somewhere in docs/getting_involved.

Second commit: this message would only be applicable if the save setting is set to non-false.

sils1297:
Art didn't see Mischa already reviewed. If my comment says something else he's right because he probably took more time than I could now.

shivanipoddariiith:
@Makman2 You are right about this confusing the user, I'd just remove it for what I feel. Should we go ahead with that then ?

Makman2:
ah yes sorry. A new pull request for my issue #367 is really better^^
Okay so just rebase and delete the last commit^^

shivanipoddariiith:
So, I have rough commits for the 3 odd issues we had. Removed the one about .coafile.
I still have the one about coverage bug and the gettext error message fix. Also added some Mac hints.
If these look okay, I can clean this up. :)

sils1297:
@shivanipoddariiith we usually make our commits perfect, so each commit is one change and we don't push a row of commits like

 * Add something
 * Ooops that was wrong, correction
 * Again
 * ...

but we squash all the "oops" commits so we have very few perfect commits.

sils1297:
and we rebase, so you shouldn't have a commit like "resolving merge commits"

sils1297:
just as GNOME does (due to bugzilla patch submittance)

sils1297:
As for the commit messages, our specification is a bit imprecise, I filed https://github.com/coala-analyzer/coala/issues/371 for this. Please use imperative language for the shortlog, thats common and also in the GNOME Commit message guidelines.

Makman2:
So far what I know for the commit messages:
- Every shortlog starts with a tag:
  `tag: ...`
- After the tag a short description follows:
  `tag: Add something`
- The shortlog has 50 chars maximum!
- If you want to be more detailed since the shortlog doesn't suffice, wrap two lines and write the detailed one:
  ```
  tag: Add something

  Add the new super-duper component that does whatever it should do.
  I write even more in more than one line since I want it...
  ```
- The detailed log has 72 chars per line maximum! (I hope it's 72.. It confuses me everytime^^)
- GitHub supports automatic linking for issues and references them online. Just type the issue into the commit message :)
  i.e. `Fixes #0815` (That would even close the issue when this commit gets merged, but that's described more detailed on github's website^^)

sils1297:
@Makman2 about commit messages: The tag is not mandantory but it's great to get some context. And please don't rely on github and provide a full link to the issue, not only the number. (Especially since we're thinking about switching to phabricator at some point)

sils1297:
looks good besides indentation

sils1297:
pushed with modified commit message

PR #364, opened 2015-03-05
Escaping helper functions v2

Makman2:
Takes over from #357.
Completely renewed all commits. Each function now comes with its complete test set and functionality.

PR #385, opened 2015-03-15
Makman2/escaping3

Makman2:
Takes over as v3 from #364.

Main changes:
- Restructurized test data/results (each function under test got an own `setUp` function), so code duplication is minimized and also test results are easier to oversee and edit.
- Tests: Added some test strings/cases
- `search_for()`: removed `max_match` parameter, since it's not needed because of the generator behaviour.
And because of that it became just a forwarding function for `re.finditer()`.
- `search_in_between()`/`unescaped_search_in_between()`: Filter generators for easier matching algorithm.

### DIFF
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -39,3 +39,59 @@ def limit(iterator, count):
             if i == count:
                 break
 
+
+def split(pattern,
+          string,
+          max_split = 0,
+          remove_empty_matches = False):
+    """
+    Splits the given string by the specified pattern. The return character (\n)
+    is not a natural split pattern (if you don't specify it yourself).
+    This function ignores escape sequences.
+
+    :param pattern:              A regex pattern that defines where to split.
+    :param string:               The string to split by the defined pattern.
+    :param max_split:            Defines the maximum number of splits. If 0 or
+                                 less is provided, the number of splits is not
+                                 limited.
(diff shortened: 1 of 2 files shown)


Makman2:
Comment for build:
`yield from` is introduced in python 3.3

sils1297:
then you can't use it or you need to do a custom branch for python 3.2 since we want to support that

Makman2:
Hm no problem, then just using
```
for elem in iter:
    yield elem
```

Makman2:
Okay changed^^

Makman2:
Man why GitHub again displays the commits in the wrong order? That bugs me somehow...

Makman2:
Ah damn syntax error, I should have run the tests again... stand by...

sils1297:
Yeah, I really hate github for that.

2015-03-16 0:34 GMT+01:00 Makman2 <notifications@github.com>:

Makman2:
I don't know why git itself tells me that I introduced an whitespace error because of the `\n` at EOF but looking at the commit 01e77f0 everything seems to be okay.

sils1297:
I *think* git tells you because you got two `\n` before EOF. Ideally your last line holds a few chars (is nonempty) followed by `\n` - which some editors show as an own line and some don't because originally every line has to end with a \n, its not like a line *seperator* but more like a line *ending*.

sils1297:
#388 takes over

PR #387, opened 2015-03-16
catching exceptions for globs

sils1297:


sils1297:
thanks for review! :)

PR #388, opened 2015-03-17
escaping v4

sils1297:


Makman2:
PS: I totally forgot to implement tests that combine `max_split`/`max_match` with the auto-trim feature. Shall we write some into these commits or do we merge it and extend them later?

Makman2:
I let you know if you can merge, I'll make a custom test locally^^

sils1297:
ack from me, waiting for go from @Makman2 to push

Issue #294, opened 2015-02-18
Add HTML printer


sils1297:
Would be nice to have some printer that does HTML on the fly and can be used as a log printer. This would be useful in order to create the HTML Interactor, see bug #277 .

sils1297:
see https://github.com/coala-analyzer/coala/blob/new/coalib/internal/output/HTMLWriter.py for related work

sils1297:
HTMLWriter exists now.

PR #395, opened 2015-03-18
Adding HTML Log printer feature

shivanipoddariiith:
This is a WIP, a basic layout for an HTML Printer.
It highlights different color for the basic log_types : 
- WARNING (yellow)
- ERROR (red) 
- DEBUG (blue)

Addresses : #294 

TODO - Adapt it into a prettier template


sils1297:
need to do #411 first

PR #418, opened 2015-03-30
Wip/strong processing

smfn3321:
A set of further useful functions to deal with escape sequences in strings.

### DIFF
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -367,3 +367,24 @@ def unescaped_rfind(string, sub, start=None, end=None):
             end = position
         else:
             return position
+
+
+def unescaped_finditer(string, sub, start=None, end=None):
+    """
+    Yields all indices in the string where substring sub is found
+    unescaped, such that sub is contained in the slice s[start:end].
+
+    :param string: Arbitrary String
+    :param sub:    Substring of which the position is to be found
+    :param start:  Begin of string slice that restricts search area
+    :param end:    End of string slice that restricts search area
+    :return:       Iterator that yields all positions of sub in string,
+                   independent of slice borders!
+    """
+    while True:
+        position = unescaped_find(string, sub, start, end)
+        if position >= 0:
+            yield position
+            start = position + 1
+        else:
+            raise StopIteration
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -294,6 +294,9 @@ def unescaped_search_in_between(begin,
                elem.group(compiled_begin_pattern.groups + 2))
 
 
+LAST_POSITION_UNESCAPED_PATTERN = re.compile(r"(?<!\\)(?:\\\\)*$")
+
+
 def position_escaped(string, position):
     """
     Checks whether a char at a specific position of the string is escaped by an
@@ -308,8 +311,7 @@ def position_escaped(string, position):
     if not isinstance(position, int) or not 0 <= position < len(string):
         raise ValueError("param position is not an int in the valid range")
 
-    pat = re.compile(r"(?<!\\)(?:\\\\)*$")
-    return not pat.search(string[:position])
+    return not LAST_POSITION_UNESCAPED_PATTERN.search(string[:position])
 
 
 def unescaped_find(string, sub, start=None, end=None):
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -311,3 +311,30 @@ def position_escaped(string, position):
     pat = re.compile(r"(?<!\\)(?:\\\\)*$")
     return not pat.search(string[:position])
 
+
+def unescaped_find(string, sub, start=None, end=None):
+    """
+    Return the lowest index in the string where substring sub is found
+    unescaped, such that sub is contained in the slice s[start:end].
+
+    :param string: Arbitrary String
+    :param sub:    Substring of which the position is to be found
+    :param start:  Begin of string slice that restricts search area
+    :param end:    End of string slice that restricts search area
+    :return:       Position of sub in string, independent of slice borders!
+    """
+    if not isinstance(string, str):
+        raise ValueError("param string is not a string")
+    if not isinstance(sub, str):
+        raise ValueError("param sub is not a string")
+    if start is not None and not isinstance(start, int):
+        raise ValueError("param start is not an int")
+    if end is not None and not isinstance(end, int):
+        raise ValueError("param end is not an int")
+
+    while True:
+        position = string.find(sub, start, end)
+        if position >= 0 and position_escaped(string, position):
+            start = position + 1
+        else:
+            return position
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -388,3 +388,18 @@ def unescaped_finditer(string, sub, start=None, end=None):
             start = position + 1
         else:
             raise StopIteration
+
+
+def unescaped_findall(string, sub, start=None, end=None):
+    """
+    Lists all indices in the string where substring sub is found
+    unescaped, such that sub is contained in the slice s[start:end].
+
+    :param string: Arbitrary String
+    :param sub:    Substring of which the position is to be found
+    :param start:  Begin of string slice that restricts search area
+    :param end:    End of string slice that restricts search area
+    :return:       List of all positions of sub in string, independent of
+                   slice borders!
+    """
+    return list(unescaped_finditer(string, sub, start, end))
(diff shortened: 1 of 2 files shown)


sils1297:
assigning @Makman2 as reviewer. Probably best if we have a meeting and talk about how we want the escaping stuff handled in the future.

Makman2:
Generally for tests: Use the test strings in the test class `self.test_strings` to cover all possible fails.
It requires some effort to implement a test but my experience has shown that it's worth.

Makman2:
BTW: I've already written the prototypes for `unescaped_search_for()` and `nested_search_in_between()`. You can look at them in the Makman2/unescaped_search_for and Makman2/search_nested branch.

smfn3321:
@Makman2 @sils1297 Ok guys, I know this discussion is getting a little messy, but I don't want to delete it by pushing something at this point. Still, I do want to talk about a couple of point.

The reason why __I__ need StringProcessing, is that I want to enable escape support for our globs. For Glob Parsing, I need to know the position of special characters `"(", ")", "/" etc.` and I need to split on `"|"`

I did make a silly mistake earlier when I tried using `unescaped_search_in_between` and re.escaped the string and not the pattern. This leaded to wrong results obviously. I checked out Makman2/unescaped_search_for now and I see that I can indeed use this in a style that fits me.

Now I'm unsure what If anything of my code is still useful, please leave your opinion on this.

This is the final `is_position_unescaped` function which doubles the performance of the old one for average strings (the older one is faster if the string contains ~50 consecutive backslashes :D ):

```python
def is_position_escaped(string, position=None):
    """
    Checks whether a char at a specific position of the string is preceded by
    an odd number of backslashes.

    :param string:   Arbitrary string
    :param position: Position of character in string that should be checked
    :return:         True if the character is escaped, False otherwise
    """
    preceding_escapes = 0
    # iterate backwards, starting one left of position.
    # Slicing provides a sane default behaviour and prevents IndexErrors
    for i in range(len(string[:position]) - 1, -1, -1):
        if string[i] == '\\':
            preceding_escapes += 1
        else:
            break
    return bool(preceding_escapes % 2)
```

This is the current finditer function for not-regex substrings:
```python
def unescaped_raw_finditer(string, sub, start=None, end=None):
    """
    Yields all indices in the string where substring sub is found unescaped,
    such that sub is contained in the slice string[start:end].

    :param string: Arbitrary String
    :param sub:    Substring of which the position is to be found
    :param start:  Begin of string slice that restricts search area
    :param end:    End of string slice that restricts search area
    :return:       Iterator that yields all positions of sub in string,
                   independent of slice borders!
    """
    while True:
        position = string.find(sub, start, end)
        if position < 0:
            break
        elif is_position_escaped(string, position):
            start = position + 1
        else:  # position is >= 0 and not escaped, therefore valid
            yield position
            start = position + 1
```
In comparison, this function uses `unescaped_search_for()` and yields exactly the same results:
```python
def unescaped_search_for_finditer(string, sub):
    sub = re.escape(sub)
    for match_elem in unescaped_search_for(sub, string):
        yield match_elem.start()
```

Performance is similar for strings with matches...
```
>>> timeit.timeit('l = list(unescaped_raw_finditer("alabama", "a"))', number=1000000, setup="from coalib.parsing.StringProcessing import unescaped_raw_finditer, unescaped_search_for_finditer")
7.796900624000045
>>> timeit.timeit('l = list(unescaped_search_for_finditer("alabama", "a"))', number=1000000, setup="from coalib.parsing.StringProcessing import unescaped_raw_finditer, unescaped_search_for_finditer")
9.062474061000103
```
... and different only for strings with a lot of escaped non-matches:
```
>>> timeit.timeit('l = list(unescaped_raw_finditer("kentucky", "a"))', number=1000000, setup="from coalib.parsing.StringProcessing import unescaped_raw_finditer, unescaped_search_for_finditer")
1.036196289000145
>>> timeit.timeit('l = list(unescaped_search_for_finditer("kentucky", "a"))', number=1000000, setup="from coalib.parsing.StringProcessing import unescaped_raw_finditer, unescaped_search_for_finditer")
6.346805084999687
```

smfn3321:
I did develop a split() function with similar functionality and performance differences, but there, too, a wrapper would do.

So, all in all:
 * should I update this PR at all or do we stay with the regex functions as-is?
 * which, if any, of 'my' regex-less functions should I include (and write tests for etc.)
 * which, if any, wrappers should I include/write
 * would it make more sense to introduce yet another parameter to the regex functions, that `re.escape()`s the pattern?

Makman2:
I like the `is_position_escaped()`. That should be work way faster than regexes (since backslash searching starts at the indicated position). I'll test myself with `timeit` a maybe little performance increasement: Don't use modulo, just toggle a boolean variable. Means:
```python
def is_position_escaped(string, position=None):
    escapes_uneven = False
    # iterate backwards, starting one left of position.
    # Slicing provides a sane default behaviour and prevents IndexErrors
    for i in range(len(string[:position]) - 1, -1, -1):
        if string[i] == '\\':
            escapes_uneven = not escapes_uneven
        else:
            break
    return escapes_uneven
```

But I really didn't thought regexes would be so slow when using non-special characters. We may could consider to write them like you completely ourselfs.

- You don't need to update this PR, just open a v2 branch. I think we should keep the discussion here^^
- I would definitely add the `is_position_escaped`, that's a nice function.
- The "wrapper" from `unescaped_raw_finditer()` (but we need to talk about the name...) is quite good, the performance tests from you are really meaningful :)
- Good idea, makes sense. But do we then implement also your performant code then or just `re.escape` the pattern?

Man this thing got quite complicated...

sils1297:
I also prefer not over modulo :)

2015-04-02 21:39 GMT+02:00 Makman2 <notifications@github.com>:

smfn3321:
To be honest, right now I think it might be best to either:
 * add another option to the existing functions that toggles whether `string` is ``regex.escape`d or not
 * Refactor the existing functions, naming them something like `unescaped_regex_find` and write wrappers called `unescaped_flat_find` or similar.

or MAYBE:
 * just call the parameter `regex` instead of `string` so people now they have to `re.escape` first

The 'raw' functions are faster, but either they will differ in functionality, which is confusing when using both of them, or we will have to apply all changes to both functions having two times the amount of work.

The is_position_escaped function should be included as proposed by Makman2 with 'bool switching'

Makman2:
Really using both, an extra param together with wrapper functions that would just change this param?

smfn3321:
No, bad sentence structure :D
Changed for clarification

Makman2:
Totally forgot to answer...
I would go with the wrapper functions. If someone knows how to use regexes he doesn't need to bother with the extra function argument that disables regexes in principle.
And why rename `string` into `regex`? `string` is the string to search in, not the matching pattern (this is `pattern` or either `begin` and `end` in the search-in-between-functions).

smfn3321:
You're right, I messed the arguments up.

Anyways, coming back from Easter and having gained a little distance, it now looks a little ridiculous to me to wrap the functions *just* to call re.escape. Also, I read some code and documentation about re.search() and match objects and I guess I'm getting why you went with it.

The only reason for me to not write a more specific wrapper in the glob module __would__ be the need for this exact kind of functionality (re.escape, starting index...) in a couple of other modules which would have do duplicate the code (which isn't a whole lot though) or import it from glob, which is messy. Anyway, this might be the way to go.

sils1297:
I do prefer having the following signatures:

```
position_is_escaped(string, position)  # Note: default value for position doesn't make sense
split(string, delimiter, max_split=0, remove_empty_matches=False, use_regex=False)
# Same for unescaped split and the other things
```

I think this is the easiest API to have and everything should be rather simple to implement.

smfn3321:
Why would a default position not make sense? 
`position=None` in `for i in range(len(string[:position]) - 1, -1, -1):` would lead to the last position of the string being checked i.e. whether or not the whole string is an escape sequence for the character afterwards. I think this is useful.

@Makman: what do you think about `use_regex=False`? As I understand this would even mean to `re.escape` the pattern by default

sils1297:
I'm in favor of `use_regex=False` because this imitates the usual `str.split` behaviour which is thus kind of the default split behaviour, also it is simpler :)

My default value opinion didn't change but if you two agree just go ahead ;)

Makman2:
Hmm yes indeed... I can't decide anymore, too many options...

smfn3321:
Well, let's commit sacrilege and follow a KDE design principle:
> simple by default, powerful when needed

From which I would derive:
 **use `use_regex=False` and `re.escape()` by default**
 * This fits the most basic usecase "find a substring where it is unescaped" for *any* pattern
 * This prevents errors that arrise from somebody forgetting to `re.escape()` his pattern
 * This provides advanced functionality through an argument which I think is appropriate

Concerning `position_is_escaped`:
I will take the freedom to leave the default behaviour as-is because I perceive it as simple, useful, and pythonic. Consider positions in slices:
```
>>>"string"[:None]
'string'
>>>"string"[:20]
'string'
>>>"string"[:-2]
'stri'
```

sils1297:
ok

2015-04-07 12:01 GMT+02:00 Fabian Neuschmidt <notifications@github.com>:

smfn3321:
@Makman2 do you agree?
If you do, would you be so kind to implement the parameter into the existing and coming functions? I will try to push a PR for `position_is_escaped` tonight.

Makman2:
Yep I'll do :)

PR #426, opened 2015-04-07
Makman2/unregexify

Makman2:
Adds the `use_regex` parameter to all existing StringProcessing functions.

### DIFF
--- a/coalib/misc/StringConverter.py
+++ b/coalib/misc/StringConverter.py
@@ -77,7 +77,8 @@ class StringConverter:
 
         self.__list = list(StringProcessing.unescaped_split(
             pattern,
-            self.value))
+            self.value,
+            use_regex=True))
 
         if remove_backslashes:
             for i in range(len(self.__list)):
--- a/coalib/parsing/StringProcessing.py
+++ b/coalib/parsing/StringProcessing.py
@@ -257,7 +257,8 @@ def unescaped_search_in_between(begin,
                                 end,
                                 string,
                                 max_matches=0,
-                                remove_empty_matches=False):
+                                remove_empty_matches=False,
+                                use_regex=False):
     """
     Searches for a string enclosed between a specified begin- and end-sequence.
     Also enclosed \n are put into the result.
@@ -278,11 +279,20 @@ def unescaped_search_in_between(begin,
                                  limited.
     :param remove_empty_matches: Defines whether empty entries should
                                  be removed from the result.
+    :param use_regex:            Specifies whether to treat the begin and end
+                                 patterns as regexes or simple strings.
     :return:                     An iterator returning the matched strings.
     """
-    # Compilation of the begin sequence is needed to get the number of
-    # capturing groups in it.
-    compiled_begin_pattern = re.compile(begin)
+    if not use_regex:
+        begin = re.escape(begin)
+        end = re.escape(end)
+        # No need to compile the begin sequence, capturing groups get escaped.
+        begin_pattern_groups = 0
+    else:
+        # Compilation of the begin sequence is needed to get the number of
+        # capturing groups in it.
+        compiled_begin_pattern = re.compile(begin)
+        begin_pattern_groups = compiled_begin_pattern.groups
 
     # Regex explanation:
     # 1. (?<!\\)(?:\\\\)* Unescapes the following char. The first part of this
@@ -308,11 +318,11 @@ def unescaped_search_in_between(begin,
 
     if remove_empty_matches:
         matches = trim_empty_matches(matches,
-                                     [compiled_begin_pattern.groups + 1,
-                                      compiled_begin_pattern.groups + 2])
+                                     [begin_pattern_groups + 1,
+                                      begin_pattern_groups + 2])
 
     matches = limit(matches, max_matches)
 
     for elem in matches:
-        yield (elem.group(compiled_begin_pattern.groups + 1) +
-               elem.group(compiled_begin_pattern.groups + 2))
+        yield (elem.group(begin_pattern_groups + 1) +
+               elem.group(begin_pattern_groups + 2))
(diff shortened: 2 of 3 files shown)


sils1297:
I assigned @smfn3321 for review, he is probably better suited. I may take a look here and there but thats only supplemental.

sils1297:
please also look at scrutinizer issues, some more duplication detected plus all those names are far too long

Makman2:
The names indicate the test structure, so I don't know how to rename them so the names makes still sense but much shorter.
Also the code duplication applies for the specific tests. I can put them into a function but they would be used only one time, so doesn't make much sense^^

sils1297:
I'm still not convinced we're writing the tests the right way because this is getting so complex and bad.

Makman2:
Give me hint *-*

sils1297:
It's nothing more than this, I don't have any clue on how to improve that. ;)

smfn3321:
Sorry I still can't really review yet, but I agree that tests are way to complicated here. And I'm not complaining about the duplication.

For example, in your last commit you added:
```
def test_unescaped_search_in_between_disabled_regex(self):
    self.assertUnescapedSearchInBetweenEquals(
        self.test_strings,
        self.test_unescaped_search_in_between_disabled_regex_expected,
        self.test_unescaped_search_in_between_disabled_regex_pattern,
        self.test_unescaped_search_in_between_disabled_regex_pattern,
        0,
        True,
        False)
```
The only thing this method does is calling another (previously defined) method with a lot of parameters that one needs to look up (somewhere else). It is really hard to understand what is going on there since you have to look up the params and then the method and it really doesn't help that everything looks so similar.

To be honest, I would actually allow more code duplication in favor of better readability. Since you got a list of test strings, I would consider making it available at the class or module level for all functions. I would define results in the tests themselves though and pass the patterns directly. This way, it is far easier to understand the tests since everything is at one place and readable.

Also, please look at the python zip command, this could probably make the iteration simpler.

Makman2:
`zip()` is definitely a good alternative for assertions. But I'll only include it if I need to refactor tests again^^

But I think the last commit you refer to is already in master, so why do you review that one? :O

Also I don't know if it's that a problem to look up the assertion functions. Once you read the little doc it's obvious what they do. It's no big deal to look up some functions, I need to do that permanently when programming. I would then really go over minimal code-duplication, since changing duplicated code involves changing it everywhere where it's duplicated^^

Btw, the test string is already at module level like any other testing parameter. I think I haven't something local there. The idea behind to put everything into module level is that many tests have similar results that can be easily derived from a "master-result".

smfn3321:
The commit I cited was https://github.com/coala-analyzer/coala/commit/d9c5c2be7cfdc98a57edba391fa469f5de52ef9c which is your second to last commit ^^


--

It really is a shame that subtests did not get introduced before python 3.4
```
class NumbersTest(unittest.TestCase):

    def test_even(self):
        """
        Test that numbers between 0 and 5 are all even.
        """
        for i in range(0, 6):
            with self.subTest(i=i):
                self.assertEqual(i % 2, 0)
```
==>
```
======================================================================
FAIL: test_even (__main__.NumbersTest) (i=1)
----------------------------------------------------------------------
  File "subtests.py", line 32, in test_even
    self.assertEqual(i % 2, 0)
AssertionError: 1 != 0

======================================================================
FAIL: test_even (__main__.NumbersTest) (i=3)
----------------------------------------------------------------------
  File "subtests.py", line 32, in test_even
    self.assertEqual(i % 2, 0)
AssertionError: 1 != 0

======================================================================
FAIL: test_even (__main__.NumbersTest) (i=5)
----------------------------------------------------------------------
  File "subtests.py", line 32, in test_even
    self.assertEqual(i % 2, 0)
AssertionError: 1 != 0
```

Makman2:
Yeah but the commit uses the assert function introduced from me in an earlier pull request.

And what are subtests?
I don't understand your example :D

smfn3321:
It's not really related to this PR, ant only got introduced in 3.4 so we probably won't use it anyway. Subtests are useful when your unit tests performs the same test over a range of inputs (as is the case here). They provide a separate test for every input set that can fail on its own.

Atm, we only know whether the test passed for the entire result list/dict or failed at some point. With subtests, we would have known which inputs worked and which not.

In the above example, a subtest is specified for each integer that checks whether or nor the integer is even. You can see that the test fails separately for every uneven number ;)

sils1297:
@smfn3321 can we somehow emulate that subtest class for python < 3.4 like we do with the barrier?

smfn3321:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest
 
I'm sure that it is *possible* but I don't think it is worth it. It doesn't add any features to coala and we should probably focus on the main program and not the tests. 

I like to use subtests during initial testing on 3.4 for debugging. It is fairly easy to add to a test if you don't understands where exactly it fails.

sils1297:
half of the code in coala is tests, if we can ease that a bit with reasonable amount of work we should do it IMO.

Other possibility is to drop support for python <3.4: https://lists.debian.org/debian-devel-announce/2015/03/msg00016.html and https://packages.debian.org/jessie/python3

sils1297:
Hey, if no one has anything against it we'll drop support for python 3.2
and 3.3 so we can use those subtests.

Ubuntu LTS does have 3.4, the Debian released later this month will also
have that, fedora 22 has it and with brew on Mac you get it. Arch is no
question.
Am 13.04.2015 13:12 schrieb "Fabian Neuschmidt" <notifications@github.com>:

Makman2:
You sure? 3.4 is quite new^^

sils1297:
Its over a year since 3.4.0
Am 13.04.2015 21:04 schrieb "Makman2" <notifications@github.com>:

PR #433, opened 2015-04-14
Wip/sils/doc

sils1297:


sils1297:
merged rebased version, no conflicting commits came in since then so avoiding rereview

PR #448, opened 2015-04-15
output: Add ClosableObject

sils1297:


sils1297:
added documentation and renamed object in #450

PR #480, opened 2015-04-25
StringProcessing: Add unescape() function

Makman2:
Function that removes all escape-sequences.
This one is a really short commit because I can only test the standard test-strings^^

I hope that's it.


Makman2:
Changed the function, should be the right one now^^

PR #519, opened 2015-05-12
setup: Use setuptools

sils1297:
As for https://docs.python.org/3/library/distutils.html we rather want
the enhanced setuptools functionality.


Makman2:
ack

PR #530, opened 2015-05-14
general: Rename "coala" to "coala.py"

Makman2:
Appends the python file extension to coala. Using the ".py" file
extension is more consequent for python files.
Also this simplifies cross platform invocation (for Windows and Linux).


Makman2:
Seems we need to distribute both, `coala` and `coala.py` to work on all systems.
I don't know if code duplication is the right choice here, need to work on that.
And why circleci tries suddenly to start tests with python 2.7... I'm getting frustrated...

Issue #574, opened 2015-05-25
sys.path in TestHelper has all directories in test folder


AbdealiJK:
In the tests, paths are added to the path and not removed. This causes the sys.path to get cluttered and makes modules clash

```
def __check_module_skip(filename):
        module_dir = os.path.dirname(filename)
        if module_dir not in sys.path:
            sys.path.insert(0, module_dir)
```

sils1297:
how about we write a preserve_sys_path contextmanager (like suppress_stdout), something like

```
@contextmanager
def preserve_sys_path():
    old = sys.path.copy()
    yield
    sys.path = old
```

So I'd say lets use this around all that?

PR #579, opened 2015-05-26
Fix #553 #574

### LINKED ISSUE #553: run_tests --html doesn't work in ubuntu
In ubuntu and debian systems, coverage isn't installed in `/usr/bin/coverage3` ... rather, it is installed as `/usr/bin/python3-coverage`. Hence, `--html` flag doesn't work. The corresponding launchpad bug is [here](https://bugs.launchpad.net/horizon/+bug/1241332)

It would be better to use `python3 -m coverage` as this is platform independent.

AbdealiJK:


AbdealiJK:
@sils1297 : I am not sure why scrutinizr is failing (It says coverage not found) ... When I tried to "retry inspection with ssh" it says I do not have access to that page. Could you give me access so I can debug ?

sils1297:
I've given you developer access, please ping me if you need more but that shouuld suffice.

PR #629, opened 2015-06-08
.gitignore: Ignore Kdevelop4 files

fneu:
This prevents files generated automatically by Kdevelop to be committed by accident. It is inline with the suppression of the PyCharm .idea files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+.directory
 .idea
+.kdev4
 __pycache__
 *~
 build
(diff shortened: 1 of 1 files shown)


fneu:
@sils1297 why is scrutinizer failing?

sils1297:
its doing this since some time now for all PRs - I have no clue, think it's
an issue on their side. I think we shouldn't let that hinder us from
pushing to master...

2015-06-08 18:19 GMT+02:00 Fabian Neuschmidt <notifications@github.com>:

PR #633, opened 2015-06-09
fnmatch and glob escape support

fneu:
This branch is wip, and there are no tests yet for the new custom fnmatch. This is because I am planning on giving the glob test a major overhaul as soon as I am done implementing escape support.

That said, my fnmatch is functionally very close to the implementation in Python 3.6 (dev) and I tested the added features extensively. Using this instead of the built-in fnmatch.fnmatch() did not break any tests.

Therefore, feel free to use this branch if you need the fnmatch feature.


fneu:
this version of glob is broken on windows for now.

Probably a drive letter issue, I will check that later today

PR #675, opened 2015-06-18
circle: Move zanata upload to rultor

sils1297:
We may rebuild old builds on master, we only want to upload that kind of
stuff when we merge to master so rultor is the option for that.


sils1297:
@rultor merge

This is a test PR to test if rultor does zanata the right way, note that this gets merged to the `test-master` branch :)

sils1297:
ok that kind of testing doesn't seem to work, merging to the real master

PR #678, opened 2015-06-18
Wip/sils/cced

sils1297:


sils1297:
@rultor config from master please

sils1297:
@rultor merge please

although I expect it to fail I'm interested in the log

AbdealiJK:
@rultor merge - Lets see if rultor merges without errors

sils1297:
@AbdealiJK it won't merge as it uses the old config with the new scripts i.e. its expected to fail

AbdealiJK:
Ah - right - the old one doesn't know that it should decrypt the file.

sils1297:
@AbdealiJK so the log outputs each command, i.e. it'll do the same for the encrypted file which we don't want, so I'll restrict access to the log

sils1297:
@rultor stop please

PR #679, opened 2015-06-18
circleci: Remove python-tmpenv

AbdealiJK:
Earlier packages like python-gi and python-dbus were installed in a
temporary python environment and then moved to the actual virtualenv.
This commit removes the need for the tmpenv and directly installs it in
the virtualenv.


AbdealiJK:
@rultor merge

Issue #596, opened 2015-05-28
Need a function similar to fnmatch in glob


AbdealiJK:
In python's glob there's a function called `fnmatch` which allows to check if a filename would be in a glob of a pattern.

Currently, in our implementation, the only way to do that is by checking the output of `glob`
So, `glob('/**/*.py')` returns all matches and to check if a file is there we do: `file_name in glob('/**/*.py')`
This is bad, as if a pattern like `/**/*.py` is given, it'd take a huge time on most harddisks - while the actual regex test is very simple and much faster.

smfn3321:
Ok, I will ad this functionality to our globbing soonish.

Issue #460, opened 2015-04-17
document what our glob supports


sils1297:

fneu:
fixed by #682

Issue #396, opened 2015-03-18
escaping for globs


sils1297:

sils1297:
marking enhancement since for paths this is at least *usually* not needed. However the globbing is incomplete without so blocker for 0.2

PR #682, opened 2015-06-24
glob rework and fnmatch introduction

fneu:
This PR adresses:
 * #396 escaping for globs
 * #460 document what our glob supports
 * #596 need a function similar to fnmatch in glob
 * #681 add examples to documentation


sils1297:
@rultor merge

Issue #500, opened 2015-05-04
Allow Keyboard interrupt to abort a single test


fneu:
We can aaccept keyboard interrupt on test execution to skip a single test if it is endlessly looping for example. I think it is a good idea to add a little timer afterwards to allow the abortion of test execution altogether. It would be slightly annoying to abort 50 tests otherwise.

AbdealiJK:
Solved in https://github.com/coala-analyzer/coala/commit/07a1c37b3e26cfca285e5e1aa09a076cfb89a04e

PR #696, opened 2015-06-27
run_tests: Add ability to skip a test manually

fneu:
Keyboard Interrupt will skip the test that is currently running.
A second Keyboard Interrupt in rapid succession stops the script.


fneu:
this addresses #500

sils1297:
rejected as per discussion :)

PR #710, opened 2015-06-29
coverage: Use future codecov

sils1297:
This is the most up to date uploader at any time and it fiddles with the
coverage data and should allow us to finally remove platform dependent
no covers.


sils1297:
not yet

AbdealiJK:
I do not think codecov is going to complete.

sils1297:
@rultor merge

PR #717, opened 2015-07-01
Make circle faster v2

AbdealiJK:


AbdealiJK:
Speed up doesnt seem to be much. But now all the dependencies are cached. I can't think of anything else to make it faster :)

sils1297:
@rultor merge

PR #736, opened 2015-07-06
appveyor

AbdealiJK:


AbdealiJK:
Tries this out on a windows computer ... it worked.
But sadly this computer is 32bit. Can someone try this on a 64bit computer ?

sils1297:
@Makman2 ?

Makman2:
I'll try it again tomorrow, the problem is that nuget does not work system-wide, you can install packages only in an existing project (from Visual Studio). For testing on windows it's okay, but users have anyway problems if they use a coala release version^^

AbdealiJK:
I don't follow - I thought nuget is just a package manager ... it installs itself in a local directory - but it's usable in the whole system ?

Any comments on chocolatey ? Apparently win10 will have a revamped version of chocolatey inbuilt as a package manager (I read)

Makman2:
okay I'll test that with the stand-alone nuget command line tool, I used the one integrated into VS2013^^

Makman2:
I thought we already had munkres on appveyor...
But: it works, the tests aren't ignored for the libclang stuff and gettext

Makman2:
But sometimes I get an error:
```
 28/78 | DiffTest
msgfmt: error while opening "build\locale\bn_BD\LC_MESSAGES\coala.mo" for writing: No such file or directory
```

sils1297:
> Sometimes it works, sometimes not :O

Try it with `-j1`, maybe it depends on the i18nTest being execute in parallel as they may be fiddling with the same files :/

Makman2:
but then we need to restructurize the test so that they don't do so^^

sils1297:
> but then we need to restructurize the test so that they don't do so^^

Does it work? Yeah, can you file a bug?

sils1297:
@Makman2 actually I can reproduce easily if I do `-j16` so that seems to be the cause. However tests aren't failing because of that so it's a testing beauty bug I'd presume.

Makman2:
yeah just wanted to say, I'll open an issue

Makman2:
https://github.com/coala-analyzer/coala/issues/746

Makman2:
I just see I only have python x86, I don't have x64 version :(
the msi installer doesn't install me one ._.

Makman2:
okay I found the x64 version, and bad news: The test aborts after 20secs because of the timeout :(

AbdealiJK:
you can use the `-T` flag to increase it. `-T 0` will make it wait indefinitly

AbdealiJK:
Can someone help out in the last commit ? It creates 3 scrutinizer issues - but I don't think they are avoidable

Makman2:
The clang tests are ignored in appveyor^^

AbdealiJK:
Yeah - I've removed them for now - I thought I'd fix the clang issue on
64bit later

Makman2:
even on x86 they are skipped^^

Makman2:
ah okay you removed clang completely temporarily^^

AbdealiJK:
@rultor merge

PR #780, opened 2015-07-15
Makman2/pylintbear fix

Makman2:


Makman2:
We can discuss the name `shellify_path_argument` :P

AbdealiJK:
Hey, the whole Shell module in coalib is really not needed - there is already a shlex in python - https://docs.python.org/3.2/library/shlex.html#shlex.quote

In 3.2.5, you'd need to use `pipe.quote()`

Makman2:
@AbdealiJK `shlex.quote` does not work
And also I don't want to check every time the python version to determine which function to use, so this would be anyway the reason to make a function that generalizes the behaviour

AbdealiJK:
Could you elaborate on why `shlex.quote` does not work? Or in what way it
does not work ? On windows ?

Makman2:
I just tried to use `shlex.quote(filename)` in the PyLintBear, but it failed again like before because of spaces in my working directory.
Btw this is the code from `shlex.quote()` (copied from python source)
```
def quote(s):
    """Return a shell-escaped version of the string *s*."""
    if not s:
        return "''"
    if _find_unsafe(s) is None:
        return s

    # use single quotes, and put single quotes into double quotes
    # the string $'b is then quoted as '$'"'"'b'
    return "'" + s.replace("'", "'\"'\"'") + "'"
```

Makman2:
This extreme `"'"'"` usage looks a bit weird to me and really not Windows proof (and Windows confirmed me that)

sils1297:
@rultor merge

Makman2:
I rebase...

Makman2:
reack for all

sils1297:
@rultor merge

PR #789, opened 2015-07-15
StringProcessingTest: Use str(key) for assert msg

Makman2:
Using repr(key) causes the output to wrap even the keywords in quotes.


sils1297:
@rultor merge

PR #815, opened 2015-07-18
zanata: Translate from coalib and bears only

AbdealiJK:
If the whole repo is taken, it has a change of taking .py files which
we did not write - and this may make it break.


AbdealiJK:
blocked by https://github.com/coala-analyzer/coala/pull/809 as this doesnt have any good way of testing until that is completed

sils1297:
@rultor merge

PR #816, opened 2015-07-18
No change

sils1297:
This commit is deliberately empty. It is needed because we switched our
deployment from CircleCI to rultor. Rultor deployment could not happen
during the merge of the switching commits because rultor uses the
configuration from master, this empty commit triggers the new
deployment. It is not needed in the future, this is a one-time action.


sils1297:
@rultor merge

PR #821, opened 2015-07-18
i18n: Use python-gettext

sils1297:
This makes translation compilation independent of gettext.


sils1297:
this fixes https://github.com/coala-analyzer/coala/issues/752 btw.

PR #870, opened 2015-07-25
ClangCloneDetectionBear: Reduce threshold

sils1297:
My evaluations have come up with 0.185 as a very good threshold value
that yields almost no false positives which is more important than a few
clones not found. If someone whishes to do a more elaborate examination
of clones he is hereby invited to raise the threshold manually.


sils1297:
@rultor merge

PR #885, opened 2015-07-29
ConsoleInteraction: Color print fix

Makman2:
The color setting overrides the ConsolePrinter default. If now
providing None, the desired default is used.

This really fixes the color-on-windows issue^^

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -200,7 +200,7 @@ def print_results(log_printer,
                   result_list,
                   file_dict,
                   file_diff_dict,
-                  color=True,
+                  color=None,
                   pre_padding=3):
     """
     Print all the results in a section.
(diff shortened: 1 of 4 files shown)


sils1297:
@Makman2 how about you do all printing related stuff from now on in the PyPrint repo so we move over asap?

Makman2:
since it integrates in our coala code I would merge also here. I think it's easier to migrate that later^^

sils1297:
@Makman2 it has almost no effect on coala and we migrate over anyway, actually I already started the migration. Now changing printers on this side just makes things incredibly complex.

Makman2:
okay, but keep the PR open so we know that we want to implement color support^^ (and how we need to integrate into coala)

Makman2:
Merged in PyPrint, see https://github.com/coala-analyzer/PyPrint/pull/12.

PR #893, opened 2015-08-05
Makman2/cancellation on pytest

sils1297:


coalib/processes/communication/InterruptProcess.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def interrupt_process(pid):
+    """
+    Interrupt a process by its pid.
+
+    Note for Windows users:
+    Be sure that the pid you pass shares the same process group as your program
+    that calls this function. If they don't share the same process group, this
+    call blocks for 10 seconds to be sure that this is the case.
```
sils1297: @Makman2 it only blocks for *up to* 10 secs right?

coalib/processes/communication/InterruptProcess.py
```diff
@@ -0,0 +1,32 @@
+import os
+import signal
+import platform
+
+if platform.system() == "Windows":
+    from ctypes import windll
+    from time import sleep
+    CTRL_C_EVENT = 0x0
```
sils1297: uncovered code I guess, we don't have cross platform coverage merging yet

coalib/tests/processes/communication/InterruptProcessTest.py
```diff
@@ -0,0 +1,56 @@
...
+        self.assertEqual("Hello coala", queue.get(timeout=0.1))
+        self.assertEqual("Hello coala", queue.get(timeout=0.1))
+        self.assertEqual("End", queue.get(timeout=0.1))
+        self.assertEqual("End2", queue.get(timeout=0.1))
+        p.terminate()
+        p.join()
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=2)
```
sils1297: I wonder if on windows, you could split off a new process group here in which you execute the test? This would be compatible with our testing infrastructure only for now but pytest is no priority so I'd be fine with that.

coalib/processes/communication/InterruptProcess.py
Makman2: If they don't share the same process group it blocks *at least* 10 secs because the interrupt will never come.
But maybe I should add a note that this call blocks until the KeyboardInterrupt was delivered...

coalib/tests/processes/communication/InterruptProcessTest.py
Makman2: hm good idea, will do that :)

coalib/processes/communication/InterruptProcess.py
sils1297: well yeah but that's nott the expected case so thats ok

sils1297:
@Makman2 how's it going? Have something mergeable today?

Makman2:
@sils1297 *cough* on the flight...

sils1297:
@Makman2 yeah, be sure to take the laptop in the cabin baggage ;)

PR #899, opened 2015-08-09
setup: Assert supported version before imports

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/733


sils1297:
@rultor merge

PR #926, opened 2015-08-14
Makman2/generate-repr-autoprop

Makman2:
`@generate_repr` includes in parameterless mode now non-private properties.


sils1297:
@rultor merge

PR #940, opened 2015-08-31
Prototype into the directon of new bear design

sils1297:
Hey guys, please take a look at this, taking some shape IMO and looks like some way we could use to at transform the usability first and do the internals later. We'd then use this decorator for everything first, create one for the global bear too (even simpler than this one), then remove the LocalBear class completely (including all the related complicated logic) and simplify the internals for the GlobalBear to not use a class anymore and use a simple multiprocessing.pool or so.

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -1,13 +1,12 @@
 # So all sections are disabled by default
 Default.enabled = false
 PythonCheck.enabled = true
-DOCS.enabled = true
 
 [Default]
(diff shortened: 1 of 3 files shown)


sils1297:
Update on october 26th, rebased so only the relevant parts are here as the rest is merged in other PRs by now.

Makman2:
@sils1297 wanna close this one but keep the branch until next-gen-bears are merged?

sils1297:
ok

PR #979, opened 2015-10-11
README: Point to real documentation

sils1297:
The link we used before was generated by readthedocs and doesn't put the
user into the documentation but into the public project view which is
rather useless.


sils1297:
@rultor merge

PR #980, opened 2015-10-11
SourcePosition: Add column

sils1297:


sils1297:
@Makman2 I've created https://github.com/coala-analyzer/coala/pull/986 as a successor of this one so the old comments can be accessed.

PR #993, opened 2015-10-13
Result: Use uuid to generate id

sils1297:
This has a higher chance of being unique since it's made for having a
high chance of uniqueness.

We use uuid4 because it generates a random UUID, other generators may
invade the users privacy by hashing in platform dependent data and
privacy is important :).


Makman2:
@rultor merge

PR #998, opened 2015-10-17
Makman2/doc-format-bear3

Makman2:
This is actually (still) not the bear, but the doc-extraction prerequisites.
Contains facilities to extract documentation from source code.
For this purpose parts of the `StringProcessing` module were rewritten to use new `Match` objects that contain more information than plain strings (for the `search_in_between` functions).


Makman2:
BTW the tests fail because the pypy build, it only supports python 3.2 and I removed support for it in some commits.

PR #1005, opened 2015-10-24
ConsoleInteraction: Move out result context printing

sils1297:
This simplifies the result printing by far because it splits out the
context printing of the result and gets rid of the state machine.

As a result functionality changes a bit, gets a bit simpler, because
we're getting rid of the complicated state machine with current file and
current line: the full context (file and the three lines before) is
printed for each result, even if it has the same file as the last one.
If no file but a line is given, it will simply be printed project wide
(way simpler: no file -> project wide. Ignore line.)

This is a preparation for getting the new results in. Plus a
simplification and code removal which is good too :).


sils1297:
@rultor merge

PR #1026, opened 2015-11-02
enable pylint

sils1297:


AbdealiJK:
We could enable pylint - but not use some of thw things we dont need. i.e.be
conservative about which options we enable in pylint.

This would atleast enable basic checks like space around operators which
pylint has.

sils1297:
sounds good to me,  but no need to keep this PR around which just enables
it with no config IMO

PR #1039, opened 2015-11-13
Prototype: BashSemicolonBear

sils1297:
Inspired by
https://github.com/skudriashev/bashlint/blob/master/bashlint.py#L50

### DIFF
--- a/.coafile
+++ b/.coafile
@@ -2,6 +2,7 @@
 Default.enabled = false
 PythonCheck.enabled = true
 DOCS.enabled = true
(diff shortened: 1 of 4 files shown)


sils1297:
ok probably we want a proper string parsing routine first, a bear that uses the language definition to get strings, comments and stuff and returns an annotated line other bears can use so they see what is actually code and what not.

sils1297:
needs #1040

sils1297:
we'll get shellcheck, removing this

PR #1045, opened 2015-11-14
Bear: Fail on inconvertible value

sils1297:
We don't want to execute a bear if its value isn't in a type he desires.

Fixes https://github.com/coala-analyzer/coala/issues/800


sils1297:
Note that travis may fail right now due to PyLint we think. Please ignore travis during review, we'll probably deactivate the builds at least temporarily.

sils1297:
@rultor merge

PR #1048, opened 2015-11-15
Constants: Pass on failure of tags dir creation

sils1297:
In certain environments access to this directory might not be given.
coala should not fail to run because of this.


sils1297:
@rultor merge

PR #1049, opened 2015-11-15
ignore files in collectors

fneu:
Not necessarily complete, i.e. a a command line options is missing, but this should work if the appropriate setting is set in the section.


sils1297:
@rultor merge

PR #1051, opened 2015-11-16
Add diff to JSON

sils1297:


sils1297:
will be migrated on top of https://github.com/coala-analyzer/coala/pull/1041 because it relates to a few other things in there; no way to drag it out. Order is important.

PR #1068, opened 2015-11-18
coala: Remove internationalization

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/1044


sils1297:
I grepped for a lot of stuff so I think I catched everything needed to remove i18n.

PR #1070, opened 2015-11-18
ClangBear: Add clang_cli_options

sils1297:
This allows customizing clang which is really basic.


sils1297:
moved over to the other PR because of the test refactoring, no way to split this up

PR #1085, opened 2015-11-24
Fneu/resfilter rubbish

fneu:


fneu:
closed because there is a successor

PR #1086, opened 2015-11-24
Result Filtering

fneu:


sils1297:
@fneu progress? please also fix the gitmate review :)

coalib/results/SourceRange.py
```diff
@@ -59,4 +59,7 @@ def file(self):
         return self.start.file
 
     def overlaps(self, other):
+        if self.file != other.file:
+            return False
+
         return self.start <= other.end and self.end >= other.start
```
sils1297: hey, just make this `self.file != other.file and ...`

coalib/results/ResultFilter.py
```diff
@@ -0,0 +1,220 @@
...
+
+    :param original_result: A result of the old files
+    :param modified_result: A result of the new files
+    :return:                Boolean value whether or not the properties match
+    """
+    # origin might be a class or class name
+    original_origin = isinstance(original_result.origin, str) and \
+        original_result.origin or original_result.origin.__name__()
+    modified_origin = isinstance(modified_result.origin, str) and \
+        modified_result.origin or modified_result.origin.__name__()
```
sils1297: the origin is always a string, you can create this with a class for convenience but internally it's always a string

coalib/results/ResultFilter.py
```diff
@@ -0,0 +1,220 @@
...
+        return False
+
+    elif original_result.severity != modified_result.severity:
+        return False
+
+    elif original_result.debug_msg != modified_result.debug_msg:
+        return False
+
+    else:
+        return True
```
sils1297: why not simply `return original_result.origin == modified_result.origin and original_result..message == ...`?

coalib/results/ResultFilter.py
```diff
@@ -0,0 +1,220 @@
...
+    :param original_result_diff_dict: diff for each file for this result
+    :param modified_result:    A result of the new files
+    :param modified_result_diff_dict: guess
+    :return:                     Boolean value whether the SourceRanges match
+    """
+    for file_name in original_file_dict:
+
+        try:  # fails if the affected range of the result get's modified
+            original_total_diff = \
+                diff_dict[file_name] + original_result_diff_dict[file_name]
```
sils1297: try to use brackets instead of backslashes, otherwise mischa will kill you.

sils1297: at least yell badly

coalib/results/ResultFilter.py
```diff
@@ -0,0 +1,220 @@
...
+    Checks whether the Diffs of two results describe the same changes
+
+    :param original_file_dict: Dict of lists of file contents before changes
+    :param modified_file_dict: Dict of lists of file contents after changes
+    :param diff_dict:          Dict of diffs describing the change in each file
+    :param original_result:    A result of the old files
+    :param modified_result:    A result of the new files
+    :return:                   Boolean value whether the Diffs match
+    """
+    return False # pragma: no cover
```
sils1297: huh? What's this :D

coalib/results/ResultFilter.py
```diff
@@ -0,0 +1,220 @@
...
+            uniques.append(m_r)
+
+        elif True not in [source_ranges_match(
+                original_file_dict,
+                modified_file_dict,
+                diffs_dict,
+                o_r,
+                orig_result_diff_dict_dict[o_r],
+                m_r,
+                mod_result_diff_dict_dict[m_r]) for o_r in original_results]:
```
sils1297: you should only consider the ones where the basics matched before here!

1. For performance
2. They both must match, there might be another result which has other basics but the same diff.

sils1297: so you probably want one iteration over all modified, inside an iteration over all originals and then do the checks sequencially.

coalib/results/ResultFilter.py
fneu: One Line to rule them all,
one line to find them,
one line to bring then all and
in the darkness bind them

Tbh. I just found this more expressive although one can make a case against 
using `!=` all that much

coalib/results/ResultFilter.py
fneu: Infrastructure... I can take it out now but I will need it anyways.

coalib/results/ResultFilter.py
fneu: Yeah, I thought I could add the second check like this... But that is 
complete rubbish isn't it? I'm amazed this works for all my tests...

coalib/results/ResultFilter.py
sils1297: maybe even a loop: `all(getattr(original_result, member) == getattr(modified_result, member) for member in ['origin', 'message', ...])`?

coalib/results/ResultFilter.py
sils1297: just fix it in a performance efficient way, this has a very bad complexity `O(n*m)`, if basics are not fit there's no need to match the rest which takes a lot of time

coalib/results/ResultFilter.py
sils1297: especially the pragma no cover is funny :P

coalib/results/ResultFilter.py
fneu: now that is pretty elegant, I'm impressed!

coalib/results/ResultFilter.py
fneu: well, otherwise I would have needed to call it in filter_results... :p

sils1297: ah interesting, I would have guessed that you need to do the pragma for the whole function... well then take it out until you need it ok?

sils1297:
@fneu I can close this, right?

sils1297:
restore and reopen if needed

PR #1090, opened 2015-11-25
Makman2/conf parser auto trim

Makman2:
Allows to parse empty iteration elements!


sils1297:
included in https://github.com/coala-analyzer/coala/pull/1093 to reduce CI builds

PR #1106, opened 2015-12-03
Clang: Take position if no range given otherwise

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/1105


sils1297:
@rultor merge

PR #1107, opened 2015-12-03
WIP Use clang package

sils1297:
this way we get rid of the cindex in our own problem. Don't merge before somewhen saturday, then the pip package for the clang python 3 bindings will probably be published on pypi.


AbdealiJK:
Just a note here, the package needs to be added to setup.py 's dependencies

sils1297:
Yes,as soon as I know the name of the package :) it won't be published by me

sils1297:
this is done in https://github.com/coala-analyzer/coala/pull/1106

PR #1116, opened 2015-12-11
Put stuff into language dirs

sils1297:
@AbdealiJK @Makman2 quick review but IMO important to make things clear.


sils1297:
@rultor merge

PR #1135, opened 2015-12-14
Lint: Check for executable availability

sils1297:
@Makman2 awesome, it works :)

```
lasse@lssteady ~/prog/coala (git)-[master] % ./coala default -f test.js -b JSHintBear
Executing section Default...
[WARNING][14:13:08] The bear JSHintBear does not fulfill all requirements. jshint is not installed.
```

this PR still needs tests though

### DIFF
--- a/coalib/bearlib/abstractions/Lint.py
+++ b/coalib/bearlib/abstractions/Lint.py
@@ -3,12 +3,13 @@ import tempfile
 import re
 import sys
 
+from coalib.bears.Bear import Bear
 from coalib.misc.Shell import escape_path_argument
 from coalib.results.Result import Result
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 
 
-class Lint():
+class Lint(Bear):
     """
     :param executable:   The executable to run the linter.
     :param arguments:    The arguments to supply to the linter, such
@@ -105,3 +106,15 @@ class Lint():
             column=groups.get("column", None),
             end_line=groups.get("end_line", None),
             end_column=groups.get("end_column", None))
+
+    def check_prerequisites(self):
+        if not self.executable:
+            return True
+
+        try:
+            subprocess.Popen([self.executable, '--version'],
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE)
+            return True
+        except OSError:
+            return self.executable+" is not installed."


AbdealiJK:
I believe this is because int he inheritance we use `LocalBear, Lint` - it should be `Lint, LocalBear` because python uses the left most class as higher priority.

sils1297:
I think we shouldn't rely on the order here, should we?

sils1297:
i.e. we can just make `Lint` a bear so the user does not have to deal with this

AbdealiJK:
Inheritance does rely on the order ...
It's not bad coding per se

On Thu, Dec 17, 2015 at 11:59 AM, Lasse Schuirmann <notifications@github.com

sils1297:
Not bad coding, bad usability.

AbdealiJK:
But ... how else do you use multiple inheritance ?
If I have same function name in class `A` and class `B` - there is a well
defined priority order.

I rarely use multiple inheritance - so, let's look at the other cases
possible:

1. Lint inherits from Bear
Even if you inherit `Bear` in `Lint` - you'd still have to order your
inheritance in the final bear as `LocalBear, Lint` and not `Lint,
LocalBear` - otherwise the Bear's `run` would be the one which is used -
which raises NotImplemented. In fact this would get complicated - because
you need `LocalBear`'s `run` and `Lint`'s `check_prerequisites` -> implying
both need different orders.

2. Lint inherits from LocalBear
The other option is to inherit LocalBear into Lint - which is not right,
because it constraints it to be a LocalBear - which is not logically
correct, as Linters are not constrained so.

Am I missing some method of doing this which makes things simpler ?



On Thu, Dec 17, 2015 at 12:23 PM, Lasse Schuirmann <notifications@github.com

sils1297:
Hm, still it's kind of weird that we want to override a method in `Lint` that is not in one of it's base classes, it's clearer if it is. We could also make `Lint` not derive from `Bear` and invoke the check in `__init__`... however we won't have a log message in that case. This feels ugly. How'd you solve this?

AbdealiJK:
I'm not sure :( - I played around with inheriting Bear into Linter earlier - and this seemed to be the best way I could think of. (Now `check_prerequisite` introduces more complexity)

AbdealiJK:
How about :
 - We make a function "check_shell_command" in `Utils.py` - which can also be used to check when tests need to be skipped.
 - Then, we make another class `LocalLintBear` which simply inherits `LocalBear` and `Lint` and overrides the `check_prerequisite` and gives the error code from `check_shell_command` ?
 - And also modify `Lint` to not take a `filename` specifically - but rather take any args. Which and `LocalLintBear` can get a filename, clean it and pass it to Lint as a arg. So, we'd be modularizing out the Lint part and LocalBear part.
 - Later : we can make a GlobalLintBear

sils1297:
That sounds good, can you look at this?

AbdealiJK:
Sure, but not until ~30th - Im travelling around until then

sils1297:
Ok, will keep this open, maybe I'll find some time until then, maybe not.

2015-12-17 8:47 GMT+01:00 AbdealiJK <notifications@github.com>:

abhsag24:
@AbdealiJK python3 manages inheritance in such a way, that the overwritten methods in any of the subclasses is used, regardless of its order in the `__mro__` (as long as there are no conflicts, then order is used) . so in our case since Lint inherits from Bear and overwrites the `check_prerequisites` function , the order of Lint doesn't matter as long as LocalBear itself doesn't overwrite the same function(which it doesn't), so Lint and LocalBear can have any order. 

The problem in our code is same as "The Diamond Problem", explained in 
 http://www.python-course.eu/python3_multiple_inheritance.php

abhsag24:
time to close this now?

sils1297:
yes @abhsag24 thanks for your help here :)

PR #1162, opened 2015-12-26
Makman2/auto-apply

Makman2:


sils1297:
From Gitter:

@Makman2 I see very much forward to this PR! But... why did you implement it in ConsoleInteraction?! This way it won't work with coala-ci, coala-json and so on.
@Makman2 didn't we agree to put it in Processing!?

Makman2:
#1170 takes over.

PR #1167, opened 2015-12-29
Collectors: Work on list copy when removing ignored

sils1297:
Otherwise nondeterministic - almost evil - things may happen.


sils1297:
@rultor merge

PR #1174, opened 2015-12-30
Move requirements to requirements.txt

sils1297:
This way the actual versions of the dependencies are fixed.


sils1297:
Note that I cancelled builds on 599126d for performance reasons, no need to worry about the red flag ;)

PR #1183, opened 2016-01-01
towards pytest - continuation

sils1297:


Makman2:
@sils1297 can't you make a skip-function for the dbus stuff? less code duplication^^

sils1297:
@Makman2 it's not really duplication, it covers different imports everywhere, hard to put into a function.

Makman2:
can't you just import dbus completely and see whether it fails or not? That could fit into a function^^

sils1297:
@Makman2 then the other imports will fail and everything will error out again

sils1297:
@rultor merge

PR #1205, opened 2016-01-10
Processing: Treat failing actions as unprocessed

Makman2:
If an auto-apply action fails due to a raised exception, it's
considered unprocessed.


sils1297:
@rultor merge

PR #1210, opened 2016-01-13
ResultFilter: Fix crashes in filter_results()

### LINKED ISSUE #1209: filter_results() crashes on certain results
The Problem lies in the `remove_result_ranges_diffs()` function:

```python
for source_range in reversed(original_result.affected_code):
            file_name = source_range.file
            new_file = remove_range(mod_file_dict[file_name],
                                    source_range)
            mod_file_dict[file_name] = new_file
```

Here, the affected code of a result gets removed from a file. The reversed sorting ensures that SourceRanges behind others get removed first so that the indices are not off.

The bane of this implementation are overlaps though: they cause the indices to be off, resulting in the wrong code being removed. This might even throw an `IndexError` if the deviation is too high:

example: removing char 2-4 and 3-6 from the string `"123456"` should result in  `1` as the chars 2-6 should be removed.

instead:
- chars 3-6 are removed, resulting in: `"12"`
- chars 2-4 from `"12"` are removed, resulting in `IndexError`.

fneu:
fixes #1209

### DIFF
--- a/coalib/results/ResultFilter.py
+++ b/coalib/results/ResultFilter.py
@@ -1,6 +1,7 @@
 import copy
 
 from coalib.results.Diff import Diff, ConflictError
+from coalib.results.SourceRange import SourceRange
 
 
 def filter_results(original_file_dict,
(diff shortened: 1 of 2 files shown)


fneu:
All Issues are addressed more cleanly in #1230

PR #1230, opened 2016-01-21
ResultFilter fix and range joining

fneu:


GitCop:
The commit message

 * Commit: d16b28d0d8751e6565209b515bfacfbb2a16436f
  * Your commit message body contains a line that is longer than 72 characters

Guidelines are available at http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/#how-to-write-good-commit-messages

---
This message was auto-generated by https://gitcop.com

Makman2:
@fneu maybe you want to implement already a specific `overlaps` for `SourceRange` too that also takes files in accounts?
```py
def overlaps(self, other).
    return self.file == other.file and TextRange.overlaps(self, other)
```

fneu:
@Makman2 the current one does take files into account.

It only evaluates `return self.start <= other.end and self.end >= other.start`,
but (unlike TextPositions) SourcePositions actually get sorted by file before line before char.

So unless two SourceRanges are from the same file, start and end of one will both be sorted either in front or behind the other SourceRanges' positions and they will not overlap.

Makman2:
@fneu oh okay but it was more meant for general purpose of this class (if you manually do `overlap` on two `SourceRange`s), nice that your algorithm works fine with that using sorting :)

sils1297:
@rultor merge

PR #1243, opened 2016-01-26
expand textrange

fneu:


sils1297:
unack 0e1ddd1

sils1297:
@fneu uhm so this needs a bit polishing from you and then we can hopefully merge it?

sils1297:
@fneu what about the None bug?

sils1297:
@rultor merge

PR #1248, opened 2016-01-29
CodeStyle: replace filehandler with the more mnemonic file

SanketDG:
In many places, filehandler(r) is used when opening a file:

with open(..) as filehandler:

This is replaced with the word file, which is more semantic:

with open(..) as file:


sils1297:
ack dd3eade6a02

sils1297:
@rultor merge

PR #1251, opened 2016-01-29
DefaultArgParser: Add bear explanation

sils1297:
Usability testing with Ana has revealed that this is a weak point in our
documentation.


sils1297:
@rultor merge

Issue #971, opened 2015-09-29
Use built-in function `vars(object)` instead of using `__dict__` directly


Makman2:
Imo this is cleaner^^

cw1:
According to this http://pythondoeswhat.blogspot.de/2012/01/dict-and-vars.html, the use of `vars(object)` is much slower than `__dict__`. Shall we ignore this performance penalty?

sils1297:
@cw1 thanks for digging that up, that'll convince @Makman2 for sure to leave this as it is and maybe even go towards using `.__len__()` instead of `len()`

Makman2:
To be honest I dont believe that. Ill test that soon locally with a venchmark^^

sils1297:
@Makman2 why? It makes totally sense. vars will first have a few
conditions, then access dict, while accessing it directly will avoid all
that unneeded checking.

2015-10-27 14:26 GMT+01:00 Makman2 <notifications@github.com>:

Makman2:
My benchmark:
```
With vars():
2.172344666005436
Directly via __dict__:
1.4370877611060155
With len():
4.380952825268093
Directly via __len__:
3.092824838883348
```
so about 50% slower.
I added another case where I just check for method delegation, and this is the bottleneck (method delegation is 100% slower), not the if-conditions or else inside `vars()` or `len()`.

sils1297:
so what do we want to do?

2015-10-27 16:19 GMT+01:00 Makman2 <notifications@github.com>:

Makman2:
I'm in favor of using built-in methods still, since they are built-in and so have a meaning to be and we use the standard interface from python.
Also `vars` is not critically used in our code, just here and there, so it wouldn't have a performance impact.

cw1:
The usage of `vars` instead of `__dict__` and `len()` instead of `.__len__()` is more consistently i think. So may we should integrate it to our code guidelines, if we ignore the performance impact.

sils1297:
why not use the faster version everywhere consistently? It has the same meaning.

Makman2:
since there're default functions for, also it makes code for newcomers more hard to read^^

sils1297:
You think so? __len__ is pretty obvious IMO...

2016-01-08 18:15 GMT+01:00 Mischa Krüger <notifications@github.com>:

Makman2:
since it's not that obvious to find in documentation I don't like using `__len__`^^

fneu:
[...]

PR #1260, opened 2016-01-31
codestyle: Use builtin vars() instead of __dict__

fneu:
this is about #971

there are 3 more usages of `__dict__` in our code, but those cannot be replaced (like `hasattr(__dict__)`).

We have 4 usages of `__len__` but all of them are function definitions, so that seems to be it


sils1297:
@rultor merge

PR #1263, opened 2016-02-01
LineLengthBear: Add ignore_length_regex

sils1297:
We probably want to allow ignoring lines that e.g. contain a URL or
something else that cannot be shortened. Checkstyle does it this way and
I think that is good.


sils1297:
@rultor merge

PR #1265, opened 2016-02-01
Sils/checkstyle

pkmm91:


sils1297:
@pkmm91 is there a reason you're creating pull requests with other peoples commits that are useless and unfinished? I am assuming you are trying to cheat the indiahacks hackathon system. If you are a participant of that hackathon and cannot explain these actions I will assume this intent and make sure you are excluded from the hackathon. We are looking for real contributions and are well aware of having own work lying around somewhere.

PR #1295, opened 2016-02-06
Pytest second try

fneu:


AbdealiJK:
@fneu - DbusTest and DbusServerTest are both creating the server by the bus name `org.coala_analyzer.v1.test`. By changing one of them, this issue doesn't come up in my local computer.

Also, a note : pytest hasn't been added to requirements.txt

AbdealiJK:
@sils1297 tat Scrutinizer issue cannot be avoided. It is required to know whether glib is installed or not (for skipping)

sils1297:
@rultor merge

PR #1300, opened 2016-02-07
Fixed find-config not searching current directory

hypothesist:
Starting search directory should be file_path, not the base directory of the file_path (which is one level behind).


sils1297:
unack  8ec70f8, also your tests seem to be failing, try the `./run_tests.py` command :)

PR #1301, opened 2016-02-07
find_user_config: handle files and directories properly

hypothesist:
Starting search directory should be file_path, not the base directory of the file_path (which is one level behind).
Also, initially, ConfigurationGatheringTest.py tested find_user_config
with a file as file_path (instead of a directory). That should not
be the case since all usages of find_user_config pass directories
check:
coalib/output/dbus/DbusDocument.py:47
coalib/settings/ConfigurationGathering.py:121

Fixes https://github.com/coala-analyzer/coala/issues/1246


hypothesist:
How about the following?
* If file, get dirname
* Otherwise, use as it is

This should do it? And I'll edit the commit message :)

PR #1316, opened 2016-02-08
Collectors: Add limit_paths() & expose it in args

AbdealiJK:
Currently there is not method to limit the files which coala would run on
in the command line arguments (without changing the coafile). This commit
adds a new argument namely "--limit-files" which ensures that coala does
not analyze any file outside the glob patterns mentioned in this argument.

This is useful specially when coala's analysis is required in only 1 file
like in the case of editor plugins.

Fixes https://github.com/coala-analyzer/coala/issues/1315

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -45,16 +45,19 @@ def icollect(file_paths):
             yield match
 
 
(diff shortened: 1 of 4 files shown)


sils1297:
@rultor merge

sils1297:
@rultor merge - really :@

Issue #1320, opened 2016-02-08
License in setup.py


AbdealiJK:
Currently we use `license="AGPL v3",` in the setup.py
This isn't the common short form for the license we are using.

[SPDX](https://en.wikipedia.org/wiki/Software_Package_Data_Exchange) is the standardized way of writing the license's short form. https://spdx.org/licenses/

According to it, we should be using `AGPL-3.0`.

Note : I came across this because `npm` kept complaining about it while I was doing coala-atom

PR #1322, opened 2016-02-08
setup.py: Standard license format

nishant-mor:
Fixed: #1320


sils1297:
@nishant-mor this needs a rebase onto the current master (IIUC @sudheesh001 is just helping you with that - thanks!), the commit message looks surprisingly good - this is your first commit to coala IIRC?

So commit message: two little things.

1. Instead of just the issue number we prefer the full URL, this way links work in Git and not only in GitHub.
2. `Standard license format` is not bad actually but doesn't fully comply to the convention. Good commit messages are *really* hard. I'd prefer `Normalize license format` since it's in imperative tense.

sudheesh001:
@nishant-mor Maybe you can close this PR now.

Issue #1317, opened 2016-02-08
Remove docs/requirements.txt


AbdealiJK:
We have a `./docs-requirements.txt` and also a `docs/requirements.txt`

I had created the `docs/requirements.txt` about half a yr ago. https://github.com/coala-analyzer/coala/commit/f14b734fbdfd330d18c4c09de05cd9593d6dea4c

I notice we're not using mkdocs anymore ? Maybe this should be removed ?

sils1297:
oh yeah, sorry

AbdealiJK:
Also, `.misc/deps.linux.sh` needs to be removed. It just says "No dependencies to install for linux" and is not used anywhere

sils1297:
@AbdealiJK can you file another bug for this? I'm doing a workshop in a month or so and need 15 newcomer suitable bugs for that :P

AbdealiJK:
sure

yash14123:
PR sent for this. ! :)

yash14123:
@AbdealiJK the second point made : removing .misc/deps.linux.sh . By that you mean .misc/deps.sh?

AbdealiJK:
@yash14123 - No, I did not mean deps.sh
Please ignore the comment aboud `.mics/deps.linux.sh`. It is invalid (check here for more details https://github.com/coala-analyzer/coala/issues/1318)

yash14123:
Can you help me @AbdealiJK ? I am stuck >.< . 
What should I do with the PR now. I committed two changes and both of them are vaguely written. Just guide me in this please.

AbdealiJK:
@yash14123 sure ! Why dont you come on to https://gitter.im/coala-analyzer/coala so its easier to chat ?

AbdealiJK:
Closed with https://github.com/coala-analyzer/coala/commit/11cdb46e3b4c267f061763bb12c8ceb7c6e21a4d

PR #1326, opened 2016-02-09
Remove docs/requirements.txt #1317

yash14123:


sils1297:
@yash14123 I've cancelled the builds for this one, I'll put your commits into a PR of mine to reduce rebases and builds because our build machines are having problems keeping up with so much action :( they really want us to pay.

Issue #1359, opened 2016-02-09
coala_delete_orig : Typo, no space


AbdealiJK:
In coalib/coala_delete_orig.py, We are giving a INFO message about deleting files using the log_printer.
After "Deleting file ..." there is not space.

We need to add the space for better formatting

iamshubh22:
@AbdealiJK I would fix this.

sils1297:
@iamshubh22 you've got mail! : ) please accept that so we can assignyou

sils1297:
@iamshubh22 also consider joining our channel at https://gitter.im/coala-analyzer/coala

iamshubh22:
@sils1297 Done :smile:

Issue #1360, opened 2016-02-09
coalib/coala_delete_orig.py - Better message when deleting file


AbdealiJK:
Currently, we say "Deleting file" which sounds scary.

We would like to say "Deleting old backup file"

iamshubh22:
@AbdealiJK I would fix this.

AbdealiJK:
@Makman2 @sils1297 could you add him to the org so that he can assign himself ?

sils1297:
@iamshubh22 cool, please do both issues (this and https://github.com/coala-analyzer/coala/issues/1359) in different commits, ideally in the same PR because they are related. (Otherwise you'll get conflicts.)

Also note that we'd love it if you don't solve more than two newcomer issues because they are really hard to come up with and we still need some for other contributors.

PR #1361, opened 2016-02-09
Fix issue #1359, #1360

iamshubh22:


sils1297:
Hey, you can use `git log` to get a feeling on how we do commit messages :) Alternatively we have a full guide on commit messages at http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/#how-to-write-good-commit-messages . (Nobody get's this right at the first time.)

So you can use the `git rebase --interactive` to batch edit your commit messages.

Other than that you see that my beloved GitMate bot has already found some minor issues an the tests seem to fail (though they automatically fail if GitMate complains so that might be the reason).

If you need help with all this ask on https://gitter.im/coala-analyzer/coala - don't be afraid, lots of people have problems with that. Over time it'll get way easier as you get accustomed to it.

PR #1382, opened 2016-02-10
Ajk/has bugs

AbdealiJK:


sils1297:
one minute 50 seconds, thanks.

PR #1429, opened 2016-02-11
[requires.io] dependency update on master branch

sils1297:


sils1297:
requires.io seems to work now, I need to configure it a bit so it doesn't do stupid stuff like this though

PR #1515, opened 2016-02-15
coala_dbus: Move lambdas to functions

AbdealiJK:
Move lambdas to functions so that they can be tested in
coalaDbusTest.py


sils1297:
@rultor merge

PR #1541, opened 2016-02-16
JSONFormatBear: Add keep_unicode argument

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/1527

### DIFF
--- a/bears/js/JSONFormatBear.py
+++ b/bears/js/JSONFormatBear.py
@@ -38,14 +38,17 @@ class JSONFormatBear(Lint, LocalBear):
             filename,
             file,
             json_sort: bool=False,
-            tab_width: int=SpacingHelper.DEFAULT_TAB_WIDTH):
+            tab_width: int=SpacingHelper.DEFAULT_TAB_WIDTH,
+            keep_unicode: bool=False):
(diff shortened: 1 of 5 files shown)


sils1297:
@Makman2 did I mention that I *hate* unicode on windows?

AbdealiJK:
@sils1297 you can simply do an if condition for the time being and mention that windows doesnt support keeping unicode ?
(As this blocks the folks a `zeroclickinfo-goodies` from using coala ... assuming they do not use windows)

AbdealiJK:
:/ I just tested this on a windows cmd and they seem to pass

AbdealiJK:
@sils1297 you can fix this by opening the file in `coalib.misc.ContextManagers.prepare_file` using something like
```
with open(filename, 'w', encoding='utf-8') as file:
```

Just tested it out and it worked.

sils1297:
ack df77299 5021200 9b95c12, apparently gitmate really cares about it being lower case...

sils1297:
alright, updated gitmate, it should handle casing now at least for ack/unack

sils1297:
@rultor merge

PR #1595, opened 2016-02-19
codestyle: Use shlex.quote() to standardize code

aayn:
`shlex.quote()` is introduced in Python 3.3. This can, and should
replace `Shell.escape_path_argument()`, which is self-written.

Fixes https://github.com/coala-analyzer/coala/issues/1588

### DIFF
--- a/bears/js/JSHintBear.py
+++ b/bears/js/JSHintBear.py
@@ -1,8 +1,8 @@
 import re
+import shlex
 
 from coalib.bearlib.abstractions.Lint import Lint
 from coalib.bears.LocalBear import LocalBear
-from coalib.misc.Shell import escape_path_argument
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 
 
@@ -29,6 +29,6 @@ class JSHintBear(LocalBear, Lint):
         self.arguments = '--verbose {filename}'
         if jshint_config:
             self.arguments += (" --config "
-                               + escape_path_argument(jshint_config))
+                               + shlex.quote(jshint_config))
 
         return self.lint(filename)
--- a/bears/perl/PerlCriticBear.py
+++ b/bears/perl/PerlCriticBear.py
@@ -1,8 +1,8 @@
 import re
+import shlex
 
 from coalib.bearlib.abstractions.Lint import Lint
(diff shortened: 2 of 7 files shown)


sils1297:
ack 2cd3c6e, tests pass so it's impossible that you forgot an occurrence

AbdealiJK:
@aayn we didn't use `shlex.quote` because it didn't escape certain characters in `cmd` IIRC.
@Makman2 will know more.

AbdealiJK:
Nevermind - he was the one who opened https://github.com/coala-analyzer/coala/issues/1588

sils1297:
@AbdealiJK either we didn't know about this or it was becayse we had 3.2 support

sils1297:
and I think the tests failed back when we didn't escape stuff properly so it should all be well

sils1297:
hah, apparently it's not so simple as the windwows build shows. CC @AbdealiJK @Makman2

aayn:
@sils1297 Appveyor failed. What does that mean?

sils1297:
@aayn apparently @Makman2 was wrong about us being simply able to replace this. Tests on windows fail.

aayn:
\*sigh\* Windows

sils1297:
I think it's best if we let @Makman2 deal with this, this is not newcomer
anymore, if you want look out for another bug in the meantime.

2016-02-19 16:39 GMT+01:00 Aayush Naik <notifications@github.com>:

Makman2:
ahh damn ...
> The quoting is compatible with UNIX shells and with split()

I've not seen this from the python docs...

sils1297:
can it be used to simplify our routine instead maybe?

2016-02-19 18:17 GMT+01:00 Mischa Krüger <notifications@github.com>:

Makman2:
Alright, then we don't replace `escape_path_argument` with `shlex.quote` but we use `shlex.quote` for unix systems so we don't have to deal with that anymore ourselves.

Makman2:
I'll rename the issue and open another one investigating `shlex.quote` on windows^^

sils1297:
Alright @aayn that means that this change isn't doing us any good, @Makman2 will probably rename the issue or create a new one and we have to reject this change. Sorry for misleading you.

PR #1617, opened 2016-02-20
Fix a few more things to get bear independency

AbdealiJK:
Earlier I had used `git grep "from bears"` to check if I had missed anything.
The integration tests do not need to import bears, so this removes the bears dependency from them


sudheesh001:
ack

sudheesh001:
@rultor merge

PR #1643, opened 2016-02-22
Lint: Save important debug info to self

AbdealiJK:
When Lint tests are happening, we need better data to be able to
debug things.
One possibility is to use DEBUG log_messages, but this would give
a lot of data when `-L DEBUG` is used in the cmdline. Hence, we
simply save these variables to `self` so that LocalBearTestHelper
can handle it in a better way.

Related to https://github.com/coala-analyzer/coala-bears/issues/72


sils1297:
@rultor merge

PR #1644, opened 2016-02-22
Globbing: Add escape function for globs

hypothesist:
When the coala directory's full path has brackets, the whole system breaks
because it is interpreted as alternatives. To fix this, the brackets are
escaped before use and then unescaped, leaving all functionality intact.

Fixes https://github.com/coala-analyzer/coala/issues/1613

### DIFF
--- a/coalib/coala_delete_orig.py
+++ b/coalib/coala_delete_orig.py
@@ -6,6 +6,7 @@ from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.parsing import Globbing
 from coalib.settings.ConfigurationGathering import get_config_directory
 from coalib.settings.Section import Section
+from coalib.parsing.Globbing import glob_escape
 
 
 def main(log_printer=None, section: Section=None):
@@ -15,8 +16,9 @@ def main(log_printer=None, section: Section=None):
     if start_path is None:
         return 255
 
-    orig_files = Globbing.glob(os.path.abspath(
-        os.path.join(start_path, '**', '*.orig')))
+    # start_path may have unintended glob characters
+    orig_files = Globbing.glob(os.path.join(
+        glob_escape(start_path), '**', '*.orig'))
 
     not_deleted = 0
     for ofile in orig_files:
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -6,7 +6,7 @@ from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.collecting.Importers import iimport_objects
 from coalib.misc.Decorators import yield_once
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
-from coalib.parsing.Globbing import fnmatch, iglob
+from coalib.parsing.Globbing import fnmatch, iglob, glob_escape
 
 
 def _get_kind(bear_class):
@@ -110,6 +110,9 @@ def icollect_bears(bear_dirs, bear_globs, kinds, log_printer):
     """
     for bear_dir, dir_glob in filter(lambda x: os.path.isdir(x[0]),
                                      icollect(bear_dirs)):
+        # Since we get a real directory here and since we
+        # pass this later to iglob, we need to escape this.
+        bear_dir = glob_escape(bear_dir)
         for bear_glob in bear_globs:
             for matching_file in iglob(
                     os.path.join(bear_dir, bear_glob + '.py')):
--- a/coalib/output/dbus/DbusDocument.py
+++ b/coalib/output/dbus/DbusDocument.py
@@ -10,7 +10,7 @@ from coalib.processes.Processing import execute_section
 from coalib.results.HiddenResult import HiddenResult
 from coalib.settings.ConfigurationGathering import (
     find_user_config, gather_configuration)
-from coalib.settings.Setting import path_list
+from coalib.settings.Setting import glob_list
 
 
 class DbusDocument(dbus.service.Object):
@@ -124,7 +124,7 @@ class DbusDocument(dbus.service.Object):
                     continue
 
                 if any([fnmatch(self.path, file_pattern)
-                        for file_pattern in path_list(section["files"])]):
+                        for file_pattern in glob_list(section["files"])]):
 
                     section["files"].value = self.path
                     section_result = execute_section(
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -3,6 +3,7 @@ import platform
 import re
 
 from coalib.misc.Decorators import yield_once
+from coalib.misc.Constants import GLOBBING_SPECIAL_CHARS
 
 
 def _end_of_set_index(string, start_index):
@@ -27,6 +28,26 @@ def _end_of_set_index(string, start_index):
         closing_index += 1
 
     return closing_index
+
+
+def glob_escape(input_string):
+    """
+    Escapes the given string with ``[c]`` pattern. Examples:
+
+    >>> from coalib.parsing.Globbing import glob_escape
+    >>> glob_escape('test (1)')
+    'test [(]1[)]'
+    >>> glob_escape('test folder?')
+    'test folder[?]'
+    >>> glob_escape('test*folder')
+    'test[*]folder'
+
+    :param input_string: String that is to be escaped with ``[ ]``.
+    :return:             Escaped string in which all the special glob characters
+                         ``()[]|?*`` are escaped.
+    """
+    return re.sub("(?P<char>[" + re.escape(GLOBBING_SPECIAL_CHARS) + "])",
+                  "[\\g<char>]", input_string)
 
 
 def _position_is_bracketed(string, position):
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -19,7 +19,7 @@ from coalib.results.result_actions.PrintDebugMessageAction import (
 from coalib.results.result_actions.ShowPatchAction import ShowPatchAction
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.results.SourceRange import SourceRange
-from coalib.settings.Setting import path_list
+from coalib.settings.Setting import glob_list
 from coalib.parsing.Globbing import fnmatch
 
 ACTIONS = [ApplyPatchAction,
@@ -335,10 +335,10 @@ def instantiate_processes(section,
                              the same for each object.
     """
     filename_list = collect_files(
-        path_list(section.get('files', "")),
+        glob_list(section.get('files', "")),
         log_printer,
-        ignored_file_paths=path_list(section.get('ignore', "")),
-        limit_file_paths=path_list(section.get('limit_files', "")))
+        ignored_file_paths=glob_list(section.get('ignore', "")),
+        limit_file_paths=glob_list(section.get('limit_files', "")))
     file_dict = get_file_dict(filename_list, log_printer)
 
     manager = multiprocessing.Manager()
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from coalib.collecting.Collectors import collect_registered_bears_dirs
 from coalib.misc.Decorators import enforce_signature, generate_repr
 from coalib.misc.DictUtilities import update_ordered_dict_key
-from coalib.settings.Setting import Setting, path_list
+from coalib.settings.Setting import Setting, glob_list
 
 
 def append_to_sections(sections,
@@ -62,7 +62,7 @@ class Section:
         self.contents = OrderedDict()
 
     def bear_dirs(self):
-        bear_dirs = path_list(self.get("bear_dirs", ""))
+        bear_dirs = glob_list(self.get("bear_dirs", ""))
         for bear_dir in bear_dirs:
             sys.path.append(bear_dir)
         bear_dirs = [
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -113,7 +113,7 @@ class Setting(StringConverter):
                                            last part will be stripped of. If
                                            you want to specify a directory as
                                            origin be sure to end it with a
-                                           directory seperator.
+                                           directory separator.
         :param strip_whitespaces:          Whether to strip whitespaces from
                                            the value or not
         :param list_delimiters:            Delimiters for list conversion
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 
 from coalib.misc.Decorators import generate_repr
 from coalib.misc.StringConverter import StringConverter
+from coalib.parsing.Globbing import glob_escape
 
 
 def path(obj, *args, **kwargs):
@@ -15,6 +16,29 @@ def path_list(obj, *args, **kwargs):
 
 def url(obj, *args, **kwargs):
     return obj.__url__(*args, **kwargs)
+
+
+def glob(obj, *args, **kwargs):
+    """
+    Creates a path in which all special glob characters in all the
+    parent directories in the given setting are properly escaped.
+
+    :param obj: The ``Setting`` object from which the key is obtained.
+    :return:    Returns a path in which special glob characters are escaped.
+    """
+    return obj.__glob__(*args, **kwargs)
+
+
+def glob_list(obj, *args, **kwargs):
+    """
+    Creates a list of paths in which all special glob characters in all the
+    parent directories of all paths in the given setting are properly escaped.
+
+    :param obj: The ``Setting`` object from which the key is obtained.
+    :return:    Returns a list of paths in which special glob characters are
+                escaped.
+    """
+    return obj.__glob_list__(*args, **kwargs)
 
 
 def typed_list(conversion_func):
@@ -112,20 +136,23 @@ class Setting(StringConverter):
         self.key = key
         self.origin = str(origin)
 
-    def __path__(self, origin=None):
+    def __path__(self, origin=None, glob_escape_origin=False):
         """
         Determines the path of this setting.
 
         Note: You can also use this function on strings, in that case the
         origin argument will be taken in every case.
 
-        :param origin:      the origin file to take if no origin is specified
-                            for the given setting. If you want to provide a
-                            directory, make sure it ends with a directory
-                            seperator.
-        :return:            An absolute path.
-        :raises ValueError: If no origin is specified in the setting nor the
-                            given origin parameter.
+        :param origin:             The origin file to take if no origin is
+                                   specified for the given setting. If you
+                                   want to provide a directory, make sure it
+                                   ends with a directory separator.
+        :param glob_escape_origin: When this is set to true, the origin of
+                                   this setting will be escaped with
+                                   ``glob_escape``.
+        :return:                   An absolute path.
+        :raises ValueError:        If no origin is specified in the setting
+                                   nor the given origin parameter.
         """
         strrep = str(self).strip()
         if os.path.isabs(strrep):
@@ -137,7 +164,30 @@ class Setting(StringConverter):
         if origin is None:
             raise ValueError("Cannot determine path without origin.")
 
-        return os.path.abspath(os.path.join(os.path.dirname(origin), strrep))
+        # We need to get full path before escaping since the full path
+        # may introduce unintended glob characters
+        origin = os.path.abspath(os.path.dirname(origin))
+
+        if glob_escape_origin:
+            origin = glob_escape(origin)
+
+        return os.path.normpath(os.path.join(origin, strrep))
+
+    def __glob__(self, origin=None):
+        """
+        Determines the path of this setting with proper escaping of its
+        parent directories.
+
+        :param origin:      The origin file to take if no origin is specified
+                            for the given setting. If you want to provide a
+                            directory, make sure it ends with a directory
+                            separator.
+        :return:            An absolute path in which the parent directories
+                            are escaped.
+        :raises ValueError: If no origin is specified in the setting nor the
+                            given origin parameter.
+        """
+        return Setting.__path__(self, origin, glob_escape_origin=True)
 
     def __path_list__(self):
         """
@@ -147,6 +197,16 @@ class Setting(StringConverter):
         :return: A list of absolute paths.
         """
         return [Setting.__path__(elem, self.origin) for elem in self]
+
+    def __glob_list__(self):
+        """
+        Splits the value into a list and creates a path out of each item in
+        which the special glob characters in origin are escaped.
+
+        :return: A list of absolute paths in which the special characters in
+                 the parent directories of the setting are escaped.
+        """
+        return [Setting.__glob__(elem, self.origin) for elem in self]
 
     @property
     def key(self):
--- a/docs/Users/Glob_Patterns.rst
+++ b/docs/Users/Glob_Patterns.rst
@@ -18,30 +18,30 @@ Syntax
 
 The special characters used in shell-style wildcards are:
 
-+-----------------+-----------------------------------------------------------+
-| PATTERN         | MEANING                                                   |
-+=================+===========================================================+
-| '[seq]'         | Matches any character in seq. Cannot be empty. Any special|
-|                 | character looses its special meaning in a set.            |
-+-----------------+-----------------------------------------------------------+
-| '[!seq]'        | Matches any character not in seq. Cannot be empty. Any    |
-|                 | special character looses its special meaning in a set.    |
-+-----------------+-----------------------------------------------------------+
-| '(seq_a|seq_b)' | Matches either sequence_a or sequence_b as a whole. More  |
-|                 | than two or just one sequence can be given.               |
-+-----------------+-----------------------------------------------------------+
-| '?'             | Matches any single character.                             |
-+-----------------+-----------------------------------------------------------+
-| '*'             | Matches everything but the directory separator            |
-+-----------------+-----------------------------------------------------------+
-| '**'            | Matches everything.                                       |
-+-----------------+-----------------------------------------------------------+
++-------------------+---------------------------------------------------------+
+| PATTERN           | MEANING                                                 |
++===================+=========================================================+
+| ``[seq]``         | Matches any character in seq. Cannot be empty. Any      |
+|                   | special character looses its special meaning in a set.  |
++-------------------+---------------------------------------------------------+
+| ``[!seq]``        | Matches any character not in seq. Cannot be empty. Any  |
+|                   | special character looses its special meaning in a set.  |
++-------------------+---------------------------------------------------------+
+| ``(seq_a|seq_b)`` | Matches either sequence_a or sequence_b as a whole. More|
+|                   | than two or just one sequence can be given.             |
++-------------------+---------------------------------------------------------+
+| ``?``             | Matches any single character.                           |
++-------------------+---------------------------------------------------------+
+| ``*``             | Matches everything but the directory separator.         |
++-------------------+---------------------------------------------------------+
+| ``**``            | Matches everything.                                     |
++-------------------+---------------------------------------------------------+
 
 Examples
 --------
 
-'[seq]'
-~~~~~~~
+``[seq]``
+~~~~~~~~~
 
     Matches any character in seq. Cannot be empty. Any special character
     looses its special meaning in a set.
@@ -73,8 +73,8 @@ brackets have to be placed at the first position.
     >>> fnmatch("a[]a", "a[]a")
     True
 
-'[!seq]'
-~~~~~~~~
+``[!seq]``
+~~~~~~~~~~
 
     Matches any character not in seq. Cannot be empty. Any special
     character looses its special meaning in a set.
@@ -96,15 +96,15 @@ brackets have to be placed at the first position.
     >>> fnmatch("a[!]a", "a[!]a")
     True
 
-'(seq\_a\|seq\_b)'
-~~~~~~~~~~~~~~~~~~
+``(seq\_a\|seq\_b)``
+~~~~~~~~~~~~~~~~~~~~
 
     Matches either sequence\_a or sequence\_b as a whole. More than two
     or just one sequence can be given.
 
 Parentheses cannot be part of an alternative, unless they are escaped by
 brackets. Parentheses that have no match are ignored as well as
-'\|'-separators that are not inside matching parentheses
+``|``-separators that are not inside matching parentheses.
 
 ::
 
@@ -135,8 +135,8 @@ brackets. Parentheses that have no match are ignored as well as
     >>> fnmatch("", "(abc|)")
     True
 
-'?'
-~~~
+``?``
+~~~~~
 
     Matches any single character.
 
@@ -155,15 +155,15 @@ brackets. Parentheses that have no match are ignored as well as
     >>> fnmatch("ac", "a?c")
     False
 
-'\*'
-~~~~
+``\*``
+~~~~~~
 
-    Matches everything but the directory separator
+    Matches everything but the directory separator.
 
 .. note::
 
-    The directory separator is platform specific. '/' is never
-    matched by '\*'. '\\' is matched on Linux, but not on Windows.
+    The directory separator is platform specific. ``/`` is never
+    matched by ``\*``. ``\\`` is matched on Linux, but not on Windows.
 
 ::
 
@@ -178,8 +178,8 @@ brackets. Parentheses that have no match are ignored as well as
     >>> fnmatch("ac", "a*c")
     True
 
-'\*\*'
-~~~~~~
+``\*\*``
+~~~~~~~~
 
     Matches everything.
 
--- a/docs/Users/Glob_Patterns.rst
+++ b/docs/Users/Glob_Patterns.rst
@@ -6,6 +6,13 @@ input of multiple files without requiring a large number of filenames,
 coala supports a number of wildcards. These are based on the unix-style
 glob syntax and they are *not* the same as regular expressions.
 
+.. note::
+
+    Any glob that does not start with a ``/`` in Linux or a drive letter
+    ``X:`` in Windows will be interpreted as a relative path. Please use comma
+    separated values instead of absolute path globs that start with a
+    glob expression.
+
 Syntax
 ------
 
--- a/tests/parsing/GlobbingTest.py
+++ b/tests/parsing/GlobbingTest.py
@@ -18,7 +18,8 @@ import os
 import unittest
 
 from coalib.parsing.Globbing import (
-    _iter_alternatives, _iter_choices, _position_is_bracketed, fnmatch, glob)
+    _iter_alternatives, _iter_choices, _position_is_bracketed, fnmatch, glob,
+    glob_escape)
 
 
 class TestFiles:
@@ -98,6 +99,35 @@ class GlobbingHelperFunctionsTest(unittest.TestCase):
         for pattern, alternatives in pattern_alternatives_dict.items():
             self.assertEqual(sorted(list(_iter_alternatives(pattern))),
                              sorted(alternatives))
+
+
+class GlobEscapeTest(unittest.TestCase):
+
+    def test_glob_escape(self):
+        input_strings = [
+            "test",
+            "test[",
+            "test []",
+            "test [[]",
+            "test ]] str [",
+            "test[][]",
+            "test(",
+            "test)",
+            "test()",
+            "test (1)"]
+        output_strings = [
+            "test",
+            "test[[]",
+            "test [[][]]",
+            "test [[][[][]]",
+            "test []][]] str [[]",
+            "test[[][]][[][]]",
+            "test[(]",
+            "test[)]",
+            "test[(][)]",
+            "test [(]1[)]"]
+        for unescaped_str, escaped_str in zip(input_strings, output_strings):
+            self.assertEqual(glob_escape(unescaped_str), escaped_str)
 
 
 class FnmatchTest(unittest.TestCase):
--- a/tests/settings/SettingTest.py
+++ b/tests/settings/SettingTest.py
@@ -40,7 +40,7 @@ class SettingTest(unittest.TestCase):
         abspath = os.path.abspath(".")
         # Need to escape backslashes since we use list conversion
         self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
-                           origin="test" + os.path.sep + "somefile")
+                           origin=os.path.join("test", "somefile"))
         self.assertEqual(path_list(self.uut),
                          [os.path.abspath(os.path.join("test", ".")), abspath])
 
--- a/tests/settings/SettingTest.py
+++ b/tests/settings/SettingTest.py
@@ -4,7 +4,9 @@ import unittest
 from collections import OrderedDict
 
 from coalib.settings.Setting import (
-    Setting, path, path_list, url, typed_dict, typed_list, typed_ordered_dict)
+    Setting, path, path_list, url, typed_dict, typed_list, typed_ordered_dict,
+    glob, glob_list)
+from coalib.parsing.Globbing import glob_escape
 
 
 class SettingTest(unittest.TestCase):
@@ -28,6 +30,12 @@ class SettingTest(unittest.TestCase):
                               origin="test" + os.path.sep),
                          os.path.abspath(os.path.join("test", "22")))
 
+    def test_glob(self):
+        self.uut = Setting("key", ".",
+                           origin=os.path.join("test (1)", "somefile"))
+        self.assertEqual(glob(self.uut),
+                         glob_escape(os.path.abspath("test (1)")))
+
     def test_path_list(self):
         abspath = os.path.abspath(".")
         # Need to escape backslashes since we use list conversion
@@ -43,6 +51,16 @@ class SettingTest(unittest.TestCase):
         with self.assertRaises(ValueError):
             uut = Setting("key", "abc")
             url(uut)
+
+    def test_glob_list(self):
+        abspath = glob_escape(os.path.abspath("."))
+        # Need to escape backslashes since we use list conversion
+        self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
+                           origin=os.path.join("test (1)", "somefile"))
+        self.assertEqual(
+            glob_list(self.uut),
+            [glob_escape(os.path.abspath(os.path.join("test (1)", "."))),
+             abspath])
 
     def test_typed_list(self):
         self.uut = Setting("key", "1, 2, 3")
--- a/coalib/misc/Constants.py
+++ b/coalib/misc/Constants.py
@@ -71,3 +71,5 @@ with open(VERSION_FILE, 'r') as ver:
 BUS_NAME = "org.coala_analyzer.v1"
 
 TAGS_DIR = appdirs.user_data_dir('coala', version=VERSION)
+
+GLOBBING_SPECIAL_CHARS = "()[]|?*"


AbdealiJK:
Hey, the coverage has decreased. That means your tests do not handle that particular case.
(because of which when testing, that specific line never got executed - and codecov marks it as bad)

You need to add another test case which would execute that line to ensure that line works as expected :)

tushar-rishav:
@hypothesist Tests are failing. Did you try testing it locally before pushing? :)

hypothesist:
Yeah, I'm fixing it. Should be done in a minute. I'm a week behind HEAD. That's the problem.

hypothesist:
@fneu Hey Fabian, wanna take a look at this and see if the latest commit is alright? I've added tests too (any extra corner cases are welcome).

coalib/coala_delete_orig.py
```diff
@@ -17,7 +18,9 @@ def main(log_printer=None, section: Section=None):
         return 255
 
     orig_files = Globbing.glob(os.path.abspath(
```
sils1297: abspath could still introduce brackets

hypothesist: escaping around that should do it, right?

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,39 @@ def _end_of_set_index(string, start_index):
...
+    return tmp.replace("(", "[(]").replace(")", "[)]")
+
+
+def unescape_brackets_and_parentheses(string):
+    """
+    Unescapes the given string
+    Note: Only for strings escaped with escape_brackets_and_parentheses
+    """
+    string = string.replace("[(]", "(").replace("[)]", ")")
+    return string.replace("[[]", "[").replace("[]]", "]")
```
sils1297: what about `|`?

sils1297: @fneu should definitely have a look and the function should rather be named `glob_escape` or so because it escapes something so that when you put in a glob it'll match what you gave it in the first place

hypothesist: This escapes just the paths *before* globs. So there shouldn't be any `|` (unless of course, the user names a directory with `(...|...)`, which is super-obscure).

sils1297: still might happen, any character can be in a path actually (even a slash
that is not a directory seperator is somehow possible)

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-02-29 18:47 GMT+01:00 Adhityaa Chandrasekar <notifications@github.com>:

hypothesist: @sils1297 If it is named `glob_escape` it would lose its primary meaning - escaping brackets and parentheses. I understand that it uses globs to escape, but I'm not really sure if it should be named as `glob_escape`.

hypothesist: @sils1297 Say there is a path `/home/user/Downloads (user | root)/coala (1)/...`. Then `unescape_brackets_and_parentheses` would escape it as `/home/user/Downloads [(]user | root[)]/coala [(]1[)]/...` which is correct, since this would not be interpreted as alternatives (that part has been taken care of in `_position_is_bracketed` - this way only those `|` inside valid brackets are considered)

hypothesist:
> Looks like you implemented a feature that can be moved into a separate commit

@Makman2 not really since that was my previous escaping technique. With the current technique, you need to escape sub-directories with `[]`.

Makman2:
Hm looks still like this could go into a separate commit :)

coalib/parsing/Globbing.py
fneu: correct

Adrianzatreanu:
"This branch is out-of-date with the base branch
Merge the latest changes from master into this branch."
i think you need a rebase :)

Makman2:
So we don't loose our discussion: https://github.com/hypothesist/coala/commit/93fd8dfe55edc352a388c9219b00a88919f52839

hypothesist:
@Makman2 @fneu @sils1297 @AbdealiJK anything you want me to change before we can merge this? I think I have mostly addressed everything that came up: fixed the absurdly long function name, added `?` support (not adding `*` because it breaks a lot of stuff; plus who has `*` in their file name?), added tests.

sils1297:
oh this is pending review? Uhm a rebase is needed, other than that I don't know right now. I fear this PR because it's so delicate so I always hope somebody else reviews it properly ;)

hypothesist:
@sils1297 Oh yeah, it's been like this for days :D I'll rebase.

sils1297:
left some comments on https://github.com/hypothesist/coala/commit/f4ebf8f24ec81a6ab1f0c79d2a8ee3a1a17146d5

sils1297:
please address those

fneu:
@hypothesist why do things break when escaping `*`? In theory this shouldn't be any different from other characters, right?

hypothesist:
@fneu I really have no idea. And it's not just a few things: I think over 10-15 tests failed when I escaped `*`.

hypothesist:
@sils1297 @fneu review again? https://github.com/coala-analyzer/coala/commit/aba2842ba02f9d9482b6cfb460c6b1b8cb8c634a

fneu:
*looks* good, I want to have a look at the `*` escaping later today though.

I feel bad ignoring it because 
a) it actually is a bug in the usability
b) it should not break in theory
c) it might be a sign that you escape valid globs somewhere, I feel bad ignoring that.

sils1297:
alright, this needs a rebase and the variable needs to be renamed before this commit can happen cleanly IMO plus the documentation comment change.

sils1297:
I know this may be frustrating but this is a very sensible thing and we need to take big care of not breaking it. Especially since we have no testcases against what we're working with (would be good if you could add a few btw.)

hypothesist:
@sils1297 Fixed the line break, created an issue to change the docs and the variable name of `bear_dirs`. Anything else?

And I absolutely agree, this is a very sensitive/delicate commit that could break a lot of stuff if it goes wrong. I don't mind any number of comments if it will make this perfect.

On a side note: have you ever had such a big PR in terms of number of amends and comments? :D

sils1297:
Certainly we had. We have been working for months on some things.

Makman2:
@hypothesist if you don't achieve to escape `*` properly, open a bug for it after merge please :+1:

sils1297:
marking wip, would love to see the simplications as suggested

sils1297:
sim-pli-fi-ca-tions

hypothesist:
https://github.com/hypothesist/coala/commit/fca161e88d28665d03a769abf3c8bb2c54f4cfb5 for discussion history.

sils1297:
unack https://github.com/hypothesist/coala/commit/178f6281a0f5b03dcc8cd0f8f4ff6c52bbfb8441

sils1297:
unack https://github.com/hypothesist/coala/commit/10aecbe398253454fe8abe52fd80ed901d3f579c#diff-e690c48863208d16808485e103164cd5R113 , careful naming needed

sils1297:
please read and comment or address all my previous comments

sils1297:
unack https://github.com/hypothesist/coala/commit/d3c70d1ef4f1d08af8625b5fabb2145482e4f2ba
https://github.com/coala-analyzer/coala/commit/ee285f6a0400ddcd40c66424141f6a98a3424921

sils1297:
unack https://github.com/hypothesist/coala/commit/d77f2f724a44c958704db29e97aa1b31c32125b3

sils1297:
https://github.com/hypothesist/coala/commit/d52e23b6e9a63a3010d771f8be4df807706cf611 needs work

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
...
+            "test []][]] str [[]",
+            "test[[][]][[][]]",
+            "test[(]",
+            "test[)]",
+            "test[(][)]",
+            "test [(]1[)]"
+        ]
+        self._test_cases(input_strings, output_strings)
+
+
```
abhsag24: much simpler than the previous implementation :)

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
...
+    """
+    output = ""
+    special_chars = "()[]|?*"
+    for char in string:
+        if char in special_chars:
+            output += "[" + char + "]"
+        else:
+            output += char
+    return output
+
```
abhsag24: much simpler than the previous implementation :)

abhsag24: though i think someone who's been following the previous versions might want to have a look at this

Makman2: Wanna see how to make out of 7 lines a single one? :D
```python
return ["[" + char + "]" if char in special_chars else char for char in string]
```
:P

sils1297: could you not edit your comment? It's very confusing for people following via emails and makes the history go bogus

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
     return closing_index
 
 
+def glob_escape(string):
+    """
+    Escapes the given string with [c] pattern
```
Makman2: period at end missing.
Also I would not say "string" but "glob" (or "glob-string" if you want), since it operates on glob expressions, yeah "glob-expression" is even better :+1:

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
...
+
+    :param string: String that is to be escaped with [ ]
```
Makman2: Period at end missing

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
...
+    :return:       Escaped string in which all the special glob characters
+                   such as ()[]!? are escaped.
```
Makman2: we have rst, I would recommend enclosing ()[]!? inside
```
``()[]!?``
```

Makman2: and be precise, not "such as", it's this set of characters you escape :)

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
                              sorted(alternatives))
 
 
+class EscapingTest(unittest.TestCase):
```
Makman2: imo it's better if you call the test class `EscapeGlobTest` so the name resembles the function better^^

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
...
+
+    def _test_cases(self, in_arr, out_arr):
+        for case in list(zip(in_arr, out_arr)):
+            self.assertEqual(glob_escape(case[0]),
+                             case[1])
```
Makman2: fits into a single line

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
                              sorted(alternatives))
 
 
+class EscapingTest(unittest.TestCase):
+
+    def _test_cases(self, in_arr, out_arr):
```
Makman2: please call this `assert_cases` or similar, at least it should start with `assert` so one directly sees this is a test assertion^^

tests/parsing/GlobbingTest.py
Makman2: and please use unpacking:
```python
for input_case, output_case in zip(in_arr, out_arr):
    self.assertEqual(glob_escape(input_case), output_case)
```

tests/parsing/GlobbingTest.py
Makman2: Oh I see since you use this only once, maybe you can inline this, such statements are quite common in tests :) But that's you decision :+1:

coalib/parsing/Globbing.py
Makman2: you will get anyway further emails, because when I've made a mistake I would post another comment that corrects it, imo this is no gain for you^^

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
...
+            "test[",
+            "test []",
+            "test [[]",
+            "test ]] str [",
+            "test[][]",
+            "test(",
+            "test)",
+            "test()",
+            "test (1)"
+        ]
```
Makman2: --> `"test (1)"]` <-- place the closing-bracket directly behind the last item :)

tests/parsing/GlobbingTest.py
```diff
@@ -100,6 +101,41 @@ def test_alternatives(self):
...
+            "test[[]",
+            "test [[][]]",
+            "test [[][[][]]",
+            "test []][]] str [[]",
+            "test[[][]][[][]]",
+            "test[(]",
+            "test[)]",
+            "test[(][)]",
+            "test [(]1[)]"
+        ]
```
Makman2: same here

coalib/settings/Setting.py
```diff
@@ -13,6 +14,22 @@ def path_list(obj, *args, **kwargs):
...
+    """
+    Creates a path in which all special glob characters in all the
+    parent directories in the given setting are properly escaped.
+    """
+    return obj.__glob__(*args, **kwargs)
+
+
+def glob_list(obj, *args, **kwargs):
+    """
+    Creates a list of paths in which all special glob characters in all the
```
Makman2: Must it be a list or is a sequence also allowed? This makes heavy differences in the return type (e.g. with lists indexing does work, other sequences may not support it)

coalib/settings/Setting.py
```diff
@@ -13,6 +14,22 @@ def path_list(obj, *args, **kwargs):
     return obj.__path_list__(*args, **kwargs)
 
 
+def glob(obj, *args, **kwargs):
+    """
+    Creates a path in which all special glob characters in all the
+    parent directories in the given setting are properly escaped.
+    """
```
Makman2: Btw please add param documentation for `obj, *args, **kwargs`, there's the question what objects can be used with `glob()`, and what `*args` and `**kwargs` can be used for.

coalib/settings/Setting.py
```diff
@@ -108,7 +125,7 @@ def __init__(self,
         self.key = key
         self.origin = str(origin)
 
-    def __path__(self, origin=None):
+    def __path__(self, origin=None, globbed_origin=False):
```
Makman2: please add documentation for this parameter!

coalib/settings/Setting.py
```diff
@@ -13,6 +14,22 @@ def path_list(obj, *args, **kwargs):
...
+    parent directories in the given setting are properly escaped.
+    """
+    return obj.__glob__(*args, **kwargs)
+
+
+def glob_list(obj, *args, **kwargs):
+    """
+    Creates a list of paths in which all special glob characters in all the
+    parent directories of all paths in the given setting are properly escaped.
+    """
```
Makman2: Same here with param documentation and return type doc

coalib/settings/Setting.py
Makman2: And don't forget return type documentation :)

coalib/settings/Setting.py
Makman2: and if this param is for setting whether to interprete `origin` as glob or not, I would rename the parameter, this name is not so obvious. Maybe a plain `is_glob`?

coalib/settings/Setting.py
```diff
@@ -129,12 +146,30 @@ def __path__(self, origin=None):
...
+        :param origin:      The origin file to take if no origin is specified
+                            for the given setting. If you want to provide a
+                            directory, make sure it ends with a directory
+                            seperator.
+        :return:            An absolute path in which the parent directories
+                            are escaped.
+        :raises ValueError: If no origin is specified in the setting nor the
+                            given origin parameter.
+        """
+        return Setting.__path__(self, origin, True)
```
Makman2: Maybe it's a better choice to do `Setting.__path__(self, glob_escape(origin))`, then you are able to remove the if condition above?

coalib/settings/Setting.py
```diff
@@ -13,6 +14,22 @@ def path_list(obj, *args, **kwargs):
...
+
+
+def glob_list(obj, *args, **kwargs):
+    """
+    Creates a list of paths in which all special glob characters in all the
+    parent directories of all paths in the given setting are properly escaped.
+    """
+    return obj.__glob_list__(*args, **kwargs)
+
+
```
Makman2: Shouldn't the functions belong to the globbing module?

coalib/settings/Setting.py
```diff
@@ -129,12 +146,30 @@ def __path__(self, origin=None):
...
 
+    def __glob__(self, origin=None):
+        """
+        Determines the path of this setting with proper escaping of its
+        parent directories.
+
+        :param origin:      The origin file to take if no origin is specified
+                            for the given setting. If you want to provide a
+                            directory, make sure it ends with a directory
+                            seperator.
```
Makman2: it's `separator` :D

tests/settings/SettingTest.py
```diff
@@ -36,6 +42,15 @@ def test_path_list(self):
         self.assertEqual(path_list(self.uut),
                          [os.path.abspath(os.path.join("test", ".")), abspath])
 
+    def test_glob_list(self):
+        abspath = os.path.abspath(".")
+        # Need to escape backslashes since we use list conversion
+        self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
```
Makman2: you can use `StringProcessing.escape()` if you want :)

tests/settings/SettingTest.py
```diff
@@ -36,6 +42,15 @@ def test_path_list(self):
...
+                           origin="test (1)" + os.path.sep + "somefile")
+        self.assertEqual(glob_list(self.uut),
+                         [os.path.abspath(os.path.join("test [(]1[)]", ".")),
+                             abspath])
```
Makman2: isn't this overindented?
not
```python
[os.path.abspath(...),
 abspath]
```
?

coalib/coala_delete_orig.py
```diff
@@ -15,8 +16,9 @@ def main(log_printer=None, section: Section=None):
     if start_path is None:
         return 255
 
-    orig_files = Globbing.glob(os.path.abspath(
-        os.path.join(start_path, '**', '*.orig')))
+    # start_path may have unintended glob characters
+    orig_files = Globbing.glob(os.path.join(
+        glob_escape(start_path), '**', '*.orig'))
```
Makman2: what about `abspath`?

coalib/processes/Processing.py
```diff
@@ -335,10 +335,10 @@ def instantiate_processes(section,
                              the same for each object.
     """
     filename_list = collect_files(
-        path_list(section.get('files', "")),
+        glob_list(section.get('files', "")),
         log_printer,
-        ignored_file_paths=path_list(section.get('ignore', "")),
-        limit_file_paths=path_list(section.get('limit_files', "")))
+        ignored_file_paths=glob_list(section.get('ignore', "")),
+        limit_file_paths=glob_list(section.get('limit_files', "")))
```
Makman2: though it seems to work I ask anyway: doesn't this prevent globs at all inside a coafile?

coalib/settings/Setting.py
AbdealiJK: Create a newcomer issue for this (dont solve it if you dont need to)

coalib/settings/Setting.py
AbdealiJK: @Makman2 it makes sense to have it here though ... because we want bear users to use it in bears and it is a setting type.

coalib/parsing/Globbing.py
AbdealiJK: Actually we dont have a docstring guide yet ...
We should make one ...

coalib/parsing/Globbing.py
AbdealiJK: Please use code tags (double backticks) in rst for things like [c] as it makes it easier to read

coalib/parsing/Globbing.py
hypothesist: Isn't the current version simpler? I agree that your solution is shorter (and I'd personally prefer it on personal projects), but I feel the current version is more readable and less prone to errors due to carelessness in the future (and more explicit in my opinion). But if this is how it is done coala wide, I'm perfectly okay with changing it.

coalib/parsing/Globbing.py
hypothesist: I think `string` is not necessarily a glob expression, the output is. In fact it escapes `string` in such a way that it loses all manners in which it *could* be a glob.

tests/parsing/GlobbingTest.py
hypothesist: Call it `GlobEscapeTest` instead since the function is `glob_escape`?

coalib/settings/Setting.py
hypothesist: Honestly, I don't know where we can use `*args and **kwargs`. @sils1297 had added it in https://github.com/coala-analyzer/coala/commit/74551ef9f707acb74b0b3cab190b68a53639ee70#diff-2663e4c825081e483f6ed8d673b78545 so, I wanted to have the same parameters for `glob` and `glob_list` since they are very similar in function to `path` and `path_list`.

But I'll add the rest of the stuff right away.

coalib/settings/Setting.py
hypothesist: @Makman2 yeah, since these functions basically get the file list (escaped, unlike the real file list returned by `path_list`) for a _given setting_, I think these should be here.

coalib/parsing/Globbing.py
AbdealiJK: @hypothesist I think @Makman2 was joking.

coalib/settings/Setting.py
hypothesist: I agree the name should be changed, but since the boolean basically tells the function to whether or not escape `origin`, I think the name should be `glob_escape_origin` as it's more explicit (plus `is_glob` feels like an assertion test more than a boolean setting). What do you think? @Makman2

coalib/settings/Setting.py
hypothesist: Actually it's there in `__path__` too (I basically copy pasted the doc here :P ), so I'll make a newcomer issue to solve both.

coalib/settings/Setting.py
hypothesist: That would not work sometimes as you may see in `__path__`, this function may be overwritten by a if condition sometimes. That would mean that the overwritten value may not be escaped properly sometimes.

tests/settings/SettingTest.py
hypothesist: I just wrote something similar to what is in `test_path_list`. I'll make a newcomer issue solving both or do I make the change myself?

coalib/parsing/Globbing.py
hypothesist: @AbdealiJK @Makman2 Ahh, I see :)

coalib/coala_delete_orig.py
hypothesist: It's actually unnecessary since `get_config_directory` always returns a full path (since `os.path.getcwd()`, `str(section["config"])` are absolute paths which are the only return values)

coalib/processes/Processing.py
hypothesist: No, `section.get('files', "")` for example would return stuff such as `coalib/**/*.py` and so on. When this is passed to `glob_list`, it escapes the origin alone (so the path to the `coafile`) and then appends `coalib/**/*.py` without any change. Then it starts globbing and getting matches, which would work proper.

coalib/settings/Setting.py
hypothesist: It must be a list since we are returning a list by construction on line 190.

hypothesist: But I think we can introduce other types of sequences too, but I fail to see the purpose: this is meant to be a function that basically returns a list of full paths (glob escaped) for the given list of globs.

coalib/settings/Setting.py
sils1297: I hadd a pass through to the function underneath because it had some other parameter concerning the absoluteness of the path or something IIRC.

coalib/parsing/Globbing.py
Makman2: huh what in fact not :O
imo this one-liner is quite good to read :O

coalib/settings/Setting.py
Makman2: then okay, just wanted to ask :)

coalib/settings/Setting.py
Makman2: `path(Setting(...), origin="xyz", is_glob=True`
vs
`path(Setting(...), origin="xyz", glob_escape_origin=True`
I think both are good to read, since the path you get in 1. `is_` a `glob` :)
and in second you say it explicitly.
Yeah your choice :+1: :)

coalib/settings/Setting.py
Makman2: wait no, the stuff you write should be fixed by you. We don't introduce issues on purpose only to file them as newcomer-issues^^
For the old one in `__path__` you can file one, but please fix yours^^

tests/settings/SettingTest.py
Makman2: since this is more a modularizing change it's up to you :)

coalib/coala_delete_orig.py
Makman2: :+1:

coalib/processes/Processing.py
Makman2: ah alright, cool :D

coalib/settings/Setting.py
hypothesist: Ah alright :+1: :)

coalib/parsing/Globbing.py
AbdealiJK: fwiw I find the one liner complicated too

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
     return closing_index
 
 
+def glob_escape(string):
+    """
+    Escapes the given string with `[c]` pattern.
```
AbdealiJK: Double backtick ... we use RST :'( Why does nobody remember ?

coalib/settings/Setting.py
```diff
@@ -17,6 +18,29 @@ def url(obj, *args, **kwargs):
     return obj.__url__(*args, **kwargs)
 
 
+def glob(obj, *args, **kwargs):
+    """
+    Creates a path in which all special glob characters in all the
+    parent directories in the given setting are properly escaped.
+
+    :param obj: The `Setting` object from which the key is obtained.
```
AbdealiJK: double backtick

coalib/settings/Setting.py
```diff
@@ -17,6 +18,29 @@ def url(obj, *args, **kwargs):
...
+    """
+    return obj.__glob__(*args, **kwargs)
+
+
+def glob_list(obj, *args, **kwargs):
+    """
+    Creates a list of paths in which all special glob characters in all the
+    parent directories of all paths in the given setting are properly escaped.
+
+    :param obj: The `Setting` object from which the key is obtained.
```
AbdealiJK: double backtick

coalib/settings/Setting.py
```diff
@@ -112,20 +136,23 @@ def __init__(self,
...
-        :return:            An absolute path.
-        :raises ValueError: If no origin is specified in the setting nor the
-                            given origin parameter.
+        :param origin:             the origin file to take if no origin is
+                                   specified for the given setting. If you
+                                   want to provide a directory, make sure it
+                                   ends with a directory seperator.
+        :param glob_escape_origin: When this is set to true, the origin of
+                                   this setting will be escaped with
+                                   `glob_escape`.
```
AbdealiJK: double backtick (you get the drift)

tests/settings/SettingTest.py
```diff
@@ -44,6 +50,15 @@ def test_url(self):
             uut = Setting("key", "abc")
             url(uut)
 
+    def test_glob_list(self):
+        abspath = os.path.abspath(".")
+        # Need to escape backslashes since we use list conversion
+        self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
+                           origin="test (1)" + os.path.sep + "somefile")
+        self.assertEqual(glob_list(self.uut),
```
AbdealiJK: Something like:

```
self.assertEqual(
    glob_list(self.uut),
    [os.path.abspath(os.path.join("test [(]1[)]", ".")), abspath])
```
would probably be easier to read, but not important and not necessary. If youre modifying the `separator` issue, then do this

coalib/settings/Setting.py
```diff
@@ -112,20 +136,23 @@ def __init__(self,
...
         origin argument will be taken in every case.
 
-        :param origin:      the origin file to take if no origin is specified
-                            for the given setting. If you want to provide a
-                            directory, make sure it ends with a directory
-                            seperator.
-        :return:            An absolute path.
-        :raises ValueError: If no origin is specified in the setting nor the
-                            given origin parameter.
+        :param origin:             the origin file to take if no origin is
```
AbdealiJK: You start with small letter and no period here, but the next param has capital and period.

We dont have a style guide for this yet - but could you use the same guide in the same function for consistency ?

tests/settings/SettingTest.py
```diff
@@ -28,6 +29,11 @@ def test_path(self):
                               origin="test" + os.path.sep),
                          os.path.abspath(os.path.join("test", "22")))
 
+    def test_glob(self):
+        self.uut = Setting("key", ".",
+                           origin="test (1)" + os.path.sep + "somefile")
```
AbdealiJK: `os.path.join("test (1)", "somefile")` would probably be simpler ? (this is used somewhere in Line ~50 too)

coalib/settings/Section.py
```diff
@@ -65,11 +66,17 @@ def bear_dirs(self):
         bear_dirs = path_list(self.get("bear_dirs", ""))
```
AbdealiJK: Shouldnt we be using glob_list here ?
Same logic as glob_list for `section.get('files', "")` and so on above ?

AbdealiJK:
There are a few issues @Makman2 brought up earlier (minor ones like "separate" which he wants solved here, plus Ive reviewed it once ...

Marking as WIP

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
     return closing_index
 
 
+def glob_escape(string):
```
AbdealiJK: `string` is a frequently used inbuilt module - https://docs.python.org/3/library/string.html

Would it be possible to rename this to something else so that it doesnt get confusing if we ever import that ?

coalib/parsing/Globbing.py
```diff
@@ -29,6 +29,24 @@ def _end_of_set_index(string, start_index):
...
+    """
+    Escapes the given string with `[c]` pattern.
+
+    :param string: String that is to be escaped with `[ ]`.
+    :return:       Escaped string in which all the special glob characters
+                   `()[]|?*` are escaped.
+    """
+    output = ""
+    special_chars = "()[]|?*"
```
AbdealiJK: If these are all the special chars in Globbing that we handle, we should make it a variable in `Constants.py` or a constant variable in this module (so that other modules can use it too ?)

something like `Globbing.SPECIAL_CHARS` ?

coalib/parsing/Globbing.py
AbdealiJK: I was looking at @Makman2 's one liner again to see if we would really want it ... and I think if you want to make it a oneliner, regex is better. (Maybe even faster)
```
>>> re.sub("(?P<char>[" + re.escape(special_chars) + "])", "[\g<char>]", "a")
'a'
>>> re.sub("(?P<char>[" + re.escape(special_chars) + "])", "[\g<char>]", "a (1)")
'a [(]1[)]'
```

coalib/parsing/Globbing.py
Makman2: We should maybe just hardcode it below^^

coalib/parsing/Globbing.py
Makman2: using regex is even better :+1:

coalib/parsing/Globbing.py
Makman2: I use `string` also as a variable name. Not sure this is so important, we don't use `string`-module at all in coala :) (at least I haven't seen it)

Makman2: At least it's not `str`^^

coalib/settings/Section.py
hypothesist: No, the `bear_dirs` key is getting escaped here: https://github.com/coala-analyzer/coala/pull/1644/files#diff-41b4f7bdbd660f8c693c7d0304602d79R79

The reason we can't use `glob_list` is because only *certain* bear dirs (the default ones, not the user specified ones) must be escaped. So we can't run `glob_list` directly on the whole `bear_dirs` key.

coalib/settings/Setting.py
hypothesist: What period? It's there at the end.

AbdealiJK: Sorry, only the capital

tests/settings/SettingTest.py
hypothesist: Yeah, I agree (again, I copy-pasted the test with a slight change from `test_path` :P)

coalib/parsing/Globbing.py
hypothesist: Does `input_string` sound better? We can't use just `input` either...

AbdealiJK: Anything you like.
As Makman said, apparently he uses string everywhere. So thats ok too

tests/settings/SettingTest.py
hypothesist: Wait, I think there might be some problem with windows related to escaping (`C:\project\...` being interpreted as "C:project...` (without the backslash)). I experienced this in an Appveyor test once and it failed. I'm not sure though.

coalib/parsing/Globbing.py
hypothesist: Agreed, there may be a `unescape_glob` in the future which might want the sequence again and `Globbing.SPECIAL_CHARS` will come in handy.

AbdealiJK:
Damn, why is this not merged yet ?!?!
We have it in the release notes for 0.5.0 ...

coalib/settings/Setting.py
```diff
@@ -89,7 +113,7 @@ def __init__(self,
                                            last part will be stripped of. If
                                            you want to specify a directory as
                                            origin be sure to end it with a
-                                           directory seperator.
+                                           directory separator.
```
AbdealiJK: For future reference, this should be in an own commit.
Im letting it go because i just want this merged soon ... (In case you are redo-ing this, please dont fix this here)

AbdealiJK:
ok. This is 
ack 752f0cc 4566fdf 59890fd

from me.
Not merging it though and keeping in pending review as @sils1297 mentioned on gitter that he wanted to take a look ?

Please lets not care about minor details ... lets just make a new newcomer issue for those. This really should get merged for 0.5, it's been open for way too long.

hypothesist:
@AbdealiJK 125 comments :O

Anyway I think @Makman2 and @fneu should take a look once too as they too gave a lot of inputs.

AbdealiJK:
ack 752f0cc 4566fdf 59890fd

coalib/parsing/Globbing.py
```diff
@@ -29,6 +30,18 @@ def _end_of_set_index(string, start_index):
     return closing_index
 
 
+def glob_escape(input_string):
+    """
+    Escapes the given string with ``[c]`` pattern.
+
```
sils1297: can you put a few doctest examples here?

coalib/settings/Setting.py
```diff
@@ -133,12 +160,30 @@ def __path__(self, origin=None):
 
         if hasattr(self, "origin") and self.origin != "":
             origin = self.origin
+        if glob_escape_origin:
+            origin = glob_escape(origin)
 
         if origin is None:
             raise ValueError("Cannot determine path without origin.")
 
         return os.path.abspath(os.path.join(os.path.dirname(origin), strrep))
```
sils1297: doing an abspath here is wrong and may introduce non globbed path component

sils1297: rather use abspath in 162 and 164?

tests/settings/SettingTest.py
```diff
@@ -28,6 +29,11 @@ def test_path(self):
...
+        self.assertEqual(glob(self.uut), os.path.abspath("test [(]1[)]"))
```
sils1297: you need to make sure that everything done by the abspath function is escaped too!

tests/settings/SettingTest.py
```diff
@@ -44,6 +50,15 @@ def test_url(self):
             uut = Setting("key", "abc")
             url(uut)
 
+    def test_glob_list(self):
+        abspath = os.path.abspath(".")
+        # Need to escape backslashes since we use list conversion
+        self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
+                           origin="test (1)" + os.path.sep + "somefile")
```
sils1297: why not use os path join?

tests/settings/SettingTest.py
```diff
@@ -44,6 +50,15 @@ def test_url(self):
...
+        self.assertEqual(
+            glob_list(self.uut),
+            [os.path.abspath(os.path.join("test [(]1[)]", ".")), abspath])
```
sils1297: same here, if you use abspath it actually *means* there are non escaped glob components!

coalib/coala_delete_orig.py
```diff
@@ -6,6 +6,7 @@
 from coalib.parsing import Globbing
```
sils1297: commit message `To use `()[]|?` in project directories` `*` i missing

coalib/coala_delete_orig.py
sils1297: not unnecessary but even possibly harmful since abspath is not designed to handle glob escaped paths

sils1297:
752f0cc needs work

sils1297:
unack 59890fd , would love to see the doctests here

coalib/settings/Setting.py
hypothesist: True. Don't know how I missed it. (BTW I think an abspath before Line 163 on the right is enough.)

hypothesist: @sils1297 How about this?

```
        strrep = str(self).strip()
        if os.path.isabs(strrep):
            return strrep

        if hasattr(self, "origin") and self.origin != "":
            origin = self.origin

        if origin is None:
            raise ValueError("Cannot determine path without origin.")

        # We need to get full path before escaping since the full path
        # may introduce unintended glob characters
        origin = os.path.abspath(os.path.dirname(origin))

        if glob_escape_origin:
            origin = glob_escape(origin)

        return os.path.normpath(os.path.join(origin, strrep))
```

We need something at the end to convert `/tmp/./test/*.py` to `/tmp/test/*.py`. And `normpath` does this perfectly.

@AbdealiJK Also I just verified this:
```
73% /tmp ❯ ls -l A
total 0
lrwxrwxrwx. 1 adhityaa adhityaa 8 Mar 19 11:47 foo -> /tmp/X/B
73% /tmp ❯ ls X
B
73% /tmp ❯ ls X/B
73% /tmp ❯ touch /tmp/A/foo/file
72% /tmp ❯ ls X/B
file
72% /tmp ❯ python
Python 2.7.10 (default, Sep  8 2015, 17:20:17) 
[GCC 5.1.1 20150618 (Red Hat 5.1.1-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.normpath("/tmp/A/foo/../foo/./file")
'/tmp/A/foo/file'
```
So this basically goes to show that `normpath` does not introduce symlinks (which was the concern you raised yesterday as the symlinked paths might have dangerous characters); it only resolves `.` and `..`

But I agree, `[/]tmp/*.py` will lead to `/[/]tmp/*.py` but that is so weird to use - we could always enforce the condition that absolute paths must start with `/` (but this will rule out stuff like `(/|~/)coala/*.py`; open to any suggestions).

coalib/settings/Setting.py
```diff
@@ -3,6 +3,7 @@
```
sils1297: `Setting: Escape glob characters in parent paths` nonono, you add the glob conversion routines, right? Say that in the commit message :)

coalib/collecting/Collectors.py
```diff
@@ -110,6 +110,9 @@ def icollect_bears(bear_dirs, bear_globs, kinds, log_printer):
     """
     for bear_dir, dir_glob in filter(lambda x: os.path.isdir(x[0]),
                                      icollect(bear_dirs)):
+        # Since we get a real directory here and since we
+        # pass this later to iglob, we need to escape this.
+        bear_dir = glob_escape(bear_dir)
```
sils1297: I think this is a bugfix and should go in an own commit, only the change to collectors.py here and the test for it

coalib/output/dbus/DbusDocument.py
```diff
@@ -124,7 +124,7 @@ def Analyze(self):
                     continue
 
                 if any([fnmatch(self.path, file_pattern)
-                        for file_pattern in path_list(section["files"])]):
+                        for file_pattern in glob_list(section["files"])]):
```
sils1297: same, own commit

coalib/processes/Processing.py
sils1297: this goes in an own commit

coalib/settings/Section.py
```diff
@@ -62,7 +62,7 @@ def __init__(self,
         self.contents = OrderedDict()
 
     def bear_dirs(self):
-        bear_dirs = path_list(self.get("bear_dirs", ""))
+        bear_dirs = glob_list(self.get("bear_dirs", ""))
```
sils1297: this goes in the same commit as the changes in processing.py

sils1297: no tests at all in this commit btw.?

docs/Users/Glob_Patterns.rst
```diff
@@ -6,6 +6,13 @@ input of multiple files without requiring a large number of filenames,
 coala supports a number of wildcards. These are based on the unix-style
 glob syntax and they are *not* the same as regular expressions.
 
+.. note::
+
+    Any glob that does not start with a '/' in Linux or a drive letter (X:) in
```
sils1297: rather use

    ``/``

to show paths as code, same for the drive letter

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1644/commits/dd7053f02dc05954bce25a96ad74b4cdd11be455 https://github.com/coala-analyzer/coala/pull/1644/commits/794a2de34183598a96e94acb451d06bc765b2ec9 https://github.com/coala-analyzer/coala/pull/1644/commits/859239e36b8d3a63e850fa35ef2ad99e02c6943e

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1644/commits/56293f729f1180450212e406294347643a0af802 https://github.com/coala-analyzer/coala/pull/1644/commits/a8a89993f5b5ce20d6e46cc9f511555620eee61b https://github.com/coala-analyzer/coala/pull/1644/commits/7384083802221e18e56c99e6d3eedca48407a9c9

coalib/output/dbus/DbusDocument.py
hypothesist: I think this should also go with the processing commit since it's almost the same (file collection, but instead of `path_list`, we use `glob_list`).

sils1297: hm possibly. yes

coalib/collecting/Collectors.py
hypothesist: So basically break this commit into three, right?

- First with this bug fix
- The `coala_delete_orig.py` one
- Then the others (as they are all file collection using `glob_list` instead of `path_list`)

sils1297: yes

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-20 14:27 GMT+01:00 Adhityaa Chandrasekar <notifications@github.com>:

coalib/settings/Section.py
hypothesist: @sils1297 How do we do tests here? This basically proofs against glob characters interfering, but none of the CI (Appveyor, Travis, Circle) have these characters such that we can test if they work properly.

hypothesist: Moreover, the only change is the use of `glob_path` instead, which has tests of its own.

tests/settings/SettingTest.py
```diff
@@ -44,6 +52,16 @@ def test_url(self):
...
 
+    def test_glob_list(self):
+        abspath = glob_escape(os.path.abspath("."))
+        # Need to escape backslashes since we use list conversion
+        self.uut = Setting("key", "., " + abspath.replace("\\", "\\\\"),
+                           origin=os.path.join("test (1)", "somefile"))
+        self.assertEqual(
+            glob_list(self.uut),
+            [glob_escape(os.path.abspath(os.path.join("test (1)", "."))),
+                abspath])
```
sils1297: four spaces indent too much right?

coalib/collecting/Collectors.py
sils1297: I think for this one I'd see a test though

sils1297: ack if it's very hard to test this :/

sils1297:
reack / ack https://github.com/coala-analyzer/coala/pull/1644/commits/dd7053f02dc05954bce25a96ad74b4cdd11be455 https://github.com/coala-analyzer/coala/pull/1644/commits/6ae984e5b5086a29a185b433e10062b524c81965 https://github.com/coala-analyzer/coala/pull/1644/commits/dc25c3c0554d97a39b6c968ee5fccacc8368407e https://github.com/coala-analyzer/coala/pull/1644/commits/ed054691eae64ea1669abca1918048647b507a64 https://github.com/coala-analyzer/coala/pull/1644/commits/efbeb9ffec86f6e68c37c319c0a980468a93983d https://github.com/coala-analyzer/coala/pull/1644/commits/a9df4a42575745f7402e3a27a819edf6660ef286 https://github.com/coala-analyzer/coala/pull/1644/commits/b6b5f8b34a17369afb667c55479fe73fb840f3c9 https://github.com/coala-analyzer/coala/pull/1644/commits/86909d74ca6e843e35f0560a11f8895852883641

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1644/commits/51d54d90a18b9bf5a74bd222ddcbab390b574781

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1644/commits/51d54d90a18b9bf5a74bd222ddcbab390b574781

docs/Users/Glob_Patterns.rst
```diff
@@ -32,7 +32,7 @@ The special characters used in shell-style wildcards are:
 +-----------------+-----------------------------------------------------------+
 | '?'             | Matches any single character.                             |
 +-----------------+-----------------------------------------------------------+
-| '*'             | Matches everything but the directory separator            |
+| '*'             | Matches everything but the directory separator.           |
 +-----------------+-----------------------------------------------------------+
 | '**'            | Matches everything.                                       |
```
sils1297: you could change this to use double backticks too fwiw

hypothesist:
reack https://github.com/coala-analyzer/coala/pull/1644/commits/dd7053f02dc05954bce25a96ad74b4cdd11be455 https://github.com/coala-analyzer/coala/pull/1644/commits/d693ee32fb5e2e2c7e021ad5bbff88a1313cff17 https://github.com/coala-analyzer/coala/pull/1644/commits/0b45cf98bb8ab9bf14af1d99be585c6399b3a987 https://github.com/coala-analyzer/coala/pull/1644/commits/fcd3b3be3116caa620b28c7b16b6add7f394a480 https://github.com/coala-analyzer/coala/pull/1644/commits/e2518d15236d0a85c79b19a0ceed48e76f9f066e https://github.com/coala-analyzer/coala/pull/1644/commits/6e92b214ac37b0cd7781c0eaf7fb68eaaa00eafe https://github.com/coala-analyzer/coala/pull/1644/commits/a52dc97d3920117c0e78b306534113fe1c8eaaae

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1644/commits/f94bca0cce74d44455736cc44d804eba719cc36d

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1644/commits/1ebb260bf3c3a59dfa0d8f22079fe579b19d9537

sils1297:
@rultor merge

PR #1682, opened 2016-02-23
Makman2/anti bash injection2

Makman2:
Please don't do something here until I've cross-tested this here and on coala-bears^^


sils1297:
@Makman2 please don't revert the revert, just cherry-pick the original commit!

Makman2:
@sils1297 this is just for testing, I'll change the messages again according to my commits, or yeah just cherry-pick them. This was faster :)

PR #1699, opened 2016-02-24
Processing: Provide line & column in ignore ranges

WJCode:
Fixes https://github.com/coala-analyzer/coala/issues/1232

### DIFF
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -408,12 +408,16 @@ def yield_ignore_ranges(file_dict):
                     yield (bears,
                            SourceRange.from_values(filename,
                                                    start,
-                                                   end_line=line_number))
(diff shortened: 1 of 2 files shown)


WJCode:
I had assumed that the line containing "stop ignoring" need not be searched.
Ah ok. There could be code right before the comment character? I'll fix that

sils1297:
yeah maybe the analytics even affects the comment itself, the whole line should be ignored IMO

WJCode:
Indeed it does. Including the entire "stop ignoring" line.

I'll also try to come up with a suitable test case for the previous behaviour.

sils1297:
ack ab87a3d3b7d1

sils1297:
@rultor merge

PR #1720, opened 2016-02-26
docs/install : do not install under proxy

srisankethu:
Fixes https://github.com/coala-analyzer/coala/issues/1715


sils1297:
didn't we not want to d this?

Makman2:
I think also a proxy configuration is something up to the user^^ And disabling the proxy is often not a choice^^

sils1297:
having a note on the env vars would be more useful

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-02-26 20:06 GMT+01:00 Mischa Krüger <notifications@github.com>:

PR #1721, opened 2016-02-27
Constants.py: Added newlines to crash message

analyticalmonk:
Constants.py: Added newlines to crash message

As of now the crash message is just one single line without any newlines, this commit adds newlines to this message, the file can be found at coala/coalib/misc/Constants.py

Fixes https://github.com/coala-analyzer/coala/issues/1708

### DIFF
--- a/coalib/misc/Constants.py
+++ b/coalib/misc/Constants.py
@@ -4,12 +4,12 @@ import platform
 THIS_IS_A_BUG = ("This is a bug. We are sorry for the inconvenience. "
                  "Please contact the developers for assistance.")
 
-CRASH_MESSAGE = ("An unknown error occurred. This is a bug. We are "
-                 "sorry for the inconvenience. Please contact the "
-                 "developers for assistance. During execution of "
-                 "coala an exception was raised. This should never "
-                 "happen. When asked for, the following information "
-                 "may help investigating:")
+CRASH_MESSAGE = ("An unknown error occurred. This is a bug. We are \n"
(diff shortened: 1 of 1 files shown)


Adrianzatreanu:
From what i know, the format of the title of the commit message should be: <fileyouarechanging>: <the change>
and also no space before the ':'

analyticalmonk:
@sils1297 hey there, I have been trying to resolve the error due to which the CircleCI test failed.
The code responsible are respectively at 
https://github.com/coala-analyzer/coala/blob/master/coalib/tests/output/dbus/DbusDocumentTest.py#L83
and 
https://github.com/coala-analyzer/coala/blob/master/coalib/tests/output/dbus/DbusTest.py#L127

They seem to be comparing the crash_message with the values stored in output and analysis variables respectively. But I am unable to properly understand where the values in those variables come from and am stuck.
Any pointers?

sils1297:
Hm I have no clue why the tests fail, sorry didn't look closely before. Actually I thought this was a totally other issue. @abhsag24 why did you want this? Why can't the terminal just break it on its own? Didn't we want to have proper newlines in the message in line 23 INteractions.py?

sils1297:
@analyticalmonk sorry for that mess, you can't avoid that sometimes...

analyticalmonk:
@sils1297 Oh okay. I will pick something else up then.

AbdealiJK:
Closing this as this is definitely not going to happen in coala.
May happen in pyprint ... which is in another repo

PR #1729, opened 2016-02-27
docs/install : do not install under proxy

srisankethu:
Fixes https://github.com/coala-analyzer/coala/issues/1715

### DIFF
--- a/docs/Users/Tutorials/Self_Correcting_Bears.rst
+++ b/docs/Users/Tutorials/Self_Correcting_Bears.rst
@@ -67,7 +67,7 @@ is used to get the final output (which may be suggestions or
 auto-corrections).
 
 Let's try this bear out. Create a new file called ``sample.cpp``. Contents of
-``sameple.cpp`` are:
+``sample.cpp`` are:
 
 ::
 


docs/Users/Install.rst
```diff
@@ -153,6 +153,18 @@ If you want to install coala to an alternate location you can e.g. call
 options are documented on
 https://docs.python.org/3.3/install/#alternate-installation.
 
+.. note::
+
+    If you are using a proxy, follow these steps:
+
+    -  Set up your proxy.
+    -  Use ``sudo -E pip install coala``(-E flag takes in the existing
```
SanketDG: When you say takes in, the first thing I think of is `-E <some_argument>`, could you say something else, like `preserves the existing environment variables`

SanketDG:
Also a little extra git knowledge can help you, instead of creating a new PR every time.

docs/Users/Install.rst
srisankethu: @SanketDG It's just the same thing

sils1297: I actually like the original more because preserving would rather mean for the invoking shell IMO. How about

    (the ``-E`` flag takes the existing environment variables into the ``sudo`` environment)

?

PR #1735, opened 2016-02-28
Newcomers

Adrianzatreanu:
This is only a prototype, and i need your review guys to make it perfect.
I repeat, its only a prototype, its not great yet, i just want to hear from you
what you think will make it better, and what things i need to change/add/remove.

Thanks !


Adrianzatreanu:
this is weird, i had 2 unwanted commits here. ill open another PR on another branch in a few mins.

PR #1740, opened 2016-02-28
Pip version control for installation

saurabhiiit:
pip version from 1.5 and above is not letting to install coala.

### DIFF
--- a/docs/Users/Install.rst
+++ b/docs/Users/Install.rst
@@ -145,6 +145,14 @@ changes easily. To do this run:
 This essentially lets you install coala in a way that allows you to make
 changes to the code and see the changes take effect immediately.
 
+.. note::
+
+pip version from 1.5 and above creates problem in installing.
+One can go through the link for installation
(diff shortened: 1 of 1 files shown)


sils1297:
that sounds pretty old - why anyway?

saurabhiiit:
While running the command pip3 install coala-bears gives the error and not letting to install 
screenshot from 2016-02-29 00 49 43

sils1297:
Ah I think though that this'll be solved when we do a regular release (will
happen this week) since it just doesn't like prereleases I guess?

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-02-28 20:17 GMT+01:00 saurabhiiit <notifications@github.com>:

AbdealiJK:
This doesn't seem to be a pip issue. You need to do a `pip install coala-bears --pre` for now as we do not have stable releases yet.
This will be fixed when we do a release withing the next week, so closing it.

Reopen if you think it is still needed

PR #1762, opened 2016-03-01
Decorators: Improves documentation

laxmena:
Documentation of get_public_members(obj) incorrectly mentions that
the method returns a list, whereas actually a dictionary is returned.
The Documentation is corrected and the key-value pair is mentioned.

Fixes coala-analyzer#1739


abhsag24:
ack fbfea3b

sils1297:
unack fbfea3b as per gitter

PR #1784, opened 2016-03-02
bear ignore: Add inline glob for bear ignore

hypothesist:
Previously a comment with `ignore` followed by
bear names in it will result in the following line
not being impacted by the bears. This commit
introduces globs for the same so that multiple matching
bears can be ignored easily.

Fixes https://github.com/coala-analyzer/coala/issues/1781


sils1297:
@rultor merge

PR #1808, opened 2016-03-05
docs: Mention that bear_dirs is a glob

hypothesist:
In `icollect_bears` and `collect_bears` the parameter
`bear_dirs` is actually a glob.

Fixes https://github.com/coala-analyzer/coala/issues/1807


AbdealiJK:
:( Can you leave this as it's a newcomer issue ?

hypothesist:
Actually @sils1297 wanted me to make this commit because it was holding up https://github.com/coala-analyzer/coala/pull/1644

But I'm okay with it (we could always make this commit after the PR).

PR #1809, opened 2016-03-05
setup.py: Parse requirements.txt using pip

tushar-rishav:
This commit uses `parse_requirements` from `pip.req` module to parse the
`requirements.txt` file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,8 @@ import locale
 import sys
 from os import getenv
 from subprocess import call
+from pip.req import parse_requirements
+from pip.download import PipSession
 
 # Start ignoring PyImportSortBear as imports below may yield syntax errors
 from coalib import assert_supported_version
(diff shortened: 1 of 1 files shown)


tushar-rishav:
@sils1297  @AbdealiJK  suggested to use pip's parser to process package dependencies from a `requirements.txt` in https://github.com/coala-analyzer/coala/issues/1806. Unfortunately, after creating PR, I realised that it's not a good idea to use pip parser because it might break for higher versions of pip. :( Moreover, I found out that pip developers, officially do not provide a programmatic api to pip.  More info here https://github.com/pypa/pip/issues/2286 
I shall cancel this PR if it's fine with @AbdealiJK?

AbdealiJK:
@tushar-rishav Please close it. It does seem to add more complications that it reduces.

PR #1814, opened 2016-03-05
Makman2/linterv2

Makman2:
There it is, the next generation `Linter`.

***Do not merge until coala-bears contains coala version limits!!! This will break coala-bears!!!***


sils1297:
@Makman2 can you make a PR for fixing the version in coala-bears?

sils1297:
It'd be cool if you could update the tutorial to the new linter

Makman2:
@sils1297 yeah I'll update the tutorials :D
and yes I try to make a PR already inside coala-bears to adapt the new changes, but tomorrow because of an exam :+1:

AbdealiJK:
@Makman2 Im just thinking aloud here ...
Do you think metaclass would be more suitable than a decorator ?
Metaclasses have more functionality than decorators. Decorators are intended to be simple ... and this decorators seems to be pretty complicated.

The reason Im bringing this up is because SublimeLinter uses metaclasses
Ive never used Metaclasses though - so Im not sure which would be better.

Makman2:
@AbdealiJK That's actually an idea... Maybe this is simpler, inside the decorator/callables it gets a bit complicated^^ I'll consider that during my fixes :+1: :)

Makman2:
yeah it looks like metaclasses are the better approach, then the linter is not decorator bound any more, so that gets more flexible (as far as I understand :D)

AbdealiJK:
@Makman2 I notice that you wrote that this fixes https://github.com/coala-analyzer/coala/issues/1703 - but that's not quite true.
All the Lint functions are still not accessible and so that issue shold remain open

AbdealiJK:
@Makman2 I notice that you wrote that this fixes https://github.com/coala-analyzer/coala/issues/1703 - but that's not quite true.
All the Lint functions are still not accessible and so that issue shold remain open

Makman2:
Good point, will change the issue id to the one you created (#1928).

sils1297:
@Makman2 this can be closed due to https://github.com/coala-analyzer/coala/pull/1990 right?

PR #1837, opened 2016-03-07
ConsoleInteraction: Catch bullet unicode error

Makman2:
When printing the bullet-character that represents a whitespace, a
`UnicodeEncodeError` can be thrown and is thrown on Windows systems
since Windows uses by default cp1252 or a similar codec that can't
handle this character.

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -128,7 +128,10 @@ def print_spaces_tabs_in_unicode(console_printer, line, tab_dict,
     """
     for char in line:
         if char == " ":
-            console_printer.print("•", color='cyan', end='')
+            try:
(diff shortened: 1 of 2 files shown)


sils1297:
dcb718b needs work

Makman2:
@sils1297 btw the index stuff doesn't work at all. This bullet-printing-function needs to be replaced completely (already working on that), sorry but I'm a bit surprised that this has gotten actually merged.
Pushing now just a hotfix so we get the underlying bug solved^^

sils1297:
really? The tests looked sane and I never had problems with it.

sils1297:
cherry picking this on my PR assuming that somebody acks it soon

sils1297:
done and merged

PR #1845, opened 2016-03-08
Update action everytime

srisankethu:
Fixes https://github.com/coala-analyzer/coala/issues/1825


sils1297:
@srisankethu as per discussion on gitter this solution isn't really applicable

srisankethu:
@sils1297 So can i added a success parameter to action and check if succeeded then don't change parameter...something like that ?

sils1297:
The success information is already there somewhere, at least somewhere we print a success message and otherwise the exception info

srisankethu:
@sils1297 Ok ! I will try to use that !

PR #1847, opened 2016-03-08
coala_html.py: Add coala-html binary

tushar-rishav:
~~Install coala-html angular app via bower. Also install the bower dependencies.~~

- [x]  run coala and dump the results in json format to a particular file if it doesn't exists else update it. This file shall be used by Angular app.

- [ ] create `file.json` if it doesn't exists else update it.

- [x] Finally, create a server and open the file. This can have some flexibility. User can decide whether to launch the webpage immediately after running coala-html or maybe just update/create the files.

Fixes https://github.com/coala-analyzer/coala-html/issues/31

### DIFF
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import sys
 
 from coalib.misc import Constants
@@ -15,7 +16,7 @@ def default_arg_parser(formatter_class=None):
 
     entry_point = sys.argv[0]
     for entry in ['coala-ci', 'coala-dbus', 'coala-format', 'coala-json',
-                  'coala-delete-orig']:
+                  'coala-html', 'coala-delete-orig']:
         if entry_point.endswith(entry):
             parser_type = entry
             break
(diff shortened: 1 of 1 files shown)


sils1297:
Wait so this essentially makes the coala HTML repository completely
superfluous?

sils1297:
OK now I looked at the code, why do we need this? It doesn't do anything
else than coala JSON .

tushar-rishav:
@sils1297  So running `coala-html` binary  would do these things:

- [ ] Install coala-html angular app via bower. Also install the bower dependencies.

- [x]  run coala and dump the results in json format to a particular file if it doesn't exists else update it. This file shall be used by Angular app.

- [ ] create `file.json` if it doesn't exists else update it.

- [ ] Finally, create a server and open the file. This can have some flexibility. User can decide whether to launch the webpage immediately after running coala-html or maybe just update/create the files.

AbdealiJK:
@sils1297 should check out https://github.com/coala-analyzer/coala-html/issues/31

AbdealiJK:
Comments at https://github.com/tushar-rishav/coala/commit/28556f55b0b8d097faba6aaa16825b7a1b68007d

sils1297:
unack     

	@tushar-rishav 	coala_html.py: Add coala-html binary … 	
28556f5

sils1297:
neat, gitmate works with newlines now :)

sils1297:
@tushar-rishav we probably want to discuss this, the user shouldn't even notice that coala-html generates a json file, the user just wants to give it a directory and expects his website to be there then.

tushar-rishav:
@sils1297  Yes. I'll remove the output option from the user's end. I noticed that late. :D

coalib/parsing/DefaultArgParser.py
```diff
@@ -119,6 +120,26 @@ def default_arg_parser(formatter_class=None):
                                 metavar='BOOL',
                                 help='Write the logs as json to a file '
                                 'where filename is specified as argument.')
+    if parser_type == 'coala-html':
```
AbdealiJK: Decreases coverage, needs a test

AbdealiJK: Now this is confusing. If the coala-html is a python package of its own ... why is the argparser for the coala-html package here ?
This wouldnt even be called right ? We probably want to add these in coala-html ?

tushar-rishav: Hm no. That's not the case. you might have seen that coala-html has coala
as one of the dependencies. So we can reuse stuffs from coala. IMO that's
neat. Hm tried running it locally?

AbdealiJK: I have not tried running it locally (sorry, Im on windows and I cant code here)

And even if its a dependency ... you shouldnt have bits of the code here that are specifically for coala-html.

Why not import DefaultArgParser in coala-html and add a few args there ? Seems to make sense to me

sils1297: @tushar-rishav he's right, if we want to use the arg parser cleanly we have to move coala-html into coala or add the arguments outside...

tushar-rishav: Hm sure. That'd be better solution.  I'll try implementing it later this
evening.

PR #1853, opened 2016-03-08
CI: Don't silence commandos

Makman2:
Silencing just hinders debugging.

### DIFF
--- a/.misc/deps.osx.sh
+++ b/.misc/deps.osx.sh
@@ -3,7 +3,7 @@ set -x
 
 # Install packages with brew
 brew update >/dev/null
-brew outdated pyenv || brew upgrade --quiet pyenv
+brew outdated pyenv || brew upgrade pyenv
 brew install espeak
 brew install libffi && brew link libffi --force
 brew install cairo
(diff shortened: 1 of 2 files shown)


AbdealiJK:
Um, I has commented 5 mins back but the commit seems to have disappeared. (Sorry about the double comment if it shows up later)

I am against the `==` as @Adrianzatreanu commented : https://github.com/coala-analyzer/coala-bears/commit/0481df2a0e06d323e8a7fb3ce473b8fefa2984f7#commitcomment-16575385
I think that commit needs to be reverted too.

If pytest uses a different version of some library or any other package uses another version of the library - both those packages will be **unusable**.

I remember you saying radon had a `==` somewhere. THat would make this a bug in radon ... and we should file an issue there to fix it

sils1297:
filed https://github.com/rubik/radon/issues/82

sils1297:
https://github.com/coala-analyzer/coala/commit/646561d169504ae4092a74f2bee4dbec30e6c7cb is ready

sils1297:
@rultor merge

PR #1863, opened 2016-03-11
coafile: Add option to ignore coafile

hypothesist:
When the user adds the flag `--no-config` to his coala command
the program will now default to the most basic coafile while
ignoring the coafile that may be present in the user diretory.

Fixes https://github.com/coala-analyzer/coala/issues/1838

### DIFF
--- a/coalib/parsing/CliParsing.py
+++ b/coalib/parsing/CliParsing.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from argparse import ArgumentParser
 from collections import OrderedDict
 
 from coalib.parsing.DefaultArgParser import default_arg_parser
@@ -87,3 +88,25 @@ def parse_custom_settings(sections,
                                origin=origin,
                                section_name=key_touple[0],
                                from_cli=True)
+
+
+def check_conflicts(sections):
+    '''
+    Checks if there are any conflicting aruments passed
+
+    :return:            True if no conflicts
+    :raises SystemExit: If there are conflicting arguments (exit code: 2)
+    '''
+    conflicts = {'no_config': {'save', 'find_config'}}
+    conflicting_keys = conflicts.keys()
+
+    for section in sections:
+        keys = set(sections[section])
+        possible_conflicts = keys & conflicting_keys
+        for key in possible_conflicts:
+            intersection = keys & conflicts[key]
+            if len(intersection) > 0:
+                ArgumentParser().exit(2,
+                                      key + " cannot be given at the same "
+                                      "time with " + (', '.join(intersection)))
+    return True
--- a/tests/parsing/CliParsingTest.py
+++ b/tests/parsing/CliParsingTest.py
@@ -1,7 +1,7 @@
 import argparse
 import unittest
 
-from coalib.parsing.CliParsing import parse_cli
+from coalib.parsing.CliParsing import parse_cli, check_conflicts
 
 
 class CliParserTest(unittest.TestCase):
@@ -52,3 +52,12 @@ class CliParserTest(unittest.TestCase):
         self.assertEqual(parsed_sections["default"].name, "Default")
         self.assertEqual(self.dict_from_sections(parsed_sections),
                          expected_dict)
+
+    def test_check_conflicts(self):
+        sections = parse_cli(arg_list=["--save", "--no-config"])
+        with self.assertRaises(SystemExit) as cm:
+            check_conflicts(sections)
+            self.assertEqual(cm.exception.code, 2)
+
+        sections = parse_cli(arg_list=["--no-config", "-S", "val=42"])
+        self.assertTrue(check_conflicts(sections))
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -50,6 +50,12 @@ def default_arg_parser(formatter_class=None):
                             const=True,
                             metavar='BOOL',
                             help=FIND_CONFIG_HELP)
+    arg_parser.add_argument('-I',
+                            '--no-config',
+                            nargs='?',
+                            const=True,
+                            metavar='BOOL',
+                            help="Run without using any config file")
     arg_parser.add_argument('-f',
                             '--files',
                             nargs='+',
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -5,7 +5,7 @@ import sys
 from coalib.misc import Constants
 from coalib.output.ConfWriter import ConfWriter
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
-from coalib.parsing.CliParsing import parse_cli
+from coalib.parsing.CliParsing import parse_cli, check_conflicts
 from coalib.parsing.ConfParser import ConfParser
 from coalib.settings.Section import Section
 from coalib.settings.SectionFilling import fill_settings
@@ -114,6 +114,7 @@ def load_configuration(arg_list, log_printer):
                         indicated after colon.)
     """
     cli_sections = parse_cli(arg_list=arg_list)
+    check_conflicts(cli_sections)
 
     if (
             bool(cli_sections["default"].get("find_config", "False")) and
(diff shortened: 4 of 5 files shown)


AbdealiJK:
unack 0050f34

hypothesist:
https://github.com/hypothesist/coala/commit/0050f341ef4e883e3b8fe700a86dc317f4925845 for discussion history.

sils1297:
https://github.com/hypothesist/coala/commit/4d3d3f059ebfa83b1a8846fee3d8065b02b500e8 needs work

sils1297:
see comments on https://github.com/coala-analyzer/coala/commit/f14ea254eac80ffec8cce1ba8585791c46f8093b , needs work

sils1297:
see comments on f14ea25

sils1297:
935b418 needs work

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +87,31 @@ def parse_custom_settings(sections,
                                origin=origin,
                                section_name=key_touple[0],
                                from_cli=True)
+
+
+def check_conflicts(cli_sections, log_printer):
```
AbdealiJK: Can we make this a little generic to check for conflicts in any section ?
And possible pass `conflicts` into it ?

We should have a fucntion in DefaultArgParser like `cli_conflicts()` which gives `conflicts`, so that the logic for arg parsing is in the same place ?

AbdealiJK: Also - isn't there a better way to do this with the argparser itself ? This seems like a common issue that the argparser should handle.

I think maybe we shuold have a `parse_args()` function in DefaultArgParser which handels these checks.
http://stackoverflow.com/questions/4466197/python-argparse-with-dependencies

Basically, I see no reason why this check should be done after the sections are created. It should throw an error at the arg parsing.

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +87,31 @@ def parse_custom_settings(sections,
...
+    conflicts["no_config"].add("find_config")
+
+    args = list(conflicts.keys())
+    for arg in args:
+        for conflict in conflicts[arg]:
+            conflicts[conflict].add(arg)
+
+    for section in cli_sections:
+        for arg in cli_sections[section]:
+            for conflict in conflicts[arg]:
```
AbdealiJK: There are nicer ways of doing this because you use `defaultdict(set)`
`conflicts[arg] & set(cli_sections[section])` is a simpler way which uses the fact that you have sets rather than a for loop and `if` which would be slower.

I think these nested for loops can be optimized.

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +87,31 @@ def parse_custom_settings(sections,
...
+                if conflict in cli_sections[section]:
+                    log_printer.err(conflict + " and " + section + " cannot "
+                                    "be given at the same time")
+                    sys.exit(2)
```
AbdealiJK: There's a better error from argparse:
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.exit

tests/parsing/CliParsingTest.py
```diff
@@ -52,3 +55,12 @@ def test_parse_cli(self):
         self.assertEqual(parsed_sections["default"].name, "Default")
         self.assertEqual(self.dict_from_sections(parsed_sections),
                          expected_dict)
+
+    def test_check_conflicts(self):
+        sections = parse_cli(arg_list=["--save", "--no-config"])
+        with self.assertRaises(SystemExit) as cm:
+            check_conflicts(sections, self.log_printer)
+        self.assertEqual(cm.exception.code, 2)
```
AbdealiJK: You shouldnt be using `cm` outside the `with`

hypothesist: I was skeptical myself until I saw an official pytest docs example. I'll make the change.

coalib/parsing/CliParsing.py
hypothesist: Yep, just found that this: https://docs.python.org/3/howto/argparse.html#conflicting-options

Basically you add

```
group = parser.add_mutually_exclusive_group()
group.add_argument("-v", "--verbose", action="store_true")
group.add_argument("-q", "--quiet", action="store_true")
```

and this automatically raises "`-v` and `-q` can't be given at the same time". ArgParser is really complete!

hypothesist: But it's not clear how to make `[-a | -b] [-a | -c]`. `mutually_exclusive_group` only allows `[-a | -b | -c]`. I'll look into it more.

AbdealiJK: Maybe you could do 2 mutually exclusive groups ?

hypothesist: That would not work since we can't define the same flag twice: we can't do 
```
parser = argparse.ArgumentParser()
parser.add_argument('-e', help='e flag')
g = parser.add_mutually_exclusive_group()
h = parser.add_argument_group()
g.add_argument('-a', help='a flag')
g.add_argument('-b', help='b flag')
g = parser.add_mutually_exclusive_group()
g.add_argument('-a', help='a flag')
g.add_argument('-c', help='c flag')
```
to achieve `[-e] [-a | -b] [-a | -c]`. It really seems they haven't added this feature yet (see [here](http://bugs.python.org/issue10984) and [here](http://bugs.python.org/issue11588)). So I think we have no choice to go with our own custom conflict checker for now.

coalib/parsing/CliParsing.py
hypothesist: I agree, I'll change this.

coalib/parsing/CliParsing.py
hypothesist: I was talking with @sils1297 yesterday. Is this http://ix.io/rrS/py better? It's simpler and removes the redundant lines [105-109].

AbdealiJK: In general if you ahve a for followed by a single if inside a set is
probably what you want with the & operator. As that's exactly what that
does i think
Ive never profiled it myself though

hypothesist: I didn't get you here. Do you mean somehow using an `and` in the for loop statement?

AbdealiJK: (from our gitter conevrsation)

& in sets means intersection:

```
In [4]: set([1,2,3]) & set([2,3,4])
Out[4]: {2, 3}
```

when you do:
```
for a in list1:
    if a in list2:
        # do something
```
it basically means you want the intersection of 2 lists
In such cases, it makes sense to use 2 sets (assuming you dont care about the ordering and duplicated)

Because `for + in` is O(n^2) while `set1 & set2` is O(2*n)

Reference: https://www.ics.uci.edu/~pattis/ICS-33/lectures/complexitypython.txt

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +88,27 @@ def parse_custom_settings(sections,
...
+def check_conflicts(sections):
+    '''
+    Checks if there are any conflicting aruments passed
+
+    :return:            True if no conflicts
+    :raises SystemExit: If there are conflicting arguments (exit code: 2)
+    '''
+    conflicts = {
+        'no_config': set(['save', 'find_config'])
+    }
```
AbdealiJK: Bracket not aligned according to our style guide

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +88,27 @@ def parse_custom_settings(sections,
...
                                section_name=key_touple[0],
                                from_cli=True)
+
+
+def check_conflicts(sections):
+    '''
+    Checks if there are any conflicting aruments passed
+
+    :return:            True if no conflicts
+    :raises SystemExit: If there are conflicting arguments (exit code: 2)
```
AbdealiJK: Documentation should be changed ?

coalib/parsing/CliParsing.py
```diff
@@ -87,3 +88,27 @@ def parse_custom_settings(sections,
                                origin=origin,
                                section_name=key_touple[0],
                                from_cli=True)
+
+
+def check_conflicts(sections):
```
AbdealiJK: Are we not using the conflicting options thing ? https://docs.python.org/3/howto/argparse.html#conflicting-options

hypothesist: No, it currently doesn't support what we need: `[-a | -b] [-a | -c]`. It only supports mutually exclusive arguments like this: `[-a | -b | -c]`

coalib/parsing/CliParsing.py
hypothesist: No, it still raises the same error code of 2 (just that it's from argparse).

coalib/parsing/CliParsing.py
hypothesist: Will fix.

AbdealiJK:
rebase needed :)

sils1297:
ack 75faf4a f27b173

sils1297:
@rultor merge

PR #1886, opened 2016-03-14
python3.coalang: Escape #

abhsag24:
python3.coalang had an unescaped # which caused it to be
unrecognisable by the parser. This PR escapes that


Adrianzatreanu:
ack 7eb196e

abhsag24:
@rultor merge

PR #1891, opened 2016-03-15
Setting: Add url validation.

vivek425ster:
Added url validation in settings. It will check if the given setting
is a valid url or not. If valid will return the settings as a string
else it will throw a Value error.

### DIFF
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -13,6 +13,10 @@ def path_list(obj, *args, **kwargs):
     return obj.__path_list__(*args, **kwargs)
 
 
+def url(obj, *args, **kwargs):
+    return obj.__url__(*args, **kwargs)
+
+
 def typed_list(conversion_func):
     """
     Creates a function that converts a setting into a list of elements each
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ PyPrint>=0.2.0
 setuptools>=19.2
 # Do *not* use 0.3 which is not backwards compatible to common clang versions
 libclang-py3==0.2
+validators>=0.10
--- a/tests/settings/SettingTest.py
+++ b/tests/settings/SettingTest.py
@@ -4,7 +4,7 @@ import unittest
 from collections import OrderedDict
 
 from coalib.settings.Setting import (
-    Setting, path, path_list, typed_dict, typed_list, typed_ordered_dict)
+    Setting, path, path_list, url, typed_dict, typed_list, typed_ordered_dict)
 
 
 class SettingTest(unittest.TestCase):
@@ -35,6 +35,14 @@ class SettingTest(unittest.TestCase):
                            origin="test" + os.path.sep + "somefile")
         self.assertEqual(path_list(self.uut),
                          [os.path.abspath(os.path.join("test", ".")), abspath])
+
+    def test_url(self):
+        uut = Setting("key", "http://google.com")
+        self.assertEqual(url(uut), "http://google.com")
+
+        with self.assertRaises(ValueError):
+            uut = Setting("key", "abc")
+            url(uut)
 
     def test_typed_list(self):
         self.uut = Setting("key", "1, 2, 3")
--- a/coalib/misc/StringConverter.py
+++ b/coalib/misc/StringConverter.py
@@ -1,4 +1,5 @@
 import re
+import validators
 from collections import Iterable, OrderedDict
 
 from coalib.misc import Constants
@@ -55,6 +56,19 @@ class StringConverter:
 
     def __float__(self):
         return float(str(self))
+
+    def __url__(self):
+        """
+        Determines the url validity of this setting.
+
+        :return:            url string
+        :raises ValueError: If the url is not valid.
+        """
+        strrep = str(self).strip()
+        if validators.url(strrep):
+            return strrep
+
+        raise ValueError(repr(strrep) + " is not a valid url.")
 
     def __iter__(self, remove_backslashes=True):
         """


abhsag24:
unack 869e90a could you please run coala over your code? and follow whatever suggestion it gives, it will resolve your issue with circle :) also the commit body doesn't have a `fixes`, it should if it fixes an issue.

vivek425ster:
@It does not fixes any issue. It is an addition which may be later used in fixing other issue.

sils1297:
please remove the trailing period from your shortlog, it was a great inspiration for me though, we'll have a bear checking this soon :P

Adrianzatreanu:
wip until you change the commit message :)

coalib/settings/Setting.py
```diff
@@ -144,6 +149,29 @@ def __path_list__(self):
...
+        """
+        domain1_re = r'(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)'
+        domain2_re = r'(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)'
+        regex = re.compile(
+            r'^(?:http|ftp)s?://'
+            r'(?:' + domain1_re + '+' + domain2_re + '|'
+            r'localhost|'
+            r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'
+            r'(?::\d+)?'
+            r'(?:/?|[/?]\S+)$', re.IGNORECASE)
```
sils1297: neat, is that the django expression? How is that licensed? Can you poin out how you copied it and from where and with which license please? Don't copy unattributed work.

sils1297:
be sure to look at https://github.com/coala-analyzer/coala/pull/1891/commits/9b6ab413f7550d5022c7a63900e35127abe98b1a#r56218492 other than that it looks sane at first glance. Like it ;)

coalib/settings/Setting.py
vivek425ster: @sils1297 As i mentioned in the gitter. I used this regex. 
http://stackoverflow.com/questions/7160737/python-how-to-validate-a-url-in-python-malformed-or-not
Should i add the link to stackoverflow. as a comment ?

sils1297: according to the SO post it comes from `django url validation regex`, i.e. you need to look where in django it's used and then link there and look up the license

vivek425ster: @sils1297 From what it is mentioned in the stackoverflow link.It is from django 1.3
core/validators.py line47
the license of django is https://github.com/django/django/blob/master/LICENSE

sils1297: cool, you need to comply to the license ;)

coalib/settings/Setting.py
```diff
@@ -1,4 +1,8 @@
+#  Copyright (c) Django Software Foundation and individual contributors.
+#  License: https://raw.githubusercontent.com/django/django/master/LICENSE
```
Makman2: huh what? :O

coalib/settings/Setting.py
```diff
@@ -13,6 +17,10 @@ def path_list(obj, *args, **kwargs):
     return obj.__path_list__(*args, **kwargs)
 
 
+def url(obj, *args, **kwargs):
+    return obj.__url__(*args, **kwargs)
```
Makman2: Looks like something general-purpose that can be put into a another module^^

Makman2: @sils1297 or is this okay?

coalib/settings/Setting.py
```diff
@@ -144,6 +152,29 @@ def __path_list__(self):
         """
         return [Setting.__path__(elem, self.origin) for elem in self]
 
+    def __url__(self):
+        """
+        Determines the url validity of this setting.
+
+        :return: url string
+        :raises ValueError: If the url is not valid.
```
Makman2: Please align docs to each other:
```
:return:            The url string.
:raises ValueError: If the ...
```
We still haven't changed to a new documentation format :/

coalib/settings/Setting.py
```diff
@@ -144,6 +152,29 @@ def __path_list__(self):
...
+            r'localhost|'
+            r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'
+            r'(?::\d+)?'
+            r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+        strrep = str(self).strip()
+        result = regex.match(strrep)
+        if result:
+            return strrep
+
+        raise ValueError(strrep + " is not a valid url.")
```
Makman2: please enclose `strrep` with `repr(strrep)`, this way the URL is enclosed inside quotes and escapes also python special characters so one can better debug the address^^

tests/settings/SettingTest.py
```diff
@@ -36,6 +36,14 @@ def test_path_list(self):
         self.assertEqual(path_list(self.uut),
                          [os.path.abspath(os.path.join("test", ".")), abspath])
 
+    def test_url(self):
+        self.uut = Setting("key", "http://google.com")
```
Makman2: no need to use `self` :)

Makman2: no need to use `self` :)

coalib/settings/Setting.py
Makman2: nvm :+1:

coalib/settings/Setting.py
```diff
@@ -144,6 +152,29 @@ def __path_list__(self):
...
+            r'^(?:http|ftp)s?://'
+            r'(?:' + domain1_re + '+' + domain2_re + '|'
+            r'localhost|'
+            r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})'
+            r'(?::\d+)?'
+            r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+        strrep = str(self).strip()
+        result = regex.match(strrep)
+        if result:
+            return strrep
```
Makman2: can be inlined:
```python
if regex.match(strrep):
    return strrep
```

coalib/settings/Setting.py
vivek425ster: @Makman2 How to comply to the license?
https://github.com/coala-analyzer/coala/pull/1891/files#r56218492
I thought that to put the copyright and license file link would be compyling to the license.
No idea how licensing works

AbdealiJK:
Hey, honetstly, rather than the regex ... we could just use http://validators.readthedocs.org/en/latest/#
It seems to be maintained and later we can allow email fields and ipv4/6 too :p  

Also, we may want to try using the "truthy" they have as it seems to be useful ?

This way all the issues about django licenses can be avoided.

vivek425ster:
@AbdealiJK Yes it seems a great idea. 
I will change it to validators then.

coalib/settings/Setting.py
```diff
@@ -144,6 +149,19 @@ def __path_list__(self):
         """
         return [Setting.__path__(elem, self.origin) for elem in self]
 
+    def __url__(self):
```
AbdealiJK: What wha ? Why is here ?

Didnt we decide it should be in `StringConverter` ?
The only reason `__path__` is here is because it needs `self.origin` which doesnt exist in the other thing

requirements.txt
```diff
@@ -2,3 +2,4 @@ PyPrint>=0.2.0
 setuptools>=19.2
 # Do *not* use 0.3 which is not backwards compatible to common clang versions
 libclang-py3==0.2
+validators==0.10
```
AbdealiJK: Please use `>=`

AbdealiJK:
ack  d01fd8a

AbdealiJK:
@rultor merge

PR #1894, opened 2016-03-16
Exceptions: Handle package version conflicts

hypothesist:
Currently, when a requirement for coala is not met, the
default crash message is printed and exited. This commits handles
this exception by displaying why coala crashed along with
information about which package is at fault.

Fixes https://github.com/coala-analyzer/coala/issues/1748

### DIFF
--- a/coalib/misc/Constants.py
+++ b/coalib/misc/Constants.py
@@ -13,6 +13,15 @@ CRASH_MESSAGE = ("An unknown error occurred. This is a bug. We are "
                  "happen. When asked for, the following information "
                  "may help investigating:")
 
+VERSION_CONFLICT_MESSAGE = ("There is a conflict in the version of a "
+                            "dependency you have installed and the "
+                            "requirements of coala. This may be resolved by "
+                            "creating a separate virtual environment for "
+                            "coala or running `pip install %s`. Be aware "
+                            "that the latter solution might break other "
+                            "python packages that depend on the currently "
+                            "installed version.")
+
 OBJ_NOT_ACCESSIBLE = "{} is not accessible and will be ignored!"
 
 TRUE_STRINGS = ['1',
--- a/coalib/misc/Exceptions.py
+++ b/coalib/misc/Exceptions.py
@@ -3,6 +3,8 @@ from pyprint.NullPrinter import NullPrinter
 from coalib.misc import Constants
 from coalib.output.printers.LogPrinter import LogPrinter
 
+from pkg_resources import VersionConflict
+
 
 def get_exitcode(exception, log_printer=None):
     log_printer = log_printer or LogPrinter(NullPrinter())
@@ -15,6 +17,10 @@ def get_exitcode(exception, log_printer=None):
         exitcode = 0
     elif isinstance(exception, SystemExit):
         exitcode = exception.code
+    elif isinstance(exception, VersionConflict):
+        log_message = Constants.VERSION_CONFLICT_MESSAGE % str(exception.req)
+        log_printer.log_exception(log_message, exception)
+        exitcode = 13
     elif isinstance(exception, BaseException):
         log_printer.log_exception(Constants.CRASH_MESSAGE, exception)
         exitcode = 255
--- a/tests/misc/ExceptionsTest.py
+++ b/tests/misc/ExceptionsTest.py
@@ -1,6 +1,7 @@
 import unittest
 
 from coalib.misc.Exceptions import get_exitcode
+from pkg_resources import VersionConflict
 
 
 class ExceptionsTest(unittest.TestCase):
@@ -9,5 +10,7 @@ class ExceptionsTest(unittest.TestCase):
         self.assertEqual(get_exitcode(KeyboardInterrupt()), 130)
         self.assertEqual(get_exitcode(AssertionError()), 255)
         self.assertEqual(get_exitcode(SystemExit(999)), 999)
+        self.assertEqual(get_exitcode(VersionConflict(
+            "libclang-py3 0.3", "libclang-py3==0.2")), 13)
         self.assertEqual(get_exitcode(EOFError()), 0)
         self.assertEqual(get_exitcode(None), 0)
--- a/docs/Users/Exit_Codes.rst
+++ b/docs/Users/Exit_Codes.rst
@@ -11,6 +11,8 @@ The following is a list of coalas exit codes and their meanings:
 -  ``4`` - coala was executed with an unsupported version of python
 -  ``5`` - coala executed successfully. Results were found but patches
    to the results were applied successfully
+-  ``13`` - There is a conflict in the version of a dependency you have
+   installed and the requirements of coala.
 -  ``130`` - A KeyboardInterrupt (Ctrl+C) was pressed during the
    execution of coala.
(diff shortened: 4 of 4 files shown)


coalib/misc/Constants.py
```diff
@@ -13,6 +13,11 @@
                  "happen. When asked for, the following information "
                  "may help investigating:")
 
+VERSION_CONFLICT_MESSAGE = ("There is a conflict in the version of a package "
+                            "you have installed and the requirements of "
+                            "coala. The following information may help "
+                            "to resolve this conflict: ")
```
sils1297: do we have a recommendation on how to solve the conflict?

tests/misc/ExceptionsTest.py
```diff
@@ -1,6 +1,9 @@
 import unittest
 
 from coalib.misc.Exceptions import get_exitcode
+from pkg_resources import ContextualVersionConflict
+from pkg_resources import ResolutionError
```
sils1297: unused import according scrutinizer even

coalib/misc/Constants.py
hypothesist: Yes, it is followed right after the colon.

tests/misc/ExceptionsTest.py
hypothesist: I'm removing this.

coalib/misc/Constants.py
sils1297: no that's info about the problem, not how to solve it

hypothesist: Ah, I think we can access the requirement package (with correct version). I wonder if `pip3 install <package_name>` would work for the string returned by pkg_resources.

sils1297: we need a smart solution there, cleanest thing for the user would be a
virtualenv, alternative force overwriting some versions or something I
think we should mention both

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-03-16 19:14 GMT+01:00 Adhityaa Chandrasekar <notifications@github.com>:

AbdealiJK: I think the user can just google it.
This is an open bug in pip - https://github.com/pypa/pip/issues/988
ideally the 2 packages that cause the version Conflict should never be installed.

coalib/misc/Exceptions.py
```diff
@@ -15,6 +17,12 @@ def get_exitcode(exception, log_printer=None):
         exitcode = 0
     elif isinstance(exception, SystemExit):
         exitcode = exception.code
+    elif (isinstance(exception, ContextualVersionConflict) or
```
AbdealiJK: You can simply do `isinstance(exception, (VersionConflict, ContextualVersionConflict))`
I think

coalib/misc/Constants.py
AbdealiJK: If you want to give a better help , maybe https://pypi.python.org/pypi/pip-conflict-checker/0.3 should be mentioned ?

coalib/misc/Exceptions.py
```diff
@@ -15,6 +17,12 @@ def get_exitcode(exception, log_printer=None):
...
+          isinstance(exception, VersionConflict)):
+        log_message = Constants.VERSION_CONFLICT_MESSAGE
+        log_message += exception.report()
+        log_printer.log_exception(log_message, exception)
+        exitcode = 4
```
AbdealiJK: A new exitcode needs documentation.
What is the normal exitcode of pip when versionConflict occurs ?

coalib/misc/Exceptions.py
```diff
@@ -3,6 +3,8 @@
 from coalib.misc import Constants
 from coalib.output.printers.LogPrinter import LogPrinter
 
+from pkg_resources import ContextualVersionConflict, VersionConflict
```
AbdealiJK: Scrutinizer fails saying the Contextual one doesnt exist.
I cant find in the docs either : https://pythonhosted.org/setuptools/pkg_resources.html#exceptions

Seems like the setuptools version is not high enough in scrutinizer ?
http://sickrage.ca/forums/forum/help-support/bug-issue-reports/33421-importerror-cannot-import-name-contextualversionconflict

coalib/misc/Constants.py
sils1297: > I think the user can just google it.

Well you see how we struggle to find a proper solution if we had a recommendation that'd be great for the users.

coalib/misc/Exceptions.py
hypothesist: I just experimented around with different stuff and I'm now fairly certain that `VersionConflict` is enough: I think `ContextualVersionConflict` is derived from `VersionConflict` and `ContextualVersionConflict` is raised sometimes in the newer versions, but `VersionConflict` is raised always in the older ones.

And when I put just `VersionConflict` with the newer version, it always capturing the exception. Go ahead with it?

coalib/misc/Exceptions.py
hypothesist: It's not really new: [Docs](http://coala.readthedocs.org/en/latest/Users/Exit_Codes.html?highlight=4) - Just felt that it's another version related problem. But I think this deserves its own exit code. How [about 13](http://www.ashout.com/fix-java-started-returned-exit-code13-eclipse/)? It's used by Java.

coalib/misc/Exceptions.py
AbdealiJK: This issue has irritated so much that I'd rather just put both "in case".

In side contextual isn't a subclass because I tried isinstance(ContextualVersionConflict, VersionConflict) and it was false.

Lets put both and ignore scrutinizer

-----Original Message-----

coalib/misc/Exceptions.py
AbdealiJK: @sils1297 ? Could you take a call on this?

-----Original Message-----

sils1297: this one's a new exitcode IMO definitely and needs an own line of docs, not that it takes long to write it does it?

hypothesist: It's not a new exit code, just an added meaning to the old one. I'll amend the docs.

coalib/misc/Exceptions.py
hypothesist: See https://github.com/jaraco/setuptools/blob/master/pkg_resources/__init__.py#L359

Still go ahead? It might fail for some people.

.scrutinizer.yml
```diff
@@ -0,0 +1,4 @@
+build:
+    dependencies:
```
AbdealiJK: we seem to be using 2 spaces for all the indents in .yml files

hypothesist: Actually, I'll remove this too in the next commit as it is no longer needed (added it to try and make scrutinizer use the latest setuptools for `ContextualVersionConflict`).

hypothesist: Remove the whole file, I mean.

AbdealiJK: marking as WIP then

AbdealiJK:
ack 4e2b940

AbdealiJK:
@rultor merge

PR #1906, opened 2016-03-17
requirements: Use latest PyPrint

sils1297:
This one uses ~= and this we should escape dependency problems a bit.


Adrianzatreanu:
ack 54ffd81 595e8f3

sils1297:
@rultor merge

PR #1924, opened 2016-03-19
ContextManagers: Add contexmanager for os.chdir()

Uran198:
Fixes https://github.com/coala-analyzer/coala/issues/1314

An issue in `coala-bears` should be open regarding:

- https://github.com/coala-analyzer/coala-bears/blob/master/tests/vcs/git/GitCommitBearTest.py#L157
- https://github.com/coala-analyzer/coala-bears/blob/master/bears/vcs/git/GitCommitBear.py#L43

### DIFF
--- a/coalib/misc/ContextManagers.py
+++ b/coalib/misc/ContextManagers.py
@@ -235,3 +235,13 @@ def prepare_file(lines,
             yield lines, filename
     else:
         yield lines, filename
+
+
+@contextmanager
+def change_directory(path):
+    old_dir = os.getcwd()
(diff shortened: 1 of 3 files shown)


sils1297:
ack https://github.com/coala-analyzer/coala/pull/1924/commits/0e6447a37a76f9bbd0db4383ed95590b2399245d, that's neat thank you :) For the future I'd rather see the addition of the function and the use of it in own commits, this one's small enough so we can take it like this.

sils1297:
@rultor merge

PR #1936, opened 2016-03-21
Bear: Add helper function name()

djkonro:
Frequently, we need to use the name of the bear
inside the bear.
This should add the name() function which allows
users to just do self.name() easily.

Fixes https://github.com/coala-analyzer/coala/issues/1893

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -2,7 +2,7 @@ import traceback
 
 from pyprint.Printer import Printer
 
-from coalib.misc.Decorators import enforce_signature
+from coalib.misc.Decorators import enforce_signature, classproperty
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
@@ -31,6 +31,13 @@ class Bear(Printer, LogPrinter):
 
     Settings are available at all times through self.section.
     """
+
+    @classproperty
+    def name(cls):
+        """
+        :return: The name of the bear
+        """
+        return cls.__name__
 
     @enforce_signature
     def __init__(self,
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -67,7 +67,7 @@ class Bear(Printer, LogPrinter):
 
         cp = type(self).check_prerequisites()
         if cp is not True:
-            error_string = ("The bear " + type(self).__name__ +
+            error_string = ("The bear " + self.name +
                             " does not fulfill all requirements.")
             if cp is not False:
                 error_string += " " + cp
@@ -91,13 +91,13 @@ class Bear(Printer, LogPrinter):
                 self.get_metadata().create_params_from_section(self.section))
         except ValueError as err:
             self.warn("The bear {} cannot be executed.".format(
-                type(self).__name__), str(err))
+                self.name), str(err))
             return
 
         return self.run(*args, **kwargs)
 
     def execute(self, *args, **kwargs):
-        name = type(self).__name__
+        name = self.name
         try:
             self.debug("Running bear {}...".format(name))
             # If it's already a list it won't change it
--- a/coalib/misc/Decorators.py
+++ b/coalib/misc/Decorators.py
@@ -298,3 +298,29 @@ def enforce_signature(function):
         return function(*args, **kwargs)
 
     return decorated
+
+
+class classproperty(property):
+    """
+    Decorator to set a class function to a class property.
+
+    Given a class like:
+
+    >>> class test:
+    ...     @classproperty
+    ...     def func(self):
+    ...         return 1
+
+    We can now access the class property using the class name:
+
+    >>> test.func
+    1
+
+    And we can still have the same behaviour with an instance:
+
+    >>> test().func
+    1
+    """
+
+    def __get__(self, obj, type_):
+        return self.fget.__get__(None, type_)(type_)
--- a/coalib/processes/BearRunning.py
+++ b/coalib/processes/BearRunning.py
@@ -136,7 +136,7 @@ def get_local_dependency_results(local_result_list, bear_instance):
                               results are picked.
     :param bear_instance:     The instance of a local bear to get the
                               dependencies from.
-    :return:                  Return none if their are no dependencies for the
+    :return:                  Return none if there are no dependencies for the
                               bear. Else return a dictionary containing
                               dependency results.
     """
--- a/coalib/processes/BearRunning.py
+++ b/coalib/processes/BearRunning.py
@@ -98,7 +98,7 @@ def run_bear(message_queue, timeout, bear_instance, *args, **kwargs):
     if kwargs.get("dependency_results", True) is None:
         del kwargs["dependency_results"]
 
-    name = bear_instance.__class__.__name__
+    name = bear_instance.name
 
     try:
         result_list = bear_instance.execute(*args, **kwargs)
@@ -363,7 +363,7 @@ def get_next_global_bear(timeout,
         if dependency_results is False:
             global_bear_queue.put(bear_id)
 
-    return bear, bear.__class__.__name__, dependency_results
+    return bear, dependency_results
 
 
 def task_done(obj):
@@ -450,11 +450,12 @@ def run_global_bears(message_queue,
     """
     try:
         while True:
-            bear, bearname, dep_results = (
+            bear, dep_results = (
                 get_next_global_bear(timeout,
                                      global_bear_queue,
                                      global_bear_list,
                                      global_result_dict))
+            bearname = bear.__class__.__name__
             result = run_global_bear(message_queue, timeout, bear, dep_results)
             if result:
                 global_result_dict[bearname] = result
--- a/docs/Users/Tutorials/Writing_Bears.rst
+++ b/docs/Users/Tutorials/Writing_Bears.rst
@@ -217,7 +217,7 @@ So let's see how you could tell *coala* which Bears to run before yours:
     class DependentBear(LocalBear):
 
         def run(self, filename, file, dependency_results):
-            results = dependency_results[OtherBear.__name__]
+            results = dependency_results[OtherBear.name]
 
         @staticmethod
         def get_dependencies():
--- a/coalib/bearlib/abstractions/Lint.py
+++ b/coalib/bearlib/abstractions/Lint.py
@@ -227,7 +227,7 @@ class Lint(Bear):
                 groups[variable] = int(groups[variable])
 
         if "origin" in groups:
-            groups['origin'] = "{} ({})".format(str(self.__class__.__name__),
+            groups['origin'] = "{} ({})".format(str(self.name),
                                                 str(groups["origin"]))
 
         return Result.from_values(
--- a/coalib/collecting/Dependencies.py
+++ b/coalib/collecting/Dependencies.py
@@ -6,7 +6,7 @@ class CircularDependencyError(Exception):
         Creates the CircularDependencyError with a helpful message about the
         dependency.
         """
-        bear_names = [bear.__name__ for bear in bears]
+        bear_names = [bear.name for bear in bears]
 
         return cls("Circular dependency detected: " + " -> ".join(bear_names))
 
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -634,7 +634,7 @@ def show_bear(console_printer, bear, sections, metadata):
     :param sections:        The sections to which the bear belongs.
     :param metadata:        Metadata about the bear.
     """
-    console_printer.print("{bear}:".format(bear=bear.__name__))
+    console_printer.print("{bear}:".format(bear=bear.name))
     console_printer.print("  " + metadata.desc + "\n")
 
     show_enumeration(console_printer,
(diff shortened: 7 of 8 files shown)


coalib/bears/Bear.py
```diff
@@ -68,6 +68,9 @@ def __init__(self,
             self.warn(error_string)
             raise RuntimeError(error_string)
 
+    def name(self):
```
sils1297: hey, can you make this a property? Then you can access it without the brackets `self.name` that'd be neat, woudln't it?

coalib/processes/BearRunning.py
```diff
@@ -136,7 +136,7 @@ def get_local_dependency_results(local_result_list, bear_instance):
                               results are picked.
     :param bear_instance:     The instance of a local bear to get the
                               dependencies from.
-    :return:                  Return none if their are no dependencies for the
+    :return:                  Return none if there are no dependencies for the
```
sils1297: can you put this in an own commit? `git add -p` and our community on gitter will help you with the splitup :)

Adrianzatreanu:
ack a50382e

coalib/processes/BearRunning.py
```diff
@@ -98,7 +98,7 @@ def run_bear(message_queue, timeout, bear_instance, *args, **kwargs):
     if kwargs.get("dependency_results", True) is None:
         del kwargs["dependency_results"]
 
-    name = bear_instance.__class__.__name__
+    name = bear_instance.name
```
sils1297: hey some occasions missing, not?

```
lasse@localhost ~/prog/coala (git)-[sils/resadd] % g g __class__.__name__
coalib/bearlib/abstractions/Lint.py:            groups['origin'] = "{} ({})".format(str(self.__class__.__name__),
coalib/processes/BearRunning.py:    name = bear_instance.__class__.__name__
coalib/processes/BearRunning.py:                 "it out...".format(bear_instance.__class__.__name__),
coalib/processes/BearRunning.py:                 .format(global_bear_instance.__class__.__name__),
coalib/processes/BearRunning.py:    return bear, bear.__class__.__name__, dependency_results
coalib/results/Result.py:            origin = origin.__class__.__name__
```

sils1297:
https://github.com/coala-analyzer/coala/pull/1936/commits/98144f8422b3e2ff1278dff1245e5ec3516dbe31 needs work

coalib/processes/BearRunning.py
sils1297: it is important to note that we sometimes do `bear.__name__` those occasions have to stay like they are since they use the *class* not the *instance*

djkonro:
@sils1297 , When I try to change the  bear.__class__.__name__  to bear.name in BearRunning.py. I get the error

`>       return bear, bear.name, dependency_results
E       AttributeError: 'str' object has no attribute 'name'`

I have tried tracing the calls to the run(...) function in this file, and I see it imported in Processing.py but I cannot find where it is used within the file. I am still studying the test to understand, how it is tested.

As for origin.__class__.__name__,  from my understanding it does not have to be a bear, I could be a string, in this case it would evaluate to 'str'.

Please I would need some help/clarification. I am sorry if some of the things I have said above do not make sense or are very trivial. I am still trying to get a better understanding of the code.

AbdealiJK:
I was thinking, and it seems making a `classmethod` rather than a property makes more sense.
Because then we can directly find name using the class and the instance.

```
In [4]: class A:
   ...:     @classmethod
   ...:     def name(cls):
   ...:         return cls.__name__
   ...:     

In [6]: A.name()
Out[6]: 'A'

In [7]: a = A()

In [9]: a.name()
Out[9]: 'A'
```
What are your thoughts ?
This seems a better alternative than doing `bear.name` for instances and `bear.__name__` for the class itself. As it is uniform, and the name **is** a class attribute - so classmethod makes logical sense.

Note, you can make a `classproperty` too incase you dont want to use `name()` and just want `name`
http://stackoverflow.com/questions/128573/using-property-on-classmethods

AbdealiJK:
@djkonro Replying to your previous message:

 > bear.name in BearRunning.py

That's weird. This should work from what I see. Maybe you've changed something somewhere else that caused this ?
Also, do we even need to return this anymore ? Because `bear.name` is so easy that I think that should not be returned anymore.

> As for origin.__class__.__name__

Yes, you're right it can be a bear or a string. In `Result` we do:
```
        if not isinstance(origin, str):
            origin = origin.__class__.__name__
```
at [line 55](https://github.com/coala-analyzer/coala/blob/master/coalib/results/Result.py#L56)

sils1297:
> I was thinking, and it seems making a classmethod rather than a property makes more sense.

like that

tests/processes/BearRunningTest.py
```diff
@@ -253,13 +253,16 @@ def setUp(self):
         self.file_name_queue.put("invalid file")
         self.local_bear_list.append(LocalTestBear(self.settings,
                                                   self.message_queue))
-        self.local_bear_list.append("not a valid bear")
+        self.local_bear_list.append(SimpleGlobalBear({},
+                                                     self.settings,
+                                                     self.message_queue))
```
sils1297: why those changes? This tests that the routine handles wrong things like this so if it fails you should rather fix the method not the test right? This is also why coverage is going down.

sils1297:
second commit message, I'd simplify it to 

    BearRunning: `their` -> `there`

tests/processes/BearRunningTest.py
djkonro: @sils1297 , What is your suggestion on how to handle this, because the error message says
'''
send_msg(message_queue,
                 timeout,
                 LOG_LEVEL.WARNING,
                 "A given local bear ({}) is not valid. Leaving "
                 "it out...".format(bear_instance.__class__.__name__),
                 Constants.THIS_IS_A_BUG)
'''
which would return 'str' for the string which was previously used in the test, instead of the "not a valid bear" message.

docs/Users/Tutorials/Writing_Bears.rst
```diff
@@ -217,7 +217,7 @@ So let's see how you could tell coala which Bears to run before yours:
     class DependentBear(LocalBear):
 
         def run(self, filename, file, dependency_results):
-            results = dependency_results[OtherBear.__name__]
+            results = dependency_results[OtherBear.name]
```
abhsag24: i guess this is one of those cases where we want the name of the class

sils1297:
Hey @djkonro need help?

djkonro:
@sils1297 , Yes, I did not get a reply to my previous question.

> @sils1297` , What is your suggestion on how to handle this, because the error message says
```
send_msg(message_queue,
                 timeout,
                 LOG_LEVEL.WARNING,
                 "A given local bear ({}) is not valid. Leaving "
                 "it out...".format(bear_instance.__class__.__name__),
                 Constants.THIS_IS_A_BUG)
```
which would return 'str' for the string which was previously used in the test, instead of the "not a valid bear" message.

coalib/bears/Bear.py
```diff
@@ -9,6 +9,12 @@
 from coalib.settings.ConfigurationGathering import get_config_directory
 
 
+class classproperty(property):
+
+    def __get__(self, obj, type_):
+        return self.fget.__get__(None, type_)()
```
Makman2: I think you can directly convert the given function to a `classmethod`, so you don't need to write like below
```python
@classproperty
@classmethod
def ...
```

Makman2: Also such a function would belong to the module `coalib.misc.Decorators` and should go into another commit + according tests^^

Makman2:
@djkonro maybe I can help you, but I don't understand the problem currently why you adapted those tests^^

coalib/processes/BearRunning.py
Makman2: the classproperty should work on both^^

djkonro:
@Makman2, Thanks, for the pointers I would update the classmethod. As for the test, the value which is been passed is of type "str" and not of type "bear" so the test would pass, but the error message would contian "str" in the place of the expected string "not a valid bear". This because you would get "str" when you invoke `__class__.__name__` on a string and also a string cannot inherit the `name()` method. I also had this same talk with @AbdealiJK and he realized what I was say after testing it out. It is really frustrating that no one getting what I am trying to say.

djkonro:
I passed instance of type bear to the list, but not the expected bear type. That is, I passed a local bear to the Gobal bear list and vice-versa. This is so the test would work properly and produce a valid error message.

djkonro:
When @AbdealiJK realized what I was trying to say his exact reply was
> Wow, interesting. So that test has been broken since ages :o ?

Makman2:
> for the pointers I would update the classmethod.

Huh what pointers? :O

But I think now I get you, an `AttributeError` is raised because `str` contains no member `name`, that's why you actually feed in a bear. Though you changed the test so that the case isn't tested any more.

Makman2:
The test isn't broken, in fact we test some case which shouldn't occur ever, something else than a bear instance shouldn't get through to this function^^

Makman2:
alright okay that's quite tricky, I'll investigate that further tomorrow @djkonro ^^

tests/processes/BearRunningTest.py
sils1297: but that's right, a `str` is not a valid bear!

djkonro: @sils1297 Thanks, I get what you are saying, but this still brings me back to the question I have been asking for more than a week now. Any suggestions on how to handle the fact that a string type can not inherit the `name()` method ?. That is, should we maintain the `bear_instance.__class__.__name__` or add a condition like  
`if not isinstance(bear_instance, Bear)` to handles other types ?.

sils1297: ah, the latter. We deliberately pass something stupid here to *test that there is proper error handling*

sils1297: sorry if I haven't responded to some questions, with the github notification system it's very hard to determine where my attention is needed.

coalib/bears/Bear.py
djkonro: @Makman2 , Please can you explain more on what you mean by
> I think you can directly convert the given function to a classmethod, so you don't need to write like below
```
@classproperty
@classmethod
def ...
```

Makman2: I think you can do directly `classmethod(self.fget). ...`, then you can declare a classproperty without declaring it as `classmethod` explicitly.
But you can google classproperties in python there're enough snippets out there^^

Makman2:
@djkonro sorry it took so long, semester has just begun :)
I would say for simplicity, you don't change some statements to using `bear.name`, so you keep the old `bear.__class__.__name__`. This will get refactored from the next-gen-design by me anyway, and I have the feeling spending too much time on it trying to integrate it for the tests is not worth it (in fact no string should ever disembark there^^). Then coverage will remain 100%. If you do so mention please inside the commit message body that some occurrences inside BearRunning aren't replaced because they deal with different types, not only bears^^
@sils1297 you agree?

sils1297:
@Makman2 that's totally fine

djkonro:
@Makman2, Ok thanks, I have also  not been able to do much, because of school work. :disappointed:

coalib/misc/Decorators.py
```diff
@@ -298,3 +298,9 @@ def decorated(*args, **kwargs):
         return function(*args, **kwargs)
 
     return decorated
+
+
+class classproperty(property):
+
```
sils1297: can you add a documentation comment with an example on how to use it as a doctest?

coalib/bears/Bear.py
```diff
@@ -32,6 +32,13 @@ class Bear(Printer, LogPrinter):
     Settings are available at all times through self.section.
     """
 
+    @classproperty
+    def name(cls):
+        """
+        :return: The name of the bear
```
sils1297: a doctest would be small, simple, illustrating and good to have IMO

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1936/commits/2ba9324de6577fa8a7c30627b0cacf8759c80c78 https://github.com/coala-analyzer/coala/pull/1936/commits/8d89bbe5a3f4f950cb386e3b15037c20e6543f71

sils1297:
unack 35f9a6e 6d3762d

sils1297:
hey @djkonro this is good stuff, really!

coalib/misc/Decorators.py
```diff
@@ -298,3 +298,24 @@ def decorated(*args, **kwargs):
...
 
     return decorated
+
+
+class classproperty(property):
+    """
+    Decorator to set a class function to a class property.
+
+    >>> class test:
+    ...
```
sils1297: remove this line, in doc comments rather be short

coalib/misc/Decorators.py
```diff
@@ -298,3 +298,24 @@ def decorated(*args, **kwargs):
...
+    ...     @classproperty
+    ...     def func(self):
+    ...         return 1
+    ...
```
sils1297: this one's unneeded too

coalib/misc/Decorators.py
```diff
@@ -298,3 +298,24 @@ def decorated(*args, **kwargs):
...
+    >>> t = test()
+    >>> t.func
+    1
```
sils1297: I'd move lines 313 to 315 below 316 for clarity

coalib/misc/Decorators.py
```diff
@@ -298,3 +298,24 @@ def decorated(*args, **kwargs):
...
+    >>> test.func
+    1
```
sils1297: so I'd write a doctest comment usually like:

```
Given a class like:

>>> class ...:
...     def ...():
...         do_something()

If we now access the property of the class:

>>> classname.func
1

And the same behaviour with an instance:

>>> class().func
1
```

(be sure to phrase it out and not use my stupid example texts ;))

coalib/bears/Bear.py
sils1297: still missing

sils1297: nah actually I think it's ok this way

sils1297:
also rebase please

sils1297:
hey you didn't address my comments :(

djkonro:
@sils1297, I was still working on it. Thanks :)

djkonro:
@sils1297, I have made the changes. Could you please check it out and let me know, if that is what you requested. Thanks.

sils1297:
ack 9a5dde3 a7951a0 8cc7989 c77ccfe

sils1297:
@rultor merge

PR #1947, opened 2016-03-23
SourceRange add position functionalities 

abhsag24:
Absolute Position to work with positions directlyd

### DIFF
--- a/coalib/results/AbsolutePosition.py
+++ b/coalib/results/AbsolutePosition.py
@@ -1,0 +1,63 @@
+from coalib.results.TextPosition import TextPosition
+from coalib.misc.Decorators import enforce_signature
+
+
+class AbsolutePosition(TextPosition):
+
+    @enforce_signature
+    def __init__(self,
+                 text: (tuple, list, None)=None,
+                 position: (int, None)=None):
+        """
+        Creates an AbsolutePosition object that represents the index of a
+        character in a string.
+
+        :param text:     The text containing the character.
+        :param position: Position identifying the index of character
+                         in text.
+        """
+        line = column = None
+        if position is not None and text is not None:
+            line, column = calc_line_col(text, position)
+        self._text = text
+        self._position = position
+        super().__init__(line, column)
+
+    @property
+    def position(self):
+        return self._position
+
+
+def calc_line_col(text, position):
+    r"""
+    Creates a tuple containing (line, column) by calculating line number
+    and column in the text, from position.
+
+    The position represents the index of a character. In the following
+    example 'a' is at position '0' and it's corresponding line and column are:
+
+    >>> calc_line_col(('a\n',), 0)
+    (1, 1)
+
+    All special characters(including the newline character) belong in the same
+    line, and have their own position. A line is an item in the tuple:
+
+    >>> calc_line_col(('a\n', 'b\n'), 1)
+    (1, 2)
+    >>> calc_line_col(('a\n', 'b\n'), 2)
+    (2, 1)
+
+    :param text:          A tuple/list of lines in which position is to
+                          be calculated.
+    :param position:      Position (starting from 0) of character to be found
+                          in the (line, column) form.
+    :return:              A tuple of the form (line, column), where both line
+                          and column start from 1.
+    """
+    for linenum, line in enumerate(text, start=1):
+        linelen = len(line)
+        if position < linelen:
+            return linenum, position + 1
+        position -= linelen
+
+    raise ValueError("Position not found in text")
--- a/coalib/results/SourceRange.py
+++ b/coalib/results/SourceRange.py
@@ -3,6 +3,7 @@ from os.path import relpath
 from coalib.misc.Decorators import enforce_signature, get_public_members
 from coalib.results.SourcePosition import SourcePosition
 from coalib.results.TextRange import TextRange
+from coalib.results.AbsolutePosition import AbsolutePosition
 
 
 class SourceRange(TextRange):
@@ -56,6 +57,25 @@ class SourceRange(TextRange):
                                range.start.column,
                                range.end.line,
                                range.end.column)
+
+    @classmethod
+    @enforce_signature
+    def from_absolute_position(cls,
+                               file: str,
+                               position_start: AbsolutePosition,
+                               position_end: (AbsolutePosition, None)=None):
+        """
+        Creates a SourceRange from a start and end positions.
+
+        :param file:           Name of the file.
+        :param position_start: Start of range given by AbsolutePosition.
+        :param position_end:   End of range given by AbsolutePosition or None.
+        """
+        start = SourcePosition(file, position_start.line, position_start.column)
+        end = None
+        if position_end:
+            end = SourcePosition(file, position_end.line, position_end.column)
+        return cls(start, end)
 
     @property
     def file(self):
--- a/tests/results/AbsolutePositionTest.py
+++ b/tests/results/AbsolutePositionTest.py
@@ -1,0 +1,75 @@
+import unittest
+
+from coalib.results.AbsolutePosition import AbsolutePosition, calc_line_col
+from coalib.misc.Constants import COMPLEX_TEST_STRING
+
+
+class AbsolutePositionTest(unittest.TestCase):
+
+    def test_calc_line_col_newlines(self):
+        # no newlines
+        text = ("find position of 'z'",)
+        z_pos = text[0].find('z')
+        self.assertEqual(
+                calc_line_col(text, z_pos), (1, z_pos + 1))
+
+        # newline
+        text = ("find position of\n", "'z'",)
+        string_text = ''.join(text)
+        z_pos = string_text.find('z')
+        self.assertEqual(calc_line_col(text, z_pos), (2, 2))
+
+    def test_calc_line_col_unicode(self):
+        uni_pos = COMPLEX_TEST_STRING.find("↑")
+        self.assertEqual(
+                calc_line_col((COMPLEX_TEST_STRING,), uni_pos),
+                (1, uni_pos + 1))
+
+    def test_calc_line_col_rawstrings(self):
+        for raw in [(r'a\b',), (r'a\n',), ('a\\n',)]:
+            pos = raw[0].find(raw[0][-1])
+            self.assertEqual(calc_line_col(raw, pos), (1, 3))
+
+    def test_calc_line_col_extremes(self):
+        # End of Line
+        text = ("Fitst Line\n", "End of sencond line z")
+        string_text = ''.join(text)
+        z_pos = string_text.find('z')
+        self.assertEqual(calc_line_col(text, z_pos),
+                         (2, len(text[1])))
+
+        # Out of text
+        with self.assertRaises(ValueError):
+            text = ("Some line")
+            calc_line_col(text, 50)
+
+        # start of line
+        text = ("First Line\n", "zEnd of sencond line")
+        string_text = ''.join(text)
+        z_pos = string_text.find('z')
+        self.assertEqual(calc_line_col(text, z_pos), (2, 1))
+
+    def test_property(self):
+        uut = AbsolutePosition(("1", "2"), 1)
+        self.assertEqual(uut.position, 1)
+        self.assertEqual(uut.line, 2)
+        self.assertEqual(uut.column, 1)
+
+        uut = AbsolutePosition()
+        self.assertEqual(uut.position, None)
+        self.assertEqual(uut.line, None)
+        self.assertEqual(uut.column, None)
+
+        uut = AbsolutePosition(("a\n", "b\n"), 0)
+        self.assertEqual(uut.position, 0)
+        self.assertEqual(uut.line, 1)
+        self.assertEqual(uut.column, 1)
+
+    def test_instantiation(self):
+        with self.assertRaises(ValueError):
+            uut = AbsolutePosition((), 0)
+
+        uut = AbsolutePosition(position=5)
+        self.assertEqual(uut.position, 5)
+        self.assertEqual(uut.line, None)
+        self.assertEqual(uut.column, None)
--- a/tests/results/SourceRangeTest.py
+++ b/tests/results/SourceRangeTest.py
@@ -3,6 +3,7 @@ from collections import namedtuple
 
 from coalib.results.SourcePosition import SourcePosition
 from coalib.results.SourceRange import SourceRange
+from coalib.results.AbsolutePosition import AbsolutePosition
 
 
 class SourceRangeTest(unittest.TestCase):
@@ -37,6 +38,19 @@ class SourceRangeTest(unittest.TestCase):
 
         uut = SourceRange.from_clang_range(ClangRange(start, end))
         compare = SourceRange.from_values("t.c", 1, 2, 3, 4)
+        self.assertEqual(uut, compare)
+
+    def test_from_absolute_position(self):
+        text = ("a\n", "b\n")
+        start = AbsolutePosition(text, 0)
+        end = AbsolutePosition(text, 2)
+
+        uut = SourceRange.from_absolute_position("F", start, end)
+        compare = SourceRange.from_values("F", 1, 1, 2, 1)
+        self.assertEqual(uut, compare)
+
+        uut = SourceRange.from_absolute_position("F", start, None)
+        compare = SourceRange(SourcePosition("F", 1, 1), None)
         self.assertEqual(uut, compare)
 
     def test_file_property(self):


coalib/results/SourceRange.py
```diff
@@ -61,6 +61,33 @@ def from_clang_range(cls, range):
...
         return self.start.file
 
+    @classmethod
+    def calc_line_col(cls, text, pos_to_find):
+        """
+        Calculate line number and column in the text, from position.
+        Uses \\n as the newline character. Lines and columns start from 1.
+
+        :param text:        A string with all the contents of file.
+        :param pos_to_find: position of character to be found in the
```
AbdealiJK: Start doc with capital ?

AbdealiJK:
Type "caloumn" in first commit

coalib/results/SourceRange.py
```diff
@@ -62,12 +69,23 @@ def from_position(cls, filename, text, start_position, end_position):
...
 
     @property
     def file(self):
         return self.start.file
 
+    @property
+    def position(self):
+        if self.text is not None:
+            return self.start_position, self.end_position
+
```
abhsag24: i guess an unnecessary newline :(

coalib/results/SourceRange.py
```diff
@@ -62,12 +69,23 @@ def from_position(cls, filename, text, start_position, end_position):
...
+        else:
+            return None
```
AbdealiJK: Not sure whether we should throw error or return None here.

coalib/results/SourceRange.py
```diff
@@ -62,12 +69,23 @@ def from_position(cls, filename, text, start_position, end_position):
...
+                   text=text,
+                   start_position=start_position,
+                   end_position=end_position)
 
     @property
     def file(self):
         return self.start.file
 
+    @property
+    def position(self):
```
AbdealiJK: But, if I do from_values with text, start and end You should calculate start_position, end_position here

coalib/results/SourceRange.py
```diff
@@ -10,7 +10,10 @@ class SourceRange(TextRange):
     @enforce_signature
     def __init__(self,
                  start: SourcePosition,
-                 end: (SourcePosition, None)=None):
```
AbdealiJK: This has not been well thought through.

`start` is required. So, giving `test, start_pos, end_pos` ONLY is not possible

coalib/results/SourceRange.py
```diff
@@ -10,7 +10,10 @@ class SourceRange(TextRange):
...
+                 end: (SourcePosition, None)=None,
+                 text=None,
+                 start_position=None,
```
AbdealiJK: If I give text, start_pos, and end_pos only, "filename" will not be known

AbdealiJK: Also, we seem to be using enforce_signature. SO you should do `start_position: (int, None)=None`

coalib/results/SourceRange.py
```diff
@@ -10,7 +10,10 @@ class SourceRange(TextRange):
...
+                 end_position=None):
```
AbdealiJK: We need to do assertions in the function to make sure `start_position` and `start` are not given together and so on. Otherwise it's confusing.

coalib/results/SourceRange.py
abhsag24: but that would require, from_values to have text as a parameter, and that'd break a lot of things? maybe an optional param there too, and create start_position and end_position, if they are given?

coalib/results/SourceRange.py
abhsag24: i don't change start, start would always be there right?, i don't provide any way in which the user could give `text, start_pos, end_pos` ONLY

coalib/results/SourceRange.py
abhsag24: @AbdealiJK  AFAIK file is picked up from the SourcePosition, we don't need to pass that in SourceRange?

coalib/results/SourceRange.py
abhsag24: We can't throw an error, since __json__ calls all functions, and always throws an error, since text defaults to `None`

coalib/results/SourceRange.py
abhsag24: thanks :) i couldn't come up with how to solve tackle this :P

coalib/results/SourceRange.py
AbdealiJK: I think you're assuming a user will always use the from_* functions. That's only for easier usage.
A user can just create an instance of this class normally

coalib/results/SourceRange.py
AbdealiJK: Sorry. I meant if I do that with the constructor without using the helper from_values()

coalib/results/SourceRange.py
AbdealiJK: Another thing ...
Position should be a property of SourcePosition - as that makes logical sense, and we should use that (like how the file property is done)

sils1297:
marking wip as per @AbdealiJK s review

coalib/results/TextPosition.py
```diff
@@ -33,6 +33,19 @@ def column(self):
         return self._column
 
     @classmethod
+    def calc_position(self, text, line, column):
+        """
+        Calculates position in a string from line number and column.
+        Uses \\n  as the newline character. Position starts from 0
```
AbdealiJK: extra space after \n ?

AbdealiJK: Period after 0

coalib/results/TextPosition.py
```diff
@@ -31,3 +44,47 @@ def line(self):
     @property
     def column(self):
         return self._column
+
+    @property
+    def position(self):
+        return self._position
+
+    @classmethod
+    def calc_position(self, text, line, column):
```
abhsag24: needs to be `cls`

coalib/results/TextPosition.py
```diff
@@ -33,6 +33,19 @@ def column(self):
         return self._column
 
     @classmethod
+    def calc_position(self, text, line, column):
+        """
+        Calculates position in a string from line number and column.
+        Uses \\n  as the newline character. Position starts from 0
+        """
+        lines = text.splitlines(keepends=True)
```
AbdealiJK: :( We *could* make this algo better and faster me thinks.

BTW: Just so you know, 58570fbc5cfcf418d5a8c8134942042c2476b7fd also has a simpler algo which is doing `.splitlines(True)` and then using cumulative sum until you get the line, and then using the left over position to get the column

We probably should `timeit` them both and check efficiency :) (But this can go into another issue)

coalib/results/TextPosition.py
```diff
@@ -33,6 +33,19 @@ def column(self):
...
+        position = 0
+        for line_nr in range(0, line-1):
+            position += len(lines[line_nr])
```
AbdealiJK: This is basically a cumsum. I'd recomment not doing a for loop (they're slow).
You can just a generator with `sum()`, it'd be faster.

PS: For more extreme optimization http://stackoverflow.com/questions/9258602/elegant-pythonic-cumsum

coalib/results/TextPosition.py
abhsag24: @AbdealiJK  this is a different algo? that was for calculating line, col from positionn, this ones the other way round ? i dont understand

coalib/results/TextPosition.py
```diff
@@ -15,12 +19,21 @@ def __init__(self, line: (int, None)=None, column: (int, None)=None):
...
                             in a line is 1.
+        :param text:        The text containing the character.
+        :param position:    Position identifying the index of character in text.
         :raises TypeError:  Raised when line or columns are no integers.
         :raises ValueError: Raised when a column is set but line is None.
         """
         if line is None and column is not None:
             raise ValueError("A column can only be set if a line is set.")
 
+        if line and column and text and not position:
```
AbdealiJK: We need an assert for a case when line, col and pos are given.

coalib/results/TextPosition.py
```diff
@@ -15,12 +19,21 @@ def __init__(self, line: (int, None)=None, column: (int, None)=None):
...
+            position = TextPosition.calc_position(text, line, column)
+
+        if (line and column) is None and (text and position) is not None:
+            line, column = TextPosition.calc_line_col(text, position)
```
AbdealiJK: We may want to do this "on demand". i.e. don't calculate it unless needed.
Basically this is a memory vs speed tradeoff of sorts.
Do we save `text` in memory and later calculate the type which is not given ? (Helps speed wise as we never need to find it sometimes)
or do we calculate it now and save up on the memory ? (Helps memory wise as we dont save it)

AbdealiJK: Based on gitter discussion, it would be better to do this lazily.

What I mean by "lazily" is that we should not do `calc_line_col` here. Rather, we should do it inside the `position` property. Basically `if self._position is None and (self.text is not None)` we calculate the position and return that.

This way, we don't do extra processing when `position` is not used at all.

@abhsag24 please confirm if you understand this

coalib/results/TextPosition.py
```diff
@@ -31,3 +31,30 @@ def line(self):
...
+        :param text:        A string with all the contents of file.
+        :param pos_to_find: Position of character to be found in the
+                            (line, column) form.
+
+        :return:            a tuple of the form (line, column).
+        """
+        line = 1
+        pos = -1
+        pos_new_line = text.find('\n')
+        while True:
```
AbdealiJK: Let's use tuple of lines everywhere.

coalib/results/TextPosition.py
```diff
@@ -31,3 +31,30 @@ def line(self):
...
+            if pos_new_line == -1:
+                return (line, pos_to_find - pos)
+
+            if pos_to_find <= pos_new_line:
+                return (line, pos_to_find - pos)
+
+            else:
+                line += 1
+                pos = pos_new_line
+                pos_new_line = text.find('\n', pos_new_line + 1)
```
Makman2: See above, would be cool if you return a `TextRange` object at the end

coalib/results/TextPosition.py
abhsag24: if the user doesn't want a position he wouldn;t specify the text in the `__init__` in this case an uknowing user may make a lot of calls to position and that might make it a huge overhead? are we sure about this?

coalib/results/TextPosition.py
abhsag24: correct me if i am wrong, but isn't `TextRange` for a range of text? here are two, almost unrelated values in the name of line and column ?

coalib/results/TextPosition.py
abhsag24: i don't follow?

coalib/results/TextPosition.py
```diff
@@ -31,3 +31,30 @@ def line(self):
     @property
     def column(self):
         return self._column
+
+    @classmethod
+    def calc_line_col(cls, text, pos_to_find):
```
abhsag24: but how do i make a text range from a single position?

coalib/results/TextPosition.py
Makman2: Eh I mean TextPosition sorry :3

coalib/results/TextPosition.py
AbdealiJK: I am not sure how to explain this in a better way so I'm just going to do it with code.

```
@property
def position(self):
    if self.text is None:
        return None
    if self._position is None:
        # _position is calculated lazily. i.e. ONLY when position is asked (for the first time)
        # If a user never tries to use `position` it is never computed.
        self._position = self.calc_position(self.text, self.line, self.column)
    return self._position
```

coalib/results/TextPosition.py
AbdealiJK: What I meant was let's not take `text` as a big string. Rather let's take it as a `tuple of lines` (Like how `file` in Bear is)
So, this would need to be like:
```
    @classmethod
    def calc_line_col(cls, text, pos_to_find):
        """
        Calculate line number and column in the text, from position.
        Uses \\n as the newline character. Lines and columns start from 1.

        :param text:        A tuple of lines.
        :param pos_to_find: Position of character to be found in the
                            (line, column) form.
        :return:            a tuple of the form (line, column).
        """
        line = col = 0
        for textline in text:
            new_position = position - len(line)
            if new_position < 0:
                break
            line += 1
            position = new_position
        col = position
        return line, col
```
PS: I have not checked for corner cases.

PPS: Now, in `calc_position` you don't need to do a `splitlines()` as both these functions use a list of lines and not a huge string chunk.

coalib/results/TextPosition.py
AbdealiJK: @Makman2 this is meant to be used inside TextPosition's `__init__` and hence I didn't understand what you're trying to say.
This is going to be used to find the `line, col` in a TextPosition if a `position` was used to create it.

coalib/results/TextPosition.py
abhsag24: @AbdealiJK  i do get it, i just want to be certain that this is the best way, so right now what you suggest is that we call `calc_position` when a user wants to use the `position` property, but what if an uknowing user makes multiple calls to the position property? it would call `calc_position` multiple times and would be a huge overhead, wouldn't calling it once if the text is specified better? it would only get called when the text is specified, meaning the user does wan't to use it, also sometimes the user might want to just store a position without the text, do we not consider those cases?

abhsag24: nvm got it! :P

coalib/results/TextPosition.py
Makman2: okay, sorry I haven't seen the modifications inside `__init__`. I thought this commit adds a function where it extracts the absolute position from the line/column from a given text (like `get_absolute_position(text)`).

coalib/results/TextPosition.py
Makman2: btw: http://stackoverflow.com/questions/3012421/python-lazy-property-decorator
maybe you can use that :+1:

sils1297: it's a oneliner anyway and not worth caching IMO

Makman2: though we shouldn't calculate it everytime, often we don't need the absolute position. A lazy property could handle such a behaviour^^ and caching would be nice, then you don't need to recalculate the stuff everytime when you need it more often^^ (storing a number does not consume much memory :O)

AbdealiJK: "oneliner" is a very wrong word in this context ... It passes the whole file atleast once :/

coalib/results/TextPosition.py
abhsag24: as discussed on gitter it'd be for when `line,col,pos and text` are given

coalib/results/TextPosition.py
abhsag24: aah i thought you meant that 58570fb has a better algo for this one :P , nvm got it!

coalib/results/TextPosition.py
abhsag24: Shouldn't we also do the same for the line and column properties if the text and position are given they can be calculated as well

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,69 @@ def line(self):
...
+    @classmethod
+    def from_line_col(cls, text, line, column, create_object=True):
+        """
+        Calculates position in a file from line number and column.
+        Uses \\n as the newline character. Position starts from 0.
+
+        :param text:   A tuple of lines in which position is to be calculated.
+        :param line:   The line through which position is identified.
+        :param column: Column through which position is idnetified.
+        :return:       Position in the text, identified by line, column
```
Uran198: Here, at the end of sentence, probably should be dot. And, maybe, `line and column` would be easier to read.

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,35 @@ def line(self):
...
+                            (line, column) form.
+        :create_object:     Boolean which determines whether to return
+                            a TextPosition Object or a tuple.
+        :return:            a TextPosition object or a tuple of the form
+                            (line, column).
+        """
+        line = 1
+        try:
+            while True:
+                textline = text[line - 1]
```
Uran198: It seems that only this line can produce an exception, so it'll be better to wrap only it with `try` clause. IMO it'll be easier to read, because you'll have to think less on what can produce an exception.

coalib/results/TextPosition.py
abhsag24: thanks

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,35 @@ def line(self):
...
         return self._column
+
+    @classmethod
+    def from_position(cls, text, position, create_object=True):
+        """
+        Makes a TextPosition object or a tuple containing (line, column)
+        by calculating line number and column in the text, from position.
+        Uses \\n as the newline character. Lines and columns start from 1.
+
+        :param text:        A string with all the contents of file.
```
Uran198: It seem that `text` is rather iterable with lines from file and not a string.

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,35 @@ def line(self):
...
+            while True:
+                textline = text[line - 1]
+                new_position = position - len(textline)
+                if new_position < 0:
+                    break
+                line += 1
+                position = new_position
+            col = position+1
+            if create_object:
+                return cls(line, col, position=position)
```
Uran198: Why would you need to pass a position here? Also, position seems to be outdated at this point.

coalib/results/TextPosition.py
abhsag24: yes but i do want the whole loop to execute don't i?

tests/results/TextPositionTest.py
```diff
@@ -30,3 +40,58 @@ def test_properties(self):
         uut = TextPosition(8, 39)
         self.assertEqual(uut.line, 8)
         self.assertEqual(uut.column, 39)
+
+
+class TextPositionLineColTest(unittest.TestCase):
+
+    def test_from_position(self):
```
Uran198: Maybe, it'll be better to separate this into several tests. Then if something breaks, you'll know exactly what it is just from looking at test results and going through them.

coalib/results/TextPosition.py
abhsag24: its characteristic of that object, so the user can call it by position or line, col, more power to the user

coalib/results/TextPosition.py
abhsag24: ahh yes, it used to be a string! :P , i'll change it thanks

coalib/results/TextPosition.py
abhsag24: yup

tests/results/TextPositionTest.py
abhsag24: it's the same function that's being tested, and it does give the exact line which failed? yeah there two many tests maybe i can seperate a few not sure though

coalib/results/TextPosition.py
Uran198: It'll stop executing on the first fail, so you wont see all the errors.

tests/results/TextPositionTest.py
Uran198: It'll stop executing on the first fail, so you wont see all the errors.

abhsag24: yeah but since it belongs to the same function i wouldn't mind if i don't get further tests on a faulty function?, i'm not saying that's always the case, but all of the tests here are very specific so it would break everything if even one fails :P

coalib/results/TextPosition.py
Uran198: In that case you will probably want to pass `text` as well.

tests/results/TextPositionTest.py
Uran198: It'll be more useful for someone modifying your function. If he breaks something, he'll see immediately that, for example he messed up something related with `create_object`.

SanketDG: I agree with @Uran198 here. The fact that they are specific is the reason why they should be separated.

coalib/results/TextPosition.py
SanketDG: No documentation for `create_object` needs to be added here.

coalib/results/TextPosition.py
```diff
@@ -7,20 +7,35 @@
...
+        :raises ValueError:     Raised when a column is set but line is None.
+        :raises AssertionError: Raised when line/column do not match
+                                corresponding text/position.
         """
         if line is None and column is not None:
             raise ValueError("A column can only be set if a line is set.")
 
+        if text and position and line and column:
+            assert (line, column) == TextPosition.from_position(
+                                            text, position, create_object=False)
```
sils1297: what `if not text and position`?

coalib/results/TextPosition.py
```diff
@@ -7,20 +7,35 @@
 class TextPosition:
 
     @enforce_signature
-    def __init__(self, line: (int, None)=None, column: (int, None)=None):
+    def __init__(self,
+                 line: (int, None)=None,
+                 column: (int, None)=None,
+                 text: (tuple, list, None)=None,
```
sils1297: text should never be none, right?

abhsag24: any reason? TextPosition holds position, so text can be None?

tests/results/TextPositionTest.py
abhsag24: @Uran198  yeah i get the last part will change it :P

coalib/results/TextPosition.py
abhsag24: there's no probem right? we've had people storing line and column with no reference. in that case we just trust the user :P ? like always?  we have no way of finding out if its correct anyways

abhsag24: we could stop them from storing positions? idk?

coalib/results/TextPosition.py
abhsag24: i guess so though i  didn't pass it since it'll call the function again, and do the same process twice(during assert) i wanted to get some insight on this

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,74 @@ def line(self):
...
+                            a TextPosition Object or a tuple.
+        :return:            a TextPosition object or a tuple of the form
+                            (line, column).
+        """
+        try:
+            position = 0
+            position = sum(len(phrase) for phrase in text[:line-1])
+            position += len(text[line-1][:column])-1
+            if create_object:
+                return cls(line, column, text=text, position=position)
```
abhsag24: I have it add the text param but i am not sure, since it'll make a redundant call to this function again because of this.

coalib/results/TextPosition.py
AbdealiJK: I agree. The user may just want to have `position` and no line nor col nor text

sils1297: if the user gives position only, there's no way the other information can be derived. The whole point of having an object is so we can rely on it doing certain things for us, you remove that value and the object *may* be of no use for us and we'd need to do checks for every information which sucks. We have an object which is unusable in some cases.

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,37 @@ def line(self):
...
+        by calculating line number and column in the text, from position.
+        Uses \\n as the newline character. Lines and columns start from 1.
+
+        :param text:        A A tuple/list of lines in which position is to
+                            be calculated.
+        :param pos_to_find: Position of character to be found in the
+                            (line, column) form.
+        :create_object:     Boolean which determines whether to return
+                            a TextPosition Object or a tuple.
+        :return:            a TextPosition object or a tuple of the form
```
AbdealiJK: Capital for the starting letter (Also, this `object` does  not start with capital, other does)

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,37 @@ def line(self):
...
+        """
+        Makes a TextPosition object or a tuple containing (line, column)
+        by calculating line number and column in the text, from position.
+        Uses \\n as the newline character. Lines and columns start from 1.
+
+        :param text:        A A tuple/list of lines in which position is to
+                            be calculated.
+        :param pos_to_find: Position of character to be found in the
+                            (line, column) form.
+        :create_object:     Boolean which determines whether to return
```
AbdealiJK: :param: ?

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,37 @@ def line(self):
...
+            try:
+                textline = text[line - 1]
+            except IndexError:
+                raise ValueError("Position not found in text")
+            new_position = position - len(textline)
+            if new_position < 0:
+                break
+            line += 1
+            position = new_position
+        col = position+1
```
AbdealiJK: space between `+`

coalib/results/TextPosition.py
```diff
@@ -31,3 +46,37 @@ def line(self):
...
         return self._column
+
+    @classmethod
+    def from_position(cls, text, pos_to_find, create_object=True):
+        """
+        Makes a TextPosition object or a tuple containing (line, column)
+        by calculating line number and column in the text, from position.
+        Uses \\n as the newline character. Lines and columns start from 1.
+
+        :param text:        A A tuple/list of lines in which position is to
```
AbdealiJK: You wrote "A" twice ?

coalib/results/TextPosition.py
```diff
@@ -48,6 +48,30 @@ def column(self):
...
+                            (line, column).
+        """
+        try:
+            position = 0
+            position = sum(len(phrase) for phrase in text[:line-1])
+            position += len(text[line-1][:column])-1
+            if create_object:
+                return cls(line, column, text=text, position=position)
+            return position
+        except IndexError:
```
AbdealiJK: This try/except is not needed for the whole block. Just for one line

coalib/results/TextPosition.py
```diff
@@ -48,6 +48,30 @@ def column(self):
...
     @classmethod
+    def from_line_col(cls, text, line, column, create_object=True):
+        """
+        Calculates position in a file from line number and column.
+        Uses \\n as the newline character. Position starts from 0.
+
+        :param text:   A tuple of lines in which position is to be calculated.
+        :param line:   The line through which position is identified.
+        :param column: Column through which position is idnetified.
+        :create_object:     Boolean which determines whether to return
```
AbdealiJK: :param: ?
And indentation seems wrong ...

coalib/results/TextPosition.py
```diff
@@ -48,6 +48,30 @@ def column(self):
...
+                            a TextPosition Object or a tuple.
+        :return:            a TextPosition object or a tuple of the form
```
AbdealiJK: Start with capital (Also, object is once with a capital O and the second time with a small O)

coalib/results/TextPosition.py
```diff
@@ -48,6 +48,30 @@ def column(self):
...
+                            (line, column).
+        """
+        try:
+            position = 0
```
AbdealiJK: Don't need to set position to 0 here

coalib/results/TextPosition.py
```diff
@@ -47,6 +47,19 @@ def line(self):
...
         return self._column
 
+    @property
+    def position(self):
+        if self._position is None:
+            # _position is calculated lazily. i.e. ONLY when position
+            # is asked (for the first time) If a user never tries to use
+            # `position` it is never computed.
+            if self._text is not None:
+                self._position = self.from_line_col(
```
AbdealiJK: I'm a little confused here (sorry).
If we have this, why do we calculate `position` in `from_line_col` with the arg `create_object=True` ?
It would be better to let this property calculate it rather than do the calculation before hand right ?
(Basically, if a user does `from_line_col(text, line, col)` and never uses `position` we unnecessarily calculate it ...)


Maybe we can just change that to:
```
 +    def from_line_col(cls, text, line, column, create_object=True):
 +        """
 +        Calculates position in a file from line number and column.
 +        Uses \\n as the newline character. Position starts from 0.
 +
 +        :param text:   A tuple of lines in which position is to be calculated.
 +        :param line:   The line through which position is identified.
 +        :param column: Column through which position is idnetified.
 +        :create_object:     Boolean which determines whether to return
 +                            a TextPosition Object or a tuple.
 +        :return:            a TextPosition object or a tuple of the form
 +                            (line, column).
 +        """
 +        if create_object:
 +            return cls(line, column, text=text)
 +        try:
 +            position = sum(len(phrase) for phrase in text[:line-1])
 +            position += len(text[line-1][:column])-1
 +        except IndexError:
 +            raise ValueError("Position not found in text")
 +        return position   
```

coalib/results/TextPosition.py
AbdealiJK: An object is also used for structured data. A text position is completely defined by just a position. Now a function can be made later which takes a TextPosition and only needs position (it doesn't really need line or column or text).

Hence its a valid use case which doesn't use all helper functions in the class - but it defines a TextPosition none the less.

-----Original Message-----

coalib/results/TextPosition.py
abhsag24: yes that's right , we shouldn'y calculate here

coalib/results/TextPosition.py
abhsag24: but so is the case with line and column right? if the user doesn't give a text, it has no meaning? only when the text is specified does this have meaning

coalib/results/TextPosition.py
abhsag24: done

coalib/results/TextPosition.py
sils1297: Before we had line and column and only that info was given. If you want an object with a traditional text position this is not the same information and would have to be an own class. If you want them to be convertible into each other you can give them conversion functions. But please *never* write a quantum state class!

abhsag24: hmm, so how about we never store text? just have conversion functions? , assert that line, col and position belong to the same character during instantiation?   but then we wont be able to calculate position, line and column, lazily, and the text would be parsed everytime?

sils1297: FWIW you can drive an object from text position that holds this info in
addition

abhsag24: that'd be better IMO i could never achieve the symmetry i would have liked with this thanks!

coalib/results/TextPosition.py
```diff
@@ -7,20 +7,35 @@
...
+        :param line:            The line in file or None, the first line is 1.
+        :param column:          The column indicating the character. The first
+                                one in a line is 1.
+        :param text:            The text containing the character.
+        :param position:        Position identifying the index of character
+                                in text.
+        :raises TypeError:      Raised when line or columns are no integers.
+        :raises ValueError:     Raised when a column is set but line is None.
+        :raises AssertionError: Raised when line/column do not match
+                                corresponding text/position.
```
sils1297: that should rather be a valueerror no?

sils1297:
hm I'm not bought on the whole "this object may have this property or not" thing... what was wrong with subclassing it? None's are always a problem...

abhsag24:
@sils1297  no problem at all :) ,I was just busy with other things today :P i'll push the subclassed version soon :P

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,58 @@
...
+class AbsolutePosition(TextPosition):
+
+    @enforce_signature
+    def __init__(self,
+                 text: (tuple, list, None)=None,
+                 position: (int, None)=None):
+        """
+        Creates an AbsolutePosition object that represents the index of a
+        character in a string.
+
```
abhsag24: what? weirdly coala gives no results for me?

AbdealiJK: Which version of coala do you have ? maybe this is a bug that has creeped in ? Can you test more ?

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,59 @@
...
+        self._text = text
+        self._position = position
+        super(AbsolutePosition, self).__init__(line, column)
+
+    @property
+    def position(self):
+        return self._position
+
+    @classmethod
+    def calc_line_col(cls, text, pos_to_find):
```
sils1297: why is this a classmethod? That doesn't make sense, it doesn't use `cls` and is pretty generic, should live outside, right?

abhsag24: yes

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,56 @@
...
+    character. Line and column start from 1.
+
+    :param text:          A tuple/list of lines in which position is to
+                          be calculated.
+    :param pos_to_find:   Position of character to be found in the
+                          (line, column) form.
+    :return:              A tuple of the form (line, column).
+    """
+    line = 1
+    position = pos_to_find
```
sils1297: why not call pos_to_find position in the first place?

abhsag24: i change the position, maybe i should interchange these two names

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def calc_line_col(text, position):
+    """
+    Creates a tuple containing (line, column) by calculating line number
+    and column in the text, from position. Uses \\n as the newline
+    character. Line and column start from 1.
+
+    :param text:          A tuple/list of lines in which position is to
+                          be calculated.
```
sils1297: does this contain `\n`s?

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,56 @@
...
+    :param position:      Position of character to be found in the
+                          (line, column) form.
```
sils1297: does the position include `\n` chars? I think it does right? So `"\n".join(text)[position:position+1]` should get that char right? Then we want position to start with 0?

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,56 @@
...
+            textline = text[line - 1]
+        except IndexError:
+            raise ValueError("Position not found in text")
+        new_position = pos_to_find - len(textline)
+        if new_position < 0:
+            break
+        line += 1
+        pos_to_find = new_position
+    col = pos_to_find + 1
+    return (line, col)
```
sils1297: Shouldn't something as simple as this suffice?

```python
for linenum, line in range(1, len(text)):
    linelen = len(line)
    if position < linelen:
       return linenum, position
    position -= linelen

raise ValueError("404 :)")
```

sils1297: be sure to test corner cases like `position == linelen` and so on

abhsag24: 404 really? :P

coalib/results/AbsolutePosition.py
abhsag24: position is basically the index, so yeah it does start with 0, although the algo doesn't use care if `\n`  is a newline character anymore, i should remove that

coalib/results/AbsolutePosition.py
abhsag24: it should, though the second item in the list will be treated as a newline regardless

coalib/results/AbsolutePosition.py
abhsag24: also how does `range` return the `line`? did you mean `enumerate`?

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @property
+    def position(self):
+        return self._position
+
+
+def calc_line_col(text, position):
+    """
+    Creates a tuple containing (line, column) by calculating line number
+    and column in the text, from position. Line and column start from 1.
```
sils1297: I think doctests would be helpful here especially to show that line and column start from one and other properties, maybe you'll be covering the function with that already even

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,48 @@
...
+        :param text:     The text containing the character.
+        :param position: Position identifying the index of character
+                         in text.
+        """
+        line = column = None
+        if position and text:
+            line, column = calc_line_col(text, position)
+        self._text = text
+        self._position = position
+        super(AbsolutePosition, self).__init__(line, column)
```
SanketDG: `super().__init__()` would work too?

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,55 @@
...
+        self._text = text
+        self._position = position
+        super().__init__(line, column)
+
+    @property
+    def position(self):
+        return self._position
+
+
+def calc_line_col(text, position):
```
SanketDG: Sorry if I missed the conversation, but why is this outside the class?

AbdealiJK: Because it is not related to the class (it doesn't use the class or any object of the class) so it has no reason to be in the class

SanketDG: uh okay

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Creates a tuple containing (line, column) by calculating line number
+    and column in the text, from position. Line and column start from 1.
+    Example:
```
sils1297: newline here, otherwise it wont be rendered correctly

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,55 @@
...
+    >>> text = ("a\\n",)
+
+    Now here we can see the index of 'a' in ''.join(text) is 0, so the
+    corresponding line and column of 'a' should be:
```
sils1297: same

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,57 @@
...
+        self._position = position
+        super().__init__(line, column)
+
+    @property
+    def position(self):
+        return self._position
+
+
+def calc_line_col(text, position):
+    """
```
sils1297: it'd be good to preceed the docstring with `r`  so you don't have to double escape the `\` which is a bit confusing for the example

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,57 @@
...
+    and column in the text, from position. Line and column start from 1.
+    Example:
+
+    >>> text = ("a\\n",)
+
+    Now here we can see the index of 'a' in ''.join(text) is 0, so the
+    corresponding line and column of 'a' should be:
+
+    >>> calc_line_col(text, 0)
+    (1, 1)
```
sils1297: I find it very important to give the user an example with e.g. the first char of the second line.  Also the definition of `text` in an own line/example is a bit verbose IMO, can we do something like this (properly phrased, I'm just dumping raw thoughts as text here):

```
The position starts with zero as you can see:

>>> calc_line_col(('a\n',), 0)
(1, 1)

Now if we enter a second line, the newline character has it's own position
and counts into the number:

>>> calc_line_col(('a\n', 'b\n'), 1)
(1, 2)
>>> calc_line_col(('a\n', 'b\n'), 2)
(2, 1)
```

Also for all inline code samples like `''.join(text)` you should surround those by double backticks

coalib/results/AbsolutePosition.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :param text:          A tuple/list of lines in which position is to
+                          be calculated.
+    :param position:      Position (starting from 0) of character to be found
+                          in the (line, column) form.
+    :return:              A tuple of the form (line, column).
```
sils1297: you should definitely mention right here that they start with 1 and not zero

sils1297:
https://github.com/coala-analyzer/coala/pull/1947/commits/bbf6dbbe85614d4a790ca6ba1f8dfb8db2c166ce needs work

coalib/results/AbsolutePosition.py
Uran198: IMO the name of the function is not very helpful. Maybe,something like `line_col_from_position`, or `to_2d_position`, or `position_2d` would be more helpful? Maybe even using a class to encapsulate `(line, col)` tuple? It would be more explicit, you'd know from the type that what you're working with is a position and not just 2 numbers.

Uran198: Oh, I see. It is solely used in the class, so it should probably,be it's "private" method. Why present it to the global name space? Even more it's so small and has only one invocation, that maybe it'll be better to just put this code into `__init__`?

sils1297: will this functionality is general purpose and shoule be outside the class IMO, better name might be possible though

Uran198: This function and the class above does the same thing. But class is named
and this function returns pair of numbers. It seems redundant to have both.
While inproducing to the class properties line and column will make code
more readable in the future.

0:41, ср, 06.04.2016 Lasse Schuirmann <notifications@github.com>:

abhsag24: @Uran198 it is small but it parses the entire file, which makes it very slow!, using it in `__init__` is not a good idea IMO

abhsag24: i am also open for changing the name, something shorter maybe. As far as encapsulating it into an object goes, this function is being used so that i can create a `TextPosition` in the first place :P

Uran198: @abhsag24 Whether it's slow or not, it gets called in the `__init__` anyway.

abhsag24: @Uran198  hmm i see, but doesn't writing it entirely in `__init__` make it less readable?, i don't know, just want to be sure

sils1297: IMO this function s useful on its own as well as the class. The function does the conversion and the class can be used to pass things around, they're not redundant and each one can be used individually if need.

Uran198: Well, the purpose of the class is to store the position and line with column derived from it, so any time you'll need a line and column, you can use class. Even more, `TextPosition` is basically the tuple containing line and column, so this function can be either method of that class which will construct an instance of `TextPosition` or part of the constructor.

abhsag24: well if we move the function into `__init__` we compromise the readability, and also the option of using it again somewhere else, i think this function can be used in a lot of places, for eg: creating a `TextPosition` object from position as you already suggested, creating `SourceRange` Objects from position and so on.

Uran198: I meant moving it into the constructor of `TextPosition`, or just making private member of it and just calling it from constructor. Anywhere you may want to use it, you'll just use `TextPosition`.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1947/commits/9d306075c35189133d3e9789d4b8e75be511ace5

sils1297:
@abhsag24 merge it if you didn't want to improve further, I think it's good

sils1297:
we can change the function later

abhsag24:
@sils1297 i want to make changes to `SourceRange` as well  i just wanted confirmation on wether this was the right aproach so far

abhsag24:
reack 428b6d6

abhsag24:
unack 428b6d6 line and column not calculated when position is 0.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1947/commits/95ae663cec9d96c277f3c1e3e6e023a56ec4e523 https://github.com/coala-analyzer/coala/pull/1947/commits/c7a5b41436e07a3c62cb3aaad6c1bd3b5fb7cb08 , merge at will

abhsag24:
@rultor merge :D

PR #1954, opened 2016-03-25
Narrow except clause when dealing with annotations

### LINKED ISSUE #1950: Annotations: handling exceptions
When writing custom annotation, user should be able to access exceptions, that it's raising, to debug his function. Currently all that he gets is the information that something went wrong, e. g.
```
[WARNING][14:09:44] The bear ClangComplexityBear cannot be executed. Unable to convert parameter 'max_complexity' into type <function func at 0x7fa1b4d12d90>
```
This can be pretty easy fixed, by specifying exception [here](https://github.com/coala-analyzer/coala/blob/master/coalib/settings/FunctionMetadata.py#L105) and specifying in the docs what exception should annotations be allowed to raise.

This is related to https://github.com/coala-analyzer/coala/issues/1219 discussion. I think, maybe we need to figure out a better way to use exceptions so that there will be no need in catching broad ones, because they can hide bugs or other useful information.

Uran198:
Fixes #1950

### DIFF
--- a/coalib/settings/FunctionMetadata.py
+++ b/coalib/settings/FunctionMetadata.py
@@ -102,7 +102,7 @@ class FunctionMetadata:
 
         try:
             return annotation(section[param])
-        except:
+        except (TypeError, ValueError):
             raise ValueError("Unable to convert parameter {} into type "
                              "{}.".format(repr(param), annotation))
 
--- a/docs/Users/Tutorials/Writing_Bears.rst
+++ b/docs/Users/Tutorials/Writing_Bears.rst
@@ -155,8 +155,9 @@ The Setting does support some very basic types:
 -  Dict of strings (``dict``, values will be split by comma and colon)
 
 If you need another type, you can write the conversion function yourself
-and use this function as the annotation. We've provided a few advanced
-conversions for you:
+and use this function as the annotation (if you cannot convert value, be
+sure to throw ``TypeError`` or ``ValueError``). We've provided a few
+advanced conversions for you:
 
 -  ``coalib.settings.Setting.path``, converts to an absolute file path
    relative to the file/command where the setting was set


coalib/settings/FunctionMetadata.py
```diff
@@ -102,7 +102,7 @@ def _get_param(param, section, annotation):
 
         try:
             return annotation(section[param])
-        except:
+        except TypeError:
```
abhsag24: are we sure this doesn't break anything on `coala-bears`?,

Uran198: I don't think that `coala` should be dependant on `coala-bears`, because if it is there is no point in keeping them in separate repositories.

abhsag24: @Uran198 you are right,  i mean't we should make the corresponding changes to `coala-bears`, if that is the case

sils1297: I think you also want to catch ValueError

```
>>> int("asd")
ValueError: invalid literal for int() with base 10: 'asd'
```

sils1297:
https://github.com/coala-analyzer/coala/pull/1954/commits/4a30aa47d394f4a75620f7dc87ce52e3c5e6a43b is ready

sils1297:
2df8e22 needs work

docs/Users/Tutorials/Writing_Bears.rst
```diff
@@ -155,8 +155,9 @@ The Setting does support some very basic types:
 -  Dict of strings (``dict``, values will be split by comma and colon)
 
 If you need another type, you can write the conversion function yourself
-and use this function as the annotation. We've provided a few advanced
-conversions for you:
+and use this function as the annotation (if you cannot convert value,
+be sure to throw ``TypeError``). We've provided a few advanced conversions
```
sils1297: or ValueError

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1954/commits/b077a75928a7a8a058a8eae40d79a44866c990b1

sils1297:
https://github.com/coala-analyzer/coala/pull/1954/commits/812ce2314fac5c56244f5fcde7082ce321bcd6a9 needs work

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1954/commits/1d15667f226dbc6264153a574c664dec7ce6e150

sils1297:
btw. @Uran198 can you run tests on the bears to check if anything breaks? Due to technical reasons we have to use `>=` for coala in coala-bears so we need to check this manually or builds will fail.

Uran198:
@sils1297 Everything seems to work on my machine.

sils1297:
@rultor merge

PR #1955, opened 2016-03-25
Section: Escape bear directories before globbing

hypothesist:
To prevent misinterpretation of bear directories as glob expressions
we need to escape these characters. Also, we need to add real paths
to `sys.path` instead.
    
Introduced an extra assertion in tests to validate if bear directories
work as expected by having a directory with an unintended glob
expression.

Also mention that bear directories cannot be globs. Instead they
must be comma separated values.
    
Fixes https://github.com/coala-analyzer/coala/issues/1932
Fixes https://github.com/coala-analyzer/coala/issues/1831

### DIFF
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -6,7 +6,8 @@ from collections import OrderedDict
 from coalib.collecting.Collectors import collect_registered_bears_dirs
 from coalib.misc.Decorators import enforce_signature, generate_repr
 from coalib.misc.DictUtilities import update_ordered_dict_key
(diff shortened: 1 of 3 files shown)


sils1297:
ack https://github.com/coala-analyzer/coala/pull/1955/commits/4e763b3f6f9f26301e15a4872f498bd6f99454f0 https://github.com/coala-analyzer/coala/pull/1955/commits/60d43cc6cbd2658cb54d6748fbd7d9c851532195

sils1297:
@rultor merge

PR #1959, opened 2016-03-25
ConfParser: Backslash escape only if it means something

hypothesist:
Convert keys and values in configurations to their raw form
with the exception of some characters. This means that a value like
`test\file` will now be interpreted as it is (not as `testfile`).
This is also the default Python behavior.

Fixes https://github.com/coala-analyzer/coala/issues/1933


coalib/parsing/StringProcessing/Core.py
```diff
@@ -328,6 +328,39 @@ def escape(string, escape_chars, escape_with="\\"):
...
+    Converts a string to its raw form, converting all backslash to double
+    backslash except when the backslash escapes a character given in
+    exceptions.
+
+    :param string:     The given string that needs to be converted
+    :param exceptions: A list of characters that if escaped with backslash
+                       should not be converted to double backslash.
+    :return:           Returns the corresponding raw string.
+    """
+    always_exceptions = ['\'', '\"', 'a', 'b', 'f', 'n', 'r', 't', 'v', '/']
```
sils1297: why `'a', 'b', 'f', 'n', 'r', 't', 'v',`?

sils1297: coala doesn't interpret `\t` ATM at least.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1959/commits/f22181c685c8c6e46359fae1f5a81d5d0337bf33

coalib/parsing/StringProcessing/Core.py
hypothesist: https://docs.python.org/2.0/ref/strings.html

These are the characters that are interpreted differently in Python. So I think we should maintain that consistency here as well.

sils1297: that would mean that the path `test\test` doesn't work as `\t` is an escape char in python, we shouldn't do that! It's inconsistent as we don't interpret `\t` as tab

hypothesist:
So @sils1297 are we going forward with this? I've been away for the last few days because of exams over here.

sils1297:
@hypothesist sorry I have an exam on tuesday and just starting preparing for it so this is usually a very intense phase for me

sils1297:
maybe @AbdealiJK or @fneu can help

coalib/parsing/StringProcessing/Core.py
sils1297: I'm not convinced that this solution is feasible and useful

tests/parsing/StringProcessing/ConvertToRawTest.py
```diff
@@ -0,0 +1,21 @@
...
+        test_data = [
+            ("test", "test"),
+            ("test_path", "test_path"),
+            ("test, path", "test, path"),
+            ("test\\ path", "test\\ path"),
+            ("test\\path", "test\\\\path"),
+            ("test\\\\path", "test\\\\path"),
+            ("test\\=path", "test\\=path"),
+            ("test=path", "test=path"),
+            ("value\\=as\\something", "value\\=as\\\\something")]
```
sils1297: use `r"some\=string"` so you don't have to double escape for readability

sils1297:
@hypothesist lets get this discussion going again please

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1992/commits/707550ae44c1e434cb827ce5406a11cdea46f456 https://github.com/coala-analyzer/coala/pull/1992/commits/e01e602b3981795ca55aba5ac66f714bcbd33584 https://github.com/coala-analyzer/coala/pull/1992/commits/111d4494f4d9357f4b73d82a7fcd49721c3f87f6

sils1297:
wrong PR but gitmate will handle it :P

coalib/parsing/LineParser.py
```diff
@@ -62,8 +62,16 @@ def parse(self, line):
         # Escapes in value might be needed by the bears
         keys, value = self.__extract_keys_and_value(line)
 
+        all_delimiters = self.key_value_delimiters
+        all_delimiters += self.key_delimiters
+        all_delimiters += self.comment_seperators
+        all_delimiters += self.section_override_delimiters
+        all_delimiters = "".join(all_delimiters)
```
sils1297: not content with this yet, some of the delimiters are dicts, some are lists, can you test delims of each group please and maybe a comment why this works on dicts and lists?

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1959/commits/2cdd097cdc5f1e28a6d44b602deb93d947577e02 https://github.com/coala-analyzer/coala/pull/1959/commits/2034ab7c1c0cb858d4100fc221ac1b39a0f7b6b3

sils1297:
https://github.com/coala-analyzer/coala/pull/1959/commits/23d8704629b3b1e7b14e5dee032d2d8939148aef needs work

coalib/parsing/LineParser.py
hypothesist: No, all four are tuples. So we can set `all_delimiters` to be `self.key_value_delimiters` first and then when we do `+= self.key_delimiters` next. This will create a new tuple concatenating `all_delimiters` and `self.key_delimiters`. And similarly the rest. So after line 68 `all_delimiters` is a tuple and then we stringify it next.

sils1297: I think then the keys and values from `section_name_surroundings` are missing too, right?

coalib/parsing/LineParser.py
```diff
@@ -62,8 +62,23 @@ def parse(self, line):
...
+
+        # Add all keys and values in section_name_surroundings, which is
+        # stored as a dict
+        all_delimiters += "".join(self.section_name_surroundings.keys())
+        all_delimiters += "".join(self.section_name_surroundings.values())
```
hypothesist: Added the keys and values from `section_name_surroundings`. Also added a comment for each group.

coalib/parsing/StringProcessing/Core.py
```diff
@@ -328,6 +328,38 @@ def escape(string, escape_chars, escape_with="\\"):
...
+    Converts a string to its raw form, converting all backslash to double
+    backslash except when the backslash escapes a character given in
+    exceptions.
+
+    :param string:     The given string that needs to be converted
+    :param exceptions: A list of characters that if escaped with backslash
+                       should not be converted to double backslash.
+    :return:           Returns the corresponding raw string.
+    """
+    exceptions += "/"
```
sils1297: forward slash?

coalib/parsing/StringProcessing/Core.py
```diff
@@ -328,6 +328,38 @@ def escape(string, escape_chars, escape_with="\\"):
     return string
 
 
+def convert_to_raw(string, exceptions=""):
```
sils1297: commit message should rather be `Add convert_to_raw`

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1959/commits/16f3a8eaa4025d97178405c33fc146c70e9d998d https://github.com/coala-analyzer/coala/pull/1959/commits/8f12ac5a521ff24d596e8af8f3dba126c49eb929

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1959/commits/778844908cc8bd0428c4788790421dda53ea6cc6

coalib/parsing/StringProcessing/Core.py
hypothesist: Yeah, it is never passed as an exception, but it should be. For example, consider the case where we `re.escape` a path. Then it becomes `\/tmp\/file` which should be interpreted as it is, not `\\/tmp\\/file`. But it is not a delimiter.

hypothesist: I'll add a comment saying the same.

sils1297: I still don't see why we need to escape the slash.

Sincerely,

Lasse Schuirmann

contact@viperdev.io
http://viperdev.io/

2016-04-11 13:08 GMT+02:00 Adhityaa Chandrasekar <notifications@github.com>:

hypothesist: Actually, on second thought, I think adding `/` to the list of exceptions would go better in the `LineParser: Convert keys and values to raw form` commit. This way we don't tamper with the function (in case somebody wants a `convert_to_raw` without forward slash?).

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -182,7 +182,7 @@ def test_back_saving(self):
             lambda *args: True,
```
hypothesist: Wanted to use `ConfigurationGatheringTest: Use StringProcessing.escape` for the commit message, but it would be too long.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1959/commits/adeb8082bd04b71374b90c86c60bfe31ec98311a https://github.com/coala-analyzer/coala/pull/1959/commits/3e03fbaeafec541401586f520e2b70f918656501 https://github.com/coala-analyzer/coala/pull/1959/commits/b33e6094c536a5b51a48635342af0c3bfd02c91c

sils1297:
awaiting co approval from @fneu before merge

sils1297:
@rultor merge

PR #1964, opened 2016-03-26
Annotations: Add function for chaining annotations

Uran198:
Fixes https://github.com/coala-analyzer/coala/issues/1859

### DIFF
--- a/coalib/misc/Annotations.py
+++ b/coalib/misc/Annotations.py
@@ -1,0 +1,53 @@
+
+
+def typechain(*args):
+    """
+    Returns function which applies the first transformation it can from args
+    and returns transformed value, or the value itself if it is in args.
+
+    >>> function = typechain(int, 'a', ord, None)
+    >>> function("10")
+    10
+    >>> function("b")
+    98
+    >>> function("a")
+    'a'
+    >>> function(int)
+    <class 'int'>
+    >>> function(None) is None
+    True
+    >>> function("str")
+    Traceback (most recent call last):
+        ...
+    ValueError: Couldn't convert value 'str' to any specified type or find it \
+in specified values.
+
+    :raises TypeError:  Raises when either no functions are specified for
+                        checking.
+    """
+    if len(args) == 0:
+        raise TypeError("No arguments were provided.")
+
+    def annotation(value):
+        """
+        Returns value either transformed with one of the function in args, or
+        casted to one of types in args, or the value itself if it is in the
+        args.
+
+        :raises ValueError: Raises when cannot transform value in any one of
+                            specified ways.
+        """
+        for arg in args:
+            if value == arg:
+                return value
+            if isinstance(arg, type) and isinstance(value, arg):
+                return value
+            try:
+                return arg(value)
+            except (ValueError, TypeError):
+                pass
+        raise ValueError(
+            "Couldn't convert value {value} to any specified type "
+            "or find it in specified values.".format(
+                value=repr(value)))
+    return annotation
--- a/tests/misc/AnnotationsTest.py
+++ b/tests/misc/AnnotationsTest.py
@@ -1,0 +1,66 @@
+from unittest import TestCase
+
+from coalib.misc.Annotations import typechain
+
+
+class AnnotationsTest(TestCase):
+
+    def test_empty(self):
+        with self.assertRaises(TypeError) as ctx:
+            typechain()
+        self.assertEqual(str(ctx.exception), "No arguments were provided.")
+
+    def test_with_lambda(self):
+        function = typechain(lambda x: int(x) > 0)
+        with self.assertRaises(ValueError):
+            function("str")
+        self.assertEqual(function("10"), True)
+
+    def test_with_function(self):
+        def positive(val):
+            val = int(val)
+            if val > 0:
+                return val
+            raise ValueError
+        function = typechain(positive, ord)
+        with self.assertRaises(ValueError):
+            function(0)
+        with self.assertRaises(ValueError):
+            function("str")
+        self.assertEqual(function("10"), 10)
+        self.assertEqual(function("0"), 48)
+
+    def test_with_function_without_arguments(self):
+        def dummy():
+            return 10
+        function = typechain(dummy)
+        with self.assertRaises(ValueError):
+            function(0)
+
+    def test_with_custom_type(self):
+        class Positive:
+
+            def __init__(self, val):
+                val = int(val)
+                if val > 0:
+                    self.val = val
+                else:
+                    raise ValueError
+
+        function = typechain(Positive, ord)
+        with self.assertRaises(ValueError):
+            function(0)
+        obj = function("10")
+        self.assertIsInstance(obj, Positive)
+        self.assertEqual(obj.val, 10)
+        self.assertEqual(function("0"), 48)
+
+    def test_with_empty_class(self):
+        class Dummy:
+            pass
+
+        function = typechain(Dummy)
+        with self.assertRaises(ValueError):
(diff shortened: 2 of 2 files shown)


tests/misc/AnnotationsTest.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def test_with_function_without_arguments(self):
+        def dummy():
+            return 10
+        function = typechain(dummy)
+        with self.assertRaises(ValueError):
+            function(0)
+
+    def test_with_custom_type(self):
+        class Positive:
```
Adrianzatreanu: running coala will fix this :p

abhsag24: hmm seeing the code it seems like this is alreaady applied, a bug?

Adrianzatreanu: resolved as per discussion on gitter

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,41 @@
+
+
+def typechain(*args):
+    """
+    Returns function which applies the first transformation it can from args
+    and returns transformed value, raises ValueError when cannot apply none.
+
+    :raises TypeError:  Raises when either no functions are specified for
+                        checking or one or more object isn't callable and not
+                        None.
```
sils1297: usually this goes below the doctest

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,41 @@
...
+
+    >>> function = typechain(int, ord)
+    >>> function("10")
+    10
+    >>> function("a")
+    97
+    >>> function("str")
+    Traceback (most recent call last):
+        ...
+    ValueError: Couldn't convert to any specified type value 'str'.
```
sils1297: neat :)

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,41 @@
...
+                return value
+            if isinstance(arg, type) and isinstance(value, arg):
+                return value
+            try:
+                return arg(value)
+            except (ValueError, TypeError):
+                pass
+        raise ValueError(
+            "Couldn't convert to any specified type value "
+            "'{value}'.".format(value=value))
```
sils1297: rather use `repr(value)` IMO

tests/misc/AnnotationsTest.py
```diff
@@ -0,0 +1,93 @@
...
+        with self.assertRaises(ValueError):
+            function("str")
+        self.assertEqual(function("10"), True)
+
+    def test_with_native_types(self):
+        function = typechain(int, ord)
+        with self.assertRaises(ValueError):
+            function("str")
+        self.assertEqual(function("10"), 10)
+        self.assertEqual(function("a"), 97)
```
sils1297: you could just delete this as this is already doctested right?

coalib/misc/Annotations.py
Uran198: Is there any benefit of doing so? If the function cannot convert it to specified type, I'm not sure you'll be interested in looking at the representation, though, maybe, it would be useful to show here the type of `value`.

sils1297: exactly `repr("asd")` will show `'asd'`, `repr(5)` will show `5` which is better than `'5'` right?

Uran198: I see. What about type? Would you find it beneficial to add here?

sils1297: repr may contain that info when it's suitable, using repr is what we do in such cases and each class should decide individually how it wants to be represented best

coalib/misc/Annotations.py
sils1297: you should add a None example, e.g. `function(ord)` will return ord, right?

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,44 @@
...
+    >>> function(int)
+    <class 'int'>
+    >>> function(None) is None
+    True
+    >>> function("str")
+    Traceback (most recent call last):
+        ...
+    ValueError: Couldn't convert to any specified type value 'str'.
+
+    :raises TypeError:  Raises when either no functions are specified for
```
sils1297: TypeError -> ValueError right?

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,44 @@
...
+                        checking or one or more object isn't callable and not
+                        None.
```
sils1297: ah sorry, you probably want a doc comment for `annotation` too I guess

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    if len(args) == 0:
+        raise TypeError("No arguments were provided.")
```
sils1297: actually my expectation would be that this'd generate a conversion function that always raises a TypeError

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,44 @@
...
+    if not all(arg is None or getattr(arg, '__call__', False) for arg in args):
+        raise TypeError("One of the provided argument is not callable.")
```
sils1297: not sure I agree, say I want to allow 3, 4 and None I'd probably do something like `typechain(3, 4, None)` right? Why limit the type matching to None?

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1964/commits/ce0e5931dde9e6f0a36239384efb004bb038e6cb

coalib/misc/Annotations.py
Uran198: Can you come up with the use case where this can be useful? For me, it looks like it would only delay someone's mistyping and, maybe, make it harder to debug.

sils1297: the types might be dynamically compiled as a list and if the list is empty that behaviour would be natural

Uran198: But, will it be the desired behaviour for someone to pass an empty list to `typechain` or unintended? In the first case, I'd agree that it's then OK, but still I cannot come up with situation when something like that would be intended. In the second case, it would be better to raise exception ASAP.

sils1297: well it's just a little bit more generic if you allow no arguments, it's less code that achieves the same thing ultimately. (Although I'm also a fan of fail fast)

Uran198: You'll still have to make this check in `annotation` I think. The ideas of error that you passed wrong value and error that you messed up while generating list of appropriate values are very different. Hence, they should be treated differently.

coalib/misc/Annotations.py
```diff
@@ -0,0 +1,53 @@
...
+        :raises ValueError: Raises when cannot transform value in any one of
+                            specified ways.
+        """
+        for arg in args:
+            if value == arg:
+                return value
+            if isinstance(arg, type) and isinstance(value, arg):
+                return value
+            try:
+                return arg(value)
```
sils1297: this shouldn't be attempted `if not isinstance(arg, type)` right?

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1964/commits/9582ccaf95e77f1b78ca0849c7f75ee3620c7346 but almost ready

coalib/misc/Annotations.py
Uran198: `arg` may be a function. And if it's not callable it will raise `TypeError`, so everything will be OK.

sils1297: indeed

sils1297:
ack 9582cca

sils1297:
@rultor merge

PR #1978, opened 2016-03-30
Constants: Import appdirs when needed

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/1957

### DIFF
--- a/coalib/output/dbus/BuildDbusService.py
+++ b/coalib/output/dbus/BuildDbusService.py
@@ -1,7 +1,7 @@
 from distutils.core import Command
 from distutils.errors import DistutilsOptionError
 
-from coalib.misc import Constants
+from coalib import BUS_NAME
 
 
 class BuildDbusService(Command):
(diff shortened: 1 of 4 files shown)


AbdealiJK:
It's also probably a better idea to read VERSION in `coalib/__init__.py` and use `coalib.__version__` in Constants.
This was version is isolated from other constants, which is actually the core problem.

sils1297:
agreed.

coalib/output/dbus/BuildDbusService.py
```diff
@@ -1,7 +1,7 @@
 from distutils.core import Command
 from distutils.errors import DistutilsOptionError
 
-from coalib.misc import Constants
+from coalib import BUS_NAME
```
Adrianzatreanu: shouldn't this be ``from coalib.__init__ import BUS_NAME`` ? or it works like this too?

sebix:
Yes, works.

coalib/output/dbus/BuildDbusService.py
sils1297: that's who you use the `__init__` so people can import from the "module"

Adrianzatreanu:
ack 906e869

Adrianzatreanu:
ce6061d is ready

Adrianzatreanu:
ack ce6061d

Adrianzatreanu:
@rultor merge

PR #1990, opened 2016-03-31
Makman2/linterv4

Makman2:
The new Linter v2 API is there :D

Just as note: Introducing the new Linter breaks API, because the old `Lint` gets removed.


coalib/misc/Decorators.py
```diff
@@ -245,7 +245,7 @@ def lt(self, other):
     return decorator
 
 
-def _assert_right_type(value, types, argname):
+def assert_right_type(value, types, argname):
```
sils1297: you should add a documentation comment

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/30988380a92913af225f854218bc856e4cfbac4e https://github.com/coala-analyzer/coala/pull/1990/commits/ada01c886b26d5775eb6cd7c2dbd2ef89de4db29

Makman2:
unack 8c327f4
commit body, the `merge()` function allows to merge an arbitrary number of metadatas^^

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/41af83c12ea48ccea52a3607753a7822501ef0dc https://github.com/coala-analyzer/coala/pull/1990/commits/7dc346b0131653a802c2bf91e967daad77c4db31 (commit message not ok)

sils1297:
Tests fail, some commits still unacked.

sils1297:
(combined with me having an exam :P)

Makman2:
Idea for regex checking: I should be able to trigger warnings when someone uses capturing-groups inside the regex `(...)` instead of non-capturing-groups `(?:...)` as they are faster, and the captured string is ignored anyway.

Makman2:
Ah yeah @sils1297, this works only with python >= 3.4, do we want to drop support for 3.3?
(I need to use `functools.partialmethod` which is new in 3.4^^)

coalib/misc/Future.py
```diff
@@ -0,0 +1,83 @@
+from functools import partial
+
+
+# Start ignoring all
```
sils1297: What?!

Makman2: As this is a copy and is not completely PEP8 conform ive deactivated all bears ^

sils1297: you should add reasoning to the comment! Never ignore without a reason.

coalib/misc/Future.py
```diff
@@ -0,0 +1,83 @@
...
+
+# FIXME: Remove `partialmethod` when dropping Python 3.3 support.
+# This code falls under the Python 3.4.3 license:
+# https://www.python.org/download/releases/3.4.3/license/
```
sils1297: what license is that? Is it AGPL compatible? -> Drop it

sils1297: It's gpl compatible, just mention where you copied it from

coalib/misc/Future.py
```diff
@@ -0,0 +1,83 @@
+from functools import partial
+
+
+# Start ignoring all
+
+# FIXME: Remove `partialmethod` when dropping Python 3.3 support.
```
sils1297: shouldn't you like import it when on py 3.4 or higher and only use your implementation in the other cases? Bugfixes might happen.

coalib/bearlib/abstractions/Lint.py
```diff
@@ -1,339 +0,0 @@
...
-
-from coalib.bears.Bear import Bear
-from coalib.misc.Decorators import enforce_signature
-from coalib.misc.Shell import escape_path_argument, run_shell_command
-from coalib.results.Diff import Diff
-from coalib.results.Result import Result
-from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
-
-
-class Lint(Bear):
```
sils1297: how about keeping legacy lint as a deprecated class and emit a warning on the debug channel upon execution? Then we don't break all bears right?

coalib/settings/FunctionMetadata.py
```diff
@@ -176,7 +176,7 @@ def merge(cls, *metadatas):
             Raised when less than 2 metadatas are provided.
         :return:
             A ``FunctionMetadata`` object containing the merged signature of
-            all given ones.
+            all given metadatas.
```
sils1297: plural of metadata should be metadata, without s

sils1297: this should go in the first commit btw.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/9a6bfaa98d4dd5d799f1d9a942bb7b6c7ca31c34

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,53 @@ def from_function(cls, func, omit=frozenset()):
...
+        if len(metadatas) < 2:
+            raise ValueError("Less than 2 metadata objects provided.")
+
+        merged_name = ("<Merged signature of " +
+                       ", ".join(repr(metadata.name)
+                                 for metadata in metadatas) +
+                       ">")
+
+        merged_desc = "\n".join("{}:\n{}".format(metadata.name, metadata.desc)
+                                for metadata in metadatas)
```
sils1297: not very convinced, so if I give several descriptions in process_.. etc you'll put process_... in front of the description which isn't nice. Shouldn't they simply be concatenated or just override themselves left to right?

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/42c97cfb699ba8609b1ce7ce68a5f010981acd35

sils1297:
ack 9a6bfaa

coalib/bearlib/abstractions/Lint.py
sils1297: at least remove it in a second commit after this one please for ease of review

Makman2: have already thought about it, will do :+1:
I want to delete it so noone gets encouraged to use the old class, as it's in fact buggy^^
But if you want I can only introduce the v2 Linter API, then we merge the adaptions on coala-bears, and then we delete the `Lint` class^^

coalib/settings/FunctionMetadata.py
Makman2: The descriptions are strings right? So I can't easily extract the param-contents (or other attributes).
But this was in fact a little "prototype", I wasn't completely sure how a clean "merge" function should look like^^ maybe we can discuss on gitter^^

coalib/bearlib/abstractions/Lint.py
SanketDG: Yes this seems sane. There are a lot of WIP bears right now, once we reach 0 PRs with old lint and all existing bears having new `Lint`, we can delete this.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/5b5846e37875a49aa626657a1cb1500fa2297581

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/6d36155ff83d669ec7801ce13ad33b8506ed44d9 as per phone discussion

sils1297:
fixing some scrutinizer issues needed as per discussion on phone

coalib/settings/FunctionMetadata.py
sils1297: you should give one of them preference, i.e. take the string of the leftmost param that has it defined

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/c786db90f4fc767d20666acb3a2d9b075b50891e

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+    http://coala.readthedocs.org/en/latest/Users/Tutorials/Linter_Bears.html
+
+    :param executable:
+        The linter tool.
+    :param use_stdin:
+        Whether the input file is sent via stdin instead of passing it over the
+        command-line-interface.
+    :param output_stream:
+        The output streams to grab from the executable. Possible values are
+        ``stdout``, ``stderr`` or ``stdout+stderr`` (or ``stderr+stdout``).
```
sils1297: why not make a bool for both?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+        ``output_regex`` with the named group ``severity``) to an actual
+        ``coalib.results.RESULT_SEVERITY`` for a result.
+
+        - ``RESULT_SEVERITY.MAJOR``
+          Mapped by ``error``, ``Error`` or ``ERROR``.
+        - ``RESULT_SEVERITY.NORMAL``
+          Mapped by ``warning``, ``Warning``, ``WARNING``, ``warn``, ``Warn``
+          or ``WARN``.
+        - ``RESULT_SEVERITY.MINOR``
+          Mapped by ``info``, ``Info`` or ``INFO``.
```
sils1297: why not map case insensitive?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+                       "prerequisite_check_command"}
+
+    output_stream_to_index_map = {"stdout": 0, "stderr": 1}
+    try:
+        stream_indices = []
+        for stream in options["output_stream"].split("+"):
+            stream_indices.append(output_stream_to_index_map[stream])
+        options["output_stream"] = tuple(stream_indices)
+    except KeyError:
+        raise ValueError("Invalid output stream: " + repr(stream))
```
sils1297: yeah... use bool

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+
```
sils1297: maybe oneline comments are good for those blocks

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+                                     diffs={filename: diff},
+                                     severity=diff_severity)
+
+            def process_output_regex(
+                    self,
+                    output,
+                    filename,
+                    file,
+                    output_regex,
+                    severity_map=MappingProxyType({
```
sils1297: what the heck is a MappingProxyType and why do we need one?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+                        "ERROR": RESULT_SEVERITY.MAJOR,
+                        "warning": RESULT_SEVERITY.NORMAL,
+                        "Warning": RESULT_SEVERITY.NORMAL,
+                        "WARNING": RESULT_SEVERITY.NORMAL,
+                        "warn": RESULT_SEVERITY.NORMAL,
+                        "Warn": RESULT_SEVERITY.NORMAL,
+                        "WARN": RESULT_SEVERITY.NORMAL,
+                        "info": RESULT_SEVERITY.INFO,
+                        "Info": RESULT_SEVERITY.INFO,
+                        "INFO": RESULT_SEVERITY.INFO})):
```
sils1297: let's really be case insensitive here...

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,690 @@
...
+                    Section settings passed from ``run()``.
+                :return:
+                    A context-manager handling the config-file.
+                """
+                content = cls.generate_config(filename, file, **kwargs)
+                if content is None:
+                    yield None
+                else:
+                    tmp_suffix = options["config_suffix"]
+                    with make_temp(suffix=tmp_suffix) as config_file:
```
sils1297: ah this also removes it on it's own huh? Neat.

sils1297:
like it!

https://github.com/coala-analyzer/coala/pull/1990/commits/c752fa1e5623a456928e548d1b5a1ca5968da329 needs work

coalib/bearlib/abstractions/Linter.py
Makman2: MappingProxyType wraps a dict so it gets readonly from outside^^

sils1297: can't we use a plain and simple dict?

Makman2: It's a mutable default parameter then^^

sils1297: ah ok

Makman2:
damn forgot scrutinizer...

coalib/bearlib/abstractions/Linter.py
Makman2: there's sadly no frozendict (except you get the third-party module), Python suggest to use this type if you want a readonly wrapper^^

Makman2:
reack 6325da3

sils1297:
aaah neat! https://github.com/coala-analyzer/coala/pull/1990/commits/6098fce9f44f556af0bf7900580b7300b0e3c00c is ready

coalib/settings/FunctionMetadata.py
sils1297: hey, you didn't address this comment

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/df4b8cbaacec2bc64baf0f004f5443cde07d4160 as per previous unaddressed review

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,684 @@
...
+
+    If you override ``process_output``, you have the same feature like above
+    (auto-forwarding of the right arguments defined in your function
+    signature).
+
+    Note when overriding ``process_output``: Providing a single output stream
+    (via ``use_stdout`` or ``use_stderr``) puts the according string attained
+    from the stream into parameter ``output``, providing both output streams
+    inputs a tuple with ``(stdout, stderr)``. Generally providing no stream
+    raises a ``ValueError``.
```
sils1297: wouldn't it be wise to have stdout by default?

Makman2: `use_stdout` is `True` by default, what I mean is if you do this:
```
use_stdout = False
use_stderr = False
```

Makman2: maybe I should rephrase...

sils1297: where do I see the defaults?

Makman2: function signature^^

Makman2:
/me : Use `options` completely for default parameters: Don't use partialmethods to populate parameters on demand, directly use `param=options["default"]`

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,688 @@
...
+
+                    process_output = partialmethod(
+                        process_output_regex, **process_output_args)
+
+                else:  # pragma: no cover
+                    # This statement is never reached.
+                    # Due to a bug in coverage we can't use `pass` here, as
+                    # the ignore-pragma doesn't take up this else-clause then.
+                    # https://bitbucket.org/ned/coveragepy/issues/483/partial-
+                    # branch-coverage-pragma-no-cover
```
sils1297: remove the comment, assert false is better for a statement that shouldn't be reached anyway

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,688 @@
...
+                    The signature specification.
+                :param kwargs:
+                    The kwargs to filter.
+                :return:
+                    The filtered kwargs.
+                """
+                return {key: kwargs[key]
+                        for key in metadata.non_optional_params.keys() |
+                        metadata.optional_params.keys()
+                        if key in kwargs}
```
sils1297: this should rather be a member function of FunctionMetadata, can you move it?

Makman2: sure :+1:

Makman2:
reack a432bc954b3b63d759630ae4a2af0b5eaac109c5

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,49 @@ def from_function(cls, func, omit=frozenset()):
...
+            Raised when less than 2 metadatas are provided.
+        :return:
+            A ``FunctionMetadata`` object containing the merged signature of
+            all given metadatas.
+        """
+        # Collect the metadatas, as we operate on them more often and we want
+        # to support arbitrary sequences.
+        metadatas = tuple(metadatas)
+        if len(metadatas) < 2:
+            raise ValueError("Less than 2 metadata objects provided.")
```
sils1297: this should work very well with one metadata object shouldn't it? Or zero. Less special cases for outsiders to handle.

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,49 @@ def from_function(cls, func, omit=frozenset()):
...
+
+        merged_name = ("<Merged signature of " +
+                       ", ".join(repr(metadata.name)
+                                 for metadata in metadatas) +
+                       ">")
+
+        merged_desc = metadatas[-1].desc
```
sils1297: nah you gotta take the lastmost *nonempty* description!

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,49 @@ def from_function(cls, func, omit=frozenset()):
...
+        merged_retval_desc = metadatas[-1].retval_desc
```
sils1297: same here

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/0fcecb364fcfcb4138d7ef7de142bdaa7dffb447

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/2880468088c94f38e1188a495be3536f02631c3c

coalib/settings/FunctionMetadata.py
```diff
@@ -161,6 +161,21 @@ def from_function(cls, func, omit=frozenset()):
...
+        this metadata (either optional or non-optional).
+
+        :param dct:
+            The dict to filter.
+        :return:
+            A new dict containing the filtered items.
+        """
+        return {key: dct[key]
+                for key in self.non_optional_params.keys() |
+                self.optional_params.keys()
```
sils1297: I think this self should be aligned with the one above, maybe make brackets around them? They belong together. Romantic.

coalib/settings/FunctionMetadata.py
```diff
@@ -161,6 +161,21 @@ def from_function(cls, func, omit=frozenset()):
                    optional_params=optional_params,
                    omit=omit)
 
+    def filter_parameters(self, dct):
+        """
+        Filters the given dict for keys that are declared as parameters inside
+        this metadata (either optional or non-optional).
+
```
sils1297: I think it'd be good to have a doctest example here:

```
blablabla, like this:

>>> def function(a, b=0):
...     pass

Now you can use this function to pass only the needed parameters to that function:

>>> args = FunctionMetadata.from_function(function).filter_parameters({'a': 1, 'b': 2, 'c': 3})

You can safely pass the args to the function now:

>>> function(**args)
```

sils1297:
https://github.com/coala-analyzer/coala/pull/1990/commits/32aa92e82fbf71349566495d98343b899fdf257a needs a little work

sils1297:
https://github.com/coala-analyzer/coala/pull/1990/commits/32aa92e82fbf71349566495d98343b899fdf257a needs work

tests/settings/FunctionMetadataTest.py
```diff
@@ -107,6 +107,25 @@ def check_function_metadata_data_set(self,
...
+
+        self.assertEqual(
+            metadata.filter_parameters(
+                {"x": 50, "y": 60, "NOT_THERE": -100, "c": 20}),
+            {"x": 50, "c": 20})
+
+        self.assertEqual(
+            metadata.filter_parameters({"a": 50, "b": 60, "c": 70}),
+            {"a": 50, "b": 60, "c": 70})
+
```
sils1297: drop this, the doctest is sufficient and better documented and easier to understand. And less redundant.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,673 @@
...
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def _prepare_options(options):
+    """
+    Prepares options for ``Linter`` for a given options dict in-place.
+
+    :param options:
+        The options dict that contains user/developer inputs.
+    """
```
sils1297: /me sees RadonBear raising a huge issue for this function as soon as we activate that one...

coalib/settings/FunctionMetadata.py
Makman2: ah this example is better, missed the code ...

coalib/bearlib/abstractions/Linter.py
Makman2: what do you mean?

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,46 @@ def from_function(cls, func, omit=frozenset()):
...
+    @classmethod
+    def merge(cls, *metadatas):
+        """
+        Merges signatures of ``FunctionMetadata`` objects.
+
+        Parameter (either optional or non-optional) and non-parameter
+        descriptions are merged from left to right, meaning the right hand
+        metadata overrides the left hand one.
+
```
sils1297: hm one doctest example showing a few cases might be really helpful

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/f3e78c66e29134abfcafdd658009da1522cd3869

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/c88e74024b8ce5f06516b686c3b6dabb25b9dbe1

tests/settings/FunctionMetadataTest.py
```diff
@@ -107,6 +107,21 @@ def check_function_metadata_data_set(self,
...
+             "y": ("-2", None),
+             "z": ("-3", None),
+             "w": ("-4", None)},
+            {"a": ("1", None, 1), "b": ("2", None, 2), "c": ("3", None, 3)},
+            {"y", })
+
+        self.assertEqual(
+            metadata.filter_parameters(
+                {"x": 50, "y": 60, "NOT_THERE": -100, "c": 20}),
+            {"x": 50, "c": 20})
```
sils1297: remove the whole test case please

sils1297: redundant tests are really evil

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/2981fbecab2df273386b309972673edcc793d5c6

coalib/bearlib/abstractions/Linter.py
sils1297: this function is huge and complicated

AbdealiJK: Hehehe, have you seen the function `Linter()` XD

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+    # Check for illegal superfluous options.
+    superfluous_options = options.keys() - allowed_options
+    if superfluous_options:
+        raise ValueError(
+            "Invalid keyword arguments provided: " +
+            ", ".join(repr(s) for s in sorted(superfluous_options)))
+
+
+@enforce_signature
+def Linter(executable: str,
```
AbdealiJK: I remember mentioning earlier that it should be lowercase `linter()` as it's a function
Was there a discussion where it was decided to be upper ?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                    omit={"filename", "file"})
+
+            @classmethod
+            def _get_process_output_metadata(cls):
+                return FunctionMetadata.from_function(
+                    cls.process_output,
+                    omit={"self", "output", "filename", "file"})
+
+            @classmethod
+            def get_non_optional_settings(cls):
```
AbdealiJK: On doing `ctrl+F` it says this is not used anywhere (except in the test).
Do we need it ?

coalib/bearlib/abstractions/Linter.py
AbdealiJK: You could actually do a tuple of tuples:

```
In [30]: dict((("a", "b"), ("c", "d")))
Out[30]: {'a': 'b', 'c': 'd'}
```
And do `dict()` inside the function. This would convert any derived type into a `dict` and throw a TypeError when it can't - which makes it easy to read + flexible IMO

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+
+                for string in output:
+                    for match in re.finditer(output_regex, string):
+                        yield self._convert_output_regex_match_to_result(
+                            match, filename, severity_map=severity_map)
+
+            if options["output_format"] is None:
+                # Check if user supplied a `process_output` override.
+                if not (hasattr(klass, "process_output") and
+                        callable(klass.process_output)):
```
AbdealiJK: Consider: `not callable(getattr(klass, 'process_output', None))`
which I think is easier to read plus we don't have to use `process_output` twice.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                    raise ValueError("`process_output` not provided by given "
```
AbdealiJK: Tiny change of `"process_output not provided by given class {}".format(<the class name>)` would be much appreciated by debuggers
As the traceback will point to the file `Linter.py` and they;d have to dig around in the traceback to find the actual file

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                                     "class.")
+                # No need to assign to `process_output` here, the class mixing
+                # below automatically does that.
+            else:
+                # Prevent people from accidentally defining `process_output`
+                # manually, as this would implicitly override the internally
+                # set-up `process_output`.
+                if hasattr(klass, "process_output"):
+                    raise ValueError("`process_output` is used by given class,"
```
AbdealiJK: `Found `process_output()` defined and `output_format={}` in class {}..format(<output_format>, <class_name>)`

Seems more concise and easier to read

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                    process_output_args = {}
+                    if "diff_severity" in options:
+                        process_output_args["diff_severity"] = (
+                            options["diff_severity"])
+                    if "diff_message" in options:
+                        process_output_args["diff_message"] = (
+                            options["diff_message"])
+
+                    process_output = partialmethod(
+                        process_output_corrected, **process_output_args)
```
AbdealiJK: Easier to read:

```
process_output_args = {
    k:v for k, v in options.items()
        if k in ("diff_message", "diff_severity")}
```
 - Removes writing "diff_message" 3 times
 - Removed multiple if conditions

BTW: You seem to be doing similar things freqquently (Like L285). Consider moving this into a function like `splice_dict(dictionary, keys)` or so ?

AbdealiJK: Maybe `splice_dict(dictionary, keys, required_keys=())` ? To also be usable at L285 and L599

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                    if "severity_map" in options:
+                        process_output_args["severity_map"] = (
+                            options["severity_map"])
+
+                    process_output = partialmethod(
+                        process_output_regex, **process_output_args)
+
+                else:  # pragma: no cover
+                    # This statement is never reached.
+                    assert False
```
AbdealiJK: Actually, it can be reached if a invalid output_format is given ...
So: `raise ValueError("Invalid value for output_format was found")` makes more sense ?

coalib/bearlib/abstractions/Linter.py
AbdealiJK: `tmp_suffix` is defined and used only once. Unnecessarily creating a new variable

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+            def run(self, filename, file, **kwargs):
+                # Get the **kwargs params to forward to `generate_config()`
+                # (from `_create_config()`).
+                generate_config_kwargs = FunctionMetadata.filter_parameters(
+                    self._get_generate_config_metadata(), kwargs)
+
+                with self._create_config(
+                        filename,
+                        file,
+                        **generate_config_kwargs) as config_file:
```
AbdealiJK:
```
 +                with self._create_config(filename, file,
 +                        **generate_config_kwargs) as config_file:
```

Sorry, but this file is huge enough ... every line counts !

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                                file,
+                                string.splitlines(keepends=True)).split_diff():
+                        yield Result(self,
+                                     diff_message,
+                                     affected_code=(diff.range(filename),),
+                                     diffs={filename: diff},
+                                     severity=diff_severity)
+
+            def process_output_regex(
+                    self,
```
AbdealiJK: Again can we simply do
```
def process_output_regex(self, output, filename, file,
    output_regex, severity_map...):
```
Make it easier to read (lesser lines)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+    options["output_format"] = output_format
+    options["use_stdin"] = use_stdin
+    options["use_stdout"] = use_stdout
+    options["use_stderr"] = use_stderr
+    options["config_suffix"] = config_suffix
+    options["prerequisite_check_command"] = prerequisite_check_command
+
+    _prepare_options(options)
+
+    def create_linter(klass):
```
AbdealiJK: Can we just declare this function outside ?
Would make the whole thing reduce by 4 spaces, making it less cramped

So it'd be:

```
def create_linter(klass, options):
    # The whole kitchen sink
    # ...

def Linter(....):
    # Options stuff
    _prepare_options(...)
    return functools.partial(create_linter, options=options)
```
or similar

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                        output = output[0]
+
+                    process_output_kwargs = FunctionMetadata.filter_parameters(
+                        self._get_process_output_metadata(), kwargs)
+                    return self.process_output(output, filename, file,
+                                               **process_output_kwargs)
+
+            def __repr__(self):
+                return "<{} linter object (wrapping {})>".format(
+                    type(self).__name__, repr(options["executable"]))
```
AbdealiJK: Use `self.get_executable()` here instead ?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                       "use_stderr",
+                       "config_suffix",
+                       "prerequisite_check_command"}
+
+    if not options["use_stdout"] and not options["use_stderr"]:
+        raise ValueError("No output streams provided at all.")
+
+    if options["output_format"] == "corrected":
+        if "diff_severity" in options:
+            if options["diff_severity"] not in RESULT_SEVERITY.reverse:
```
AbdealiJK: You have 2 cascaded `if` statements. Consider using `and`
OR
A much easier formulation:
```
if options.get("diff_severity", None) not in RESULT_SEVERITY.reverse:
```

AbdealiJK: You could make this into a function like `assert_right_type()` called `assert_in(val, iterable)`.
Used here and L78

coalib/misc/Future.py
```diff
@@ -0,0 +1,83 @@
+from functools import partial
```
AbdealiJK: Put this in it's own commit please !

coalib/bearlib/abstractions/Linter.py
AbdealiJK: Just realized it's already checked in `_process_output`.

Either repeat the error here (if you think it'd be called even though it's checked in _process_output)
Or remove it completely.

`assert False` doesn't help much. Better to have a proper debug statement

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                raise TypeError("Invalid value for `diff_severity`: " +
+                                repr(options["diff_severity"]))
+
+        if "diff_message" in options:
+            assert_right_type(options["diff_message"], str, "diff_message")
+
+        allowed_options |= {"diff_severity", "diff_message"}
+    elif options["output_format"] == "regex":
+        if "output_regex" not in options:
+            raise ValueError("No `output_regex` specified.")
```
AbdealiJK: `output_regex` needed when output_format is given as "regex"

Is a better message ?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                raise ValueError("Provided `severity_map` but named group "
+                                 "`severity` is not used in `output_regex`.")
+            assert_right_type(options["severity_map"], dict, "severity_map")
+
+            for key, value in options["severity_map"].items():
+                try:
+                    assert_right_type(key, str, "<severity_map dict-key>")
+                except TypeError:
+                    raise TypeError("The key " + repr(key) + " inside given "
+                                    "severity-map is no string.")
```
AbdealiJK: Um, why not do:
`assert_right_type(key, str, "severity-map's key " + repr(key))`

instead of this try-catch ?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+
+            # Auto-convert keys to lower-case. This creates automatically a new
+            # dict which prevents runtime-modifications.
+            options["severity_map"] = {
+                key.lower(): value
+                for key, value in options["severity_map"].items()}
+
+        allowed_options |= {"output_regex", "severity_map"}
+    elif options["output_format"] is not None:
+        raise ValueError("Invalid `output_format` specified.")
```
AbdealiJK: Rather than doing this, preferably use the `assert_in` which I'd mentioned above. Makes it cleaner ? (Here I need to go up and check all the values it can take in the `if`s - which is slightly irritating)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+            def get_metadata(cls):
+                merged_metadata = FunctionMetadata.merge(
+                    cls._get_process_output_metadata(),
+                    cls._get_generate_config_metadata(),
+                    cls._get_create_arguments_metadata())
+                merged_metadata.desc = inspect.getdoc(cls)
+                return merged_metadata
+
+            @classmethod
+            def _execute_command(cls, args, stdin=None):
```
AbdealiJK: Do we really need this ? It's a function that calls and returns another function with minor modifications.
And the function is used only once (and I cannot think of any scenario where the bear writer would need to overwrite this)

It just consumes an unnecessary documentation comment which is fairly easy to understand already

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                # Pre process the groups
+                groups = match.groupdict()
+
+                try:
+                    groups["severity"] = severity_map[
+                        groups["severity"].lower()]
+                except KeyError:
+                    self.warn(
+                        "No correspondence for " + repr(groups["severity"]) +
+                        " found in given severity map. Assuming "
```
AbdealiJK: `repr(groups["severity"]) + "Not found in severity-map.` is simpler to read

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+
+                for variable in ("line", "column", "end_line", "end_column"):
+                    groups[variable] = (None
+                                        if groups.get(variable, "") == "" else
+                                        int(groups[variable]))
+
+                if "origin" in groups:
+                    groups["origin"] = "{} ({})".format(
+                        str(klass.__name__),
+                        str(groups["origin"]))
```
AbdealiJK: This line can be moved above (after the comma)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+
+                # Construct the result.
+                return Result.from_values(
+                    origin=groups.get("origin", self),
+                    message=groups.get("message", ""),
+                    file=filename,
+                    severity=int(groups.get("severity",
```
AbdealiJK: I think L446 ensures that the severity will definitely have a value and this default value of NORMAL is not needed

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,680 @@
...
+                    An iterator returning results containing patches for the
+                    file to correct.
+                """
+                if isinstance(output, str):
+                    output = (output,)
+
+                for string in output:
+                    for diff in Diff.from_string_arrays(
+                                file,
+                                string.splitlines(keepends=True)).split_diff():
```
AbdealiJK: Why is this `splitlines()` done ? Isn't output already a list of lines, making `string` a single line ?

coalib/misc/Future.py
AbdealiJK: Just realized that I'm commenting on the whole diff `>_>`
Ignore this

coalib/bearlib/abstractions/Linter.py
Makman2: It does help as if this one occurs there's definitely a fatal logic error. An AssertionError ensures we doesn't catch it inside coala^^

sils1297: @Makman2 can't you just remove the whole else branch?

sils1297: you'll rather want to put an else in 598

coalib/bearlib/abstractions/Linter.py
Makman2: I think inlining teh generator is better, this is more a specific task. But this one is better I believe:
```
process_output_args = {key: options[key] for key in ("diff_message", "diff_severity") if key in options}
```

coalib/bearlib/abstractions/Linter.py
Makman2: cool :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: yeah this is good

coalib/bearlib/abstractions/Linter.py
Makman2: :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: but this still allows to modify the default argument in some place. I know `MappingProxyType` is quite unreadable, though it's really the preferred way if you want a readonly proxy for a dict, I remember such a discussion on the python issue tracker^^

coalib/bearlib/abstractions/Linter.py
Makman2: shouldn't i be aligned like this then:
```python
def process_output_regex(self, output, filename, file,
                         output_regex, severity_map=...):
```
and then severity map doesn't fit that nicely :D

coalib/bearlib/abstractions/Linter.py
Makman2: nope output is now a plain string

coalib/bearlib/abstractions/Linter.py
Makman2: yes thanks

coalib/bearlib/abstractions/Linter.py
Makman2: :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: yeah I think I can inline it into `run`

coalib/bearlib/abstractions/Linter.py
Makman2: yes we need it, it's used by our bear infrastructure to get information on which settings need to be provided
See `Bear.py`

coalib/bearlib/abstractions/Linter.py
Makman2: nope we can't since I use the `options` variable that only exists inside the `def Linter(...)` scope

coalib/bearlib/abstractions/Linter.py
Makman2: I wanted to change that, but somehow I recognized that it makes the class it decorates a new "Linter" class, that's why I uppercased it to resemble some new "class generator". Hope you understand what I mean :D

coalib/bearlib/abstractions/Linter.py
Makman2: But this is less redundant, I need the if-s anyway. When I need to assert with `in`, I need to predefine an allowed set out output formats above and another checking line.

coalib/bearlib/abstractions/Linter.py
Makman2: Because the error message from `assert_right_type` would be misleading then, as a key inside the severity_map is no parameter^^

coalib/bearlib/abstractions/Linter.py
Makman2: :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: I don't think creating a function extra for that is needed^^

coalib/bearlib/abstractions/Linter.py
Makman2: :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: but this is just codestyle :O
and I like mine more than yours :/

coalib/bearlib/abstractions/Linter.py
Makman2: yes that's actually better

coalib/settings/FunctionMetadata.py
Makman2: :+1:

tests/settings/FunctionMetadataTest.py
Makman2: this is not redundant, it does way more testing than the doctest. It accounts also for omits^^

sils1297: should be easy to change doctest to implicitly test omits with only a few additional chars

coalib/bearlib/abstractions/Linter.py
Makman2: And btw it must be 
```
"diff_severity" in options and
options["diff_severity"] not in RESULT_SEVERITY.reverse)
```
otherwise the body can't access `diff_severity` if it's not inside the options dict :)

tests/settings/FunctionMetadataTest.py
```diff
@@ -106,3 +106,61 @@ def check_function_metadata_data_set(self,
...
+
+    def test_filter_parameters(self):
+        metadata = FunctionMetadata(
+            "A", "DESC", "RETURNS SOMETHING",
+            {"x": ("-1", None),
+             "y": ("-2", None),
+             "z": ("-3", None),
+             "w": ("-4", None)},
+            {"a": ("1", None, 1), "b": ("2", None, 2), "c": ("3", None, 3)},
+            {"y", })
```
sils1297: testing omit functionality is superfluous because this is an unrelated property that is handled at a lower level. It makes your testcase overly complicated.

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,72 @@ def from_function(cls, func, omit=frozenset()):
...
+        >>> def a(x):
+        ...     '''
+        ...     desc of *a*
+        ...     :param x: x of a
```
sils1297: can you add another param here to show that it not only takes values from the `b` function?

coalib/settings/FunctionMetadata.py
```diff
@@ -160,3 +160,72 @@ def from_function(cls, func, omit=frozenset()):
...
+
+        merged_desc = next((m.desc for m in reversed(metadatas) if m.desc), "")
+        merged_retval_desc = next(
+            (m.retval_desc for m in reversed(metadatas) if m.retval_desc), "")
+        merged_non_optional_params = {}
+        merged_optional_params = {}
+
+        for metadata in metadatas:
+            merged_non_optional_params.update(metadata.non_optional_params)
+            merged_optional_params.update(metadata.optional_params)
```
sils1297: you merge the omit, for the optional and non optional params you shouldn't use the properties but the private variables!

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/d065796578cdddbd3ac7be67b2ff439a675f51de

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/32f3a4cb6c5b6cfa451e528c95f76dbc2682ef6a

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/c036c15dcca15968af05518440e7445efedfbabf

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/9c427a4aba9573288892261337abdf5de514ec99

coalib/bearlib/abstractions/Linter.py
sils1297: actually I agree, a decorator is always a function and should start lowercase. How much work is it to change that? I assume like 10 mins because of all the commits on coala-bears?

sils1297: hm I'm not fully sure though... let's go the way of less work and take it as it is

coalib/bearlib/abstractions/Linter.py
sils1297: can't you just inherit from Bear and get it for free?

sils1297: yeah the function should be already there, inherited

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,651 @@
...
+                # Construct the result.
+                return Result.from_values(
+                    origin=groups.get("origin", self),
+                    message=groups.get("message", ""),
+                    file=filename,
+                    severity=groups["severity"],
+                    line=groups["line"],
+                    column=groups["column"],
+                    end_line=groups["end_line"],
+                    end_column=groups["end_column"])
```
sils1297: can you add a default field for `additional_info`?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,651 @@
...
+                else:
+                    assert options["output_format"] == "regex"
+
+                    process_output_args = {
+                        key: options[key]
+                        for key in ("output_regex", "severity_map")
+                        if key in options}
+
+                    process_output = partialmethod(
+                        process_output_regex, **process_output_args)
```
sils1297: can you put all this in a function `validate_class_overrides(options, klass)`?

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -0,0 +1,711 @@
...
+        with self.assertRaises(ValueError) as cm:
+            Linter("some-executable", invalid_arg=88, ABC=2000)
+        self.assertEqual(
+            str(cm.exception),
+            "Invalid keyword arguments provided: 'ABC', 'invalid_arg'")
+
+        with self.assertRaises(ValueError) as cm:
+            Linter("some-executable", diff_severity=RESULT_SEVERITY.MAJOR)
+        self.assertEqual(str(cm.exception),
+                         "Invalid keyword arguments provided: 'diff_severity'")
```
sils1297: can you make each of those an own function? They all test a different invalidity of params, right? Let's seperate them.

sils1297: scratch that it's ok

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -0,0 +1,711 @@
...
+        process_output_mock.assert_called_once_with(("hello stdout\n",
+                                                     "hello stderr\n"), "", [])
+
+    def test_process_output_corrected(self):
+        uut = (Linter(sys.executable, output_format="corrected")
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        original = [s + "\n" for s in ["void main()  {", "return 09;", "}"]]
+        fixed = [s + "\n" for s in ["void main()", "{", "return 9;", "}"]]
```
sils1297: adding `\n` right in the string is easier, better readable and shorter

sils1297:
unack https://github.com/coala-analyzer/coala/pull/1990/commits/e1053bb6e9eaad007ef8f7b141d5626d05cc4728 but we're totally almost there

coalib/bearlib/abstractions/Linter.py
AbdealiJK: "severity-map's key <key> must be an instance of one of (str) (provided value: <invalid_val>)"

Which looks quite alright

coalib/bearlib/abstractions/Linter.py
AbdealiJK: Yes, but it would be clearer to see all the acceptable values (to me)
Not a major one, and a subjective opinion

coalib/bearlib/abstractions/Linter.py
AbdealiJK: Did you see what I wrote above ? We can make a arg to create_linter and do functools.partial inside Linter to give that arg

coalib/bearlib/abstractions/Linter.py
AbdealiJK: I verified what @sils1297 said. This doesn't seem needed.

coalib/bearlib/abstractions/Linter.py
AbdealiJK: Please let's keep that a new issue ?

coalib/bearlib/abstractions/Linter.py
AbdealiJK:
```
def process_output_regex(
    self, output, filename, file, output_regex,
    severity_map...):
```

coalib/bearlib/abstractions/Linter.py
AbdealiJK: alright

coalib/bearlib/abstractions/Linter.py
AbdealiJK: Sure (yep, it's minor and not important)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,658 @@
...
+            raise ValueError("`output_regex` needed when specified "
+                             "output-format 'regex'.")
+
+        options["output_regex"] = re.compile(options["output_regex"])
+
+        # Don't setup severity_map if one is provided by user or if it's not
+        # used inside the output_regex. If one is manually provided but not
+        # used in the output_regex, throw an exception.
+        if "severity_map" in options:
+            if "severity" not in options["output_regex"].groupindex:
```
AbdealiJK: assert_in would be useful here

coalib/bearlib/abstractions/Linter.py
```diff
@@ -0,0 +1,658 @@
...
+            raise TypeError("Invalid value for `diff_severity`: " +
+                            repr(options["diff_severity"]))
+
+        if "diff_message" in options:
+            assert_right_type(options["diff_message"], str, "diff_message")
+
+        allowed_options |= {"diff_severity", "diff_message"}
+    elif options["output_format"] == "regex":
+        if "output_regex" not in options:
+            raise ValueError("`output_regex` needed when specified "
```
AbdealiJK: assert_in would be useful here

coalib/bearlib/abstractions/Linter.py
Makman2: omg sorry... yeah that looks actually good

coalib/bearlib/abstractions/Linter.py
Makman2: okay okay :3

coalib/bearlib/abstractions/Linter.py
Makman2: oh indeed understand now what you mean :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: hm yeah, let's make it lowercase :+1:

coalib/bearlib/abstractions/Linter.py
Makman2: what should be written inside `additional_info` ?

coalib/bearlib/abstractions/Linter.py
Makman2: you think that makes things easier? :O
setting the `process_output` is quite embedded into the class, needs options, class functions and the class to mixin later^^

tests/bearlib/abstractions/LinterTest.py
Makman2: hm...
`original = "void main()  {\nreturn 09;\n}"`

Makman2: oh or do you mean:
`["void main()  {\n", "...\n", ...]`
?

sils1297: I meant `["void main() {\n", "...\n", ...]`

coalib/bearlib/abstractions/Linter.py
sils1297: see Result documentation comment,

```
 46         :param additional_info: A long description holding additional
 47                                 information about the issue and/or how to fix
 48                                 it. You can use this like a manual entry for a
 49                                 category of issues.
```

Makman2:
reack 38c2439, 3031129, d3eac19, 1cdd60b

sils1297:
reack https://github.com/coala-analyzer/coala/pull/1990/commits/3d4689bbb6dc8aa80a47f9d6346bc69850b06c7a

sils1297:
reack https://github.com/coala-analyzer/coala/pull/1990/commits/dc5c2bf1872cf1b2d5ba03f1c596111651648fe8

coalib/bearlib/abstractions/Linter.py
sils1297: @AbdealiJK file an issue? That's a low one

sils1297: @AbdealiJK file an issue? That's a low one and I want this merged ASAP, it's not a big issue is it? Just a simplification.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/1990/commits/493b93172b9864bb8f905c445323ea6387971fe4

sils1297:
coverage checked manually, is full

sils1297:
@rultor merge

sils1297:
@rultor merge

PR #1996, opened 2016-04-07
GlobbingTest: Ignore .pyc files

srisankethu:
Fixes https://github.com/coala-analyzer/coala/issues/1994

### DIFF
--- a/tests/parsing/GlobbingTest.py
+++ b/tests/parsing/GlobbingTest.py
@@ -15,6 +15,7 @@ GlobTestDir
 └── File3.z
 """
 import os
+import re
 import unittest
 
 from coalib.parsing.Globbing import (
@@ -218,7 +219,8 @@ class GlobTest(unittest.TestCase):
     def _test_glob(self, pattern, file_list):
         results = sorted([os.path.normcase(g) for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
+        self.assertEqual([i for i in results
+                          if re.search(r"(__pycache__|\.pyc)", i) is None],
                          file_list)
 
     def test_collect_files(self):
@@ -331,7 +333,8 @@ class GlobTest(unittest.TestCase):
         results = sorted([os.path.normcase(os.path.join(os.curdir, g))
                           for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
+        self.assertEqual([i for i in results
+                          if re.search(r"(__pycache__|\.pyc)", i) is None],
                          file_list)
         os.curdir = old_curdir
 


tests/parsing/GlobbingTest.py
```diff
@@ -218,8 +218,8 @@ def setUp(self):
     def _test_glob(self, pattern, file_list):
         results = sorted([os.path.normcase(g) for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
-                         file_list)
+        self.assertEqual([i for i in results if not any(x in i
+                          for x in ("__pycache__", ".pyc"))], file_list)
```
srisankethu: if i correct this there would be max_line_length error

AbdealiJK: Your code seems quite complex ...
Can't you use regex ? Would make it simple
... If re.search(r"(__pycache__|\.pyc)", I) ...

Also, try indenting it differently so that it doesn't cause max line length error

srisankethu: :+1:

sils1297:
marking WIP due to tests and gitmate

srisankethu:
ready is guess !

tests/parsing/GlobbingTest.py
```diff
@@ -331,7 +333,8 @@ def test_no_dirname_recursive(self):
         results = sorted([os.path.normcase(os.path.join(os.curdir, g))
                           for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
+        self.assertEqual([i for i in results if re.search(r"(__pycache__|.pyc)",
```
Makman2: superfluos group in regex, just `__pycache__|.pyc` suffices :)
aand: please escape the `.`, as this means any character^^

tests/parsing/GlobbingTest.py
```diff
@@ -218,7 +219,8 @@ def setUp(self):
     def _test_glob(self, pattern, file_list):
         results = sorted([os.path.normcase(g) for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
+        self.assertEqual([i for i in results if re.search(r"(__pycache__|.pyc)",
```
hypothesist: BWT did we not a unspoken policy of not having single character variables? I remember reading this somewhere. Not sure though.

tests/parsing/GlobbingTest.py
```diff
@@ -218,7 +219,8 @@ def setUp(self):
...
+                                                          i) is None],
```
Makman2: just for aesthetics: I would wrap the list like this:
```python
[i for i in results
 if re.search(r"...", i) is None]
```
same below :D

Makman2:
unack 0c2bebf

tests/parsing/GlobbingTest.py
Makman2: as this was the same before this won't belong into this commit. But indeed in this case we could change it, `i` normally means a number for me^^

tests/parsing/GlobbingTest.py
srisankethu: ``if re.search(...) is None`` will exceed 80 characters

tests/parsing/GlobbingTest.py
srisankethu: what is a superflous group?

Makman2: you enclose the regex inside braces, so you declare a new regex group. You don't need it :)

srisankethu:
@Makman2 Changed!

tests/parsing/GlobbingTest.py
srisankethu: What about g and f variables ?

tests/parsing/GlobbingTest.py
```diff
@@ -218,7 +219,8 @@ def setUp(self):
     def _test_glob(self, pattern, file_list):
         results = sorted([os.path.normcase(g) for g in glob(pattern)])
         file_list = sorted([os.path.normcase(f) for f in file_list])
-        self.assertEqual([i for i in results if "__pycache__" not in i],
+        self.assertEqual([i for i in results
+                          if re.search("(__pycache__|\.pyc)", i) is None],
```
Makman2: you removed the `r` prefix for the string, you need to keep that in this case (as you do `\.`, without the `r` prefix you need to do `\\.`)

same below

srisankethu: so should i ``\\``

Makman2:
ack 4abddc9

Makman2:
@rultor merge my dear

PR #2005, opened 2016-04-12
docs: Add Testing_Bears.rst

srisankethu:
Fixes https://github.com/coala-analyzer/coala/issues/1916

### DIFF
--- a/docs/Users/Tutorials/Testing_Bears.rst
+++ b/docs/Users/Tutorials/Testing_Bears.rst
@@ -1,0 +1,144 @@
+How to use LocalBearTestHelper to test your bears
+=================================================
+
+*coala* has an awesome testing framework so you can write tests for bears
+with ease.
+
+You can use the following to test your bears:
+
+- ``LocalBearTestHelper.check_validity``
+- ``LocalBearTestHelper.check_results``
+- ``verify_local_bears``
+
+Understanding through examples
+------------------------------
+
+Let us understand how to write tests for ``TooManyLinesBear`` in ``some_dir``.
+``TooManyLinesBear`` checks if a file has less than or equal to
+``max_number_of_lines`` lines. ``max_number_of_lines`` by default is 10.
+
+.. code::
+
+    from coalib.results.Result import Result
+    from coalib.bears.LocalBear import LocalBear
+
+
+    class TooManyLinesBear(LocalBear):
+
+        def run(file,
+                filename,
+                max_number_of_lines: int=10):
+            """
+            Detects if a file has more than "max_number_of_lines" lines
+
+            :param max_number_of_lines    Maximum number of lines to be
+                                          allowed for a file. Default is 10.
+            """
+
+            if(len(file)>max_number_of_lines):
+                yield Result(self, "Too many lines")
+
+**EXAMPLE 1** using ``verify_local_bear``
+
+.. code::
+
+    from bears.some_dir.TooManyLinesBear import TooManyLinesBear
+    from tests.LocalBearTestHelper import verify_local_bear
+
+    good_file = '1\n2\n3\n4\n'.splitlines()
+    bad_file = '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n'.splitlines()
+
+    TooManyLinesBearTest = verify_local_bear(TooManyLinesBear,
+                                             valid_files=(good_file,),
+                                             invalid_files=(bad_file,))
+
+``good_file`` is a file which your bear considers as non-style-violating
+and a ``bad_file`` is one which has at least one error/warning/info.
+We need to write a ``good_file`` which has less than or equal to
+``max_number_of_lines`` lines and a ``bad_file`` which has more than
+``max_number_of_lines`` lines and feed them to ``verify_local_bear`` as input
+along with your bear (TooManyLinesBear in this case) and a few additional
+arguments.
+
+.. note::
+
+    ``good_file`` and ``bad_file`` are sequences just like ``file``. A ``file``
+    is a sequence of an input file.
+
+**EXAMPLE 2** using ``LocalBearTestHelper.check_validity``
+
+.. code::
+
+    from queue import Queue
+    from bears.some_dir.TooManyLinesBear import TooManyLinesBear
+
+    from tests.LocalBearTestHelper import LocalBearTestHelper
+    from coalib.settings.Section import Section
+    from coalib.settings.Setting import Setting
+
+
+    class TooManyLinesBearTest(LocalBearTestHelper):
+
+        def setUp(self):
+            self.section = Section('name')
+            self.section.append(Setting('max_number_of_lines', '10'))
+            self.uut = TooManyLinesBear(self.section, Queue())
+
+        def test_valid(self):
+            self.check_validity(self.uut, ["import os"])
+
+        def test_invalid(self):
+            self.check_validity(self.uut, bad_file, valid=False)
+
+.. note::
+
+    ``bad_file`` here is same as ``bad_file`` in the above example.
+
+``check_validity`` asserts if your bear yields any results for a particular
+check with a list of strings. First a *Section* and your Bear
+(in this case ``TooManyLinesBear``) is ``setUp``. Now your *Section* consists
+by default *Settings*. You can append any *Setting* depending on your test.
+Validate a check by passing your bear, lines to check as parameters
+(pass a few other parameters if necessary) to ``check_validity``. The method
+``self.check_validity(self.uut, ["import os"])`` asserts if your bear
+``self.uut`` yields a result when a list of strings ``["import os"]`` is
+passed.
+
+**EXAMPLE 3** using ``LocalBearTestHelper.check_results``
+
+.. code::
+
+    from queue import Queue
+
+    from bears.some_dir.TooManyLinesBear import TooManyLinesBear
+    from tests.LocalBearTestHelper import LocalBearTestHelper
+    from coalib.results.Result import Result
+    from coalib.settings.Section import Section
+
+
+    class TooManyLinesBearTest(LocalBearTestHelper):
+
+        def setUp(self):
+            self.uut = TooManyLinesBear(Section('name'), Queue())
+
+        def test_run(self):
+            self.check_results(
+                self.uut,
+                file,
+                Result.from_values("TooManyLinesBear",
+                                   "Too many lines"
+                                   settings={'max_number_of_lines': int=20}))
+
+``check_results`` asserts if your bear results match the actual
+results on execution on CLI. Just like the above example, we need to ``setUp``
+a *Section* and your Bear with some *Settings*. ``check_results`` validates
+your results by giving your local bear, lines to check and expected results
+as input. ``check_results`` asserts if your bear's results on checking the
+``file`` match with ``Results.from_values(...)``.
+
+A Final Note
+------------
+
+``LocalBearTestHelper`` is written to ease off testing for bears. Make sure
+that your tests have 100% coverage and zero redundancy. Use ``check_results``
+as much as possible to test your bears.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -37,6 +37,7 @@ Welcome to the coala documentation!
 
    First Step <Users/Tutorials/Tutorial>
    Writing Bears <Users/Tutorials/Writing_Bears>
+   Testing Bears <Users/Tutorials/Testing_Bears>
    Linter Bears <Users/Tutorials/Linter_Bears>
    Linter Bears - Suggest corrections <Users/Tutorials/Self_Correcting_Bears>
    Git Help <Users/Tutorials/Git_Help>


docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,146 @@
...
+- ``LocalBearTestHelper.check_validity``
+- ``LocalBearTestHelper.check_results``
+- ``verify_local_bears``
+
+Understanding through examples
+------------------------------
+
+**EXAMPLE 1** using ``verify_local_bear``
+
+PyDocStyleBearTest
```
abhsag24: could we have simpler `custom bears`  for the example?, this seems like a lot of work, for a newcomer, where they'd have to go through what is `PyDocStyleBear` in the first place

Makman2: Yeah thats a good point^^

srisankethu: Sure! I chose ``PyDocStyleBear`` in the first place because i felt it is one of the simpliest( i could find :) )

Makman2: you can just write "MyBear" for example everywhere, and for the good file contents you write some text like "This is considered good code", same principle for bad code^^

srisankethu: I was thinking explaning through already written bears is better because he can understand the output from a bear and how its corresponding test is written.

Makman2: yeah that's right, but most of time the bears can already do "some" stuff and the tests aren't that obvious everytime :)
but yeah if you have a minimal example using our bears that's easy and short (`PyDocStyleBear` does not count :P) then sure use that instead, explaining by reallife examples is everytime better than using fictive ones :D :+1:

srisankethu: Ok :+1:

abhsag24:
marking wip, since tests fail, and for the simplification suggested

srisankethu:
@abhsag24 added your suggestions

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
...
+
+    class MyBearTest(LocalBearTestHelper):
+
+        def setUp(self):
+            self.section = Section(some section)
+            self.section.append(Setting(some minimal setting))
+            self.uut = MyBear(self.section, Queue())
+
+        def test_valid(self):
+            self.check_validity(self.uut, [something that asserts true])
```
abhsag24: it is worth mentioning that this is a list of strings

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Understanding through examples
+------------------------------
+
+Let us to understand how to write tests for ``MyBear`` written in
+``somelanguage``.
+
+.. code::
+
+    import something_necessary
```
abhsag24: how about something totally simple and custom but not vague, like 
```
from coalin.results.Result import Result

class MyBear(LocalBear):

    def run(file, filename):
        if(len(file) > 10):
          yield Result(self, "Too many lines")
``` 
this is just an example off the top of my head , but i believe an example like this would make it easier to understand, and this what i meant by `simpler custom bears`

srisankethu: I had the same doubt in mind so i mentioned you

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
+How to use LocalBearTestHelper to test your bears
+=================================================
+
+**coala** has an awesome testing framework so you can write tests for bears
```
Adrianzatreanu: apparently, from what @sils1297 said, we don't bold coala, due to typographic disruption, so we decided to italic it.

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
...
+You can use the following to test your bears:
+
+- ``LocalBearTestHelper.check_validity``
+- ``LocalBearTestHelper.check_results``
+- ``verify_local_bears``
+
+Understanding through examples
+------------------------------
+
+Let us to understand how to write tests for ``MyBear`` written in
```
Adrianzatreanu: ``Let us understand`` is correct

docs/Users/Tutorials/Testing_Bears.rst
Adrianzatreanu: @abhsag24's example is good! i disagree with the vague example as well

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
...
+    import something_necessary
+
+    class MyBear:
+
+        def do_something_cool():
+            does_something_cool
+
+**EXAMPLE 1** using ``verify_local_bear``
+
+.. code::
```
Adrianzatreanu: did you test this? does this work and it encapsulates with the code tag until line 52? i'm not sure how it works, thats why im asking you

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,134 @@
...
+        def test_run(self):
+            self.check_results(
+                self.uut,
+                list of lines to check,
+                list of expected results)
+
+``check_results`` asserts if your bear results match the actual
+results on execution on CLI. Just like the above example, we need to ``setUp``
+a *Section* and your Bear with some *Settings*. ``check_results`` validates
+your results by giving your local bear, lines to check and expected results
```
Adrianzatreanu: i think the comma is redundant

docs/Users/Tutorials/Testing_Bears.rst
abhsag24: i think all the indented code is encapsulated

srisankethu: its encapsulated

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: sorry! a typo

docs/Users/Tutorials/Testing_Bears.rst
Adrianzatreanu: great then :)

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: ??

Adrianzatreanu: i'm saying there shouldnt be a comma here

srisankethu: i mean't why not ! three parameters for the method - param1, param2 and param3

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+
+    '''.splitlines(keepends=True)
+
+    MyBearTest = verify_local_bear(MyBear,
+                                   valid_files=(good_file,),
+                                   invalid_files=(bad_file,))
+
+``good_file`` is a file which your bear considers as non-style-violating
+and a ``bad_file`` is one which has at least one error/warning/info.
+``My Bear`` checks if a file has less than or equal 10 lines. So you need to
```
Adrianzatreanu: 1 line later you used this correctly, here you didn't :P it's ``equal to``

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+.. note::
+
+    ``file`` contains more than 10 lines.
+
+``check_validity`` asserts if your bear yields any results for a particular
+check with a list of strings. First a *Section* and your Bear
+(in this case MyBear) is ``setUp``. Now your *Section* consists on default
+*Settings*. You can append any *Setting* depending on your test. Validate a
+check by passing your bear, lines to check as parameters
+(pass few others parameters if necessary) to ``check_validity``. The method
```
Adrianzatreanu: 1. it's other parameters, not others
2. i think it should be ``a few`` instead of ``few``, since those 2 have different meanings, and ``a few`` means ``some/ a small number``, which is the case.

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+``self.check_validity(self.utt, ["import os"]`` asserts if your bear
+``self.utt`` yields a result if a list of strings ``["import os"]`` is passed.
```
Adrianzatreanu: you have ``if [..] if`` ..the phonetics here is pretty bad. i'd replace the second ``if`` with ``supposing`` or something like this, do you agree?

docs/Users/Tutorials/Testing_Bears.rst
Adrianzatreanu: yeah i understand it now, you're right.

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+                Result.from_values("MyBear",
+                                   "Too many lines")
+                )
+
+``check_results`` asserts if your bear results match the actual
+results on execution on CLI. Just like the above example, we need to ``setUp``
+a *Section* and your Bear with some *Settings*. ``check_results`` validates
+your results by giving your local bear, lines to check and expected results
+as input. ``check_results`` asserts if your bear's results on checking the
+``file`` matches with the ``Results.from_values(...)``.
```
Adrianzatreanu: ``results [...] match`` i think

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+        def test_invalid(self):
+            self.check_validity(self.uut, file, valid=False)
+
+.. note::
+
+    ``file`` contains more than 10 lines.
+
+``check_validity`` asserts if your bear yields any results for a particular
+check with a list of strings. First a *Section* and your Bear
+(in this case MyBear) is ``setUp``. Now your *Section* consists on default
```
Adrianzatreanu: earlier you used Mybear with double backticks, here you didn't. please folloow the same pattern throughout the doc

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+results on execution on CLI. Just like the above example, we need to ``setUp``
+a *Section* and your Bear with some *Settings*. ``check_results`` validates
+your results by giving your local bear, lines to check and expected results
+as input. ``check_results`` asserts if your bear's results on checking the
+``file`` matches with the ``Results.from_values(...)``.
+
+A Final Note
+------------
+
+LocalBearTestHelper is written to ease off testing for bears. Make sure that
```
Adrianzatreanu: ``LocalBearTestHelper`` might be used in italic or with double backticks, i'm not so sure if it should be since you haven't used it before, probably it shouldn't. if noone else comments regarding this, ignore this.

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: ``when`` is appropriate is guess

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: i did not get you! did you mean
`` ...on checking the file match with Results.from_values(...)`` ?

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+
+    from coalib.results.Result import Result
+    from coalib.bears.LocalBear import LocalBear
+
+
+    class MyBear(LocalBear):
+
+        def run(file,filename):
+            if(len(file)>10):
+                yield Result(self, "Too many lines")
```
abhsag24: It is important here too mention when this bear will give results

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+
+**EXAMPLE 1** using ``verify_local_bear``
+
+.. code::
+
+    from bears.python.MyBear import MyBear
+    from tests.LocalBearTestHelper import verify_local_bear
+
+
+    good_file = '''
```
abhsag24: IMO you could just do 
```
good_file = '1\n2\n3\n'.splitlines
bad_file = '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n'.splitlines()
```
makes it readable and clear that the here the bad file is bad since it has 11 lines

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+
+    '''.splitlines(keepends=True)
+
+    MyBearTest = verify_local_bear(MyBear,
+                                   valid_files=(good_file,),
+                                   invalid_files=(bad_file,))
+
+``good_file`` is a file which your bear considers as non-style-violating
+and a ``bad_file`` is one which has at least one error/warning/info.
+``My Bear`` checks if a file has less than or equal to 10 lines. So you need
```
abhsag24: you can just move this above, so people can know about the bear before reading the example

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+
+    ``good_file`` and ``bad_file`` are sequences just like ``file``. A ``file``
+    is a sequence of an input file.
+
+**EXAMPLE 2** using ``LocalBearTestHelper.check_validity``
+
+.. code::
+
+    from queue import Queue
+    from bears.python.MyBear import MyBear
```
abhsag24: `MyBear` isn't in the bear/python directory right? you should just do
`from somePathTo.MyBear import MyBear`

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+    from tests.LocalBearTestHelper import LocalBearTestHelper
+    from coalib.settings.Section import Section
+    from coalib.settings.Setting import Setting
+
+
+    class MyBearTest(LocalBearTestHelper):
+
+        def setUp(self):
+            self.section = Section('name')
+            self.section.append(Setting('max_number_of_lines', '10'))
```
abhsag24: `MyBear` doesn't have this setting right? i think you should add this where you define it

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,158 @@
...
+            self.uut = MyBear(self.section, Queue())
+
+        def test_valid(self):
+            self.check_validity(self.uut, ["import os"])
+
+        def test_invalid(self):
+            self.check_validity(self.uut, file, valid=False)
```
abhsag24: could you not use the word `file`? it's `invalid_file` and define it somewhere so it's easier to believe that it has more than 10 lines

srisankethu: ``bad_file`` is better i guess

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: Isn't it obvious from the ``if`` condition. Anyway I am adding the line ``MyBear checks if a file...``
which you have mentioned below.

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,137 @@
...
+                filename,
+                max_number_of_lines: int=10):
+            """
+            Detects if a file has more than 10 lines
+
+            :param max_number_of_lines    Maximum number of lines to be
+                                          allowed for a file.
+            """
+
+            if(len(file)>10):
```
abhsag24: inconsequential but better to use `max_number_of_lines` instead of `10` :P

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,137 @@
...
+            self.section = Section('name')
+            self.section.append(Setting('max_number_of_lines', '10'))
+        def setUp(self):
+            self.uut = MyBear(self.section, Queue())
+
+        def test_valid(self):
+            self.check_validity(self.uut, ["import os"])
+
+        def test_invalid(self):
+            self.check_validity(self.uut, bad_file, valid=False)
```
abhsag24: give `bad_file` a value so it is understandable why it is bad

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,137 @@
...
+
+``check_validity`` asserts if your bear yields any results for a particular
+check with a list of strings. First a *Section* and your Bear
+(in this case ``MyBear``) is ``setUp``. Now your *Section* consists on default
+*Settings*. You can append any *Setting* depending on your test. Validate a
+check by passing your bear, lines to check as parameters
+(pass a few other parameters if necessary) to ``check_validity``. The method
+``self.check_validity(self.utt, ["import os"]`` asserts if your bear
+``self.utt`` yields a result when a list of strings ``["import os"]`` is passed.
```
abhsag24: self.uut*

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: same  as  the above ``good_file`` in example1 ? because it does not yield any results right ?

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: :+1:

docs/Users/Tutorials/Testing_Bears.rst
abhsag24: i don't understand what you ask, but `check_validity` with `valid = False` expects the bear to yield results, so the `bad_file` should be such that `MyBear` yields results

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,139 @@
...
+``My Bear`` checks if a file has less than or equal to ``max_number_of_lines``
+lines. ``max_number_of_lines`` by default is 10.
+
+.. code::
+
+    from coalib.results.Result import Result
+    from coalib.bears.LocalBear import LocalBear
+
+
+    class MyBear(LocalBear):
```
abhsag24: can we change the name from `MyBear` to maybe `TooManyLinesBear` or something else? it's upto you though i think the Bear isn't generic enough for `MyBear` anymore :P

Adrianzatreanu: this is actually an amazing idea, i agree with @abhsag24 :) since we give concrete names to our bears, why shouldnt this one have a name too :) ?

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: so is it like the ``good_file`` in example1 ?

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,141 @@
...
+
+    from bears.some_dir.tooManyLinesBear import tooManyLinesBear
+    from tests.LocalBearTestHelper import verify_local_bear
+
+    good_file = '1\n2\n3\n4\n'.splitlines()
+    bad_file = '1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n'.splitlines()
+
+    tooManyLinesBearTest = verify_local_bear(tooManyLinesBear,
+                                             valid_files=(good_file,),
+                                             invalid_files=(bad_file,))
```
abhsag24: we can subtly pass settings here as well to let users know that settings can be changed in `verify_local_bear`

docs/Users/Tutorials/Testing_Bears.rst
abhsag24: ok, so i think it would still be worth mentioning that `bad_file` remains the same as in previous example? maybe down below where you explain everything?

srisankethu: i will a note then! is that fine ?

abhsag24: that'd be fine

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,148 @@
...
+    from bears.some_dir.tooManyLinesBear import tooManyLinesBear
+
+    from tests.LocalBearTestHelper import LocalBearTestHelper
+    from coalib.settings.Section import Section
+    from coalib.settings.Setting import Setting
+
+
+    class tooManyLinesBearTest(LocalBearTestHelper):
+
+            self.section = Section('name')
```
abhsag24: indentation is weird?

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: is it ``setting=Setting('max_number_of_lines', '10')`` ?

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,148 @@
...
+        def setUp(self):
+            self.uut = tooManyLinesBear(self.section, Queue())
+
+        def test_run(self):
+            self.check_results(
+                self.uut,
+                file,
+                Result.from_values("tooManyLinesBear",
+                                   "Too many lines")
+                )
```
abhsag24: this is taking an extra line, can be moved up

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: looks good to me -> https://github.com/srisankethu/coala/blob/master/docs/Users/Tutorials/Testing_Bears.rst

abhsag24: shouldn't this be inside `setUp` ?  also you can't use `self` outside of a function

docs/Users/Tutorials/Testing_Bears.rst
abhsag24: have a look at tests for `ESLintBear`

abhsag24:
3a9b55f needs work

docs/Users/Tutorials/Testing_Bears.rst
srisankethu: so its ``setting= { 'max_number_of_lines' : int=20 }``

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,144 @@
...
+``tooManyLinesBear`` checks if a file has less than or equal to
+``max_number_of_lines`` lines. ``max_number_of_lines`` by default is 10.
+
+.. code::
+
+    from coalib.results.Result import Result
+    from coalib.bears.LocalBear import LocalBear
+
+
+    class tooManyLinesBear(LocalBear):
```
Adrianzatreanu: just out of curiosity. why is it tooManyLinesBear and not TooManyLinesBear?

docs/Users/Tutorials/Testing_Bears.rst
```diff
@@ -0,0 +1,144 @@
...
+
+    ``bad_file`` here is same as ``bad_file`` in the above example.
+
+``check_validity`` asserts if your bear yields any results for a particular
+check with a list of strings. First a *Section* and your Bear
+(in this case ``tooManyLinesBear``) is ``setUp``. Now your *Section* consists
+by default *Settings*. You can append any *Setting* depending on your test.
+Validate a check by passing your bear, lines to check as parameters
+(pass a few other parameters if necessary) to ``check_validity``. The method
+``self.check_validity(self.uut, ["import os"]`` asserts if your bear
```
sils1297: missing `)` in code sample

docs/Users/Tutorials/Testing_Bears.rst
Adrianzatreanu: from what i checked, all out bears start with a capital letter.

Adrianzatreanu:
ack b5b52ae

Adrianzatreanu:
reack faf22a4

Adrianzatreanu:
@rultor merge

PR #2029, opened 2016-04-19
Diff: Convert newvalue to tuple if needed

sils1297:
Only tuples can be concatenated with tuples and lists with lists, mixing
is a bad idea and gives exceptions.

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -69,6 +69,7 @@ class Diff:
         :param file:  A list of lines in the file to apply the fixit to.
         :return:      The corresponding Diff object.
         """
+        assert isinstance(file, (list, tuple))
 
         oldvalue = '\n'.join(file[fixit.range.start.line-1:
                                   fixit.range.end.line])
@@ -78,7 +79,7 @@ class Diff:
                     fixit.value +
                     oldvalue[endindex:])
         new_file = (file[:fixit.range.start.line-1] +
-                    newvalue.splitlines(True) +
+                    type(file)(newvalue.splitlines(True)) +
                     file[fixit.range.end.line:])
 
         return cls.from_string_arrays(file, new_file)


coalib/results/Diff.py
```diff
@@ -78,7 +78,7 @@ def from_clang_fixit(cls, fixit, file):
                     fixit.value +
                     oldvalue[endindex:])
         new_file = (file[:fixit.range.start.line-1] +
-                    newvalue.splitlines(True) +
+                    type(file)(newvalue.splitlines(True)) +
                     file[fixit.range.end.line:])
```
Uran198: I think it may be better to convert file to list. In this way we'll be sure of what type is new_file. Also, some issues may arise if type of file will be something obscure, like neither list nor tuple.

sils1297: not sure, we switched to tuples for files because they're immutable, still it might be that a user gives a list I think this is rather safe. If the user passes something obscure this'll most definitely fail

Uran198: I'm not that this will fail with arbitrary class. If that class have constructor from list, has subrange and plus operaors everything seems to be OK. Here should be at least assert to validate file's type.

sils1297: assert is neat, let's take that

Uran198:
ack 6e9f3b1

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -285,39 +282,6 @@ def test_no_config(self):
                     self.log_printer)
                 self.assertEqual(cm.exception.code, 2)
```
Uran198: There should be at least simple test

coalib/settings/ConfigurationGathering.py
```diff
@@ -200,15 +200,47 @@ def find_user_config(file_path, max_trials=10):
...
-    try:
-        path = str(section["config"])
-        return path if os.path.isdir(path) else os.path.dirname(path)
-    except IndexError:
-        if os.path.isfile(os.path.join(os.getcwd(), '.coafile')):
-            return os.getcwd()
-        return None
+
+    path = os.path.abspath(section.get('files', '').origin)
+    return path if os.path.isdir(path) else os.path.dirname(path)
```
Uran198: What if path is `None`?

Uran198: Ooups.. Nevermind.

tests/settings/ConfigurationGatheringTest.py
sils1297: there is, in the documentation

coalib/settings/ConfigurationGathering.py
```diff
@@ -200,15 +200,47 @@ def find_user_config(file_path, max_trials=10):
...
         return os.getcwd()
-    try:
-        path = str(section["config"])
-        return path if os.path.isdir(path) else os.path.dirname(path)
-    except IndexError:
-        if os.path.isfile(os.path.join(os.getcwd(), '.coafile')):
-            return os.getcwd()
-        return None
+
+    path = os.path.abspath(section.get('files', '').origin)
```
Uran198: If `section.get` will return `''` it'll raise exception on `.origin`

tests/settings/ConfigurationGatheringTest.py
Uran198: Oopus.. I didn't saw doctests.

coalib/settings/ConfigurationGathering.py
sils1297: no, section.get is smart and automatically creates a setting with the key and `''` as value and this a valid origin that'll auto default to cwd

sils1297:
@rultor merge

PR #2039, opened 2016-04-21
calculate square root of number

fneu:


change.py
```diff
@@ -0,0 +1,6 @@
+import math
+import this
```
fneu: this import is unneccessary

fneu: unneeded import

change.py
```diff
@@ -0,0 +1,6 @@
...
+
+def square_root(number):
+	sr = math.sqrt(number)
+        return sr
```
fneu: inconsistent whitespace

sils1297:
@fneu what are you doing?

PR #2047, opened 2016-04-24
ConsoleInteraction: Missing lang-spec in braces

Makman2:
If language specification is missing for a bear, the message is
displayed inside braces for better readability.

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -641,8 +641,8 @@ def show_bear(console_printer, bear, sections, metadata):
                      "Supported languages:",
                      bear.LANGUAGES,
                      "  ",
-                     "The bear does not provide information about which "
-                     "languages it can analyze.")
+                     "(The bear does not provide information about which "
+                     "languages it can analyze.)")
     show_enumeration(console_printer,
                      "Used in:",
                      sections,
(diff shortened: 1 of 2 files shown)


Makman2:
Help me with commit message... :D

coalib/output/ConsoleInteraction.py
```diff
@@ -641,8 +641,8 @@ def show_bear(console_printer, bear, sections, metadata):
                      "Supported languages:",
                      bear.LANGUAGES,
                      "  ",
-                     "The bear does not provide information about which "
-                     "languages it can analyze.")
+                     "(The bear does not provide information about which "
+                     "languages it can analyze.)")
```
sils1297: this commit makes this message inconsistent with the others below

AbdealiJK: I don't know why the brackets are being added. This seems fine to me

sils1297: essentially this string is shown if there's no info about this property, we have that below for the used sections, needed parametres etc. they should have all the same style which is why I made a function that shows it. If we want them in brackets that function should add the brackets consistently automatically.

sils1297:
ping @Makman2 ?

Makman2:
Sorry missed this one, what function should add brackets? And where's the problem with hardcoding it?

sils1297:
Hardcoding is always bad. Look at `show_enumeration`, the `no_items_text` should either *always* or *never* be surrounded by brackets, not *sometimes* like your solution

Makman2:
isn't it everytime surrounded with my commit? :O

sils1297:
look into the function, its used muktiple times

Makman2:
Do we want that actually? Looks like we don't gain anything from it and nobody wants it :)

sils1297:
great

PR #2066, opened 2016-04-25
Linter: Accept non-matches when converting

sils1297:
If there's no match, groups.get will return None and not an empty string
which isn't int convertible.

CC @Makman2 for review


Adrianzatreanu:
ack f5065a2

tushar-rishav:
ack f5065a2

sils1297:
coverage is 100%, merging

sils1297:
@rultor merge

PR #2100, opened 2016-05-01
Udayan/res filter

Udayan12167:


sils1297:
CC @fneu for feedback

coalib/results/ResultFilter.py
```diff
@@ -210,15 +211,31 @@ def remove_result_ranges_diffs(result_list, file_dict):
...
 
     :param original_file_dict: Dict of lists of file contents before  changes
     :param modified_file_dict: Dict of lists of file contents after changes
     """
     affected_files = set(original_file_dict.keys()).union(
         set(modified_file_dict.keys()))
+    original_unique_files = affected_files - set(modified_file_dict.keys())
+    renamed_files_dict = {}
     for file in affected_files:
         if file not in original_file_dict:
```
sils1297: can you make another commit converting this if into a filter

coalib/results/ResultFilter.py
```diff
@@ -210,15 +211,31 @@ def remove_result_ranges_diffs(result_list, file_dict):
...
-            original_file_dict[file] = []
+            renamed = 0
+            for comparable_file in original_unique_files:
+                s = SequenceMatcher(
+                    None,
+                    ''.join(modified_file_dict[file]),
+                    ''.join(original_file_dict[comparable_file]))
+                if s.real_quick_ratio() >= 0.5:
+                    if s.ratio() > 0.5:
```
sils1297: just use `and`, that should be run from left to right. (But confirm that by using your favourite search engine :))

coalib/results/ResultFilter.py
```diff
@@ -210,15 +211,31 @@ def remove_result_ranges_diffs(result_list, file_dict):
...
+                        renamed_files_dict[comparable_file] = file
+                        renamed = 1
+                        break
+            if renamed == 0:
```
sils1297: just use `for ... else`, the else clause will be run if you *dont break out of the loop*, this way you can get rid of this `renamed` variable

tests/results/ResultFilterTest.py
```diff
@@ -562,3 +562,84 @@ def test_delete_file_with_result(self):
...
+        tf1_new = abspath('tf1_new')
+
+        original_file_dict = {tf1: testfile_1, tf2: testfile_2}
+        modified_file_dict = {tf1_new: testfile_1}
+
+        renamed_files = ensure_files_present(original_file_dict,
+                                             modified_file_dict)
+
+        self.assertEqual({tf1: tf1_new}, renamed_files)
+
```
sils1297: better make the stuff below an own testcase, they are individual cases aren't they?

tests/results/ResultFilterTest.py
```diff
@@ -562,3 +562,84 @@ def test_delete_file_with_result(self):
...
+        tf2_new = abspath('tf2_new')
+
+        original_file_dict = {tf1: testfile_1, tf2: testfile_2}
+        modified_file_dict = {tf2_new: testfile_2_new}
+
+        renamed_files = ensure_files_present(original_file_dict,
+                                             modified_file_dict)
+
+        self.assertEqual({}, renamed_files)
+
```
sils1297: and so on, make them seperate

sils1297:
CC @fneu for review

coalib/results/ResultFilter.py
```diff
@@ -77,17 +85,22 @@ def basics_match(original_result,
...
 
     :param original_file_dict: Dict of lists of file contents before changes
     :param diff_dict:          Dict of diffs describing the changes per file
     :param original_result_diff_dict: diff for each file for this result
     :param modified_result_diff_dict: guess
+    :param renamed_files:   A dictionary containing file renamings across runs
     :return:                     Boolean value whether the SourceRanges match
     """
     for file_name in original_file_dict:
+        mod_file_name = file_name
```
sils1297: put this in else

coalib/results/ResultFilter.py
Udayan12167: What do you mean by a filter... a separate method?

tests/results/ResultFilterTest.py
```diff
@@ -368,6 +368,288 @@ def test_affected_code(self):
...
+        new_result_list = [res1_pre_change,       # FALSE POSITIVE (in-line)
+                           res1_change,           # correctly kept
+                           res1_post_change,      # FALSE POSITIVE (in-line)
+                           res1_around_change,    # FALSE POSITIVE (in-line)
+                           res1_with_change,      # correctly kept
+                           res1_whole_change,     # correctly kept
+
+                           res1_pre_remove,       # correctly filtered out
+                           res1_post_remove,      # FALSE POSITIVE (in-line)
```
sils1297: ahm did you copy another test? :/ with the false positives? We only want to test renaming here, not everything, right?

coalib/results/ResultFilter.py
sils1297: `for ... in filter(lambda file: file not in original_file_dict, affected_files):`

tests/results/ResultFilterTest.py
Udayan12167: They are just evaluating different conditions of the ensure_files_present. They aren't really different test cases.

sils1297: Disagree. You test certain aspects concerinng file renaming, every aspect should have an own case or it doesn't deserve a test. You're testing several things sequentially here which is bad.

tests/results/ResultFilterTest.py
Udayan12167: I wanted to test the full functionality of the ResultFilter with a renamed file. Hence, I reused the test.

sils1297: copying large chunks of code is bad, can we take the other test and give it a parameter? This tests totally unrelated stuff and makes things really complicated. http://pytest.org/latest/parametrize.html#parametrized-test-functions

Udayan12167: http://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests

tests/results/ResultFilterTest.py
sils1297: this is *so much easier* to understand with every test having an own name describing it's purpose! :)

coalib/results/ResultFilter.py
```diff
@@ -210,15 +211,28 @@ def remove_result_ranges_diffs(result_list, file_dict):
 
 def ensure_files_present(original_file_dict, modified_file_dict):
     """
-    Ensures that all files are available as keys in both dicts.
+    Ensures that all files are available as keys in both dicts. Return a
+    dictionary of renamed files.
```
sils1297: we'd rather want a `:return` annotation below the params right?

coalib/results/ResultFilter.py
```diff
@@ -210,15 +211,28 @@ def remove_result_ranges_diffs(result_list, file_dict):
...
 
     :param original_file_dict: Dict of lists of file contents before  changes
     :param modified_file_dict: Dict of lists of file contents after changes
     """
     affected_files = set(original_file_dict.keys()).union(
         set(modified_file_dict.keys()))
+    original_unique_files = affected_files - set(modified_file_dict.keys())
```
sils1297: I think it'd be good to do something like:

```
original_files = set(original_file_dict.keys())
modified_files = ...
affected_files = original_files | modified_files
original_unique_files = original_files - modified_files
```

and later instead of `if file not in original_file_dict` you can just `if file not in original_files` as the set might be better optimized for that operation

tests/results/ResultFilterTest.py
```diff
@@ -562,3 +562,87 @@ def test_delete_file_with_result(self):
                                      [old_result_tf1, old_result_tf2],
                                      [new_result])
         self.assertEqual(new_results, [new_result])
+
```
sils1297: can you maybe put all tests that test the `ensure_files_present` in an own class? Then we can rely on them not depending on anything that's needed for the others, less complexity to think about.

tests/results/ResultFilterTest.py
```diff
@@ -562,3 +562,87 @@ def test_delete_file_with_result(self):
...
+        testfile_2_new = ['1\n', '2\n', '0\n', '1\n', '2\n', '1\n', '2\n']
+        tf2_new = abspath('tf2_new')
+
+        original_file_dict = {tf1: testfile_1, tf2: testfile_2}
+        modified_file_dict = {tf2_new: testfile_2_new}
+
+        renamed_files = ensure_files_present(original_file_dict,
+                                             modified_file_dict)
+
+        self.assertEqual({}, renamed_files)
```
sils1297: this commit is awesome!

coalib/results/ResultFilter.py
```diff
@@ -20,12 +20,19 @@ def filter_results(original_file_dict,
...
+                                         modified_file_dict)
     # diffs_dict[file] is a diff between the original and modified file
     diffs_dict = {}
     for file in original_file_dict:
-        diffs_dict[file] = Diff.from_string_arrays(original_file_dict[file],
-                                                   modified_file_dict[file])
+        if file in renamed_files:
+            diffs_dict[file] = Diff.from_string_arrays(
+                original_file_dict[file],
+                modified_file_dict[renamed_files[file]])
```
sils1297: why not

```
            diffs_dict[file] = Diff.from_string_arrays(
                original_file_dict[file],
                modified_file_dict[renamed_files.get(file, file)])
```

that's shorter and should yield at least the same performance

coalib/results/ResultFilter.py
```diff
@@ -98,7 +112,7 @@ def source_ranges_match(original_file_dict,
         # original file with file_diff and original_diff applied
         original_total_file = original_total_diff.modified
         # modified file with modified_diff applied
-        modified_total_file = modified_result_diff_dict[file_name].modified
+        modified_total_file = modified_result_diff_dict[mod_file_name].modified
```
sils1297: I'd use the same `renamed_files.get(file_name, file_name)` here, no need for the one variable for one use

tests/results/ResultFilterTest.py
```diff
@@ -368,6 +368,72 @@ def test_affected_code(self):
...
+                                               end_line=7,
+                                               end_column=5)
+
+        original_result_list = [res0_pre_change]
+
+        new_result_list = [res1_pre_change,
+                           res1_change,
+                           res1_whole_remove]
+
+        unique_new_result_list = [res1_pre_change,
```
sils1297: shouldn't `res1_pre_change` not be there? It was only in a renamed file, right?

sils1297: ah the line changes, but your testcase doesn't really test this, does it? You should rather add a result that stays the same due to the renaming, hm? Make a result in line 7 (8 in the original file)! That one's shifted and renamed, if we can detect that properly that should work out fine.

coalib/results/ResultFilter.py
```diff
@@ -235,18 +235,21 @@ def ensure_files_present(original_file_dict, modified_file_dict):
...
-                    None,
-                    ''.join(modified_file_dict[file]),
-                    ''.join(original_file_dict[comparable_file]))
-                if s.real_quick_ratio() >= 0.5 and s.ratio() > 0.5:
-                    renamed_files_dict[comparable_file] = file
-                    break
-            else:
-                original_file_dict[file] = []
-        if file not in modified_file_dict:
-            modified_file_dict[file] = []
```
sils1297: ouh there are two ifs, yeah drop this commit, that was a bad idea of mine

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2100/commits/292665f40c472014a9e0ec32ee0fbbc4d4b3ab0f and the other commits as well. This also needs a rebase but it's close to merge!

tests/results/EnsureFilesPresentTest.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+from os.path import abspath
+
+from coalib.results.ResultFilter import ensure_files_present
+
+
+class EnsureFilesPresentTest(unittest.TestCase):
+
```
sils1297: /me feels joy overcoming his body while reading this test class

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2100/commits/4aa810c664ec11434608057d4fdd9967152562e0

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2100/commits/79b4cb40bfc6fc49032f981828c7dc68296e640e

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2100/commits/de5cf0a76531ee6d0f144ebbcda5e2142e65e1c7?diff=split

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2100/commits/5dfbe2f43a72445038400c986a0c913ea10b264c

sils1297:
@rultor merge

PR #2135, opened 2016-05-12
Sanketdg/docfuse

SanketDG:
CC @Makman2

### DIFF
--- a/coalib/settings/DocstringMetadata.py
+++ b/coalib/settings/DocstringMetadata.py
@@ -4,12 +4,12 @@ from collections import OrderedDict
 from coalib.misc.Enum import enum
 
 
-class DocumentationComment:
+class DocstringMetadata:
     _ParseMode = enum("DESCRIPTION", "PARAM", "RETVAL")
 
     def __init__(self, desc, param_dict, retval_desc):
         """
-        Represents a documentation comment of a python class or function.
+        Represents a docstring of a python class or function.
 
         :param desc:        A description as string.
         :param param_dict:  A dictionary containing parameter names as key and
--- a/coalib/settings/FunctionMetadata.py
+++ b/coalib/settings/FunctionMetadata.py
@@ -3,7 +3,7 @@ from copy import copy
 from inspect import getfullargspec, ismethod
 
 from coalib.misc.Decorators import enforce_signature
-from coalib.settings.DocumentationComment import DocumentationComment
+from coalib.settings.DocstringMetadata import DocstringMetadata
 
 
 class FunctionMetadata:
@@ -145,7 +145,7 @@ class FunctionMetadata:
             return metadata
 
         doc = func.__doc__ or ""
-        doc_comment = DocumentationComment.from_docstring(doc)
+        doc_comment = DocstringMetadata.from_docstring(doc)
 
         non_optional_params = OrderedDict()
         optional_params = OrderedDict()
--- a/tests/settings/DocstringMetadataTest.py
+++ b/tests/settings/DocstringMetadataTest.py
@@ -1,9 +1,9 @@
 import unittest
 
-from coalib.settings.DocumentationComment import DocumentationComment
+from coalib.settings.DocstringMetadata import DocstringMetadata
 
 
-class DocumentationCommentParserTest(unittest.TestCase):
+class DocstringMetadataTest(unittest.TestCase):
 
     def test_from_docstring(self):
         self.check_from_docstring_dataset("")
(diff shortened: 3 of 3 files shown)


sils1297:
@SanketDG why is this wip? Because of travis? We're removing travis right now.

setup.cfg
```diff
@@ -14,7 +14,6 @@ norecursedirs =
     .env
 testpaths =
     coalib
-    docs
```
sils1297: I'm against this: https://github.com/coala-analyzer/coala/issues/1790#issuecomment-219034544

SanketDG:
@Makman2 upon much thinking I realized that the documentation parsing functions used for `FunctionMetadata` are fairly simple. The way we parse them right now, (as discussed on Gitter) is too complex for that usecase. So we have to find a way to transform the parsed docstring into a simple namedtuple.

Right now I am just creating a new function that parses the doc first, finds the relevant parts using `isinstance()` and then returns a namedtuple based on that.

Another thing we could do is to add a `simple: bool=False`(could use a better name) param to the function. In this way, when I pass `simple=True`,I could get a simple version of the parsed documentation.

Thoughts?

sils1297:
@SanketDG the way I see the doc class thing: they are different classes and one of them should be renamed.

Makman2:
Haven't we said we wanted a sequence/list of namedtuple?
And what's a "simple" version? :O

But I also agree with @sils1297, we should not merge stuff from `FunctionMetadata` and so on, they both deal with documentations, but the one is for general purpose documentation inside source code, while the other is purely Python specific and used in our coala core.
I think you should concentrate on getting the parsing function into the `coalib.bearlib.languages.documentation.DocumentationComment` only, and renaming the other one to maybe `PythonDocComment`/`PythonDocumentationComment` first^^

setup.cfg
SanketDG: CC @Makman2 @AbdealiJK @tushar-rishav @abhsag24 your thoughts on this?

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -24,3 +24,69 @@ def test_fields(self):
...
+    def check_docstring(self,
+                        docstring,
+                        expected=[]):
+        self.assertIsInstance(docstring,
+                              str,
+                              "docstring needs to be a string for this test.")
+
+        self.assertIsInstance(expected,
+                              list,
+                              "docstring needs to be a list for this test.")
```
SanketDG: TODO needs to be expected

coalib/settings/DocstringMetadata.py
```diff
@@ -24,54 +23,21 @@ def __init__(self, desc, param_dict, retval_desc):
...
-            if parse_mode == cls._ParseMode.RETVAL:
-                retval_desc = concat_doc_parts(retval_desc, line)
-            elif parse_mode == cls._ParseMode.PARAM:
-                param_dict[cur_param] = concat_doc_parts(param_dict[cur_param],
-                                                         line)
-            else:
-                desc = concat_doc_parts(desc, line)
-
-        return (cls(desc=desc.strip(),
+        for part in parsed_doc:  # pragma: no cover
```
SanketDG: Reason why this is not being covered:

By the current documentation parsing, there can be only one `Desc` section. When the loop passes through, it only finds it once and fails to go back to the for loop again. This causes a missed partial branch.

Need solutions :stuck_out_tongue_winking_eye: 

CC @Makman2

sils1297:
unack

- https://github.com/coala-analyzer/coala/pull/2135/commits/eb51294974055e1ec11fc11eb94a5a2a81a3b42f as per discussion
- https://github.com/coala-analyzer/coala/pull/2135/commits/b00cc768507243226b340571d1bfb3ba775387a2 since it's just test duplication, that code is already in the docs where it belongs.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2135/commits/c3d034eb9bbffabc625de3533069c237884fd4b5

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2135/commits/c03248078768993f1ba82847c4f55b70a7a87876 and https://github.com/coala-analyzer/coala/pull/2135/commits/ec3549809cfd95a86c861244ee4a61d9f67e1826

I don't think merging those classes is meaningful, you're moving a language independent internal routine over to another class where it isn't needed and then you use that method in the original location.

I'd really just take the one commit I acked from this PR and be done with it.

sils1297:
unack c032480 and ec35498

SanketDG:
reack 01fdf48

@sils1297 I will get a new PR for the new functions and all after some more refactoring

SanketDG:
reack 01fdf48

sils1297:
@rultor merge then

PR #2136, opened 2016-05-12
Decorators.py: Move to own package

Adrianzatreanu:
Decorators are now in a Pypi package named
coala-decorators, which is installed in the
requirements.

Fixes https://github.com/coala-analyzer/coala/issues/2027


SanketDG:
TODO change imports

PR #2137, opened 2016-05-13
Sanketdg/bugfix

SanketDG:


coalib/bearlib/abstractions/Linter.py
```diff
@@ -447,10 +447,13 @@ def run(self, filename, file, **kwargs):
...
 
-                arguments = (self.get_executable(),) + tuple(
-                    self.create_arguments(
-                        filename, file, config_file,
-                        **create_arguments_kwargs))
+                args = self.create_arguments(filename, file, config_file,
+                                             **create_arguments_kwargs)
+
+                if args is None:  # pragma: no cover
+                    self.warn("The tuple of arguments cannot be None.")
```
sils1297: 1. This should be tested!
2. DOn't you want to do a return here or something instead of just running into 456 throwing an exception? ALso this should be an err IMO.

CC @Makman2

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2137/commits/27c448f13e0030763f2e67a096e6ee9172b90725

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2137/commits/7ec15084a80e377ae15668766dd7240cfbd1e1fe

sils1297:
CC @AbdealiJK @Udayan12167 something's wrong with travis ATM, what do we want to do about it? Affects other PRs too! We need a quick fix for this.

sils1297:
please rebase, your first commit was already merged

coalib/bearlib/abstractions/Linter.py
Makman2: yep this should be an error.
and yes please return otherwise it still crashes.

And I think you should change the check to a try-error construct:
```
try:
    arguments = tuple(args)
except TypeError:
    self.err("Arguments are not iterable") # <-- make this a bit nicer
    # + you could also fire a debug message here too with additional info on the linter,
    # and what was exactly returned etc.
    return
```

Makman2: using try-catch has the advantage that you can return any sequence, you can even use `yield` then which is very cool and convenient sometimes (maybe some bears shall actually use them...)

SanketDG:
How it looks:
```
❯ coala -I -b JuliaLintBear -f hello.jl -L DEBUG
[DEBUG][13:50:12] Platform Linux -- Python 3.4.3, pip 7.0.3, coalib 0.7.0.dev99999999999999
Executing section Default...
[DEBUG][13:50:13] Files that will be checked:
/home/sanket/coala-bears/hello.jl
[DEBUG][13:50:15] Running bear JuliaLintBear...
[ERROR][13:50:15] The given arguments `None` are not iterable.
```

coalib/bearlib/abstractions/Linter.py
```diff
@@ -447,10 +447,17 @@ def run(self, filename, file, **kwargs):
...
 
-                arguments = (self.get_executable(),) + tuple(
-                    self.create_arguments(
-                        filename, file, config_file,
-                        **create_arguments_kwargs))
+                args = self.create_arguments(filename, file, config_file,
+                                             **create_arguments_kwargs)
+
+                try:
+                    args = tuple(args)
```
sils1297: why not put the whole thing into the try except? No need to split it up in args, right?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -447,10 +447,17 @@ def run(self, filename, file, **kwargs):
...
+                except TypeError:
+                    self.err("The given arguments "
+                             "`{}` are not iterable.".format(args))
```
sils1297: rather use `{0!r}` to let python use repr on args instead of putting single quotes around it manually

coalib/bearlib/abstractions/Linter.py
sils1297: I think I'd inline line 450 here and keep the rest

sils1297:
unack acd7500

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2137/commits/c3c2cfebe7da1e0b694d0acb70bef6a943bfa60c

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2137/commits/c277ad0f74213a44cd0bf8ed7f457b5f30834917 merge at will :)

SanketDG:
@rultor merge

coalib/bearlib/abstractions/Linter.py
Makman2: it's good that it wasn't inlined, if `create_argument` raises a TypeError this would be catched too, but in this case the bear shall crash^^

PR #2141, opened 2016-05-14
Diff: Add distance param

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/2117

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -173,18 +173,40 @@ class Diff:
         :param filename: The filename to associate the SourceRange's to.
         :return:         A list of all related SourceRange objects.
         """
-        return list(diff.range(filename) for diff in self.split_diff())
+        return list(diff.range(filename)
+                    for diff in self.split_diff(distance=0))
 
-    def split_diff(self):
+    def split_diff(self, distance=1):
         """
         Splits this diff into small pieces, such that several continuously
         altered lines are still together in one diff. All subdiffs will be
         yielded.
+
+        A diff like this with changes being together closely won't be splitted:
+
+        >>> diff = Diff.from_string_arrays([     'b', 'c', 'e'],
+        ...                                ['a', 'b', 'd', 'f'])
+        >>> len(list(diff.split_diff()))
+        1
+
+        If we set the distance to 0, it will be splitted:
+
+        >>> len(list(diff.split_diff(distance=0)))
+        2
+
+        If a negative distance is given, every change will be yielded as an own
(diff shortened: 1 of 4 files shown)


sils1297:
CC @Uran198

coalib/results/Diff.py
```diff
@@ -175,16 +175,31 @@ def affected_code(self, filename):
...
+        >>> len(list(diff.split_diff()))
+        1
+
+        If we set the distance to 0, it will be splitted:
+
+        >>> len(list(diff.split_diff(distance=0)))
+        2
+
+        :param distance: Number of unchanged lines that are allowed in between
+                         two changed lines so they get yielded as one diff.
```
abhsag24: shouldn't we assert that distance > 0, i don't think it matters i'm just trying to be logical, we should have a generic test case in case we are allowing that

sils1297: IMO it's ok, if the user provides <0 then every line will end up in an own change even if they're right beneath each other which could be even intended :P behaviour still makes sense so I don't see posing a restriction would be much needed

abhsag24: so in the above test case we should state: `If we set the distance to anything less than 1, it will be splitted:`

sils1297:
argh tests fail...

coalib/results/Diff.py
sils1297: changed

coalib/results/Diff.py
```diff
@@ -173,18 +173,40 @@ def affected_code(self, filename):
         :param filename: The filename to associate the SourceRange's to.
         :return:         A list of all related SourceRange objects.
         """
-        return list(diff.range(filename) for diff in self.split_diff())
+        return list(diff.range(filename)
+                    for diff in self.split_diff(distance=0))
 
-    def split_diff(self):
+    def split_diff(self, distance=1):
```
Makman2: not using `distance=0` as default?

Makman2:
ack ce538f097b8c66cdeba0387993ff9f8ee2a7a3e2

coalib/results/Diff.py
Makman2: at least it's git behaviour then :)

sils1297: having one unchanged line in between makes a bit of sense since those changes likely still belong together. This is about splitting semantical changes after all!

Makman2:
ack b8add6a

Makman2:
@rultor merge

sils1297:
yayyyy!

PR #2154, opened 2016-05-15
Simplify `str.format` usages

Makman2:


SanketDG:
ack 1b315f4

tests/misc/BuildManPageTest.py
```diff
@@ -75,8 +75,8 @@ def test_mk_options(self):
     def test_mk_footer(self):
         uut = ManPageFormatter(app_name, ext_sections=sections)
         self.assertEqual(uut._mk_footer(),
-                         ".SH {0}\n"
-                         " {1}".format(section_name.upper(), section_text))
+                         ".SH {}\n {}".format(
```
tushar-rishav: @Makman2 What's up here? I vaguely remember, we were accepting `{0} {1}` instead of `{ } { }` style.
 :D

Makman2: yeah but I thought the line break was not so good to read, and I directly removed the numbers inside {} as they are superfluous in this case :)
You think I should keep as it is?

tushar-rishav: @Makman2  Naah it's fine. It looks actually better now.  :+1:
I just wanted to confirm. Because, I happen to be in such situation often when I need to choose between `convention` and `asthetics` :D

sils1297:
@rultor merge

PR #2166, opened 2016-05-21
Add PackageRequirements

Adrianzatreanu:
Requirements is a tuple that holds the requirements
that are to be installed for each bear. This helps
not maintaining a requirements.txt file.

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -9,6 +9,8 @@ from appdirs import user_data_dir
 from pyprint.Printer import Printer
 
 from coala_decorators.decorators import enforce_signature, classproperty
+from coalib.bears.requirements.PackageRequirement import PackageRequirement
+from coalib.bears.requirements.PythonRequirement import PythonRequirement
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
@@ -47,6 +49,29 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To indicate the requirements of the bear, assign ``REQUIREMENTS`` a tuple
+    with instances of ``PackageRequirements``.
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (
+    ...         PackageRequirement('pip', 'coala_decorators', '0.2.1'),)
+
+    If your bear uses requirements from a manager we have a subclass from,
+    you can use the subclass, such as ``PythonRequirement``, without specifying
+    manager:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (PythonRequirement('coala_decorators', '0.2.1'),)
+
+    To specify multiple requirements using ``pip``, you can use the multiple
+    method. This can receive both tuples of strings, in case you want a specific
+    version, or a simple string, in case you want the latest version to be
+    specified.
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple(
+    ...         ('colorama', '0.1'), 'coala_decorators')
+
     Every bear has a data directory which is unique to that particular bear:
 
     >>> class SomeBear(Bear): pass
@@ -56,6 +81,7 @@ class Bear(Printer, LogPrinter):
     """
 
     LANGUAGES = ()
+    REQUIREMENTS = ()
 
     @classproperty
     def name(cls):
--- a/coalib/bears/requirements/PackageRequirement.py
+++ b/coalib/bears/requirements/PackageRequirement.py
@@ -1,0 +1,51 @@
+from coala_decorators.decorators import generate_eq, generate_repr
+
+
+@generate_eq("manager", "package", "version")
+@generate_repr()
+class PackageRequirement:
+    """
+    This class helps keeping track of bear requirements. It should simply
+    be appended to the REQUIREMENTS tuple inside the Bear class.
+
+    Two ``PackageRequirements`` should always be equal if they have the same
+    manager, package and version:
+
+    >>> pr1 = PackageRequirement('pip', 'coala_decorators', '0.1.0')
+    >>> pr2 = PackageRequirement('pip', 'coala_decorators', '0.1.0')
+    >>> pr1 == pr2
+    True
+    """
+
+    def __init__(self, manager: str, package: str, version=""):
+        """
+        Constructs a new ``PackageRequirement``.
+
+        >>> pr = PackageRequirement('pip', 'colorama', '0.1.0')
+        >>> pr.manager
+        'pip'
+        >>> pr.package
+        'colorama'
+        >>> pr.version
+        '0.1.0'
+
+        :param manager: A string with the name of the manager (pip, npm, etc).
+        :param package: A string with the name of the package to be installed.
+        :param version: A version string. Leave empty to specify latest version.
+        """
+        self.manager = manager
+        self.package = package
+        self.version = version
+
+    def check(self):
+        """
+        Check if the requirement is satisfied.
+
+        >>> PackageRequirement('pip', 'coala_decorators', '0.2.1').check()
+        Traceback (most recent call last):
+        ...
+        NotImplementedError
+
+        :return: Returns True if satisfied, False if not.
+        """
+        raise NotImplementedError
--- a/coalib/bears/requirements/PythonRequirement.py
+++ b/coalib/bears/requirements/PythonRequirement.py
@@ -1,0 +1,26 @@
+from coalib.bears.requirements.PackageRequirement import PackageRequirement
+
+
+class PythonRequirement(PackageRequirement):
+    """
+    This class is a subclass of ``PackageRequirement``, and helps specifying
+    requirements from ``pip``, without using the manager name.
+    """
+
+    def __init__(self, package, version=""):
+        """
+        Constructs a new ``PythonRequirement``, using the ``PackageRequirement``
+        constructor.
+
+        >>> pr = PythonRequirement('setuptools', '19.2')
+        >>> pr.manager
+        'pip'
+        >>> pr.package
+        'setuptools'
+        >>> pr.version
+        '19.2'
+
+        :param package: A string with the name of the package to be installed.
+        :param version: A version string. Leave empty to specify latest version.
+        """
+        PackageRequirement.__init__(self, 'pip', package, version)
--- a/coalib/bears/requirements/PythonRequirement.py
+++ b/coalib/bears/requirements/PythonRequirement.py
@@ -24,3 +24,65 @@ class PythonRequirement(PackageRequirement):
         :param version: A version string. Leave empty to specify latest version.
         """
         PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
+        Creates a tuple of multiple ``PythonRequirements``.
+
+        You should use the ``multiple`` method if you have more
+        requirements from the same manager. This can receive both tuples of
+        strings, in case you want a specific version, or a simple string, in
+        case you want the latest version to be specified.
+
+        This is the case where you would provide strings only, to specify the
+        latest version automatically:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     'coala_decorators', 'setuptools')
+
+        And if you choose to mix them, specifying version for some and for some
+        not:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     'coala_decorators', ('setuptools', '19.2'))
+
+        In case you provide too many arguments into the tuple, an error will be
+        raised:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     'coala_decorators', ('setuptools', '19.2', 'colorama'))
+        Traceback (most recent call last):
+        ...
+        TypeError: The tuple must have 2 elements.
+
+        The same would happen in case you provide something different than a
+        string or a tuple:
+
+        >>> x = [1, 2, 3, 4]
+        >>> REQUIREMENTS = PythonRequirement.multiple(x)
+        Traceback (most recent call last):
+        ...
+        TypeError: The arguments need to be tuples or strings.
+
+        :param args:       Should be tuples of strings: ``('packageName',
+                           'version')`` or strings: ``'packageName'`` if latest
+                           version is wanted.
+        :return:           A tuple containing ``PythonRequirements``.
+        :raises TypeError: In case the tuples contain more or less than two
+                           elements. Also raised when arguments are neither
+                           tuples nor strings.
+        """
+        reqs = []
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs.append(cls(requirement),)
+            elif isinstance(requirement, tuple):
+                try:
+                    name, version = requirement
+                    reqs.append(cls(name, version),)
+                except ValueError:
+                    raise TypeError('The tuple must have 2 elements.')
+            else:
+                raise TypeError('The arguments need to be tuples or strings.')
+        return tuple(reqs)


coalib/bears/Bear.py
```diff
@@ -47,6 +47,13 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To see what requirements the bear has, give the bear a value for
+    ``REQUIREMENTS``, which will be a tuple of strings:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = ('yamllint~=1.0.1')
```
Makman2: redundant braces, if you want a tuple don't forget the comma at the end :)
--> `('yamllint~=1.0.1',)`

coalib/bears/Bear.py
```diff
@@ -47,6 +47,13 @@ class Bear(Printer, LogPrinter):
...
+    ...     REQUIREMENTS == SomeBear.requirements
```
Makman2: yes this is true :P
why do you have this?

coalib/bears/Bear.py
SanketDG: if you take this as a tuple, there needs to be a comma for an one element tuple

coalib/bears/Bear.py
```diff
@@ -47,6 +47,13 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To see what requirements the bear has, give the bear a value for
+    ``REQUIREMENTS``, which will be a tuple of strings:
```
Makman2: I would explicitly mention pip packages for python :+1:

coalib/bears/Bear.py
Adrianzatreanu: but you said redundant braces and then you placed them. shouldn't it ideally be ``'yamllint~=1.0.1,'`` ?

coalib/bears/Bear.py
Adrianzatreanu: for coverage. if i wouldnt use the method, it would be 99.80%

coalib/bears/Bear.py
```diff
@@ -57,6 +64,8 @@ class Bear(Printer, LogPrinter):
 
     LANGUAGES = ()
 
+    REQUIREMENTS = ()
```
Makman2: as we're on it, I think you can write this directly beneath `LANGUAGES`, so it's a bit more dense :+1:

coalib/bears/Bear.py
```diff
@@ -72,6 +81,13 @@ def supported_languages(cls):
         return (cls.LANGUAGES if isinstance(
             cls.LANGUAGES, tuple) else (cls.LANGUAGES,))
 
+    @classproperty
+    def requirements(cls):
+        """
+        :return: The requirements needed by the bear.
```
Makman2: I would mention return type here: `tuple`
Or even better: `Iterable`, as you are (presumably) able to support any iterable type, not only tuples (though they are encouraged instead of other mutable collections, frozensets are in fact also possible :D)

coalib/bears/Bear.py
```diff
@@ -72,6 +81,13 @@ def supported_languages(cls):
...
+        """
+        return cls.REQUIREMENTS
```
Makman2: hm I think this property is actually not needed, you have a field which you can access like the property, imo this is redundant :+1:

coalib/bears/Bear.py
Makman2: this would make a string with a comma. So either you want a plain string (--> no braces at all) or a tuple (braces + comma)

coalib/bears/Bear.py
Makman2: though I would remove the prop below, you can actually test this with writing below `True`
-->
```python
>>> SomeBear.REQUIREMENTS == SomeBear.requirements
True
```
but then this needs to be outside of the class :+1:

coalib/bears/Bear.py
Adrianzatreanu: i understand then!

coalib/bears/Bear.py
```diff
@@ -47,6 +47,12 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To see what ``pip`` packages requirements the bear has, give the bear a
```
Makman2: not "package requirements"?

Adrianzatreanu: im not sure. its pip packages that are requirements.

Adrianzatreanu: i think you are right though

coalib/bears/Bear.py
```diff
@@ -47,6 +47,12 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To see what ``pip`` package requirements the bear has, give the bear a
+    value for ``REQUIREMENTS``, which will be a tuple of strings:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = ('yamllint~=1.0.1',)
```
sils1297: how will coala enforce this requirement? Should we show it at the detailed bear information?

Adrianzatreanu: hm I guess. I haven't thought of that.

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,37 @@
...
+
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package: str, version):
+        """
+        :param package: A string with the name of the package to be installed.
+        :param version: A number that contains the version.
+        """
+        self.package = package
+        self.version = version
```
sils1297: nonononono! You have to invoke `PackageRequirement(package, 'pip', version)` to do the parental construction

sils1297: and no ndeed to override str

Adrianzatreanu: what's the difference? isn't it just a different order in instantiating?

sils1297: you're not initializing the base class here at all

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,32 @@
...
+        raise NotImplementedError
+
+
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package, version):
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    def __str__(self):
+        return "{.package} version {.version}".format(self)
```
SanketDG: where is the manager here?

Can you try:

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,32 @@
...
+        :param manager: A string with the name of the manager (pip, npm, etc).
+        :param package: A string with the name of the package to be installed.
+        :param version: A number that contains the version.
+        """
+        self.manager = manager
+        self.package = package
+        self.version = version
+
+    def __str__(self):
+        return "{.package} version {.version} (for {.manager})".format(self)
```
SanketDG: same as below

coalib/bears/PackageRequirement.py
Adrianzatreanu: sure, trying now

coalib/bears/PackageRequirement.py
SanketDG: use self

coalib/bears/PackageRequirement.py
Adrianzatreanu: i tried this on the packagerequirement class and it still fails

Adrianzatreanu:
@justuswilhelm the __str__ method fails in the doctest and i tend to this it is bad

Makman2:
`PackageRequirements: Add it`
An `s` too much :)

Adrianzatreanu:
ah sure, will fix this in a sec, let me see the tests pass first :p i might change to need more stuff

coalib/bears/PackageRequirement.py
Makman2: I think you don't need `__str__` any more :+1:

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,15 @@ def check(self):
...
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package, version):
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    def __str__(self):
+        return "{.package} version {.version}".format(self)
+
+    def multiple(self, *args):
+        return args
```
Makman2: Shouldn't this be a classmethod that returns a tuple of `PythonRequirement` instances?

coalib/bears/PackageRequirement.py
Adrianzatreanu: i don't, forgot about this.

coalib/bears/Bear.py
```diff
@@ -47,6 +48,25 @@ class Bear(Printer, LogPrinter):
...
+    manager:
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (PythonRequirement('coala_decorators', '0.2.1'),)
+    Alternatively, you should use the ``multiple`` method if you have more
+    requirements from the same manager.
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (PythonRequirement.multiple('coala_decorators',
+    ...                                                '0.2.1',
+    ...                                                'setuptools',
+    ...                                                '19.2'))
```
Makman2: For better readability I would place an empty line between normal text and a doctest^^

coalib/bears/Bear.py
```diff
@@ -47,6 +48,25 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To indicate the requirements of the bear, give to the ``REQUIREMENTS``
+    tuple instances of the ``PackageRequirement`` class
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (PackageRequirement('pip',
+    ...                                        'coala_decorators',
+    ...                                        '0.2.1'),)
```
Makman2: I think
```python
REQUIREMENTS = (
    PackageRequirement('pip', 'coala_decorators', '0.2.1'),)
```
looks a bit cleaner and saves a line :)

coalib/bears/Bear.py
Makman2: And we need to design if someone wants to not specify a version, meaning he wants the latest.
Maybe if plain strings are provided, this means latest version, and if the user wants to specify it explicitly, pack it inside a tuple:
```
.multiple(('coala_decorators', '0.2.1'), ('setuptools', '19.2'), 'latest-package-version')
```

Makman2: or just providing `latest` as a version string is a special value that omits the version :)

Adrianzatreanu: sure, I'll open an issue about this and fix as I think it should be in another commit too, is this okay :)?

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,30 @@
...
+        """
+        raise NotImplementedError
+
+
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package, version):
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    def multiple(self, *args):
```
justuswilhelm: Would it make sense to turn this into a `@classmethod` ?

Adrianzatreanu: sure, but why?

justuswilhelm: I was looking at the example above calling for `PythonRequirement.multiple` this might not work if `multiple` is not a classmethod - if I am correct. You will most likely get a `TypeError` because you are missing 1 positional argument -- `self`.
Let's say we have this code:
```
class PythonRequirement:
    def multiple(self, *args):
        return args

PythonRequirement.multiple('test', 'hey')
```
you will get a `TypeError` telling you
```
TypeError: unbound method multiple() must be called with PythonRequirement instance as first argument (got str instance instead)
```

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,13 @@ def check(self):
...
         """
         raise NotImplementedError
+
+
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package, version):
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
```
Adrianzatreanu: @justuswilhelm there it is :D

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,13 @@ def check(self):
...
+    def multiple(cls, *args):
+        return args
```
Makman2: still not instantiating `PythonRequirement`s from it :)

Adrianzatreanu: yep this method still needs some work

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
         Check if the requirement is satisfied.
         """
         raise NotImplementedError
+
+
+class PythonRequirement(PackageRequirement):
+
+    def __init__(self, package, version = ""):
```
Makman2: no spaces around `=` for default arguments
+
docstring :)

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
```
Makman2: also here docstring

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += cls(requirement)
+            elif isinstance(requirement, tuple):
+                if len(requirement == 2):
+                    reqs += cls(requirement[0], requirement[1])
+                else
+                    raise NotImplementedError
```
Makman2: I would use `TypeError` here too, `NotImplementedError` is for not implemented functions.
or if you want `ValueError`, but if someone wants to catch them he needs two `except` clauses inside a `try`^^
+ provide a message which states what exactly went wrong ("invalid numbers of args provided in tuple" or so)

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+            else
+                raise TypeError
```
Makman2: please provide a small message here too :+1:

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += cls(requirement)
+            elif isinstance(requirement, tuple):
+                if len(requirement == 2):
```
Makman2:
```
>>> len(True)
TypeError: object of type 'bool' has no len()
```
:)

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+                    reqs += cls(requirement[0], requirement[1])
+                else
+                    raise NotImplementedError
+            else
+                raise TypeError
+        return reqs
```
Makman2: and I would move that class into a separate file^^

Adrianzatreanu:
currently returning a tuple instead of a cls() as it was giving the following error: 
``TypeError: can only concatenate tuple (not "PythonRequirement") to tuple`` in doctests
will fix this tomorrow

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+            if isinstance(requirement, str):
+                reqs += (requirement)
+            elif isinstance(requirement, tuple):
+                if (len(requirement) == 2):
+                    reqs += (requirement[0], requirement[1])
+                else:
+                    raise NotImplementedError
+            else:
+                raise TypeError
+        return reqs
```
Makman2: please move this class into separate file

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+    def multiple(cls, *args):
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (requirement)
+            elif isinstance(requirement, tuple):
+                if (len(requirement) == 2):
+                    reqs += (requirement[0], requirement[1])
+                else:
+                    raise NotImplementedError
```
Makman2: raise a TypeError here too + provide some kind of message

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+            else:
+                raise TypeError
```
Makman2: some message would be good

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (requirement)
+            elif isinstance(requirement, tuple):
+                if (len(requirement) == 2):
```
Makman2: redundant braces

coalib/bears/PackageRequirement.py
Makman2: you don't instantiate PythonRequirement classes any more :O

coalib/bears/PackageRequirement.py
Makman2: maybe this is more elegant too btw:
```
try:
    name, version = requirement
    reqs += (name, version) # <-- just seeing you don't use cls any more...
except ValueError:
    raise TypeError("...")
```

coalib/bears/PackageRequirement.py
```diff
@@ -19,3 +19,24 @@ def check(self):
...
+
+    def __init__(self, package, version=""):
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (requirement)
```
Makman2: does not work, you can't concat `str` + `tuple`

Adrianzatreanu: does it work if i do ``reqs += (requirement,)`` ? if not, how should I add the strings then?

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,22 @@
+from coala_decorators.decorators import generate_repr
+
+
+@generate_repr()
+class PackageRequirement:
```
sils1297: maybe make an own directory, coalib/bears/requirements? This'll grow over time, right? This is the small start of something that may grow really huge over the time

coalib/bears/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
...
+                reqs += (requirement,)
+            elif isinstance(requirement, tuple):
+                try:
+                    name, version = requirement
+                    reqs += (name, version)
+                except ValueError:
+                    raise TypeError('The tuple must have 2 elements.')
+            else:
+                raise TypeError('The arguments need to be tuples or strings.')
+        return reqs
```
Makman2: you still need to return instances of this class, this is what I would expect^^

coalib/bears/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
+from coalib.bears.PackageRequirement import PackageRequirement
+
+
+class PythonRequirement(PackageRequirement):
```
Makman2: maybe a small docstring for the class too?

coalib/bears/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        :param package: A string with the name of the package to be installed.
+        :param version: A number that contains the version. Leave empty to
+                        install latest.
+        """
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
```
Makman2: Please also a brief description what this method is for with a doctest that demonstrates the usage.
Also add `:return:` description^^

coalib/bears/PackageRequirement.py
Makman2: I think this would be good too^^

coalib/bears/PackageRequirement.py
```diff
@@ -0,0 +1,22 @@
...
+                        install the latest.
+        """
+        self.manager = manager
+        self.package = package
+        self.version = version
+
+    def check(self):
+        """
+        Check if the requirement is satisfied.
+        """
```
Makman2: Oh I think you should add a `:return:` spec, so definitely stating "returns `True` if satisfied, `False` if not", as this function gets overriden it :+1:

coalib/bears/PackageRequirement.py
Makman2: And a little class description does also not harm :)

coalib/bears/Bear.py
```diff
@@ -47,6 +49,67 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+    To indicate the requirements of the bear, give to the ``REQUIREMENTS``
+    tuple instances of the ``PackageRequirement`` class:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (
+    ...             PackageRequirement('pip', 'coala_decorators', '0.2.1'),)
```
Makman2: indentation not right, unindent by 1 level^^

coalib/bears/Bear.py
```diff
@@ -47,6 +49,67 @@ class Bear(Printer, LogPrinter):
...
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple('coala_decorators',
+    ...                                               'setuptools')
+
+    And if you choose to mix them, specifying version for some and for some not:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple('coala_decorators',
+    ...                                            ('setuptools', '19.2'))
```
Makman2: indentation not right, you could do
```
...multiple(
    'coala-decorators', ('setuptools', '19.2'))
```

coalib/bears/Bear.py
```diff
@@ -47,6 +49,67 @@ class Bear(Printer, LogPrinter):
...
+    requirements from the same manager. This can receive both tuples of strings,
+    in case you want a specific version, or a simple string, in case you want
+    the latest version to be installed.
+
+    This is the case where you would provide strings only, to install the
+    latest version automatically:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple('coala_decorators',
+    ...                                               'setuptools')
```
sils1297: those tests belong into the requirement class, you can hint it here but a full case isn't in order here!

coalib/bears/Bear.py
```diff
@@ -47,6 +49,67 @@ class Bear(Printer, LogPrinter):
...
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple('coala_decorators',
+    ...                                            ('setuptools', '19.2'))
+
+    In case you provide too many arguments into the tuple, an error will be
+    raised:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = PythonRequirement.multiple('coala_decorators',
+    ...                                 ('setuptools', '19.2', 'colorama'))
```
Makman2: this is actually not a correct usage of multiple, `colorama` needs to be outside of tuple + indentation not right

justuswilhelm: I want to see this whole thing in action, perhaps with a more thorough integration test. Then the inconsistencies might clean themselves up.

Adrianzatreanu: @Makman2 this doctest is exactly proving that colorama should be outside the tuple ...

Adrianzatreanu: it tests the case where it has a different number of tuple members than 2 ^^

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,28 @@
...
+        :param version: A number that contains the version. Leave empty to
+                        install the latest.
+        """
+        self.manager = manager
+        self.package = package
+        self.version = version
+
+    def check(self):
+        """
+        Check if the requirement is satisfied.
```
Makman2: one empty line between brief description and `:xxx:` annotations :)

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,28 @@
+from coala_decorators.decorators import generate_repr
+
+
+@generate_repr()
+class PackageRequirement:
+
```
Makman2: unneeded empty line

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
+from coalib.bears.PackageRequirement import PackageRequirement
+
+
+class PythonRequirement(PackageRequirement):
```
Makman2: no docstring? :)

coalib/bears/requirements/PackageRequirement.py
Adrianzatreanu: PEP says so .. :/

Adrianzatreanu: ah actually i might be wrong, it might be unneeded, but i think i applied coala and it happened, i dont bet

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
...
+                        install latest.
+        """
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
+        :param args: Either strings if the latest version is wanted, or tuples
+                     of strings, where the strings consist of the package and
+                     the version number (which is also a string).
```
Makman2: okay this is confusing. I would put some kind of example, like `("package_name", "version")` and remove the part with `where the string consist of the package and ...`

coalib/bears/requirements/PythonRequirement.py
Adrianzatreanu: most docstrings are in the last commit :)

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (requirement,)
+            elif isinstance(requirement, tuple):
+                try:
+                    name, version = requirement
+                    reqs += (name, version)
```
Makman2: please do `cls(name, version)`

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,33 @@
...
+    def multiple(cls, *args):
+        """
+        :param args: Either strings if the latest version is wanted, or tuples
+                     of strings, where the strings consist of the package and
+                     the version number (which is also a string).
+        """
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (requirement,)
```
Makman2: -> `cls(requirement)`

coalib/bears/requirements/PythonRequirement.py
Makman2: this definitely belongs to this commit^^

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -14,9 +57,14 @@ def __init__(self, package, version=""):
     @classmethod
     def multiple(cls, *args):
         """
+        Creates a tuple of multiple ``pip`` requirements, consisting of
+        either tuples or strings.
+
```
Makman2: this one could get a doctest too :+1:

Makman2: or move the one from above regarding `multiple` here

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,45 @@
...
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
+        Creates a tuple of multiple ``pip`` requirements, consisting of
+        either tuples or strings.
+
+        :param args: Should be tuples of strings: ('packageName', 'version')
+                     or strings: 'packageName' if latest version is wanted.
```
Makman2: you could enclose `packageName` and the tuple before inside double ticks to highlight this as code :)

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,45 @@
...
+        :return:     A tuple containing tuples (package names and versions) and
+                     strings (package names only).
```
Makman2: nope :3

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,45 @@
+from coalib.bears.requirements.PackageRequirement import PackageRequirement
+
+
+class PythonRequirement(PackageRequirement):
+    """
+    This class is a subclass of ``PackageRequirement``, and helps specifying
+    more requirements from ``pip``, without writing the name of the class
```
Makman2: "more"? more than what? :)

Adrianzatreanu: hm I guess you're right, the method ``multiple`` helps you do that, not the class itself.

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,45 @@
...
+        :param version: A number that contains the version. Leave empty to
+                        install latest.
+        """
+        PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
+        Creates a tuple of multiple ``pip`` requirements, consisting of
+        either tuples or strings.
```
Makman2: "consisting of either tuples or string"? It should consist of `PythonRequirements`, right?

coalib/bears/requirements/PythonRequirement.py
Adrianzatreanu: not anymore**

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,45 @@
...
+
+        :param args: Should be tuples of strings: ('packageName', 'version')
+                     or strings: 'packageName' if latest version is wanted.
+        :return:     A tuple containing tuples (package names and versions) and
+                     strings (package names only).
+        """
```
Makman2: Please add also a `:raises TypeError:` annotation

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -25,6 +25,41 @@ def multiple(cls, *args):
...
+
+        The same would happen in case you provide something different than a
+        string or a tuple:
+
+        >>> x = [1, 2, 3, 4]
+        >>> REQUIREMENTS = PythonRequirement.multiple(x)
+        Traceback (most recent call last):
+        ...
+        TypeError: The arguments need to be tuples or strings.
+
```
Makman2: this should be moved into the commit which add `multiple`.
As this is quite big, maybe move the adding of `multiple` also into a separate commit?

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -25,6 +34,12 @@ def check(self):
         """
         Check if the requirement is satisfied.
 
+        >>> PR = PackageRequirement('pip', 'coala_decorators', '0.2.1')
+        >>> PR.check()
+        Traceback (most recent call last):
+        ...
+        NotImplementedError
+
```
Makman2: add both doctests to the according commit please that adds the class and function :+1:

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -25,6 +34,12 @@ def check(self):
         """
         Check if the requirement is satisfied.
 
+        >>> PR = PackageRequirement('pip', 'coala_decorators', '0.2.1')
+        >>> PR.check()
```
Makman2: if it fits into a single line, you could inline:
`PackageRequirement(...).check()`

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -25,6 +25,41 @@ def multiple(cls, *args):
...
+        You should use the ``multiple`` method if you have more
+        requirements from the same manager. This can receive both tuples of
+        strings, in case you want a specific version, or a simple string, in
+        case you want the latest version to be installed.
+
+        This is the case where you would provide strings only, to install the
+        latest version automatically:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...                                     'coala_decorators','setuptools')
```
Makman2: indentation

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -25,6 +25,41 @@ def multiple(cls, *args):
...
+
+        And if you choose to mix them, specifying version for some and for some
+        not:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...                         'coala_decorators',('setuptools', '19.2'))
```
Makman2: also indentation

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -25,6 +25,41 @@ def multiple(cls, *args):
...
+
+        In case you provide too many arguments into the tuple, an error will be
+        raised:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...             'coala_decorators',('setuptools', '19.2', 'colorama'))
```
Makman2: indentation

Makman2: and a missing space after the first `,`

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -25,6 +25,41 @@ def multiple(cls, *args):
...
+        Traceback (most recent call last):
+        ...
+        TypeError: The tuple must have 2 elements.
+
+        The same would happen in case you provide something different than a
+        string or a tuple:
+
+        >>> x = [1, 2, 3, 4]
```
Makman2: we should maybe consider to allow lists too, as they are unpackable^^ but this support can be added later too^^

coalib/bears/requirements/PythonRequirement.py
Makman2: don't you want to assert against another PythonRequirement?
```
>>> REQUIREMENTS == (PythonRequirement('coala-decorators'), PythonRequirement('setuptools'))
True

```

coalib/bears/requirements/PythonRequirement.py
Makman2: and also here some kind of comparing would be good to see what `multiple` actually does

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -6,6 +7,14 @@ class PackageRequirement:
     """
     This class helps keeping track of bear requirements. It should simply
     be appended to the REQUIREMENTS tuple inside the Bear class.
+
+    Two PackageRequirements should always be equal if they have the same
+    manager, package and version:
+
+    >>> PR1 = ('pip', 'coala_decorators', '0.1.0')
+    >>> PR2 = ('pip', 'coala_decorators', '0.1.0')
+    >>> PR1 == PR2
```
Makman2: I think it's generally a good habbit to make variables lower-case, also inside doctests, though this is just preference^^

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -6,6 +7,14 @@ class PackageRequirement:
     """
     This class helps keeping track of bear requirements. It should simply
     be appended to the REQUIREMENTS tuple inside the Bear class.
+
+    Two PackageRequirements should always be equal if they have the same
```
Makman2: please enclose `PackageRequirement`s inside double ticks (except the `s`) :)

coalib/bears/requirements/PackageRequirement.py
Makman2: oh and btw this test does not belong to this class at all, you just compare two tuples

Makman2: using `generate_eq` here as discussed would turn this into a meaningful test when you create `PackageRequirement`s from the above params^^

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,43 @@
+from coala_decorators.decorators import generate_eq
+from coala_decorators.decorators import generate_repr
```
Makman2: --> `from coala_decorators.decorators import generate_eq, generate_repr`

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,43 @@
...
+
+@generate_eq("manager", "package", "version")
+@generate_repr()
+class PackageRequirement:
+    """
+    This class helps keeping track of bear requirements. It should simply
+    be appended to the REQUIREMENTS tuple inside the Bear class.
+
+    Two ``PackageRequirements`` should always be equal if they have the same
+    manager, package and version:
```
Makman2: empty line missing between this and doctest

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    This class is a subclass of ``PackageRequirement``, and helps specifying
+    requirements from ``pip``, without using the manager name.
+    """
+
+    def __init__(self, package, version=""):
+        """
+        Constructs a new PythonRequirement, using the PackageRequirement
+        constructor.
+
```
Makman2: in fact `__init__` needs to be tested here, but there's a meaningful test you can use:
```
PythonRequirement("pack", version="1.2.3") == PackageRequirement("pip", "pack", version="1.2.3")
```

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -1,6 +1,7 @@
 from coala_decorators.decorators import generate_eq
 from coala_decorators.decorators import generate_repr
 
+
```
Makman2: fixup with commit that introduces PackageRequirement

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,66 @@ def __init__(self, package, version=""):
                         install latest.
         """
         PackageRequirement.__init__(self, 'pip', package, version)
+
+    @classmethod
+    def multiple(cls, *args):
+        """
+        Creates a tuple of multiple ``pip`` requirements, consisting of
+        ``PythonRequirements``.
```
Makman2: I think this would be easier + more precise:
`Creates a tuple of multiple ''PythonRequirements''`

coalib/bears/Bear.py
```diff
@@ -56,6 +73,7 @@ class Bear(Printer, LogPrinter):
     """
 
     LANGUAGES = ()
+    REQUIREMENTS = ()
 
     @classproperty
     def name(cls):
```
Makman2: commit message: I think you don't need to mention the doctests, docs and tests are normal so a reader would understand why you added docs and doctests ;)

coalib/bears/requirements/PythonRequirement.py
Adrianzatreanu: great idea

coalib/bears/requirements/PythonRequirement.py
Adrianzatreanu: agreed

coalib/bears/requirements/PackageRequirement.py
Adrianzatreanu: coala has done this. i ran coala when i had it ``from coala_decorators.decorators import generate_repr, generate_eq`` and it done it this way :( dammit coala, should i file an issue about this, it shouldnt work like this?

Makman2: you can wrap it using braces:
```
from ... import (generate_eq,
        generate_repr)
# or
from ... import (
    generate_eq, generate_repr)
```
coala shouldn't corrected it automatically then^^

Makman2: hm wait, this statement does not exceed the 80 char limit at all!
this is a bug in coala, ignore that one^^

coalib/bears/requirements/PythonRequirement.py
Adrianzatreanu: what would I possibly fix at indentation? i think it's okay?

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,81 @@
...
+                           tuples nor strings.
+        """
+        reqs = ()
+        for requirement in args:
+            if isinstance(requirement, str):
+                reqs += (cls(requirement),)
+            elif isinstance(requirement, tuple):
+                try:
+                    name, version = requirement
+                    reqs += (cls(name, version),)
```
justuswilhelm: Appending to a tuple repeatedly feels like an anti-pattern. If you dig going all-tuple you might want to write an inner generator function and then just create a tuple from the generator it returns.

Adrianzatreanu: Hm i dont really know how to do that yet, but i think a tuple suffices for now, doesnt it? I may change this later, or do you think its a must right now :) ?

Makman2: possible but too complicated, you can keep it as it is @Adrianzatreanu 

also exception throwing is a statement and not supported in such expressions^^ we want that behaviour. The question is more: is using a list more performant and then turning it into a tuple?
But anyway, keep it :+1:

Makman2: oh shit, using a list and then collecting is way faster:
```python
>>> from timeit import *
>>> def list_then_collect():
...  x = []
...  for i in range(1000):
...   x.append(i)
...  return tuple(x)
...
>>> def direct_tuple():
...  x = ()
...  for i in range(1000):
...   x += (i,)
...  return x
...
>>> dt = Timer(direct_tuple)
>>> ltc = Timer(list_then_collect)
>>> dt.timeit(1000)
1.2906605905547812
>>> ltc.timeit(1000)
0.10712832800553862
>>> ltc.timeit(1000)
0.10652603528101068
>>> dt.timeit(1000)
1.1920436711211142
```
12x slower with using tuples, so we have learnt as a design consideration: appending to tuples sucks! Better using a list and then collecting!

justuswilhelm: @mischa Great thinking there:) The generator I was thinking about is the following
```python
>>> def slow():
...     result = []
...     for a in range(100):
...             result.append(a)
...     return tuple(result)
... 
>>> def fast():
...     def gen():
...             for a in range(100):
...                     yield a
...     return tuple(gen())
... 
>>> from timeit import Timer
>>> slow = Timer(slow)
>>> fast = Timer(fast)
>>> slow.timeit(1000)
0.018393283997284016
>>> fast.timeit(1000)
0.014460056998359505
```

Adrianzatreanu: ugh @justuswilhelm you tagged someone else,haha :D

justuswilhelm: That's why normal people just sleep at this time😅

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,44 @@
...
+    >>> pr1 = PackageRequirement('pip', 'coala_decorators', '0.1.0')
+    >>> pr2 = PackageRequirement('pip', 'coala_decorators', '0.1.0')
+    >>> pr1 == pr2
+    True
+    """
+
+    def __init__(self, manager: str, package: str, version=""):
+        """
+        Constructs a new PackageRequirement.
```
Makman2: wanna use double ticks around `PackageRequirement`? :)

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,19 @@
...
+class PythonRequirement(PackageRequirement):
+    """
+    This class is a subclass of ``PackageRequirement``, and helps specifying
+    requirements from ``pip``, without using the manager name.
+    """
+
+    def __init__(self, package, version=""):
+        """
+        Constructs a new PythonRequirement, using the PackageRequirement
+        constructor.
```
Makman2: also no double ticks around class names? :)

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -0,0 +1,44 @@
...
+
+        :param manager: A string with the name of the manager (pip, npm, etc).
+        :param package: A string with the name of the package to be installed.
+        :param version: A number that contains the version. Leave empty to
+                        install the latest.
```
Makman2: I think you should not write "number", as this suggerates an integral type. Maybe just: "A version string. Leave empty to specify latest version." (also not that I rephrased the last part, because you don't necessary install stuff with this class, essentially this just defines a requirement^^)

coalib/bears/requirements/PackageRequirement.py
Makman2: oh maybe to test members, you could write a little doctest:
```
>>> pr = PackageRequirement(...)
>>> pr.manager
'pip'
>>> pr.package
'coala_decorators'
>>> pr.version
'0.1'
```

coalib/bears/requirements/PythonRequirement.py
Makman2: maybe also a little doctest like in the commit before, that tests the members of this class :+1:, this also demonstrates the attribute access (meaning how members are named)

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -0,0 +1,19 @@
...
+
+        :param package: A string with the name of the package to be installed.
+        :param version: A number that contains the version. Leave empty to
+                        install latest.
```
Makman2: See previous commit^^

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+        """
+        Creates a tuple of multiple ``PythonRequirements``.
+
+        You should use the ``multiple`` method if you have more
+        requirements from the same manager. This can receive both tuples of
+        strings, in case you want a specific version, or a simple string, in
+        case you want the latest version to be installed.
+
+        This is the case where you would provide strings only, to install the
+        latest version automatically:
```
Makman2: please don't use "install", this class doesn't install currently, just say "specify" or "define"

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...             'coala_decorators', 'setuptools')
```
Makman2: ah indentation thingy: indent by 4 spaces suffices:
```
REQUIREMENTS = PythonRequirement.multiple(
    'coala_decorators', 'setuptools')
```

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+
+        And if you choose to mix them, specifying version for some and for some
+        not:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...             'coala_decorators', ('setuptools', '19.2'))
```
Makman2: here too

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+
+        In case you provide too many arguments into the tuple, an error will be
+        raised:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...             'coala_decorators', ('setuptools', '19.2', 'colorama'))
```
Makman2: and here too :D

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+        >>> x = [1, 2, 3, 4]
+        >>> REQUIREMENTS = PythonRequirement.multiple(x)
+        Traceback (most recent call last):
+        ...
+        TypeError: The arguments need to be tuples or strings.
+
+        :param args:       Should be tuples of strings: ``('packageName',
+                           'version')`` or strings: ``'packageName'`` if latest
+                           version is wanted.
+        :return:           A tuple containing ``PythonRequirements``.
```
Makman2: okay to make docs even more polished (denoting ' now as a single quote)--> ''PythonRequirement''s
technically the class name is written without *s* :P

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -17,3 +17,65 @@ def __init__(self, package, version=""):
...
+        :raises TypeError: In case the tuples contain with more or less than
```
Makman2: ~~with~~

coalib/bears/requirements/PythonRequirement.py
Makman2: we aren't normal, we are cool :P
yeah right I know what you meant, but like said: we can't raise exceptions inside generator expressions, and we do this if we receive something non-stringy or non-tupely. And it makes sense to raise a `TypeError` here :+1:

coalib/bears/Bear.py
```diff
@@ -47,6 +49,21 @@ class Bear(Printer, LogPrinter):
     >>> class SomeBear(Bear):
     ...     LANGUAGES = "Java"
 
+
```
Makman2: a newline too much

coalib/bears/Bear.py
```diff
@@ -47,6 +49,21 @@ class Bear(Printer, LogPrinter):
...
+    To indicate the requirements of the bear, give to the ``REQUIREMENTS``
+    tuple instances of the ``PackageRequirement`` class:
```
Makman2: hmm... give to a tuple ... rather saying "Assign ``REQUIREMENTS`` a tuple with instances of ..."?

coalib/bears/Bear.py
```diff
@@ -47,6 +49,21 @@ class Bear(Printer, LogPrinter):
...
+    ...     REQUIREMENTS = (
+    ...         PackageRequirement('pip', 'coala_decorators', '0.2.1'),)
+
+    If your bear uses requirements from a manager we have a subclass from,
+    you can use the subclass, such as ``PythonRequirement``, without specifying
+    manager:
+
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = (PythonRequirement('coala_decorators', '0.2.1'),)
+
```
Makman2: oh please demonstrate the `multiple` method, really really useful here :+1:

coalib/bears/requirements/PythonRequirement.py
Makman2: okay you can use exceptions when you write an inner function using `yield` (haven't thought about that), but that's somehow uncool imo^^ I think there's no great speed improvement also^^

coalib/bears/Bear.py
Adrianzatreanu: done, a few lines under this :D

Makman2:
ack 4325d0271b5f113c56dc9a28e7d22e55177dd6a9

Makman2:
ack c6b61dcfc95ee743e172b292d73014a425c2a4d9

Makman2:
ack b01d61fb4b7a32ea636648824d615f4cc790c7d9

Makman2:
ack da8db85ef7d3e7f09aa5e4b999aedeec380bebf4

Makman2:
@rultor merge

nice one, lgtm

Makman2:
@rultor merge

PR #2179, opened 2016-05-26
PythonRequirement: Add lists to multiple

Adrianzatreanu:
Multiple method supports taking lists
as arguments.

Fixes https://github.com/coala-analyzer/coala/issues/2178

### DIFF
--- a/coalib/bears/requirements/PythonRequirement.py
+++ b/coalib/bears/requirements/PythonRequirement.py
@@ -47,6 +47,11 @@ class PythonRequirement(PackageRequirement):
         >>> REQUIREMENTS = PythonRequirement.multiple(
         ...     'coala_decorators', ('setuptools', '19.2'))
 
+        Lists are also valid arguments:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     ['coala_decorators', '19.2'],)
+
         In case you provide too many arguments into the tuple, an error will be
         raised:
 
@@ -54,16 +59,7 @@ class PythonRequirement(PackageRequirement):
         ...     'coala_decorators', ('setuptools', '19.2', 'colorama'))
         Traceback (most recent call last):
         ...
-        TypeError: The tuple must have 2 elements.
-
-        The same would happen in case you provide something different than a
-        string or a tuple:
-
-        >>> x = [1, 2, 3, 4]
-        >>> REQUIREMENTS = PythonRequirement.multiple(x)
-        Traceback (most recent call last):
-        ...
-        TypeError: The arguments need to be tuples or strings.
+        TypeError: Too many elements provided.
 
         :param args:       Should be tuples of strings: ``('packageName',
                            'version')`` or strings: ``'packageName'`` if latest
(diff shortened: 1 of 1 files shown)


coalib/bears/requirements/PythonRequirement.py
```diff
@@ -83,6 +103,15 @@ def multiple(cls, *args):
...
                     raise TypeError('The tuple must have 2 elements.')
+            elif isinstance(requirement, list):
+                if len(requirement) == 1:
+                    reqs.append(cls(requirement),)
+                else:
+                    try:
+                        name, version = requirement
+                        reqs.append(cls(name, version),)
+                    except ValueError:
+                        raise TypeError('The list must have max 2 elements.')
```
sils1297: nonononono this is code duplication!

Just remove the whole isinstance checking, it's not pythonic. Just invoke the len and do the pattern matching and if that fails make a TypeError that you need any iterable nongenerator with twoish elements.

sils1297:
7d351b3 needs work

coalib/bears/requirements/PythonRequirement.py
justuswilhelm: You could try turning the `elif` branch for tuples into just an `else` and catch any errors when unpacking the requirement into `name, version`. It would make the whole interface a bit easier for the developer using this. Basically, I want two options for every element in `args`: 1) Requirement without version, 2) Requirement with version. It should also work when I am passing generators, e.g. (('a', 'b') for _ in range(2)), to `multiple`

sils1297: basically when writing python you usually don't want to know what an object is, you would rather do stuff with it and if it can't do it you expect it to raise an exception. This way people could write new objects with the same capabilities that are then supported generically since they have the capabilities.

Makman2: I thought of it too, but I'm a bit struggling that it's even possible to unpack sets, which yields in unknown unpacking behaviour, because items inside sets are not positional fixed. So imo it makes sense to restrict types here to tuples and lists.
But I agree, code duplication is bad, just do `if isinstance(..., tuple) or isinstance(..., list)` and change the TypeError message to something more general

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -47,6 +47,26 @@ def multiple(cls, *args):
...
+        Lists are also valid arguments:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     ['coala_decorators', '19.2'],)
+
+        Specifying a list with only one member will also work, implying the
+        latest version should be considered:
+
+        >>> REQUIREMENTS = PythonRequirement.multiple(
+        ...     "setuptools", ['coala_decorators'])
```
Makman2: hmm... just forbid that :3

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -77,12 +91,14 @@ def multiple(cls, *args):
         for requirement in args:
             if isinstance(requirement, str):
                 reqs.append(cls(requirement),)
-            elif isinstance(requirement, tuple):
+            elif isinstance(requirement, tuple) or isinstance(requirement,
```
AbdealiJK: You can do `isinstance(requirement, (tuple, list))`:
```
>>> isinstance((), (tuple, list))
True
>>> isinstance([], (tuple, list))
True
```

Adrianzatreanu: im not anymore, lasse doesnt like ``isinstance``, im doing it without

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -77,12 +91,14 @@ def multiple(cls, *args):
...
+                                                              list):
                 try:
                     name, version = requirement
                     reqs.append(cls(name, version),)
```
AbdealiJK: Probably not related to this PR ... but why is there `reqs.append(cls(name, version),)` ? Why is there a comma ?

Adrianzatreanu: because initially it was a tuple, now its a list, good observation.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2179/commits/29a1526eee3924b2206b79911858a32a35d31950

sils1297:
@rultor merge

PR #2185, opened 2016-05-26
PackageRequirement and stuff

Adrianzatreanu:
Some needed changes for the requirement classes.


sils1297:
ack https://github.com/coala-analyzer/coala/pull/2185/commits/58a2e2b20308ed8d95df33558861926c26ca57cd

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -49,3 +49,38 @@ def check(self):
         :return: Returns True if satisfied, False if not.
         """
         raise NotImplementedError
+
+    @classmethod
+    def multiple(cls, *args):
```
sils1297: nah I don't think this is good. So this class is abstract, i.e. you're not supposed to instantiate it!

If you did something like:

```python
class PackageRequirement:
    ...
    @classmethod
    def multiple(cls, *args):
        manager = cls.manager
        ...
```

and then

```python
class PythonRequirement:
    ...
    manager = 'pip'
```

then you would never run `PackageRequirement.multiple` because that's an abstract class but rather `PythonRequirement.multiple(('perlcritic', '6.2.3'))`

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2185/commits/e4fcaba521fe7c492c9776ebde5fac96353a984f

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2185/commits/6cba5e9d69a28549b0e6d135ff6f3c13d040edd1

coalib/bears/requirements/PackageRequirement.py
AbdealiJK: @Adrianzatreanu What @sils1297 is saying (I think) is that if you have let's say a class like `Bear` ... this is a template right ? It has no features but does boiler plate code for all other bears. So, if a user were to create an object of Bear `b = Bear()` it is actually not valid, as it's just meant as a template.

Now, this classmethod provides a way to create an object of an abstract class, which is non-intuitive, and confusing.
i.e. `PackageRequirement.multiple(('pip', 'coala_decorators', '0.1.0'))` can create an object of `PackageRequirement`.

There are many possible solutions (assuming I understood the issue @sils1297 is trying to tell us):
 - I personally don't mind this. It's hacky, but it works. And if a user does try to create an object of a class we say is abstract in the docs, they should expect unusual outputs.
 - You can check that `cls` is not this class itself, but is a derived class. This can be done using the `issubclass` builtin function or even with a simple `if cls == PackageRequirement:` -> `raise NotImplementedError`
 - Finally, python has an "abstract base class" module - https://docs.python.org/3/library/abc.html I've never used it myself, but it's the official way to do it. It has something called `abstractclassmethod` ... but I don't think it's what you want here.

2nd option seems best to me.

Adrianzatreanu:
reack 274ba1a 6115dbb

coalib/bears/requirements/PythonRequirement.py
```diff
@@ -5,6 +5,37 @@ class PythonRequirement(PackageRequirement):
     """
     This class is a subclass of ``PackageRequirement``, and helps specifying
     requirements from ``pip``, without using the manager name.
+
```
Adrianzatreanu: this was added here as there was no way to test it in the base class, and this solves coverage.

AbdealiJK: Isnt an easier way to simply do:
```
>>> class PyRequirement(PackageRequirement): pass
>>> # Then these doctests 
```
in the earlier place ?
If you need to do it here, it shouldn't be a doctest, as the documentation is confusing rather than helping.

Adrianzatreanu: why is it confusing? it explains all cases in which you would use the method and how it would work, imo its useful?

AbdealiJK: It's confusing because it's not in the right place where it "should" be.
It's like reading Geography in a History book :P

Adrianzatreanu: Where do you think it should be then?

AbdealiJK: Under `def multiple(...)`. You can do:

```
 +        Should not be instances of ``PackageRequirement``, as this is an
 +        abstract class:
 +
 +        >>> PackageRequirement.multiple(('pip', 'coala_decorators', '0.1.0'),)
 +        ...
 +        NotImplementedError
 +
 +        It can only be used for requirements of the same manager. For example,
 +        consider a manager ``XYZRequirement`` that inherits from
 +        PackageRequirement:
 +        >>> class XYZRequirement(PackageRequirement): pass
+    This is the case where you would provide strings only, to specify the
 +    latest version automatically:
 +
 +    >>> REQUIREMENTS = XYZRequirement.multiple(
 +    ...     'package1', 'package2')
 +
 +    And if you choose to mix them, specifying version for some and for some
 +    not:
 +
 +    >>> REQUIREMENTS = XYZRequirement.multiple(
 +    ...     'package1', ('package2', '2.0'))
 +
 +    Lists are also valid arguments:
 +
 +    >>> REQUIREMENTS = XYZRequirement.multiple(
 +    ...     ['package1', '1.0'],)
 +
 +    In case you provide too many arguments into the tuple, an error will be
 +    raised:
 +
 +    >>> REQUIREMENTS = XYZRequirement.multiple(
 +    ...     'package1', ('package2', '2.0', '# some comment'))
 +    Traceback (most recent call last):
 +    ...
 +    TypeError: Too many elements provided.
```

AbdealiJK: I write that pretty quickly. Please modify appropriately to correct errors

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -49,3 +49,38 @@ def check(self):
...
+        """
+        Creates a tuple of multiple instances of a class.
+
+        Should not be instances of ``PackageRequirement``, as this is an
+        abstract class:
+
+        >>> PackageRequirement.multiple(('pip', 'coala_decorators', '0.1.0'),)
+        Traceback (most recent call last):
+        ...
+        NotImplementedError
```
sils1297: you would rather make an example here of a subclass and how it has to be implemented so this works

sils1297: as abdeali pointed out below :)

Adrianzatreanu: great, just done that, check it out @AbdealiJK :)

coalib/bears/requirements/PackageRequirement.py
justuswilhelm: I'm personally a fan of the 3rd option as it has pretty good run time checks included that you would have to implement yourself otherwise. Embrace Python 3 👍

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2185/commits/8258f9cafed499f4894f62ee81c8d2862cad5726

Adrianzatreanu:
reack 7cfef78

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -49,3 +49,73 @@ def check(self):
...
+
+        It can only be used for requirements of the same manager. For example,
+        consider a manager ``XYZRequirement`` that inherits from
+        PackageRequirement. This subclass will have the manager set to XYZ:
+
+        >>> class XYZRequirement(PackageRequirement):
+        ...     manager = 'xyz'
```
sils1297: I think this manager variable doesn't need to be exposed to the user, can we make it `_manager` at least? Also it's a constant so it should rather be `_MANAGER` I think. This doesn't belong into this PR but maybe file an issue?

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2185/commits/9b40c5c6a712e377d5c76b6780c4fc745a91ea1d

sils1297:
Rebase and issue filing needed.

coalib/bears/requirements/PackageRequirement.py
Adrianzatreanu: Sure, can do that

Adrianzatreanu: https://github.com/coala-analyzer/coala/issues/2203

Adrianzatreanu:
done :)

Adrianzatreanu:
reack 703e920 4e007e2

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2185/commits/76d41f037f571bab9e418925c868a3b2a18f59c6

Adrianzatreanu:
@rultor merge

Adrianzatreanu:
@rultor merge

sils1297:
included in https://github.com/coala-analyzer/coala/pull/2204

PR #2198, opened 2016-05-30
ExternalBearWrap: write external_bear_wrap decorator base

Redridge:
external_bear_wrap decorator that gets an executable name as argument.

### DIFF
--- a/coalib/bearlib/abstractions/ExternalBearWrap.py
+++ b/coalib/bearlib/abstractions/ExternalBearWrap.py
@@ -1,0 +1,221 @@
+import json
+import inspect
+from functools import partial
+from collections import OrderedDict
+
+from coalib.bears.LocalBear import LocalBear
+from coala_decorators.decorators import enforce_signature
+from coalib.misc.Shell import run_shell_command
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def _prepare_options(options):
+    """
+    Checks for illegal options and raises ValueError.
+
+    :param options:
+        The options dict that contains user/developer inputs.
+    :raises ValueError:
+        Raised when illegal options are specified.
+    """
+    allowed_options = {"executable",
+                       "settings"}
+
+    # Check for illegal superfluous options.
+    superfluous_options = options.keys() - allowed_options
+    if superfluous_options:
+        raise ValueError(
+            "Invalid keyword arguments provided: " +
+            ", ".join(repr(s) for s in sorted(superfluous_options)))
+
+    if not 'settings' in options:
+        options['settings'] = {}
+
+
+def _create_wrapper(klass, options):
+    NoDefaultValue = object()
+
+    class ExternalBearWrapBase(LocalBear):
+
+        @staticmethod
+        def create_arguments():
+            """
+            This method has to be implemented by the class that uses
+            the decorator in order to create the arguments needed for
+            the executable.
+            """
+            return ()
+
+        @classmethod
+        def get_executable(cls):
+            """
+            Returns the executable of this class.
+
+            :return:
+                The executable name.
+            """
+            return options["executable"]
+
+        @staticmethod
+        def _normalize_desc(description, setting_type,
+                            default_value=NoDefaultValue):
+            """
+            Normalizes the description of the parameters only if there
+            is none provided.
+
+            :param description:
+                The parameter description to be modified in case it is empty.
+            :param setting_type:
+                The type of the setting. It is needed to create the final
+                tuple.
+            :param default_value:
+                The default value of the setting.
+            :return:
+                A value for the OrderedDict in the ``FunctionMetadata`` object.
+            """
+            if description == "":
+                description = FunctionMetadata.str_nodesc
+
+            if default_value is NoDefaultValue:
+                return (description, setting_type)
+            else:
+                return (description + " " +
+                        FunctionMetadata.str_optional.format(default_value),
+                        setting_type, default_value)
+
+        @classmethod
+        def get_non_optional_params(cls):
+            """
+            Fetches the non_optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                ``FunctionMetadata`` object.
+            """
+            non_optional_params = {}
+            for setting_name, description in options['settings'].items():
+                if len(description) == 2:
+                    non_optional_params[
+                        setting_name] = cls._normalize_desc(description[0],
+                                                            description[1])
+            return OrderedDict(non_optional_params)
+
+        @classmethod
+        def get_optional_params(cls):
+            """
+            Fetches the optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                ``FunctionMetadata`` object.
+            """
+            optional_params = {}
+            for setting_name, description in options['settings'].items():
+                if len(description) == 3:
+                    optional_params[
+                        setting_name] = cls._normalize_desc(description[0],
+                                                            description[1],
+                                                            description[2])
+            return OrderedDict(optional_params)
+
+        @classmethod
+        def get_metadata(cls):
+            metadata = FunctionMetadata(
+                'run',
+                optional_params=cls.get_optional_params(),
+                non_optional_params=cls.get_non_optional_params())
+            metadata.desc = inspect.getdoc(cls)
+            return metadata
+
+        @classmethod
+        def _prepare_settings(cls, settings):
+            """
+            Adds the optional settings to the settings dict in-place.
+
+            :param settings:
+                The settings dict.
+            """
+            opt_params = cls.get_optional_params()
+            for setting_name, description in opt_params.items():
+                if setting_name not in settings:
+                    settings[setting_name] = description[2]
+
+        @staticmethod
+        def get_severity(result):
+            """
+            Returns the severity corresponding to the one from the
+            parsed JSON.
+
+            :param result:
+                A result dict extracted from the parsed JSON.
+            :return:
+                A RESULT_SEVERITY value.
+            :raises KeyError:
+                Raised when an invalid severity is specified.
+            """
+            return (RESULT_SEVERITY.str_dict[result['severity']]
+                    if 'severity' in result else RESULT_SEVERITY.NORMAL)
+
+        def parse_output(self, out, filename):
+            """
+            Parses the output JSON into Result objects.
+
+            :param out:
+                Raw output from the given executable (should be JSON).
+            :param filename:
+                The filename of the analyzed file. Needed to
+                create the Result objects.
+            :return:
+                An iterator yielding ``Result`` objects.
+            """
+            output = json.loads(out)
+
+            for result in output:
+                yield Result.from_values(
+                    origin=self,
+                    message=result['message'],
+                    file=filename,
+                    line=result['line'],
+                    end_line=result.get('end_line', None),
+                    column=result.get('column', None),
+                    end_column=result.get('end_column', None),
+                    severity=self.get_severity(result),
+                    debug_msg=result.get('debug_msg', ""),
+                    additional_info=result.get('additional_info', ""))
+
+        def run(self, filename, file, **settings):
+            self._prepare_settings(settings)
+            json_string = json.dumps({'filename': filename,
+                                      'file': file,
+                                      'settings': settings})
+
+            args = self.create_arguments()
+            try:
+                args = tuple(args)
+            except TypeError:
+                self.err("The given arguments "
+                         "{!r} are not iterable.".format(args))
+                return
+
+            shell_command = (self.get_executable(),) + args
+            out, err = run_shell_command(shell_command, json_string)
+
+            return self.parse_output(out, filename)
+
+    result_klass = type(klass.__name__, (klass, ExternalBearWrapBase), {})
+    result_klass.__doc__ = klass.__doc__ if klass.__doc__ else ""
+    return result_klass
+
+
+@enforce_signature
+def external_bear_wrap(executable: str, **options):
+
+    options["executable"] = executable
+    _prepare_options(options)
+
+    return partial(_create_wrapper, options=options)
--- a/tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
+++ b/tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
@@ -1,0 +1,197 @@
+import os
+import sys
+import json
+import unittest
+
+from coalib.bearlib.abstractions.ExternalBearWrap import external_bear_wrap
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def get_testfile_path(name):
+    """
+    Gets the full path to a testfile inside the same directory.
+
+    :param name: The filename of the testfile to get the full path for.
+    :return:     The full path to given testfile name.
+    """
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                        name)
+
+
+class ExternalBearWrapComponentTest(unittest.TestCase):
+
+    class Dummy:
+        pass
+
+    class TestBear:
+
+        @staticmethod
+        def create_arguments():
+            return (os.path.join(
+                os.path.dirname(__file__),
+                "test_external_bear.py"),)
+
+    class WrongArgsBear:
+
+        @staticmethod
+        def create_arguments():
+            return 1
+
+    def setUp(self):
+        self.section = Section("TEST_SECTION")
+
+        self.test_program_path = get_testfile_path("test_external_bear.py")
+
+        self.testfile_path = get_testfile_path("test_file.txt")
+        with open(self.testfile_path, mode="r") as fl:
+            self.testfile_content = fl.read().splitlines(keepends=True)
+
+    def test_decorator_invalid_parameters(self):
+        with self.assertRaises(ValueError) as cm:
+            external_bear_wrap("exec", invalid_arg=88)
+        self.assertEqual(
+            str(cm.exception),
+            "Invalid keyword arguments provided: 'invalid_arg'")
+
+    def test_decorator_invalid_parameter_types(self):
+        # Provide some invalid severity maps.
+        with self.assertRaises(TypeError):
+            external_bear_wrap(executable=1337)
+
+    def test_get_executable(self):
+        uut = (external_bear_wrap("exec")(self.TestBear))
+        self.assertEqual(uut.get_executable(), "exec")
+
+    def test_get_severity(self):
+        uut = (external_bear_wrap("exec")(self.Dummy))
+        result_major = {'severity': "MAJOR"}
+        result_normal = {'severity': "NORMAL"}
+        result_info = {'severity': "INFO"}
+        result_none = {}
+        result_wrong = {'severity': "Info"}
+
+        self.assertEqual(uut.get_severity(result_major),
+                         RESULT_SEVERITY.MAJOR)
+        self.assertEqual(uut.get_severity(result_normal),
+                         RESULT_SEVERITY.NORMAL)
+        self.assertEqual(uut.get_severity(result_info),
+                         RESULT_SEVERITY.INFO)
+        self.assertEqual(uut.get_severity(result_none),
+                         RESULT_SEVERITY.NORMAL)
+        with self.assertRaises(KeyError):
+            uut.get_severity(result_wrong)
+
+    def test_create_arguments_fail(self):
+        uut = (external_bear_wrap("exec")(self.Dummy))
+        self.assertEqual(uut.create_arguments(), ())
+
+    def test_create_arguments_non_iterable(self):
+        uut = (external_bear_wrap("exec")
+               (self.WrongArgsBear)
+               (self.section, None))
+        with self.assertRaises(TypeError):
+            res = list(uut.run(self.testfile_path, self.testfile_content))
+
+    def test_invalid_output(self):
+        broken_json = json.dumps([{'broken': "JSON"}])[:-1]
+        uut = (external_bear_wrap("exec")(self.Dummy)(self.section, None))
+        with self.assertRaises(ValueError):
+            # Something needs to be done with the result otherwise
+            # parse_output will not yield and thus will not raise the ValueError
+            list(uut.parse_output(broken_json, "some_file"))
+
+    def test_setting_desc(self):
+        uut = (external_bear_wrap("exec",
+                                  settings={
+                                     "asetting": ("", bool),
+                                     "bsetting": ("", bool, True),
+                                     "csetting": ("My desc.", bool, False),
+                                     "dsetting": ("Another desc", bool),
+                                     "esetting": ("", int, None)
+                                     })(self.Dummy))
+        metadata = uut.get_metadata()
+        self.assertEqual(metadata.non_optional_params["asetting"][0],
+                         FunctionMetadata.str_nodesc)
+        self.assertEqual(metadata.optional_params["bsetting"][0],
+                         FunctionMetadata.str_nodesc + " " +
+                         FunctionMetadata.str_optional.format(True))
+        self.assertEqual(metadata.optional_params["csetting"][0], "My desc." +
+                         " " + FunctionMetadata.str_optional.format(False))
+        self.assertEqual(metadata.non_optional_params["dsetting"][0],
+                         "Another desc")
+        self.assertEqual(metadata.optional_params["esetting"][0],
+                         FunctionMetadata.str_nodesc + " " +
+                         FunctionMetadata.str_optional.format(None))
+
+    def test_optional_settings(self):
+        uut = (external_bear_wrap(sys.executable,
+                                  settings={"a": ("", bool),
+                                            "b": ("", bool, False),
+                                            "c": ("", bool, True)})
+               (self.TestBear)
+               (self.section, None))
+        results = list(uut.run(self.testfile_path, self.testfile_content,
+                               a=False))
+        expected = [
+            Result.from_values(
+                origin=uut,
+                message="This is wrong",
+                file=self.testfile_path,
+                line=1,
+                severity=RESULT_SEVERITY.MAJOR
+                ),
+            Result.from_values(
+                origin=uut,
+                message="This is wrong too",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.INFO)]
+        self.assertEqual(results, expected)
+
+        results = list(uut.run(self.testfile_path, self.testfile_content,
+                               a=True))
+        expected = [
+            Result.from_values(
+                origin=uut,
+                message="This is wrong",
+                file=self.testfile_path,
+                line=1,
+                severity=RESULT_SEVERITY.NORMAL
+                ),
+            Result.from_values(
+                origin=uut,
+                message="This is wrong too",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.NORMAL)]
+        self.assertEqual(results, expected)
+
+    def test_settings(self):
+        uut = (external_bear_wrap(sys.executable,
+                                  settings={"a": ("", bool),
+                                            "b": ("", bool, False),
+                                            "c": ("", bool, True)})
+               (self.TestBear)
+               (self.section, None))
+        results = list(uut.run(self.testfile_path, self.testfile_content,
+                               a=False, b=True, c=False))
+        expected = [
+            Result.from_values(
+                origin=uut,
+                message="This is wrong",
+                file=self.testfile_path,
+                line=1,
+                severity=RESULT_SEVERITY.MAJOR,
+                debug_msg="Sample debug message"
+                ),
+            Result.from_values(
+                origin=uut,
+                message="Different message",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.INFO)]
+        self.assertEqual(results, expected)
--- a/tests/bearlib/abstractions/external_bear_wrap_testfiles/test_external_bear.py
+++ b/tests/bearlib/abstractions/external_bear_wrap_testfiles/test_external_bear.py
@@ -1,0 +1,32 @@
+import sys
+import json
+
+if __name__ == "__main__":
+
+    line = sys.stdin.read()
+    args = json.loads(line)
+    settings = args['settings']
+
+    out = [{
+            'message': "This is wrong",
+            'file': args['filename'],
+            'severity': "MAJOR",
+            'line': 1
+        }, {
+            'message': "This is wrong too",
+            'file': args['filename'],
+            'severity': "INFO",
+            'line': 3}]
+
+    if settings['a']:
+        for res in out:
+            res['severity'] = "NORMAL"
+
+    if settings['b']:
+        out[0]['debug_msg'] = "Sample debug message"
+
+    if not settings['c']:
+        out[1]['message'] = "Different message"
+
+    json_dump = json.dumps(out)
+    sys.stdout.write(json_dump)
--- a/tests/bearlib/abstractions/external_bear_wrap_testfiles/test_file.txt
+++ b/tests/bearlib/abstractions/external_bear_wrap_testfiles/test_file.txt
@@ -1,0 +1,3 @@
+1st line
+2nd line
+3rd line


coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,69 @@
...
+
+        @staticmethod
+        def get_executable():
+            """
+            Returns the executable of this class.
+
+            :return:
+                The executable name.
+            """
+            return options["executable"]
```
sils1297: don't you rather wanna make it a property?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,69 @@
...
+        def _get_optional_metadata(cls):
+            if hasattr(cls, 'optional_args'):
+                return FunctionMetadata.from_function(
+                    cls.optional_args,
+                    omit={"self"})
+            return None
+
+        @classmethod
+        def get_metadata(cls):
+            metadata = cls._get_optional_metadata()
```
sils1297: metadata might be None here

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,69 @@
...
+            metadata.desc = (
+                    "{}\n\nThis bear uses the {!r} executable.".format(
+                        inspect.getdoc(cls), cls.get_executable()))
```
sils1297: I wouldn't do this, we're just removing it for `linter` as well

Redridge:
Yes, I just pushed it here for the time being.
I forgot to label it as WIP. It is done now

Redridge:
For some reason appveyor seems to fail :(

abhsag24:
@Redridge  so its a file not found error probably caused by not escaping backslashes in windows path right?

Redridge:
I joined the paths using `os.path.join` and generally the `os` module which I though handles these problems

sils1297:
ping @Makman2 , maybe you can spot the problem

Makman2:
yeah windows doesn't allow to start processes just by their extension, you need to provide an executable or cmd/batch file. If you want to start python scripts, use a executable `sys.executable` and provide as first argument the name of the script^^

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,126 @@
...
+from coalib.results.Result import Result
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def _prepare_options(options):
+    """
+    Prepares options for a given options dict in-place.
+
+    :param options:
+        The options dict that contains user/developer inputs.
```
Udayan12167: The doc comment should start in the same line

Udayan12167:
Proper documentation required for a lot of methods

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,105 @@
...
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                        name)
+
+
+class ExternalBearWrapComponentTest(unittest.TestCase):
+
+    class Dummy:
+        pass
+
+    class TestBear:
```
Udayan12167: The purpose of this class is to only fetch the executable?

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,105 @@
...
+import json
+import unittest
+
+from coalib.bearlib.abstractions.ExternalBearWrap import external_bear_wrap
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.settings.Section import Section
+
+
+def get_testfile_name(name):
```
Udayan12167: Correct name of the method

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def test_invalid_output(self):
+        broken_json = json.dumps([{'broken': "JSON"}])[:-1]
+        uut = (external_bear_wrap("exec")(self.Dummy)(self.section, None))
+        with self.assertRaises(ValueError):
+            # Something needs to be done with the result otherwise
+            # parse_output will not yield and thus will not raise the ValueError
+            list(uut.parse_output(broken_json, "some_file"))
+
+    def test_setting_desc(self):
```
Udayan12167: What happens when a non optional param is not available? Could you add a test for that.

Redridge: In that case coala asks you for that param, right?

Udayan12167: I would suggest just test if that works the same way here.

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
Redridge: Yes, but I refactored the code a bit. Now the class will be responsible for overwriting `create_arguments()` similar to linter. I found this to be a better way since it makes it easier to use languages that use interpreters.

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
Redridge: Yes I can try to write a test but I don't know how that would be handled since by default coala would prompt you for the value of the setting.

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def _prepare_options(options):
+    """
+    Prepares options for a given options dict in-place.
+
+    :param options: The options dict that contains user/developer inputs.
+    """
+    allowed_options = {
+                        "executable",
```
sils1297: move this up a line and unindent by one space

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def _prepare_options(options):
+    """
+    Prepares options for a given options dict in-place.
+
+    :param options: The options dict that contains user/developer inputs.
```
sils1297: add `:raises` as this is what this actually is about right?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+                'run',
+                optional_params=cls.get_optional_params(),
+                non_optional_params=cls.get_non_optional_params())
+            return metadata
+
+        def parse_output(self, out, filename):
+            try:
+                output = json.loads(out)
+            except ValueError:
+                raise ValueError
```
Makman2: hm... then don't do a try except :)

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+    class ExternalBearWrapBase(LocalBear):
+
+        @staticmethod
+        def create_arguments():
+            """
+            This method has to be implemented by the class that uses
+            the decorator in order to create the arguments needed for
+            the executable
+            """
+
```
Makman2: I think this newline is not necessary :)

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+            desc = "No description given."
+            if len(val) == 2:
+                return (desc, val[1])
+            else:
+                desc += " Defaults to " + str(val[2])
+                return (desc, val[1], val[2])
+
+        @classmethod
+        def get_non_optional_params(cls):
+            dict = {}
```
Makman2: please don't override the built-in `dict` function :+1:

Makman2: call it `dct` or something, or even better `non_optional_param` or something that describes the purpose of that variable :+1:

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+        def get_executable(cls):
+            """
+            Returns the executable of this class.
+
+            :return: The executable name.
+            """
+            return options["executable"]
+
+        @staticmethod
+        def normalize_desc(val):
```
Makman2: public API, please add docstring

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+        def get_non_optional_params(cls):
+            dict = {}
+            for key, value in options['settings'].items():
+                if len(value) == 2:
+                    dict[key] = cls.normalize_desc(value)
+            return OrderedDict(dict)
+
+        @classmethod
+        def get_optional_params(cls):
+            dict = {}
```
Makman2: same here

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+                    dict[key] = cls.normalize_desc(value)
+            return OrderedDict(dict)
+
+        @classmethod
+        def get_metadata(cls):
+            metadata = FunctionMetadata(
+                'run',
+                optional_params=cls.get_optional_params(),
+                non_optional_params=cls.get_non_optional_params())
+            return metadata
```
Makman2: I would inline the variable :+1:

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+                    origin=self,
+                    message=result['message'],
+                    file=filename,
+                    line=result['line'])
+
+        def run(self, filename, file, **kwargs):
+            json_string = json.dumps({
+                                    'filename': filename,
+                                    'file': file
+                                    })
```
Makman2: -->
```
json_string = json.dumps({'filename': filename,
                          'file': file})
```

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+            except TypeError:
+                self.err("The given arguments "
+                         "{!r} are not iterable.".format(args))
+                return
+
+            shell_command = (self.get_executable(),) + args
+            out, err = run_shell_command(shell_command, json_string)
+
+            for result in self.parse_output(out, filename):
+                yield result
```
Makman2: --> `yield from self.parse_output(out, filename)`

Makman2: or just do `return self.parse_output(...)` :+1:

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,134 @@
...
+
+    result_klass = type(klass.__name__, (klass, ExternalBearWrapBase), {})
+    result_klass.__doc__ = klass.__doc__ if klass.__doc__ else ""
+    return result_klass
+
+
+@enforce_signature
+def external_bear_wrap(executable: str,
+                       **options):
```
Makman2: I would put the signature into a single line, makes it easier to read (at least for me :D)

Makman2: \+ documentation missing, but this can be written after merge^^

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+from coalib.misc.Shell import run_shell_command
+from coalib.results.Diff import Diff
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def _prepare_options(options):
+    """
+    Prepares options for a given options dict in-place.
```
sils1297: this function doesn't modify the dict!

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: this param description doesn't really tell me how I have to structure my dict

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: you could give `val` a better name, this one is useless and I still don't fully understand what this function is about

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+            :return: A value for the OrderedDict in the FunctionMetada object.
+            """
+            if val[0] != "":
+                return val
+            desc = "No description given."
+            if len(val) == 2:
+                return (desc, val[1])
+            else:
+                desc += " Defaults to " + str(val[2])
+                return (desc, val[1], val[2])
```
sils1297: this code looks similar to something we already have somewhere, duplication?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+            :param result: A result dict extracted from the parsed JSON
+
+            :return: A RESULT_SEVERITY
+            """
+            if not 'severity' in result or result['severity'] == "NORMAL":
+                return RESULT_SEVERITY.NORMAL
+            elif result['severity'] == "MAJOR":
+                return RESULT_SEVERITY.MAJOR
+            elif result['severity'] == "INFO":
+                return RESULT_SEVERITY.INFO
```
sils1297: ahm it's an Enum, right? It should have some kind of reverse_dict automatically generated that does this translation for you :)

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+
+        def parse_output(self, out, filename):
+            """
+            Parses the output JSON into Result objects
+
+            :param out: Raw output from the given executable (should be JSON)
+            :param filename: The filename of the analyzed file. Needed to
+                create the Result objects.
+
+            :return: The Result objects.
```
sils1297: you don't return them, you yield them!Difference.

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+            """
+            output = json.loads(out)
+
+            for result in output:
+                yield Result.from_values(
+                    origin=self,
+                    message=result['message'],
+                    file=filename,
+                    line=result['line'],
+                    severity=self.get_severity(result))
```
sils1297: you don't allow other fields? debug message, additional info, column, ...?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,174 @@
...
+
+        def run(self, filename, file, **kwargs):
+            json_string = json.dumps({'filename': filename,
+                                      'file': file})
```
sils1297: no settings?

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: This function is similar to `_prepare_options` in linter. I figured I would let it the same but I didn't need to modify the options in the end. I will modify the docstring to be more relevant to what the function does.

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: The settings here are passed in the decorator (I will make a test for that too), so I have to manually craft the FunctionMetadata object.

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I am looking into it, but I don't know how to do this at the moment :D

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I will add them right now, basically there is nothing hard, I used the minimum necessary fields in order to have the decorator working.

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: They should be in `kwargs` adding them right away.

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: yes, the enum is a little complicated indeed, I think you can use the `RESULT_SEVERITY.str_dict` which is a dict that maps strings to the actual severity objects

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,197 @@
...
+
+            :return: The executable name.
+            """
+            return options["executable"]
+
+        @staticmethod
+        def normalize_desc(val):
+            """
+            Normalizes the description of the parameters only if there
+            is none provided.
```
fneu: yeah, please find a better name for val and also please add a parameter description to the docstring!

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,200 @@
...
+    class ExternalBearWrapBase(LocalBear):
+
+        @staticmethod
+        def create_arguments():
+            """
+            This method has to be implemented by the class that uses
+            the decorator in order to create the arguments needed for
+            the executable.
+            """
+            raise NotImplementedError
```
Makman2: do you actually need this function? it seems that settings are passed via stdin as json^^

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,200 @@
...
+            opt_params = dict(cls.get_optional_params())
+            for key, val in opt_params.items():
+                if key not in settings:
+                    settings[key] = val[2]
+
+        @staticmethod
+        def get_severity(result):
+            """
+            Returns the severity corresponding to the one from the
+                parsed JSON
```
Makman2: indentation

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,200 @@
...
+
+        def parse_output(self, out, filename):
+            """
+            Parses the output JSON into Result objects
+
+            :param out: Raw output from the given executable (should be JSON)
+            :param filename: The filename of the analyzed file. Needed to
+                create the Result objects.
+
+            :yield: The Result objects.
```
Makman2: is this a valid tag? :O
what we currently do is
```
:return: An iterator yielding ``Result`` objects.
```

coalib/bearlib/abstractions/ExternalBearWrap.py
Makman2: nvm, we discuss that somewhen later

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,200 @@
...
+    return result_klass
+
+
+@enforce_signature
+def external_bear_wrap(executable: str, **options):
+
+    options["executable"] = executable
+    _prepare_options(options)
+
+    return partial(_create_wrapper, options=options)
```
Makman2: one thing I would improve for this file is the general documentation style, either use the "old" or the "new" one:
```
# old style:
:param x:    Hello world x
:param xyzw: xyzw

#new style
:param x:
    x-desc
:param xyzw:
    xyzw
```

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,200 @@
...
+        def get_executable(cls):
+            """
+            Returns the executable of this class.
+
+            :return: The executable name.
+            """
+            return options["executable"]
+
+        @staticmethod
+        def normalize_desc(param_entry):
```
Makman2: I think this is not public API, you should maybe put an underscore in front of the name :+1:

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+            """
+            Returns the executable of this class.
+
+            :return:
+                The executable name.
+            """
+            return options["executable"]
+
+        @staticmethod
+        def _normalize_desc(param_entry):
```
sils1297: can you doctest this function? Think that would shed a little light

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+
+            :param param_entry:
+                A FunctionMetadata OrderedDict entry. It is a
+                tuple, (description, type, default_value). The default_value
+                is used to distinguish optional from non_optional parameters.
+            :return:
+                A value for the OrderedDict in the FunctionMetada object.
+            """
+            if param_entry[0] != "":
+                return param_entry
```
sils1297: what if the description is `!= ""` but it has a default? Don't you want that added to the description?

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I prefered to keep it like that so it looks like linter. Maybe in the future be merged.

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+            """
+            Fetches the non_optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                FunctionMetada object.
+            """
+            non_optional_params = {}
+            for key, value in options['settings'].items():
```
sils1297: instead of key value, use `setting_name` and `description`?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+            """
+            Fetches the optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                FunctionMetada object.
+            """
+            optional_params = {}
+            for key, value in options['settings'].items():
```
sils1297: same

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+        @classmethod
+        def _prepare_settings(cls, settings):
+            """
+            Adds the optional settings to the settings dict in-place.
+
+            :param settings:
+                The settings dict.
+            """
+            opt_params = dict(cls.get_optional_params())
+            for key, val in opt_params.items():
```
sils1297: careful with naming ;)

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+            Returns the severity corresponding to the one from the
+            parsed JSON
+
+            :param result:
+                A result dict extracted from the parsed JSON
+
+            :return:
+                A RESULT_SEVERITY
+            """
+            severity_map = {'MAJOR': RESULT_SEVERITY.MAJOR,
```
sils1297: there should be a map already somewhere generated by the enum, look into the Enum class!

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+                            'NORMAL': RESULT_SEVERITY.NORMAL,
+                            'INFO': RESULT_SEVERITY.INFO}
+
+            if not 'severity' in result:
+                return RESULT_SEVERITY.NORMAL
```
sils1297: a `.get` with a default will just do fine

sils1297: ah sorry

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+            else:
+                sev = result['severity']
+                return severity_map[sev]
```
sils1297: you should use a get with a default here though

Redridge: I wanted that if the user specifies `severity=blablabla` that he would get an Error because that is not a valid severity

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: will look for it

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I don't really understand what do you mean.

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: How would that go? Maybe help me with an example?

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I wanted to use the default only when description is not provided by the user.

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,213 @@
...
+        @staticmethod
+        def _normalize_desc(param_entry):
+            """
+            Normalizes the description of the parameters only if there
+            is none provided.
+
+            :param param_entry:
+                A FunctionMetadata OrderedDict entry. It is a
+                tuple, (description, type, default_value). The default_value
+                is used to distinguish optional from non_optional parameters.
```
sils1297: why don't you just take three parameters btw.?

sils1297: that'd be way easier

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: why? You're trying to mimic how the documentation comment class does it, right? Maybe extract the cleanup it does from there and use the function there and here? less duplication?

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: do you think so? I liked it more that way

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I am leaving it like this ok? So that the user gets an error if he supplies an unknown severity

sils1297: ok

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net
http://coala-analyzer.org/ - http://viperdev.io/ - http://gitmate.com/

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: see https://duckduckgo.com/?q=doctest&t=ffab&ia=about

coalib/bearlib/abstractions/ExternalBearWrap.py
sils1297: yes because they have a name each instead of a number and you every time have to look up the docs

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: looking at DocumentationComment.py I see that it is used to parse the descriptions from the docstring. In my case the descriptions are not in the docstring but rather passed in the decorator.

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,220 @@
...
+                The default value of the setting.
+            :return:
+                A value for the OrderedDict in the FunctionMetada object.
+            """
+            if description != "":
+                if default_value is not None:
+                    return (description, setting_type, default_value)
+                else:
+                    return (description, setting_type)
+            desc = "No description given."
```
sils1297: can you use the constant `FunctionMetadata.str_nodesc`?

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,220 @@
...
+            if default_value is None:
+                return (desc, setting_type)
+            else:
+                desc += " Defaults to " + str(default_value)
```
sils1297: use `FunctionMetadata.str_optional`

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,220 @@
...
+                The type of the setting. It is needed to create the final
+                tuple.
+            :param default_value:
+                The default value of the setting.
+            :return:
+                A value for the OrderedDict in the FunctionMetada object.
+            """
+            if description != "":
+                if default_value is not None:
+                    return (description, setting_type, default_value)
```
sils1297: don't you want to do the `defaults to` thing here?

sils1297: FunctionMetadata does it and I think you should too

coalib/bearlib/abstractions/ExternalBearWrap.py
Makman2: I think `@classproperty` is not really working (at least I remember so, that's why I used a function too in `@linter`), let's discuss after merge indeed

coalib/bearlib/abstractions/ExternalBearWrap.py
Makman2: oh just thinking about it, by default this function shouldn't be necessary to override, right? As it's only needed when interpreters and stuff need to be invoked. In this case I would return an empty tuple^^

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,220 @@
...
+        def parse_output(self, out, filename):
+            """
+            Parses the output JSON into Result objects
+
+            :param out:
+                Raw output from the given executable (should be JSON)
+            :param filename:
+                The filename of the analyzed file. Needed to
+                create the Result objects.
+
```
Makman2: superfluous newline

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
Makman2: it's easier to test with writing a little python script that acts as the executable. though I don't care for this PR, we can open an issue for that after merge^^

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,197 @@
...
+                                     "asetting": ("", bool),
+                                     "bsetting": ("", bool, True),
+                                     "csetting": ("My desc", bool, False),
+                                     "dsetting": ("Another desc", bool)
+                                     })(self.Dummy)
+        metadata = uut.get_metadata()
+        self.assertEqual(metadata.non_optional_params[
+                         "asetting"][0], "No description given.")
+        self.assertEqual(metadata.optional_params["bsetting"][
+                         0], "No description given. Defaults to True")
```
Makman2: that looks weird :3
-->
```
        self.assertEqual(metadata.non_optional_params[
+                         "asetting"][0], "No description given.")
+        self.assertEqual(metadata.optional_params["bsetting"][
+                         0], "No description given. Defaults to True")
```

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,197 @@
...
+                severity=RESULT_SEVERITY.MAJOR
+                ),
+            Result.from_values(
+                origin=uut,
+                message="This is wrong too",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.INFO
+                )
+        ]
```
Makman2: please do
```
severity=RESULT_SEVERITY.INFO)]
```

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,197 @@
...
+                severity=RESULT_SEVERITY.NORMAL
+                ),
+            Result.from_values(
+                origin=uut,
+                message="This is wrong too",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.NORMAL
+                )
+        ]
```
Makman2: same here

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,197 @@
...
+            list(uut.parse_output(broken_json, "some_file"))
+
+    def test_setting_desc(self):
+        uut = external_bear_wrap("exec",
+                                 settings={
+                                     "asetting": ("", bool),
+                                     "bsetting": ("", bool, True),
+                                     "csetting": ("My desc", bool, False),
+                                     "dsetting": ("Another desc", bool)
+                                     })(self.Dummy)
```
Makman2: I would be consistent:
```
uut = (external_bear_wrap("exec",
                          settings={
                          ...
                          "dsetting": (...)})
       (self.Dummy))
```

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
```diff
@@ -0,0 +1,197 @@
...
+                debug_msg="Sample debug message"
+                ),
+            Result.from_values(
+                origin=uut,
+                message="Different message",
+                file=self.testfile_path,
+                line=3,
+                severity=RESULT_SEVERITY.INFO
+                )
+        ]
```
Makman2: same here

tests/bearlib/abstractions/external_bear_wrap_testfiles/test_external_bear.py
```diff
@@ -0,0 +1,37 @@
...
+            'severity': "MAJOR",
+            'line': 1
+        }, {
+            'message': "This is wrong too",
+            'file': args['filename'],
+            'severity': "INFO",
+            'line': 3
+        }
+
+    ]
```
Makman2: please shrink that a bit too :)

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: I am not sure, should I overwrite the description if the user has given one already?

Redridge: Nvm, I understood now

coalib/bearlib/abstractions/ExternalBearWrap.py
Redridge: Yeah, better

tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
Redridge: magic of coala :))

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,219 @@
...
+        @classmethod
+        def get_non_optional_params(cls):
+            """
+            Fetches the non_optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                FunctionMetada object.
```
Makman2: typo :3
\+ you could enclose into double backticks:
```
``FunctionMetadata``
```

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,219 @@
...
+
+        @classmethod
+        def get_optional_params(cls):
+            """
+            Fetches the optional_params from ``options['settings']``
+            and also normalizes their descriptions.
+
+            :return:
+                An OrderedDict that is used to create a
+                FunctionMetada object.
```
Makman2: typo too + double backticks :)

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,219 @@
...
+            """
+            Parses the output JSON into Result objects
+
+            :param out:
+                Raw output from the given executable (should be JSON)
+            :param filename:
+                The filename of the analyzed file. Needed to
+                create the Result objects.
+            :return:
+                An iterator yielding ``Result`` objects.
```
Makman2: could you put a period behind each documentation sentence?

Makman2:
ack 74e4209

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,220 @@
...
+
+            :param result:
+                A result dict extracted from the parsed JSON.
+            :return:
+                A RESULT_SEVERITY value.
+            """
+            if not 'severity' in result:
+                return RESULT_SEVERITY.NORMAL
+            else:
+                sev = result['severity']
```
sils1297: sev variale unneeded

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,219 @@
...
+            "Invalid keyword arguments provided: " +
+            ", ".join(repr(s) for s in sorted(superfluous_options)))
+
+    if not 'settings' in options:
+        options['settings'] = {}
+
+
+def _create_wrapper(klass, options):
+    class NoDefaultValue:
+        pass
```
Makman2: use unique object instances for this, without creating a whole type:
```
NoDefaultValue = object()
```

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,219 @@
...
+
+        @classmethod
+        def _prepare_settings(cls, settings):
+            """
+            Adds the optional settings to the settings dict in-place.
+
+            :param settings:
+                The settings dict.
+            """
+            opt_params = dict(cls.get_optional_params())
```
Makman2: no need to create another dict^^

coalib/bearlib/abstractions/ExternalBearWrap.py
```diff
@@ -0,0 +1,218 @@
...
+
+            :param result:
+                A result dict extracted from the parsed JSON.
+            :return:
+                A RESULT_SEVERITY value.
+            """
+            if not 'severity' in result:
+                return RESULT_SEVERITY.NORMAL
+            else:
+                return RESULT_SEVERITY.str_dict[result['severity']]
```
Makman2: maybe this?
```
return (RESULT.SEVERITY.str_dict[results['severity']]
        if 'severity' in result else
        RESULT_SEVERITY.NORMAL)
```
(note I've changed the order so you don't need to negate the condition afterwards)

Makman2: ouh even better:
```
return RESULT_SEVERITY.std_dict.get(results['severity'], RESULT_SEVERITY.NORMAL)
```

Redridge: I want an error to be raised if an invalid severity is specified (Ex: severity=blabla) so the first one would achieve that but the second one would just put a normal severity.

Do you think it is better with an error or without?

Makman2: ah okay, then better with, but then use the first version of my snippet and mention that you don't want to use `.get()` inside a comment to get this error behaviour. Maybe you should add a `:raises KeyError:` specification?

Makman2:
ack a355730

Makman2:
@rultor merge finally

PR #2209, opened 2016-06-01
Globbing: Fix glob for patterns with ** wildcard

fneu:
previously, when globbing for a pattern "/dir/**.xyz",
glob would check all files in dir with fnmatch(pattern).

It did not, however, recurse into sub directories and would
therefore miss matching files in those sub dirs. This commit
enables fnmatch matching against all those files when the
double-star wildcard is present in the pattern.

### DIFF
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -280,7 +280,10 @@ def relative_wildcard_glob(dirname, pattern):
     if not dirname:
         dirname = os.curdir
     try:
-        names = os.listdir(dirname)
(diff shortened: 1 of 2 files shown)


sils1297:
hey doesn't this fix an issue or so? Can you add a note?

sils1297:
also rebase

sils1297:
@fneu this needs attention

sils1297:
rebased and moved to https://github.com/coala-analyzer/coala/pull/2246 for merge

PR #2236, opened 2016-06-03
ConfigurationGatheringTest: Do not import bears

sils1297:
Some of the ConfigurationGatheringTest test cases will fail if there are
any bears installed locally. To get around this for now we can simply
disable importing of all bears for this entire test suite.

Fixes https://github.com/coala-analyzer/coala/issues/2210


mtomwing:
ack bed4992

sils1297:
@rultor merge

PR #2337, opened 2016-06-13
show_bears and filter_by_language in coala-json

arafsheikh:


coalib/coala_json.py
```diff
@@ -28,12 +33,51 @@ def main():
...
+                                                   ],
+                               "Optional settings": [param for param in
+                                                     metadata.optional_params]
+                               })
+
+            print(json.dumps({"bears": retval},
+                             cls=JSONEncoder,
+                             sort_keys=True,
+                             indent=2,
+                             separators=(',', ': ')))
```
sils1297: naaah, I'd put this into an output variable or so, so it'll be printed to file if the user requests so

coalib/coala_json.py
```diff
@@ -28,12 +33,51 @@ def main():
...
+        try:
+            sections, _ = load_configuration(arg_list=None,
+                                             log_printer=log_printer)
+            local_bears, global_bears = collect_all_bears_from_sections(
+                sections, log_printer)
+            if args.filter_by_language:
+                local_bears = filter_section_bears_by_languages(
+                    local_bears, args.filter_by_language)
+                global_bears = filter_section_bears_by_languages(
+                    global_bears, args.filter_by_language)
```
sils1297: this is code duplication between coala and coala-json I think we should avoid that

sils1297: see also https://github.com/coala-analyzer/coala/issues/2321 as we ight make the `--json` an argument of the main coala app, not sure how we wanna do this yet

arafsheikh: I need this now for coala-eclipse.

arafsheikh: How do you propose to de-duplicate?

sils1297: maybe a function that does this stuff? Not fully sure, possibly ask others for ideas

sils1297:
marking WIP in the meantimg

coalib/coala_json.py
arafsheikh: I'm planning to make a function out of line 40-48, but in which file should the function reside? `CliParsing.py`?

hypothesist: IMO Collectors.py sounds more fitting (you're collecting bears that are for a particular language).

I'm also for making this a function as I'm doing the exact same thing in coala-quickstart unnecessarily.

coalib/parsing/DefaultArgParser.py
```diff
@@ -187,23 +187,23 @@ def default_arg_parser(formatter_class=None):
...
-    # The following are "coala" specific arguments
-    if parser_type == 'coala':
-        SHOW_BEARS_HELP = ("Display bears and its metadata with the sections "
-                           "that they belong to")
-        outputs_group.add_argument('-B',
-                                   '--show-bears',
-                                   const=True,
-                                   action='store_const',
-                                   help=SHOW_BEARS_HELP)
+    SHOW_BEARS_HELP = ("Display bears and its metadata with the sections "
```
AbdealiJK: If you remove the if condition, then this is even there for coala-format, coala-html. etc etc...
You want the if condition to check for coala and coala-json only

coalib/coala_json.py
```diff
@@ -37,26 +35,19 @@ def main():
...
-                    global_bears, args.filter_by_language)
+            local_bears, global_bears = get_filtered_bears(
+                args.filter_by_language, log_printer)
 
             bears = inverse_dicts(local_bears, global_bears)
             results = []
-            for bear, _sections in sorted(bears.items(),
-                                          key=lambda bear_tuple:
-                                          bear_tuple[0].name):
+            for bear, sections in sorted(bears.items(),
```
sils1297: put the sections rename in an own commit

coalib/coala_json.py
```diff
@@ -37,26 +35,19 @@ def main():
...
+                                         key=lambda bear_tuple:
+                                         bear_tuple[0].name):
                 metadata = bear.get_metadata()
                 results.append({"Name": bear.name,
                                 "Description": metadata.desc,
-                                "Supported languages": bear.supported_languages,
-                                "Used in": _sections,
+                                "Supported languages": bear.LANGUAGES,
```
sils1297: the switch to LANGUAGES is an own commit

coalib/collecting/Collectors.py
```diff
@@ -228,6 +228,28 @@ def collect_all_bears_from_sections(sections, log_printer):
     return local_bears, global_bears
 
 
+def get_filtered_bears(language, log_printer):
+    """
+    Fetch bears and filter them based on given language.
```
sils1297: languages please, plural. This allows a list as well.

coalib/collecting/Collectors.py
```diff
@@ -228,6 +228,28 @@ def collect_all_bears_from_sections(sections, log_printer):
...
+
+    :param language:    Filter bears based on the given language.
+    :param log_printer: The log_printer to handle logging.
+    :return:            Tuple containing dictionaries of local bears
+                        and global bears.
+    """
+    from coalib.settings.ConfigurationGathering import load_configuration
```
sils1297: why the import here? Circular import? Sounds evil. That shouldn't be!

arafsheikh: Oh, was circular import the reason why I was getting an import error if I imported it with other stuff above?

sils1297: guess so, that's evil can we try finding a way around it?

sils1297: to me it looks like this function doesn't at all belong into this file

coalib/coala_json.py
```diff
@@ -28,12 +31,38 @@ def main():
...
-    retval = {"results": results}
+            bears = inverse_dicts(local_bears, global_bears)
+            results = []
+            for bear, sections in sorted(bears.items(),
+                                         key=lambda bear_tuple:
+                                         bear_tuple[0].name):
+                metadata = bear.get_metadata()
+                results.append({"Name": bear.name,
+                                "Description": metadata.desc,
+                                "Supported languages": bear.LANGUAGES,
```
sils1297: bears have more attributes by now, can't you just append the bear and let the JSONEncoder do the job? It should do it just fine extracting also all the other info.

arafsheikh: @sils1297 If I append bear directly I get `TypeError: <property object at 0x7f71baf57318> is not JSON serializable`. Although `set([type(x) for x in dir(bear)])` says all the attributes are strings, so they should be serializable.

sils1297: uhm some interactive debugging might help

coalib/collecting/Collectors.py
arafsheikh: There's another cyclic import [here](https://github.com/coala-analyzer/coala/blob/master/coalib/collecting/Collectors.py#L196). We can move both functions to a different module. Or maybe a new module.

sils1297: move them to ConfigurationGathering would make sense wouldn't it?

arafsheikh: But these functions have nothing to do with config files. I was thinking something like a `Provider`(like content/info provider) module in Collectors dir?

coalib/coala_json.py
arafsheikh: @AbdealiJK You wrote the `JSONEncoder`, can you help me with this?

AbdealiJK: Possibly add a try/catch and catch TypeErrors. And print which property is causing this ?
Also, `bear`s have sets right ? For LANGUAGE, etc etc ... So, I don't see how everything can be a string.
`set()` is not JSON serializable:
```
In [2]: a = {"a": set([1,2])}
In [3]: a
Out[3]: {'a': {1, 2}}
In [4]: import json
In [5]: json.dumps(a)
...
TypeError: set([1, 2]) is not JSON serializable
```

arafsheikh: `set` is not a problem as [this](https://github.com/coala-analyzer/coala/blob/master/coalib/output/JSONEncoder.py#L23) takes care of it.
`<class 'property'>` is the problem. More specifically https://github.com/coala-analyzer/coala/blob/master/coalib/bears/Bear.py#L386

FWIW I tired commenting out the whole function but then I got another TypeError stating `<class '_sre.SRE_Pattern'>`(which is the compiled regex object) can't be serialized

arafsheikh: @sils1297 I think that appending `bear` isn't a good idea. The JSONEncoder will recursively go through all attributes in a bear, but some attributes like [this](https://github.com/coala-analyzer/coala-bears/blob/master/bears/c_languages/ClangComplexityBear.py#L23) will break the JSONEncoder. We can handle such cases in JSONEncoder but then we'll have to do this every time something isn't serializable.
I think we should stick to only outputting the stuff we want (like in `coala`). What do you say?

sils1297: The then the JSONEncoder is buggy, the stuff *should* be serializable, if it is't that's a bug!

arafsheikh: `CursorKind.IF_STMT, CursorKind.WHILE_STMT, ...` are external classes. We can't predict if they are serializable.

sils1297: then they should be private or so and not be serialized

arafsheikh: @sils1297 @AbdealiJK Okay so almost everything is fixed, except I wanted to know the `non_optional_params` for each bear for coala-eclipse. This info can be accessed using `get_metadata()` classmethod.
Is it possible to somehow append this data to each bear?

sils1297: we could make the metadata a property so it gets serialized as well, sounds like something useful in general, huh?

coalib/bears/Bear.py
```diff
@@ -261,6 +262,29 @@ def get_metadata(cls):
...
+            "desc": metadata.desc,
+            "non_optional_params": [({param: non_optional_params[param][0]}
+                                     for param in non_optional_params)],
+            "optional_params": [({param: optional_params[param][0]}
+                                 for param in optional_params)]}
+
+        # Delete attributes that cannot be serialized
+        unserializable_attributes = ["new_result", "printer"]
+        for attribute in unserializable_attributes:
+            if attribute in _dict:
```
arafsheikh: Coverage drops as the branch will always be true in the current state. Should I remove this check and directly delete the element from the dict?

tests/coalaJSONTest.py
```diff
@@ -51,6 +53,29 @@ def test_fail_acquire_settings(self):
...
+            output = json.loads(output)
+            self.assertEqual(len(output["bears"]), 4)
+
+    def test_show_language_bears(self):
+        with bear_test_module():
+            retval, output = execute_coala(
+                coala_json.main, 'coala-json', '-B', '-l', 'java')
+            self.assertEqual(retval, 0)
+            output = json.loads(output)
+            self.assertEqual(len(output["bears"]), 2)
```
tushar-rishav: @arafsheikh Could you please a test for the other attributes too? Like name, desc etc?

tushar-rishav:
ack 460a8c190345aa29e47ac65ceb27b20a3f7e2e9b

tests/coalaJSONTest.py
arafsheikh: @tushar-rishav Done (see below).

tushar-rishav:
ack f9e46c8

tushar-rishav:
unack f9e46c8

tushar-rishav:
ack https://github.com/coala-analyzer/coala/commit/f9e46c89794035d4dad7dc2a149faf935c28a8c4

tushar-rishav:
@rultor merge

PR #2342, opened 2016-06-14
LANGUAGES: Change from tuple to set

Adrianzatreanu:
There's no point in having a tuple, and
('C','C++') should be the same with ('C++','C').

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -185,11 +185,11 @@ def filter_section_bears_by_languages(bears, languages):
     """
     new_bears = {}
     # All bears with "all" languages supported shall be shown
-    languages = set(x.lower() for x in languages) | {'all'}
+    languages = set(language.lower() for language in languages) | {'all'}
     for section in bears.keys():
         new_bears[section] = tuple(
             bear for bear in bears[section]
-            if {x.lower() for x in bear.LANGUAGES} & languages)
+            if {language.lower() for language in bear.LANGUAGES} & languages)
     return new_bears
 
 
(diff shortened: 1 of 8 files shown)


coalib/collecting/Collectors.py
```diff
@@ -189,7 +189,7 @@ def filter_section_bears_by_languages(bears, languages):
     for section in bears.keys():
         new_bears[section] = tuple(
             bear for bear in bears[section]
-            if set(x.lower() for x in bear.supported_languages) & languages)
+            if set(x.lower() for x in bear.LANGUAGES) & languages)
```
hypothesist: While not related, `{x.lower() for x in bear.LANGUAGES}` is simpler IMO

Adrianzatreanu: fair point

coalib/collecting/Collectors.py
```diff
@@ -189,7 +189,7 @@ def filter_section_bears_by_languages(bears, languages):
     for section in bears.keys():
         new_bears[section] = tuple(
             bear for bear in bears[section]
-            if set(x.lower() for x in bear.supported_languages) & languages)
+            if {x.lower() for x in bear.LANGUAGES} & languages)
```
tushar-rishav: IMO it'd be nice to change the variable name here. Instead of `x` use something that gives more info about the items being iterated. 
```python
for fruits in ['Apple', 'Mango']  # Good
for x in ['Apple', 'Mango']  ['Apple', 'Mango']  Bad
```

tushar-rishav:
ack 751ea6d 9b22b21

Feel free to merge

Adrianzatreanu:
@rultor merge

PR #2391, opened 2016-06-25
Uploadmethod

Adrianzatreanu:


coalib/bears/requirements/PackageRequirement.py
```diff
@@ -1,4 +1,5 @@
 from coala_decorators.decorators import generate_eq, generate_repr
+import os
```
tushar-rishav: @Adrianzatreanu coala module imports should come after inbuilt modules. :stuck_out_tongue_winking_eye:

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -119,3 +120,19 @@ def multiple(cls, *args):
...
+    @classmethod
+    def upload(self):
+        """
+        An upload method which uploads a conda package containing the command
+        to install the corresponding package.
+        """
+        name = self.package
+        if not os.path.exists(name):
+            os.mkdir(name)
+        file = open(name + '/' + name + '.py', 'a')
```
tushar-rishav: @Adrianzatreanu Can we not use `os.path.join(name, name + '.py' )` ?

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -119,3 +120,19 @@ def multiple(cls, *args):
...
+        file.write(self.manager + ' ' + self.package)
```
tushar-rishav: Also I think the current file I/O approach can be made better. This is more pythonic and you won't have to close the file descriptor explicitly :
```python
with open(os.path.join('name','name.py'), 'a') as package_file:
    package_file.write("{manager} {package}".format(manager=self.manager, 
                                                    package=self.package))
```

coalib/bears/requirements/PackageRequirement.py
```diff
@@ -119,3 +120,19 @@ def multiple(cls, *args):
...
+        file.close()
+        meta = open(name + '/meta.yaml', 'w')
```
tushar-rishav: Same thing here

Adrianzatreanu:
thx @tushar-rishav but this is going to be dropped most likely :D

PR #2409, opened 2016-06-30
logging

Uran198:
So, this is how the logging can look like. It seems that logging may be more dependant on printers than was initially thought, so the printer will probably have to be passed in either way.
Not sure, how much this is better.

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -282,7 +282,9 @@ class Bear(Printer, LogPrinter):
         """
         Override JSON export of ``Bear`` object.
         """
-        _dict = get_public_members(cls)
+        # json cannot serialize properties, so drop them
+        _dict = {key: value for key, value in get_public_members(cls).items()
+                 if not isinstance(value, property)}
         metadata = cls.get_metadata()
         non_optional_params = metadata.non_optional_params
         optional_params = metadata.optional_params
@@ -292,11 +294,6 @@ class Bear(Printer, LogPrinter):
                                     for param in non_optional_params),
             "optional_params": ({param: optional_params[param][0]}
                                 for param in optional_params)}
-
-        # Delete attributes that cannot be serialized
-        unserializable_attributes = ["new_result", "printer"]
-        for attribute in unserializable_attributes:
-            _dict.pop(attribute, None)
         return _dict
 
     @classmethod
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -14,14 +14,14 @@ from coala_utils.decorators import (enforce_signature, classproperty,
 
 from coalib.bears.requirements.PackageRequirement import PackageRequirement
 from coalib.bears.requirements.PipRequirement import PipRequirement
-from coalib.output.printers.LogPrinter import LogPrinter
+from coalib.output.printers.LogPrinter import LogPrinterMixin
 from coalib.results.Result import Result
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
 from coalib.settings.ConfigurationGathering import get_config_directory
 
 
-class Bear(Printer, LogPrinter):
+class Bear(Printer, LogPrinterMixin):
     """
     A bear contains the actual subroutine that is responsible for checking
     source code for certain specifications. However it can actually do
@@ -200,7 +200,6 @@ class Bear(Printer, LogPrinter):
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
         """
         Printer.__init__(self)
-        LogPrinter.__init__(self, self)
 
         if message_queue is not None and not hasattr(message_queue, "put"):
             raise TypeError("message_queue has to be a Queue or None.")
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -16,6 +16,7 @@ import functools
 from pyprint.ConsolePrinter import ConsolePrinter
 
 from coalib.coala_main import run_coala
+from coalib.misc.Constants import configure_logging
 from coalib.misc.Exceptions import get_exitcode
 from coalib.collecting.Collectors import filter_capabilities_by_languages
 from coalib.output.ConsoleInteraction import (
@@ -27,6 +28,8 @@ from coalib.settings.ConfigurationGathering import get_filtered_bears
 
 
 def main():
+    configure_logging()
+
     try:
         console_printer = ConsolePrinter()
         log_printer = LogPrinter(console_printer)
--- a/coalib/coala_delete_orig.py
+++ b/coalib/coala_delete_orig.py
@@ -7,9 +7,12 @@ from coalib.parsing import Globbing
 from coalib.settings.ConfigurationGathering import get_config_directory
 from coalib.settings.Section import Section
 from coalib.parsing.Globbing import glob_escape
+from coalib.misc.Constants import configure_logging
 
 
 def main(log_printer=None, section: Section=None):
+    configure_logging()
+
     start_path = get_config_directory(section)
     log_printer = (LogPrinter(ConsolePrinter()) if log_printer is None
                    else log_printer)
--- a/coalib/coala_main.py
+++ b/coalib/coala_main.py
@@ -14,6 +14,7 @@ from coalib.settings.ConfigurationGathering import gather_configuration
 from coalib.misc.Caching import FileCache
 from coalib.misc.CachingUtilities import (
     settings_changed, update_settings_db, get_settings_hash)
+from coalib.misc.Constants import configure_logging
 
 do_nothing = lambda *args: True
 
@@ -54,6 +55,8 @@ def run_coala(log_printer=None,
     :return:                        A dictionary containing a list of results
                                     for all analyzed sections as key.
     """
+    configure_logging()
+
     log_printer = (
         LogPrinter(ConsolePrinter(), LOG_LEVEL.DEBUG) if log_printer is None
         else log_printer)
--- a/coalib/misc/Constants.py
+++ b/coalib/misc/Constants.py
@@ -143,24 +143,28 @@ def configure_logging():
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
-            'raw': {
+            'colored': {
                 'class': 'logging.StreamHandler',
-                'formatter': 'raw',
+                'formatter': 'colored',
                 'stream': sys.stdout
             }
         },
         'root': {
-            'level': 'DEBUG'
-        },
-        'loggers': {
-            'coala.raw': {
-                'handlers': ['raw']
-            }
+            'level': 'DEBUG',
+            'handlers': ['colored']
         },
         'formatters': {
-            'raw': {
-                'format': '[%(levelname)s][%(asctime)s] %(message)s',
-                'datefmt': '%X'
+            'colored': {
+                '()': 'colorlog.ColoredFormatter',
+                'format': '%(log_color)s[%(levelname)s]%(reset)s[%(asctime)s]'
+                          ' %(message)s',
+                'datefmt': '%X',
+                'log_colors': {
+                    'ERROR': 'red',
+                    'WARNING': 'yellow',
+                    'INFO': 'blue',
+                    'DEBUG': 'green'
+                }
             }
         }
     })
--- a/coalib/misc/Constants.py
+++ b/coalib/misc/Constants.py
@@ -132,3 +132,35 @@ URL_REGEX = re.compile(
     r'(?::\d+)?'  # optional port number
     r'(?:/?|[/?]\S+)$',  # path
     re.IGNORECASE)
+
+
+def configure_logging():
+    """
+    Configures the logging with hard coded dictionary.
+    """
+    import sys
+    import logging.config
+    logging.config.dictConfig({
+        'version': 1,
+        'handlers': {
+            'raw': {
+                'class': 'logging.StreamHandler',
+                'formatter': 'raw',
+                'stream': sys.stdout
+            }
+        },
+        'root': {
+            'level': 'DEBUG'
+        },
+        'loggers': {
+            'coala.raw': {
+                'handlers': ['raw']
+            }
+        },
+        'formatters': {
+            'raw': {
+                'format': '[%(levelname)s][%(asctime)s] %(message)s',
+                'datefmt': '%X'
+            }
+        }
+    })
--- a/coalib/output/printers/LOG_LEVEL.py
+++ b/coalib/output/printers/LOG_LEVEL.py
@@ -1,6 +1,10 @@
+import logging
 from coalib.misc.Enum import enum
 
-LOG_LEVEL = enum("DEBUG", "INFO", "WARNING", "ERROR")
+LOG_LEVEL = enum(DEBUG=logging.DEBUG,
+                 INFO=logging.INFO,
+                 WARNING=logging.WARNING,
+                 ERROR=logging.ERROR)
 LOG_LEVEL_COLORS = {LOG_LEVEL.ERROR: "red",
                     LOG_LEVEL.WARNING: "yellow",
                     LOG_LEVEL.INFO: "blue",
--- a/coalib/output/printers/LogPrinter.py
+++ b/coalib/output/printers/LogPrinter.py
@@ -6,48 +6,12 @@ from coalib.output.printers.LOG_LEVEL import LOG_LEVEL, LOG_LEVEL_COLORS
 from coalib.processes.communication.LogMessage import LogMessage
 
 
-class LogPrinter:
+class LogPrinterMixin:
     """
-    The LogPrinter class allows to print log messages to an underlying Printer.
-
-    This class is an adapter, means you can create a LogPrinter from every
-    existing Printer instance.
+    Provides access to the logging interfaces (e.g. err, warn, info) by routing
+    them to the log_message method, which should be implemented by descendants
+    of this class.
     """
-
-    def __init__(self,
-                 printer,
-                 log_level=LOG_LEVEL.INFO,
-                 timestamp_format="%X"):
-        """
-        Creates a new log printer from an existing Printer.
-
-        :param printer:          The underlying Printer where log messages
-                                 shall be written to. If you inherit from
-                                 LogPrinter, set it to self.
-        :param log_level:        The minimum log level, everything below will
-                                 not be logged.
-        :param timestamp_format: The format string for the
-                                 datetime.today().strftime(format) method.
-        """
-        self._printer = printer
-        self.log_level = log_level
-        self.timestamp_format = timestamp_format
-
-    @property
-    def printer(self):
-        """
-        Returns the underlying printer where logs are printed to.
-        """
-        return self._printer
-
-    def _get_log_prefix(self, log_level, timestamp):
-        datetime_string = timestamp.strftime(self.timestamp_format)
-
-        if datetime_string != "":
-            datetime_string = "[" + datetime_string + "]"
-
-        return '[{}]{}'.format(LOG_LEVEL.reverse.get(log_level, "ERROR"),
-                               datetime_string)
 
     def debug(self, *messages, delimiter=" ", timestamp=None, **kwargs):
         self.log_message(LogMessage(LOG_LEVEL.DEBUG,
@@ -119,6 +83,57 @@ class LogPrinter:
                        "Exception was:" + "\n" + traceback_str,
                        timestamp=timestamp),
             **kwargs)
+
+    def log_message(self, log_message, **kwargs):
+        """
+        It is your reponsibility to implement this method, if you're using this
+        mixin.
+        """
+        raise NotImplementedError
+
+
+class LogPrinter(LogPrinterMixin):
+    """
+    The LogPrinter class allows to print log messages to an underlying Printer.
+
+    This class is an adapter, means you can create a LogPrinter from every
+    existing Printer instance.
+    """
+
+    def __init__(self,
+                 printer,
+                 log_level=LOG_LEVEL.INFO,
+                 timestamp_format="%X"):
+        """
+        Creates a new log printer from an existing Printer.
+
+        :param printer:          The underlying Printer where log messages
+                                 shall be written to. If you inherit from
+                                 LogPrinter, set it to self.
+        :param log_level:        The minimum log level, everything below will
+                                 not be logged.
+        :param timestamp_format: The format string for the
+                                 datetime.today().strftime(format) method.
+        """
+        self._printer = printer
+        self.log_level = log_level
+        self.timestamp_format = timestamp_format
+
+    @property
+    def printer(self):
+        """
+        Returns the underlying printer where logs are printed to.
+        """
+        return self._printer
+
+    def _get_log_prefix(self, log_level, timestamp):
+        datetime_string = timestamp.strftime(self.timestamp_format)
+
+        if datetime_string != "":
+            datetime_string = "[" + datetime_string + "]"
+
+        return '[{}]{}'.format(LOG_LEVEL.reverse.get(log_level, "ERROR"),
+                               datetime_string)
 
     def log_message(self, log_message, **kwargs):
         if not isinstance(log_message, LogMessage):
--- a/coalib/output/printers/LogPrinter.py
+++ b/coalib/output/printers/LogPrinter.py
@@ -94,7 +94,8 @@ class LogPrinterMixin:
 class LogPrinter(LogPrinterMixin):
     """
     This class is deprecated and will be soon removed. To get logger use
-    logging.getLogger('coala.raw').
+    logging.getLogger(__name__). Make sure that you're getting it when the
+    logging configuration is loaded.
 
     The LogPrinter class allows to print log messages to an underlying Printer.
 
@@ -117,8 +118,7 @@ class LogPrinter(LogPrinterMixin):
         :param timestamp_format: The format string for the
                                  datetime.today().strftime(format) method.
         """
-        self.logger_name = 'coala.raw'
-        self.logger = logging.getLogger(self.logger_name)
+        self.logger = logging.getLogger()
 
         self._printer = printer
         self.log_level = log_level
@@ -159,4 +159,4 @@ class LogPrinter(LogPrinterMixin):
     def __setstate__(self, newdict):
         self.__dict__.update(newdict)
         # restore logger by name
-        self.logger = logging.getLogger(self.logger_name)
+        self.logger = logging.getLogger()
--- a/coalib/output/printers/LogPrinter.py
+++ b/coalib/output/printers/LogPrinter.py
@@ -1,8 +1,7 @@
 import traceback
+import logging
 
-from pyprint.ColorPrinter import ColorPrinter
-
-from coalib.output.printers.LOG_LEVEL import LOG_LEVEL, LOG_LEVEL_COLORS
+from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.processes.communication.LogMessage import LogMessage
 
 
@@ -94,6 +93,9 @@ class LogPrinterMixin:
 
 class LogPrinter(LogPrinterMixin):
     """
+    This class is deprecated and will be soon removed. To get logger use
+    logging.getLogger('coala.raw').
+
     The LogPrinter class allows to print log messages to an underlying Printer.
 
     This class is an adapter, means you can create a LogPrinter from every
@@ -101,8 +103,8 @@ class LogPrinter(LogPrinterMixin):
     """
 
     def __init__(self,
-                 printer,
-                 log_level=LOG_LEVEL.INFO,
+                 printer=None,
+                 log_level=LOG_LEVEL.DEBUG,
                  timestamp_format="%X"):
         """
         Creates a new log printer from an existing Printer.
@@ -115,9 +117,26 @@ class LogPrinter(LogPrinterMixin):
         :param timestamp_format: The format string for the
                                  datetime.today().strftime(format) method.
         """
+        self.logger_name = 'coala.raw'
+        self.logger = logging.getLogger(self.logger_name)
+
         self._printer = printer
         self.log_level = log_level
         self.timestamp_format = timestamp_format
+
+    @property
+    def log_level(self):
+        """
+        Returns current log_level used in logger.
+        """
+        return self.logger.getEffectiveLevel()
+
+    @log_level.setter
+    def log_level(self, log_level):
+        """
+        Sets log_level for logger.
+        """
+        self.logger.setLevel(log_level)
 
     @property
     def printer(self):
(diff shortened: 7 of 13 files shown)


Uran198:
All tests except tests for logging pass.
This is one way of how logging may work. Any thoughts? @coala-analyzer/coala-contributors 
So, if you'll want yourself a logger you'll either create a new one and it will return existing (or create one), or , in future needed loggers will be initialized and configured in the beginning and you'll only call `logging.getLogger` with the appropriate name.

Uran198:
Basically you'll have one logger for every type of `Printer`.

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,135 @@
...
     """
+    _loggers = {}
+
+    @staticmethod
+    def _get_logger(printer, logger_name, time_fmt=None):
+        """
+        Returns existing or newly created and configured logger.
+        :param logger_name: name of the logger to be returned.
+        :param printer:     printer to initialize logger with if it doesn't
+                            exist, yet.
```
tushar-rishav: @Uran198 Forgot to mention about `time_fmt` in docstring  I guess?

Uran198: @tushar-rishav Yes, for now I'm more interested in general approach. :smile: What's your opinion?

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,161 @@
+import sys
 import traceback
+import logging
+from colorama import Style
```
Uran198: It seems to comment on the very wrong line. Don't know if this issue is already reported. @sils1297

sils1297: It isn't, can you report it to gitlab gitmate coala-incremental?

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,159 @@
...
-from coalib.output.printers.LOG_LEVEL import LOG_LEVEL, LOG_LEVEL_COLORS
+from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.processes.communication.LogMessage import LogMessage
 
 
+class PrinterHandler(logging.StreamHandler):
+    """
+    Handler for logging which uses printer to print messages and falls back to
+    writing to stdout if no printer is provided.
+    """
```
sils1297: Can you delete irrelevant comments? Phone ui doesn't let me do it

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,194 @@
...
+    symbols where appropriate.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        :param use_color:  If True highlights records levelname with respective
+                           color.
+        """
+        self.use_color = kwargs.pop('use_color', False)
+        super(ColoredFormatter, self).__init__(*args, **kwargs)
```
sils1297: Since we use python 3 IIUC super() suffices?

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,194 @@
...
+        if not LogPrinter._loggers.get(logger_name):
+            if time_fmt:
+                format_str = '%(levelname)s[%(asctime)s] %(message)s'
+            else:
+                format_str = '%(levelname)s %(message)s'
+
+            use_color = isinstance(printer, ColorPrinter)
+            formatter = ColoredFormatter(format_str,
+                                         time_fmt,
+                                         use_color=use_color)
```
sils1297: With the formatter we don't need the color printer anymore do we? Also we would want to avoid printing copies when piping strong I guess...

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,194 @@
...
-                 printer,
+                 printer=None,
                  log_level=LOG_LEVEL.INFO,
-                 timestamp_format="%X"):
+                 timestamp_format="%X",
+                 of_type=True):
         """
-        Creates a new log printer from an existing Printer.
+        Creates a new log printer from an existing Printer or returns existing
+        one if it already exist.
```
sils1297: So on long term this class will be removed, right? Add deprecation note?

coalib/output/printers/LogPrinter.py
```diff
@@ -1,54 +1,194 @@
...
-                                 LogPrinter, set it to self.
+                                 Printer, set it to self.
         :param log_level:        The minimum log level, everything below will
                                  not be logged.
         :param timestamp_format: The format string for the
                                  datetime.today().strftime(format) method.
+        :param of_type:          If False returns new logger object every time.
+                                 Else may return logger which uses the same
+                                 printer type or will use printer to initialize
+                                 new one.
```
sils1297: Feels weird, kind of singleton but not really, smells a bit to me. Is it needed or can we just drop the singleton  aspect?

sils1297: Not sure on this one

Uran198: The logging works as singletons accessed by name, but coala occasionally uses LogPrinters with StringPrinters, IIRC only in tests, so in time this will change. Basically because of this we need this part, kind of delaying correcting it.

coalib/output/printers/LogPrinter.py
Uran198: Yes, I think we'll be able to fully remove dependence on pyprint in the future. I think colored output was the only reason, why it stayed.

I didn't get the second part...

coalib/output/printers/LogPrinter.py
Uran198: Sure, I'll leave this comments for reference.

coalib/output/printers/LogPrinter.py
Uran198: It may be a little early for the note, cause there is no alternative configured printing solution, yet. Not sure, if it's better to add in this PR configuring logging from files for different kinds of printers.

coalib/bears/Bear.py
```diff
@@ -267,6 +267,9 @@ def __json__(cls):
         Override JSON export of ``Bear`` object.
         """
         _dict = get_public_members(cls)
+        # json cannot serialize properties, so drop them
+        _dict = {key: value for key, value in _dict.items()
```
sils1297: just use `get_public_members(cls).items()`, otherwise you kind of change the meaning of a variable which is worse then just making a new variable

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/bad34a8614f7848783e53d5c2505238b15ac3b95 https://github.com/coala-analyzer/coala/pull/2409/commits/be9acd0e2fac3a5f87b84e417377b08e59b36e2d

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2409/commits/ca195fc5df9a8c195fca2a66c58ae789114c87a8

coalib/bears/Bear.py
```diff
@@ -266,7 +266,9 @@ def __json__(cls):
         """
         Override JSON export of ``Bear`` object.
         """
-        _dict = get_public_members(cls)
+        # json cannot serialize properties, so drop them
```
sils1297: shouldn't the JSONEncoder use the value of the property? It makes a getattr on it which retrieves the value, right?

coalib/coala_main.py
```diff
@@ -52,6 +54,10 @@ def run_coala(log_printer=None,
     :return:                        A dictionary containing a list of results
                                     for all analyzed sections as key.
     """
+    # Configure logging
+    with open('logging_config.json') as fd:
+        logging.config.dictConfig(json.load(fd))
```
sils1297: can you add a commit adding user documentation for this?

coalib/output/printers/LOG_LEVEL.py
```diff
@@ -1,7 +1,12 @@
...
 
-LOG_LEVEL = enum("DEBUG", "INFO", "WARNING", "ERROR")
-LOG_LEVEL_COLORS = {LOG_LEVEL.ERROR: "red",
-                    LOG_LEVEL.WARNING: "yellow",
-                    LOG_LEVEL.INFO: "blue",
-                    LOG_LEVEL.DEBUG: "green"}
+LOG_LEVEL = enum(DEBUG=logging.DEBUG,
+                 INFO=logging.INFO,
+                 WARNING=logging.WARNING,
+                 ERROR=logging.ERROR)
```
sils1297: changing the log level could be easily separately done in an own commit right? (This one's too big IMO.)

coalib/output/printers/LOG_LEVEL.py
```diff
@@ -1,7 +1,12 @@
...
+LOG_LEVEL_COLORS = {LOG_LEVEL.ERROR: Fore.RED,
+                    LOG_LEVEL.WARNING: Fore.YELLOW,
+                    LOG_LEVEL.INFO: Fore.BLUE,
+                    LOG_LEVEL.DEBUG: Fore.GREEN}
```
sils1297: this change of using the actual values here could be done as well in an own commit I think

coalib/output/printers/LogPrinter.py
```diff
@@ -1,21 +1,37 @@
...
+
+    def emit(self, record):
+        msg = self.format(record)
+        sys.stdout.write(msg + "\n")
+
+
 class LogPrinter:
     """
+    This class is deprecated and will be soon removed. To get logger use
+    loggin.getLogger('coala.colored') for colored logger and
```
sils1297: loggin? typo?

coalib/bears/Bear.py
Uran198: No, JSONEncoder, gets output of `__json__` which operates on class, so no values of properties here. Then it only filters out some values from a dict and that's it. The only thing that it gets is class attributes, nothing related to instances.

coalib/output/printers/LogPrinter.py
sils1297: the user of the logger shouldn't care about colored or not IMO, the coala logger should just not use color if the output is piped e.g. into a file or so

logging_config.json
```diff
@@ -0,0 +1,31 @@
...
+        }
+    },
+    "formatters": {
+        "raw": {
+            "format": "[%(levelname)s][%(asctime)s] %(message)s",
+            "datefmt": "%X"
+        }
+    }
+}
+
```
sils1297: a newline too much

coalib/coala_main.py
sils1297: btw. you want to load in a global config or should the user write it? The one below looks global but here youre loading it from current directory

coalib/output/printers/LOG_LEVEL.py
Uran198: It may be not a good idea to separate this one in separate, cause we will have one commit after which there will be not a working version. Putting this change before or after the commit it is introduced in will break things

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2409/commits/4f8422f569041c4aa49873a5869634260c9e2a50

coalib/coala_main.py
Uran198: It should be global, users shouldn't care about it existence even.

coalib/bears/Bear.py
sils1297: why not make `__json__` a method and instead of `key: value` use `key: getattr(self, key)`?

coalib/coala_main.py
Uran198: I'll change that.

As for the docs, what exactly should go there. It's just a file with the configurations, nothing fancy. Or you're saying documentation on how to use logging, i. e. names of the loggers to use?

sils1297: the you should add a constant for the filenme in Constants and use the calib root constant for getting the filename

sils1297: as for the docs: not needed if its not for the user

coalib/output/printers/LogPrinter.py
Uran198: How to know if the output is piped somewhere?
IMO, when you're writing something, you know whether it will be piped or no, so you can use what will be the most appropriate for you.

sils1297: I don't know but it's not something you should decide when getting a logger but the logger should decide it on its own

coalib/bears/Bear.py
Uran198: The usecases that I encountered worked with classes only, so I think it was the original purpose. I think you cannot have two methods with the same name, one as a class method and the other as regular one.

sils1297: if the JSONEncoder has an object you can simply make a normal method and be fine. Or we could patch the JSONEncoder to do the getattr on its own.

coalib/output/printers/LogPrinter.py
Uran198: The logger has much less context. Also, as it is right now, it always writes to sys.stdout, so it cannot get pipped. I'll configure list log printer, and it will not be used with colours, and can be used if you want something to be piped.

One idea here, is to configure loggers to be called as `getLogger(__name__)`. Then configure the respective "module level" loggers, this way it can decide which kind of logger needs to be used based on the module it is in.

coalib/bears/Bear.py
Uran198: Basically as it is right now, it is working the way as before, if there was an instance of the object passed to JSONEncoder, we would get value from `get_public_member` and we will not filter that out. Basically what you're saying is to make json a regular method, or to modify JSONEncoder. This I think is more a separate issue. Here I only correct the current behaviour, and you want to introduce a new one.

arafsheikh: @Uran198 If `_dict` doesn't contain any property object then [this](https://github.com/coala-analyzer/coala/pull/2409/files#diff-7b888ece875aa379bf8cfcebcd7cdef1R282) should be removed as well. As it was there only to filter out un-serializable properties.

Uran198: Nice catch @arafsheikh. I did not notice that it was hard coded before.

coalib/coala.py
```diff
@@ -27,6 +30,10 @@
 
 
 def main():
+    # Configure logging
```
sils1297: how about `# Loads logging configuration from JSON`?

coalib/coala.py
```diff
@@ -27,6 +30,10 @@
...
+    with open(log_settings_file) as fd:
```
sils1297: fd isn't a good variable name

coalib/coala.py
```diff
@@ -27,6 +30,10 @@
...
+        logging.config.dictConfig(json.load(fd))
```
sils1297: I feel like this thing belongs into a function, it's going to be needed in other main functions as well right?

tests/output/ConsoleInteractionTest.py
```diff
@@ -3,9 +3,9 @@
 from unittest.mock import patch
 from collections import OrderedDict
 from os.path import abspath, relpath
+import logging
```
sils1297: typo in commit message

tests/output/ConsoleInteractionTest.py
```diff
@@ -758,8 +758,7 @@ def test_show_bears_capabilities(self):
 class PrintFormattedResultsTest(unittest.TestCase):
 
     def setUp(self):
-        self.printer = StringPrinter()
-        self.logger = LogPrinter(self.printer)
+        self.logger = LogPrinter()
```
sils1297: is th logger only used below? Then it doesn't need to appear in setup.

coalib/coala_delete_orig.py
```diff
@@ -7,9 +9,14 @@
 from coalib.settings.ConfigurationGathering import get_config_directory
 from coalib.settings.Section import Section
 from coalib.parsing.Globbing import glob_escape
+from coalib.misc.Constants import log_settings_file
 
 
 def main(log_printer=None, section: Section=None):
+    # Configure logging
+    with open(log_settings_file) as fd:
+        logging.config.dictConfig(json.load(fd))
```
sils1297: yeah make it a function :) or a decorator? like `@configure_logging(config_file_path)`?

coalib/logging_config.json
```diff
@@ -0,0 +1,32 @@
...
+        "colored": {
+            "class": "logging.StreamHandler",
+            "formatter": "raw",
+            "stream": "ext://sys.stdout"
+        },
+        "raw": {
+            "class": "logging.StreamHandler",
+            "formatter": "raw",
+            "stream": "ext://sys.stdout"
+        }
```
sils1297: hm those two handlers look pretty similar to each other... the same. How does the colored one get the color?

coalib/logging_config.json
```diff
@@ -0,0 +1,32 @@
...
+        },
+        "coala.colored": {
+            "handlers": ["colored"]
+        }
+    },
+    "formatters": {
+        "raw": {
+            "format": "[%(levelname)s][%(asctime)s] %(message)s",
+            "datefmt": "%X"
+        }
```
sils1297: no color formatter?

coalib/output/printers/LogPrinter.py
```diff
@@ -29,26 +34,38 @@ def __init__(self,
...
+        self.logger = logging.getLogger(self.logger_name)
+
         self._printer = printer
         self.log_level = log_level
         self.timestamp_format = timestamp_format
 
     @property
+    def log_level(self):
+        """
+        Returns current log_level used is logger.
```
sils1297: grammar

coalib/output/printers/LogPrinter.py
```diff
@@ -29,26 +34,38 @@ def __init__(self,
...
+        """
+        return self._log_level
```
sils1297: can't you get the log level from the logger directly?

tests/collecting/CollectorsTest.py
```diff
@@ -202,20 +200,18 @@ def setUp(self):
...
-                                           self.log_printer), ([],))
-            self.assertRegex(sio.getvalue(),
-                             ".*\\[WARNING\\].*No bears were found matching "
-                             "'invalid_name'.\n")
+        self.assertEqual(collect_bears(["invalid_paths"],
+                                       ["invalid_name"],
+                                       ["invalid kind"],
+                                       self.log_printer), ([],))
+        self.assertEqual(''.join(log.message for log in self.log_printer.logs),
+                         "No bears were found matching 'invalid_name'.")
```
sils1297: how about `self.assertEqual([log.message for log in self.log_printer], ['No ...'])`?

tests/processes/LogPrinterThreadTest.py
```diff
@@ -1,12 +1,13 @@
...
 from coalib.misc.ContextManagers import retrieve_stdout
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.processes.LogPrinterThread import LogPrinterThread
 
 
-class TestPrinter(LogPrinter):
+class TestPrinter(Printer, LogPrinter):
 
     def __init__(self):
         LogPrinter.__init__(self, self)
```
sils1297: you should call the init methods of all superclasses!

tests/processes/ProcessingTest.py
```diff
@@ -51,7 +52,7 @@ def is_alive(self):
         return not self.control_queue.empty() and not self.starts_dead
 
 
-class ProcessingTestLogPrinter(LogPrinter):
+class ProcessingTestLogPrinter(Printer, LogPrinter):
 
     def __init__(self, log_queue):
         LogPrinter.__init__(self, self)
```
sils1297: same

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/f507e6cd3e9a126c9d4c386f61b94373e7aa5649 https://github.com/coala-analyzer/coala/pull/2409/commits/26aa4a5f624d7da2e749de94d9c564fa370d8c3e

coalib/logging_config.json
```diff
@@ -3,7 +3,7 @@
     "handlers": {
         "colored": {
             "class": "logging.StreamHandler",
-            "formatter": "raw",
+            "formatter": "colored",
```
sils1297: yeah the last commit makes more sense if this is squashed into i

tests/output/ConsoleInteractionTest.py
Uran198: It's used in almost every test case.

tests/processes/LogPrinterThreadTest.py
Uran198: Actually, there is no need in these changes anymore. :)

sils1297:
did you read my comment about squashing https://github.com/coala-analyzer/coala/pull/2409/commits/3e1708bd5bc6ebc13c3d8fb9dfa120494f15895e ?

Uran198:
reack 377bf5b 2f90389

Uran198:
@sils1297 Sorry, I've misinterpreted that one. Why do you want it squashed? Previous commit is already very big and this change makes sense to keep separate, though, I changed it a little, so right now it should make more sense. One possibility to make more logical can be to add `coala.colored` handler with colour formatter.

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/3454d2fc19aed8fe78b4f6c7adef4faa487a1374

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/30b9593f2a13c887e274a0687a502b5a871897ae

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/6820399c9c8bbbbc14134d6e00fe83a01a31d320

sils1297:
@Uran198 can you make an asciinema ? How does the new logging look like? Also, is there a way to tell it to pause logging and resume it later, flushing messages that came in during that time?

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/6901813cec6a1239bf651d7d7a74165c87a68ea9

sils1297:
maybe we want to wait with merging this until the release for stability?

Uran198:
@sils1297 Basically, it looks exactly the same. Cause of that I only made asciinema to show that it works: https://asciinema.org/a/3fkds56dxj5fm6qpvrxi860rw . But it would be good if someone else would validate that it is working as well, as I may have forgotten to push some dependencies or something.

There is possibility not to flush messages right away, take a look at [MemoryHandler](https://docs.python.org/3/library/logging.handlers.html#memoryhandler), though it has limited capacity and flushes when it's full. But if needed there will be no problem in inheriting it and overriding default behavior to needed.

It's probably better to wait for a release before merging, IMO.

Uran198:
reack 26ff23ecbf7e9b8af2b0c85f580cbab7527062ea 85500050a57ce1064453ad8dc12fa63593874370  ac25b1cc3eaddd955051963513f8c0ba77407888  5bd4e21b0d9e054d619cae3d85109194be44d6b5 ae69a9595efd1e974731139158f05d645407f933 49b8c5d2974d1417f63e3b5e8853b8531fde57ac

sils1297:
@Uran198 I don't see how this would break our release, my opinion: let's just rebase and merge it now so we can profit from it.

Uran198:
reack 410ace8 fb09f49 c90f25d 0ef2d65 9a86ed5

Uran198:
reack cd2b8c4

tests/output/printers/LogPrinterTest.py
```diff
@@ -66,12 +65,8 @@ def test_set_state(self):
         uut = LogPrinter()
         state = uut.__getstate__()
         uut.__setstate__(state)
-        self.assertEqual(uut.logger, logging.getLogger('coala.raw'))
-
-    def test_colored_printer(self):
-        uut = LogPrinter(printer=ColorPrinter())
-        self.assertIs(uut.logger, logging.getLogger('coala.colored'))
+        self.assertEqual(uut.logger, logging.getLogger('coalib'))
```
hypothesist: Why would this be `self.assertEqual` but the other be `self.assertIs`? IIUC both should be `self.assertIs`?

hypothesist:
So https://github.com/coala-analyzer/coala/pull/2409/commits/f206732a312d99c0038ed40f7b613407e91d60e6 looks alright to me, except for that one question

Uran198:
reack 40dedbe716dd3ea82ad1c2784c0039436b789589

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/d84184b223369c4d45c0fe96eaf89ec50b4cf796

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/4b223ae2471a675ece71690acf49ab5721e6a2a3

sils1297:
hm @Uran198 when I run coala with your PR it doesn't show debug log messages of bears it seems

Uran198:
@sils1297 This will require a lot of testing and may break a lot of things, cause LogPrinter is created in a lot of places, and log_level can be changed in any one of them. I'm thinking of switching already to the `logging.getLogget(__name__)` style, where this happens. This probably should go into d84184b commit, which will bloat it even more. Another alternative, is to go to the compromise and have intermediate version where it's working, but buggy.

Uran198:
reack 0f9c8c0 e68cefa 7ce0fcb b0972e1 bbf7281 4e3bd71 417c40e

Uran198:
@sils1297 Could you verify that it still doesn't show debug messages? Make sure that the logging level is set up correctly, as by default it is info, so it may have been correct behavior when it did not show debug messages.

sils1297:
now it always shows debug messages

sils1297:
even if I run coala -L WARNING

Uran198:
@sils1297 This is really strange cause I did not change the original code, only rebased it.

sils1297:
Unable to reproduce my issues in a clean venv

Uran198:
reack 6e962b321f7e6e4fb7ffc8e7ceb6814ece0c9fd9 048573274dceb873917919af48c1b322f78ada06 f6367751ed286e4cff891cabedbdb1a9f37c31e8 189e6c7027184b229af1a842a6a5ada989d631c7 7dd39e1e5b3a9c34def41b941e602fc0b04bba8e 718b7289788d69bb66b4a45c38e6b6d1601b086b 7a3532b796aa3fbd49e3723a16929ba20afee35a

Uran198:
reack 7fd00141c98e7d003ec524538c6fd25e2c5aa626 c18996e57fc1518fae061b4bf278fe8e97f089b0 8c261cef9f1a652371a2c20d4a3d21ea30c165e4 7fb344c19abcc42926feeb3e74838f2f44e4cb8e 08d0d76f61b5894ab02d3fe0442b6fb67320e1d1 6aba75fada2a0a0f257fbcc526cd422e64523d2b 96972f2249b1b8b47060256f398489bb54c230c9

Uran198:
There were some changes in 8c261ce, but in got reacked anyway. So this one also need reviewing.

sils1297:
unack https://github.com/coala-analyzer/coala/commit/8c261cef9f1a652371a2c20d4a3d21ea30c165e4 then as it needs review

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2409/commits/96972f2249b1b8b47060256f398489bb54c230c9

coalib/settings/ConfigurationGathering.py
```diff
@@ -122,6 +124,16 @@ def warn_config_absent(sections, argument, log_printer):
...
 
 
+def configure_logging_from_json(filename):
+    """
+    Configures logging module based on parameters from JSON file.
+
+    :param filename:    Name of a file with logging configuration.
+    """
+    with open(filename) as cnf_file:
+        logging.config.dictConfig(json.load(cnf_file))
```
underyx: What's the benefit of having a separate JSON file instead of just keeping the config somewhere as a Python dictionary?

Uran198: Separation of concerns. When settings are separated, it's easier to find
and edit them. In particular, if you will want to add, say, the ability for
users to have their own logging settings, it will much easier to change
from configuration in the file. What are your thoughts? Do you think it'll
be better to have it somewhere in code?

sils1297: I think putting it in constants is simpler and also better for performance.
It looks same...

underyx: The separation of concerns requirement can be satisfied by just moving this to its own module. That would still save us the trouble of
- Managing files in the code manually
- Having to bring in JSON serialization
- Having to write JS code

But anyway, I don't think 'separation of concerns' applies at the level. Loading configuration (`ConfigurationGathering.py`) and logging configuration (`logging_config.json`) are the same concern, so it might be okay. The issue would be if the config gatherer suddenly started installing bears that are mentioned in the config file or something :)

sils1297:
oh no conflicts :/

sils1297:
can you rebase? I think it would be good to move the config into code as well.

Uran198:
reack a218bce 7ccd6b2 eb7859e 80dca05

sils1297:
ack https://github.com/coala/coala/pull/2409/commits/578fe472bb3acc50e6ea98bd68d30945ce18fc03#diff-f9283b08ae99b5a2f6d1781935a9b8d5L10

coalib/misc/Constants.py
```diff
@@ -90,3 +90,31 @@
...
+                'handlers': ['raw']
+            }
+        },
+        'formatters': {
+            'raw': {
+                'format': '[%(levelname)s][%(asctime)s] %(message)s',
+                'datefmt': '%X'
+            }
+        }
+    }
```
sils1297: why is this a function and not a constant?

coalib/output/printers/LogPrinter.py
```diff
@@ -1,6 +1,8 @@
 import traceback
 import logging
 
+from pyprint.ColorPrinter import ColorPrinter
```
sils1297: isn't that a recursive import then?

sils1297:
ack https://github.com/coala/coala/pull/2409/commits/17155e3985a2ed8a2c8c115469cb467d9f56250b - beautiful solution!

coalib/output/printers/LogPrinter.py
```diff
@@ -117,7 +120,10 @@ def __init__(self,
         :param timestamp_format: The format string for the
                                  datetime.today().strftime(format) method.
         """
-        self.logger_name = 'coala.raw'
+        if isinstance(printer, ColorPrinter):
+            self.logger_name = 'coala.colored'
+        else:
+            self.logger_name = 'coala.raw'
```
sils1297: this is a bad design pattern, ideally the ColorPrinter does what it needs on its own and this LogPrinter shouldn't have anything to do with it

coalib/misc/Constants.py
sils1297: Sys is usually used anyway, problems are imports from non builtins

sils1297:
@Uran198 you left my comments unanswered

sils1297:
besides those it looks good

Uran198:
@sils1297 sorry for not responding, I agree with your comments but had no time to fix it. And I just forgot to label this WIP.

coalib/misc/Constants.py
Uran198: I'll add the comment for this.

Uran198:
reack c48944e 98fb89d 149e322 6f27f70

sils:
ack https://github.com/coala/coala/pull/2409/commits/6f27f70737f3f50056f5f9c0c86ccdf03cf7f93f

sils:
ack https://github.com/coala/coala/pull/2409/commits/6f27f70737f3f50056f5f9c0c86ccdf03cf7f93f

sils:
ack https://github.com/coala/coala/pull/2409/commits/f41c4179fbbd2fd14ad285963e21289dde2a0571

sils:
ack https://github.com/coala/coala/pull/2409/commits/bc515e990712456b1f7a17246e7a6a451ed1f0e5

coalib/coala.py
```diff
@@ -27,6 +29,8 @@
 
 
 def main():
+    logging.config.dictConfig(get_log_config_dict())
```
sils: maybe that function should actually do the dictConfig invocation as well - less duplication cross all those files

tests/coalaCITest.py
```diff
@@ -60,6 +60,7 @@ def test_fix_patchable_issues(self):
                 "-b", "SpaceConsistencyTestBear",
                 "--settings", "autoapply=true", "use_spaces=True",
                 "default_actions=SpaceConsistencyTestBear:ApplyPatchAction")
+            print(output)
```
sils: debug print?

sils:
can't find anything but those

Uran198:
reack 0312153809527d29358304d7cba7091c6005d381 2646d7e658dfc421f95cc0f9cae347306c5de952 8442f0375964c2c5b97c3f66757ded5a3557aac7 b4c4e04eee149f3f392e7ee9c7c00cfcb4e3a62b b9b8a9d25ccadaea6d55da976c305ce19a357a4c d5c0c3912282bf26cfa5e24dd9b17e9355c791ef

sils:
ack https://github.com/coala/coala/pull/2409/commits/d5c0c3912282bf26cfa5e24dd9b17e9355c791ef

sils:
ack https://github.com/coala/coala/pull/2409/commits/c657d44b62e8fac62c2504c96f6848f302a11c35

sils:
@rultor merge

sils:
@Uran198 can you file issues about what needs to be done for the full switch and maybe close the original one or make it a master issue linking to the others?

sils:
Also this is great work :)

PR #2415, opened 2016-07-02
Syntax highlighting

mr-karan:
Need help
So I am trying to bring in syntax highlighting for the output lines. I am able to get the basic thing done but it's without the dots and also I am unsure on where to place this thing. Currently, I have placed it [here]( https://github.com/coala-analyzer/coala/blob/master/coalib/output/ConsoleInteraction.py#L165)

`print(highlight(line, PythonLexer(), TerminalFormatter()))`

Python Lexer highlights the line. I'll make a more generic version where lexer will be detected by which language is it by taking that from bear property. But before going that step, I want to get this thing correct.

So I want to know which function should I replace with my custom print function?

Because right now, you can see the 2 outputs I want my syntax highlighted version with the dots (i.e. basically remove the function which uses red or blue color highlights)

screenshot from 2016-07-02 13-44-36

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -1,3 +1,5 @@
+from termcolor import colored
+
 try:
     # This import has side effects and is needed to make input() behave nicely
     import readline  # pylint: disable=unused-import
@@ -9,7 +11,6 @@ from pyprint.ConsolePrinter import ConsolePrinter
 
 from coalib.misc.DictUtilities import inverse_dicts
 from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
-from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.results.Result import Result
 from coalib.results.result_actions.ApplyPatchAction import ApplyPatchAction
 from coalib.results.result_actions.OpenEditorAction import OpenEditorAction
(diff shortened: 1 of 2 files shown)


Adrianzatreanu:
thats not how you mark a WIP PR :P

mr-karan:
@Makman2 @sils1297 needs some discussion here.

sils1297:
additional point to consider: we would probably want to add a background color to affected piece of code

So you'll wanna do the highlighting and then do the dots thing to the highlighted text, right? The escape characters have to be ignored for tab width stuff though... complicatd :/

Makman2:
I think we should have some kind of `ConsoleText` class, which allows to set attributes on certain parts of the text.
```python
text = ConsoleText("my very cool text")
text.color(0, 2, "blue")
text.background_color(3, 7, "red")
text.print()
```
or similar^^

The advantage is then that we can apply step by step certain colorings depending on their meaning:
```python
# Highlight whitespaces and replace them with bullets
text = ConsoleText(text.replace(" ", bullet_character))
# this is just pseudocode, but you should get the idea:
for pos in text.find(bullet_character):
    text.color(pos, pos+1, "cyan")

# then we color syntax:
for start, end, color in highlight_syntax(text):
    text.color(start, end, color)

# then we apply background colors:
for result in results:
    text.background_color(result.start, result.end, "red") # <-- we can change colors depending on severity :D

text.print()
```

just a simplification idea :+1:

Much talk, but what explicitly shall be discussed? :P :P

sils1297:
didn't we discuss something like this earlier?

Makman2:
possible :O

mr-karan:
@Makman2 So, I had discovered this function `highlight` from Pygments, which takes a string and a lexer as input. 
highlight("var x=41;', JavascriptLexer()) would implement syntax highlighting on this line. But I am not able to figure out where that result string is. The one without any bullet marks, just plain text, so that I can wrap on this `highlight` class, and that further has bullet marks on it or background color etc. 

Should we go in this direction or make a `ConsoleText` class like you suggested?

Makman2:
@mr-karan what you think makes less work :D though it looks like `highlight` from Pygments is not really usable as it directly outputs to console :/

mr-karan:
(copying from gitter PM):
>so the highlight was wrapped in print() hence it directly outputted to console.
>If I can get just the text without bullets in them (spaces/tabs), I can pass to highlight and it will do the >job
>I am not able to find that thing anywhere in ConsoleInteraction

mr-karan:
@Makman2 Delighted to tell you, that I have got it working somehow. So with Pygments, we don't even need our `print_spaces_tabs_in_unicode` as we can apply a [VisibleWhitespaceFilter](http://pygments.org/docs/filters/?highlight=visiblewhitespace#VisibleWhitespaceFilter) which does **exactly** what we do. 

We can also have background styles etc. I need some more time to clean it up and do further testing before sending a PR :+1:

sils1297:
Nice!

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net
http://coala-analyzer.org/ - http://viperdev.io/ - http://gitmate.com/

Makman2:
yeah that sounds really good :D

mr-karan:
closing this in favour of https://github.com/coala-analyzer/coala/pull/2569 

This branch got really messed up with few wrong rebase and has merge conflicts which I am unable to resolve.

PR #2423, opened 2016-07-03
Sanketdg/assemble

SanketDG:


sils1297:
ack https://github.com/coala-analyzer/coala/pull/2423/commits/047c63408111a6c868e9bc559f83324d65d29521

sils1297:
hey, coverage is lowered?

SanketDG:
@sils1297 yes I will need some help on that. I am not sure why.

Adrianzatreanu:
01e42e2 imo the commit body could use some work, especially the tense used :D

SanketDG:
reack 199b4fe

Makman2:
ack 7df0651

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -130,3 +130,20 @@ def _parse_documentation_with_symbols(self, param_identifiers,
                 parsed.append(self.Description(desc=desc))
 
         return parsed
+
+    @classmethod
+    def assemble_documentation(cls, doccomment, param, ret):
```
Makman2: just `assemble` should suffice :)

Makman2: \+ docstring missing

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -130,3 +130,20 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+                assembled_doc += section_desc[0]
+            elif isinstance(section, cls.Parameter):
+                assembled_doc += (param[0] + section.name + param[1] +
+                                  section_desc[0])
+            elif isinstance(section, cls.ReturnValue):
+                assembled_doc += ret + section_desc[0]
+            for desc in section_desc[1:]:
+                if desc != "":
+                    assembled_doc += desc
+        return assembled_doc
```
Makman2: why not making this function an instance method?

Makman2: and what about indentation?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -14,9 +14,11 @@ class DocstyleDefinition:
     documentation comment (for which language, documentation style/tool used
     etc.).
     """
+    Metadata = namedtuple("Metadata", "param_start, param_end, return_sep")
```
Makman2: though `namedtuple` allows it, I'm not a fan of using a single string to specify attribute names. On module load they get splitted anyway (though the performance impact is minor, but there is one), so let's do this ourselves with using a tuple `("param_start", "param_end", "return_sep")`. Imo the single string function exists for easier typing in the interactive shell^^

But: This is just my preference. If you want you can keep this :P

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -14,9 +14,11 @@ class DocstyleDefinition:
...
 
     @enforce_signature
-    def __init__(self, language: str, docstyle: str, markers: (Iterable, str)):
+    def __init__(self, language: str, docstyle: str, markers: (Iterable, str),
+                 metadata: Metadata):
```
Makman2: missing documentation for `metadata`

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -106,6 +110,18 @@ def markers(self):
         """
         return self._markers
 
+    @property
+    def metadata(self):
+        """
+        A namedtuple of documentation symbols.
+
+        These documentation symbols are used to define certain symbols that
+        define a documentation comment.
```
Makman2: hm they don't really define the comment, they contain more information about special attributes inside the documentation (I think). Imo you should rephrase :+1:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -106,6 +110,18 @@ def markers(self):
...
+
+        Currently consists of ``param_start``, ``param_end`` and ``return_sep``.
```
Makman2: if someone is interested, he/she will look into the `Metadata` namedtuple and see all members. Especially if you change `Metadata` you also have to maintain this little sentence :D

Makman2: --> remove this sentence

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +168,29 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+                    setting = setting.value
+                except AttributeError:
+                    setting = ""
+                values.append(setting)
+            return values
+
+        values = get_from_docstyle_settings('param_start', 'param_end',
+                                            'return_sep')
+
+        metadata = cls.Metadata._make(values)
```
Makman2: this should look better and not so "internal-API-hacking" :D
```
cls.Metadata(*values)
```
if it's a dict:
```
cls.Metadata(**values)
```

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +168,29 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+
+        ignored_settings = ("comment", "param_start", "param_end", "return")
```
Makman2: you should make out of `param_start`, `param_end` and `return` a separate tuple, which you can use above for
`get_from_docstyle_settings(*special_values)` and here you concat them with `"comment"` :

```
ignored_settings = ("comment",) + special_settings
```

`special_settings` is maybe a bad name, but I hope you get the principle :D

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +168,29 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+
         marker_sets = (tuple(value)
                        for key, value in
-                       filter(lambda kv: not kv[0].startswith("comment"),
-                              docstyle_settings.contents.items()))
+                       filter(lambda kv: not kv[0].startswith(
+                           ignored_settings),
+                       docstyle_settings.contents.items()))
```
Makman2: if you change these line, you can use list comprehension:
```
... for key, value in docstyle_settings.contents.items() if not key.startswith(ignored_settings)
```
:+1:

coalib/bearlib/languages/documentation/default.coalang
```diff
@@ -1,9 +1,18 @@
...
+param_start = :param\ # here's a space
+param_end = :
+return_sep = :return:
 
 [JAVA]
 doc-marker1 = /**, \ *, \ */
 doc-marker2 = /**, , \ */
+param_start = @param\ \ # here's a space
+param_end = \ # here's a space
+return_sep = @return\ # space
```
Makman2: why not `here's a space`? :P :P

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +168,29 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+            values = list()
+            for req_setting in args:
+                setting = dict(
+                    docstyle_settings.contents.items()).get(req_setting)
+                try:
+                    setting = setting.value
+                except AttributeError:
+                    setting = ""
+                values.append(setting)
+            return values
```
Makman2: hm this looks like some `get_values_or_default`, maybe you should just hardcode that below?
```
param_start_value = section.get("param_start", "")
```

Makman2: or you use a loop:
```
special_settings = ('param_start', ...)
values = [section.get(setting, "") for setting in special_settings]
```

Makman2:
ack a47ac62

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -30,8 +30,9 @@ def __init__(self, documentation, language,
         :param range:         The position range of type TextRange.
         """
         self.documentation = documentation
-        self.language = language.lower()
-        self.docstyle = docstyle.lower()
+        self.metadata = docstyle_definition.metadata
+        self.language = docstyle_definition.language.lower()
+        self.docstyle = docstyle_definition.docstyle.lower()
```
Makman2: naaaah now use properties for both that forward to `docstyle_definition` ;D ;D

```
@property
def language(self):
     return self.docstyle.language
```

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -60,10 +61,6 @@ def parse(self):
         elif self.language == "java" and self.docstyle == "default":
             return self._parse_documentation_with_symbols(
                 ("@param  ", " "), "@return ")
-        else:
-            raise NotImplementedError(
-                "Documentation parsing for {0.language!r} in {0.docstyle!r}"
-                " has not been implemented yet".format(self))
```
Makman2: hm you don't use the values from the `docstyle_definition` yet, I would keep this until you change this in the next commits :+1:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
SanketDG: no this omits any form of magic or strange behaviour :+1:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
SanketDG: but I am doing `setting.value` so I am not sure how I would fit this in ^

SanketDG: I think there is no need to specify a default value, because I do it later anyway.

coalib/bearlib/languages/documentation/DocumentationComment.py
SanketDG: what difference does it make? :smile:

Makman2: we don't want to duplicate information. As the information resides already inside the docstyle, we just need to refer to it instead of copying the string reference.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
Makman2: but you do inside `get_from_docstyle_settings`:
```python
try:
    setting = setting.value
except AttributeError:
    setting = ""
```
I don't understand completely what goes into this function, so I'm a bit confused about it and I have the feeling that it is not really necessary^^ (like said, looks like some "get-it-or-get-default" function :D)

SanketDG:
reack aec16c2 8f0a067 36ee174

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -36,11 +36,13 @@ def _extract_doc_comment_simple(content, line, column, markers):
     while line < len(content):
         pos = content[line].find(markers[2])
         if pos == -1:
-            doc_comment += content[line][align_column:]
+            if content[line] == "\n":
+                doc_comment += "\n"
+            else:
+                doc_comment += content[line][align_column:]
```
Makman2: maybe we should check whether `conent[line][align_column:]` is empty? In this case even not right indented code can't fail (the examples have 4-spaces indentation, I'm highlighting them explicitly with dots):

```
....my documentation

..x

.it will start here
```

would then extract:
```
my documentation



will start here
```

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -36,11 +36,13 @@ def _extract_doc_comment_simple(content, line, column, markers):
...
         else:
             doc_comment += content[line][align_column:pos]
             return line, pos + len(markers[2]), doc_comment
-
```
Makman2: just keep that newline :)

Makman2:
don't know whether we should do 	de59748 , as such load functions are often quite specific to their directories. If you think this modularization makes sense, you can keep it :+1:
Or maybe we can just move this function outside the class, your choice :D

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    assembled_doc = assemble_documentation(doccomment, param_sym, ret_sym)
+    assembled_doc = assembled_doc.splitlines(keepends=True)
+    assembled = indent + markers[0]
+    for section in assembled_doc:
+        if section == "\n":
+            assembled += '\n'
+        else:
+            assembled += indent + markers[1] + section
+    return assembled + indent + markers[2]
```
Makman2: I think we should make this function an instance method of `DocumentationComment`, while `assemble_documentation` above takes `param_sym`, `ret_sym` and so on and generates a new `DocumentationComment` (like discussed :D).

So it would work like this:
```python
docs = list(extract_documentation(my_file))
doc = docs[0]  # just get some documentation

metadata = doc.metadata
# do whatever you want to do with the metadata

new_doc = assemble_documentation(str(doc), metadata.param, metadata.return)
new_doc.assemble()  # <-- this is the `full_assemble` function
```

You think that makes sense? :D

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,71 @@
...
+ReturnValue = DocumentationComment.ReturnValue
+
+
+def assemble_documentation(doccomment, param_sym, ret_sym):
+    """
+    Assembles a list of parsed documentation comments.
+
+    This function just assembles the documentation comment
+    itself, without the markers and indentation.
+
```
Makman2: a doctest would be really helpful, so the user sees how the format of the newly assembled documentation looks like :+1:

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,71 @@
...
+        comment.
+    """
+    assembled_doc = assemble_documentation(doccomment, param_sym, ret_sym)
+    assembled_doc = assembled_doc.splitlines(keepends=True)
+    assembled = indent + markers[0]
+    for section in assembled_doc:
+        if section == "\n":
+            assembled += '\n'
+        else:
+            assembled += indent + markers[1] + section
```
Makman2: --> `section += '\n' if section == '\n' else indent + markers[1] + section`

Makman2: you can go even farther with taking in the outer loop:
```python
assembled = ''.join('\n' if section == '\n' else indent + markers[1] + section
                    for section in assembled_doc)
```

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,71 @@
...
+        A string that defines how a return statement definition occurs in
+        a documentation comment.
+    :param markers:
+        The markers to be used.
+    :param indent:
+        The indentation level to be used for each line of the documentation
+        comment.
+    """
+    assembled_doc = assemble_documentation(doccomment, param_sym, ret_sym)
+    assembled_doc = assembled_doc.splitlines(keepends=True)
```
Makman2: I would do the splitlines on the same line like the first assignment:
```python
assembled_doc = assemble_documentation(occomment, param_sym, ret_sym).splitlines(keepends=True)
```
if you need to linebreak it -->

Makman2:
```python
assembled_doc = assemble_documentation(
    doccomment, param_sym, ret_sym).splitlines(keepends=True)
```

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,71 @@
...
+    assembled_doc = ""
+    for section in doccomment:
+        section_desc = section.desc.splitlines(keepends=True)
+        if isinstance(section, Description):
+            assembled_doc += section_desc[0]
+        elif isinstance(section, Parameter):
+            assembled_doc += (param_sym[0] + section.name + param_sym[1] +
+                              section_desc[0])
+        elif isinstance(section, ReturnValue):
+            assembled_doc += ret_sym + section_desc[0]
```
Makman2: I think because of aesthethic reasons it would be good to introduce a newline here, so the `for`-loop distinguishes better from the if-else cases :)

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -29,6 +32,8 @@ def __init__(self, language: str, docstyle: str, markers: (Iterable, str)):
                          or a single marker/delimiter string iterable that
                          identify a documentation comment. See ``markers``
                          property for more details on markers.
+        :param metadata: A namedtuple consisting of certain attributes that
+                         are present in the documentation comment.
```
Makman2: hmm I don't like that :D :D

it's super generic I think, "namedtuple consisting of certain attributes that are present in the documentation comment".

The question I would still have:
- what are "attributes", that are present in the documentation comment?

Also you should try to mention that it contains more descriptive properties about the layout/format of such a metadata field that appears inside a doccomment.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +168,30 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+                    setting = setting.value
+                except AttributeError:
+                    setting = ""
+                values.append(setting)
+            return values
+
+        metadata_settings = ("param_start", "param_end", "return")
+
+        values = get_from_docstyle_settings('param_start', 'param_end',
+                                            'return_sep')
```
Makman2: `return_sep` does not match `return` from above

also you should use
```python
values = get_from_docstyle_settings(*metadata_settings)
```
to avoid duplication^^

Makman2:
ack ea30ec8

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -30,15 +30,26 @@ def __init__(self, documentation, language,
...
         self.indent = indent
         self.marker = marker
         self.range = range
 
     def __str__(self):
         return self.documentation
 
+    @property
+    def language(self):
+        return self.docstyle_definition.language.lower()
```
Makman2: I think the docstyle_definition is already lowered, just return `language` itself :)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -30,15 +30,26 @@ def __init__(self, documentation, language,
...
+
+    @property
+    def docstyle(self):
+        return self.docstyle_definition.docstyle.lower()
```
Makman2: same here

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -30,15 +30,26 @@ def __init__(self, documentation, language,
...
+
+    @property
+    def metdata(self):
+        return self.docstyle_definition.metadata
```
Makman2: 1. typo --> `metdata` --> `metadata`
2. you sure you want to expose this? someone could get the idea that `doc.parse()` is the same like `doc.metadata`^^

Makman2:
unack fd8d43f

Makman2:
I think you've accidentally acked fd8d43f but you meant ea30ec8 instead :P

SanketDG:
ack 281d6ef d258890 0396625

Makman2:
ack  a228120

coalib/bearlib/languages/documentation/AssembledDocumentation.py
Makman2: docstyle still missing :3

coalib/bearlib/languages/documentation/AssembledDocumentation.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
+            assert isinstance(section, ReturnValue)
+            assembled_doc += docstyle_definition.metadata.return_sep + \
+                section_desc[0]
+
+        for desc in section_desc[1:]:
+            assembled_doc += desc
+
+    return DocumentationComment(assembled_doc, docstyle_definition, indent,
+                                marker, range)
```
Makman2: hm not sure if this function is that useful^^

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,16 @@ def _parse_documentation_with_symbols(self, param_identifiers,
                 parsed.append(self.Description(desc=desc))
 
         return parsed
+
+    def assemble(self):
+        """
+        Assembles a list of parsed documentation comments.
```
Makman2: hm not a list :D

coalib/bearlib/languages/documentation/AssembledDocumentation.py
Makman2: and I would put this one into an own commit, `assemble` below is independent from this function :)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,16 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+
+        This function assembles the whole documentation comment,
+        with the given markers and indentation.
+        """
+        lines = self.documentation.splitlines(keepends=True)
+        assembled = self.indent + self.marker[0]
+        assembled += ''.join('\n' if line == '\n' else self.indent +
```
Makman2: I think you need to add to the if: `and not self.marker[1]`, as only if there's no middle-marker you can leave the line completely empty :)

e.g.
```python
# hello
#
# world
```

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -106,6 +117,20 @@ def markers(self):
...
+    def metadata(self):
+        """
+        A namedtuple of certain attributes present in the documentation.
+
+        These attributes are used to define parts of the documentation.
+        """
+        escaped_metadata_dict = self._metadata._asdict()
+        unescaped_metadata_dict = {key: unescape(value)
+                                   for (key, value) in
+                                   escaped_metadata_dict.items()}
```
Makman2: shouldn't `unescape` be done inside the coafile-parser?

coalib/bearlib/languages/documentation/AssembledDocumentation.py
SanketDG: Does it really need one though? I mean this isn't one independent class, it depends on a lot of other classes, plus I'd have to set up a stuff. Even if I write one, it won't be succinct for sure.

coalib/bearlib/languages/documentation/AssembledDocumentation.py
SanketDG: > hm not sure if this function is that useful^^

What do you mean?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
SanketDG: Yes but for this case this is a namedtuple, so unescaping does happen within every coafile-parser, just not the stuff inside a namedtuple.

coalib/bearlib/languages/documentation/AssembledDocumentation.py
Makman2: you take a list of doc-comments and copy their descriptions and other metadata together.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
Makman2: don't understand that :O

SanketDG:
reack 6eb9ee0 1b4a102 edf9f3e 31a8008

tests/bearlib/languages/documentation/TestUtils.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+
+def load_testdata(filename):
+    filename = os.path.join(os.path.dirname(
+        os.path.realpath(__file__)),
+        os.path.join("documentation_extraction_testdata",
+                     filename))
+
+    with open(filename, "r") as test_file:
```
Makman2: you can omit `"r"` :D

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +172,25 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
             raise KeyError("Language {!r} is not defined for docstyle {!r}."
                            .format(language, docstyle))
 
+        def get_from_docstyle_settings(*args):
+            values = list()
+            for req_setting in args:
+                setting = str(
+                    docstyle_settings.get(req_setting, ""))
+                values.append(setting)
+            return values
```
Makman2: can be shortened :)
```python
return [str(docstyle_settings.get(req_setting, "")) for req_setting in args]
```
btw dou you need to invoke `str()` on `setting.get()`? isn't already a string returned?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +172,25 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+
+        metadata_settings = ("param_start", "param_end", "return_sep")
+
+        values = get_from_docstyle_settings(*metadata_settings)
+
+        metadata = cls.Metadata(*values)
```
Makman2: I would inline `values` :)
-->
```python
metadata = cls.Metadata(*get_from_docstyle_settings(*metadata_settings))
```

coalib/bearlib/languages/documentation/DocstyleDefinition.py
Makman2: maybe you could inline this then below where you use this function ;)

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +172,25 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+
+        ignored_settings = ("comment",) + metadata_settings
+
         marker_sets = (tuple(value)
                        for key, value in
-                       filter(lambda kv: not kv[0].startswith("comment"),
-                              docstyle_settings.contents.items()))
+                       docstyle_settings.contents.items()
+                       if not key.startswith(ignored_settings))
```
Makman2: maybe -->
```python
if key not in metadata_settings and not key.startswith("comment")
```
?
so it performs a full-match for your special settings, and something like `param_start_custom` does not get matched :)

coalib/bearlib/languages/documentation/doxygen.coalang
```diff
@@ -34,10 +34,16 @@ doc-marker6 = //!, //!, //!
 [PYTHON]
 doc-marker1 = """, , """
 doc-marker2 = \#\#, \#, \#
+param_start = @param\ # here's a space
+param_end = \ # space again
+return_sep = @return\ # not again
```
Makman2: "not again"? :O

tests/bearlib/languages/documentation/DocstyleDefinitionTest.py
```diff
@@ -8,49 +8,68 @@
 
 class DocstyleDefinitionTest(unittest.TestCase):
 
+    Metadata = DocstyleDefinition.Metadata
+    dummy_metadata = Metadata._make([":param ", ":", ":return:"])
```
Makman2: --> `Metadata(":param ", ":", ":return:")`

tests/bearlib/languages/documentation/DocstyleDefinitionTest.py
```diff
@@ -73,7 +92,13 @@ def test_load(self):
         self.assertEqual(result.docstyle, "default")
         self.assertEqual(result.markers, (('"""', '', '"""'),))
 
+        self.assertEqual(result.metadata, self.dummy_metadata)
+        self.assertEqual(result.metadata.param_start, ":param ")
+        self.assertEqual(result.metadata.param_end, ":")
+        self.assertEqual(result.metadata.return_sep, ":return:")
+
     def test_load_external_coalang(self):
+        empty_metadata = self.Metadata._make(['', '', ''])
```
Makman2: --> `self.Metadata('', '', '')`

coalib/parsing/LineParser.py
```diff
@@ -125,10 +127,19 @@ def __separate_by_first_occurrence(string,
...
+
+        first_part_stripped = unescaped_rstrip(first_part)
+        second_part_stripped = unescaped_rstrip(second_part)
+
+        if position_is_escaped(second_part, len(second_part) - 1):
+            return (first_part.lstrip().strip("\n"),
+                    second_part.lstrip().strip("\n"))
+
+        return (first_part_stripped.lstrip().strip("\n"),
+                second_part_stripped.lstrip().strip("\n"))
```
Makman2: is the commit order right? it needs to be applied before 37b58ec79cec702e1b2c40ecfda30181a53118d3

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -60,7 +71,7 @@ def parse(self):
         elif self.language == "java" and self.docstyle == "default":
             return self._parse_documentation_with_symbols(
                 ("@param  ", " "), "@return ")
-        else:
+        else:  # pragma: no cover
```
Makman2: why that pragma? :O

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -227,6 +227,8 @@ def extract_documentation_with_markers(content, markers, language, docstyle):
     # begin sequence we initially want to search for in source code. Then
     # the possible found documentation match is processed further with the
     # rest markers.
+    markers = docstyle_definition.markers
```
Makman2: maybe inline that? your choice :)

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -281,5 +283,4 @@ def extract_documentation(content, language, docstyle):
     """
     docstyle_definition = DocstyleDefinition.load(language, docstyle)
     return extract_documentation_with_markers(content,
-                                              docstyle_definition.markers,
-                                              language, docstyle)
+                                              docstyle_definition)
```
Makman2: should fit now into a single line together with `return extract_...` :)

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -47,12 +54,7 @@ def test_fields(self):
         self.assertEqual(str(uut), "qwertzuiop")
         self.assertEqual(uut.marker, ("##", "#", "#"))
         self.assertEqual(uut.range, None)
-
-    def test_not_implemented(self):
-        not_implemented = DocumentationComment(
-            "some docs", "nolang", "doxygen", None, None, None)
-        with self.assertRaises(NotImplementedError):
-            not_implemented.parse()
```
Makman2: why do you remove this? :O

Makman2: why do you remove this? :O

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -47,12 +54,7 @@ def test_fields(self):
...
+        self.assertEqual(uut.metadata, python_doxygen_metadata)
```
Makman2: this test would belong to the `DocstyleDefinitionTest` imo^^

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,67 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+                      marker, indent, range):
+        r"""
+        Assembles a list of parsed documentation comments.
+
+        This function just assembles the documentation comment
+        itself, without the markers and indentation.
+
+        >>> from coalib.bearlib.languages.documentation.DocumentationComment \
+        ...     import DocumentationComment
+        >>> from coalib.bearlib.languages.documentation.DocstyleDefinition \
```
Makman2: is the backslash needed? do you need a raw-string too?

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,67 @@ def _parse_documentation_with_symbols(self, param_identifiers,
                 parsed.append(self.Description(desc=desc))
 
         return parsed
+
+    @classmethod
+    def from_metadata(cls, doccomment, docstyle_definition,
+                      marker, indent, range):
+        r"""
+        Assembles a list of parsed documentation comments.
```
Makman2: maybe "... of parsed documentation comment metadata." ?

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,67 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+        :param docstyle_definition:
+            The ``DocstyleDefinition`` instance that defines what docstyle is
+            being used in a documentation comment.
+        :param marker:
+            The markers to be used in the documentation comment.
+        :param indent:
+            The indentation to be used in the documentation comment.
+        :param range:
+            The range of the documentation comment.
+
```
Makman2: no need for this empty line :)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,67 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+        >>> parsed_doc = [Description(desc='\nDescription\n'),
+        ...               Parameter(name='age', desc='Age\n')]
+        >>> str(DocumentationComment.from_metadata(
+        ...                  parsed_doc, python_default,
+        ...                  python_default.markers[0], 4,
+        ...                  TextRange.from_values(0, 0, 0, 0)))
+        '\nDescription\n:param age:Age\n'
+
+        :param doccomment:
+            The list of parsed documentation comments.
```
Makman2: --> "The list of parsed documentation comment metadata."?

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,67 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+            elif isinstance(section, cls.Parameter):
+                assembled_doc += (docstyle_definition.metadata.param_start +
+                                  section.name +
+                                  docstyle_definition.metadata.param_end +
+                                  section_desc[0])
+
+            else:
+                assert isinstance(section, cls.ReturnValue)
+                assembled_doc += docstyle_definition.metadata.return_sep + \
+                    section_desc[0]
```
Makman2: use braces :D
```
assembled_doc += (docstyle_definition.metadata.return_sep +
    section_desc[0])
```

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -168,3 +168,25 @@ def test_java_default(self):
                          desc='     the concatenated string\n')]]
 
         self.assertEqual(expected, parsed_docs)
+
+
+class AssembledDocumentationTest(unittest.TestCase):
+
+    def test_python_assembly(self):
```
Makman2: wanna call it `test_from_metadata` and inline it into the above test class?

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -190,3 +190,6 @@ def test_python_assembly(self):
                           for doc in parsed_docs]
 
         self.assertEqual([str(doc) for doc in assembled_docs], original)
+
+        for assembled_doc in assembled_docs:
+            self.assertIn(assembled_doc.assemble(), "".join(data))
```
Makman2: join the data before the loop :)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -205,3 +205,17 @@ def from_metadata(cls, doccomment, docstyle_definition,
...
+
+        This function assembles the whole documentation comment, with the
+        given markers and indentation.
+        """
+        lines = self.documentation.splitlines(keepends=True)
+        assembled = self.indent + self.marker[0]
+        assembled += ''.join('\n' if line == '\n' and not self.marker[1]
+                             else self.indent + self.marker[1] + line
+                             for line in lines)
+        return assembled + self.indent + self.marker[2]
```
Makman2: what if the last line contains no `\n`? then you overindent it inside the doccomment:
e.g.
```
    /** this is my comment **/
```
would be assembled to
```
    /** this is my comment     **/
```
\+ please test for this case :)

coalib/bearlib/languages/documentation/DocstyleDefinition.py
SanketDG: Yes I need to do str() because that's where the "unescaping" happens.

coalib/bearlib/languages/documentation/doxygen.coalang
SanketDG: sorry thats just me trying to show some bad humor.

tests/bearlib/languages/documentation/DocumentationCommentTest.py
SanketDG: this is to make sure that the `.metadata` is working properly. All of the other properties have been done the same too.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
SanketDG: I feel like its a to long name. :P

coalib/bearlib/languages/documentation/DocumentationComment.py
SanketDG: well it fails without them. :P

tests/bearlib/languages/documentation/DocumentationCommentTest.py
SanketDG: um, assembling documentation is a different thing, wouldn't it be better if it was a different class.

coalib/parsing/LineParser.py
Makman2: \+ wanna do a test for that?

for this case -->
```
hello = world\ # yes here's a space
```

coalib/parsing/LineParser.py
```diff
@@ -125,10 +127,19 @@ def __separate_by_first_occurrence(string,
...
+        first_part = string[:delim_pos]
+        second_part = string[delim_pos + (
+            len(used_delim) if strip_delim else 0):]
+
+        first_part_stripped = unescaped_rstrip(first_part)
+        second_part_stripped = unescaped_rstrip(second_part)
+
+        if position_is_escaped(second_part, len(second_part) - 1):
+            return (first_part.lstrip().strip("\n"),
+                    second_part.lstrip().strip("\n"))
```
Makman2: you can move this if stuff above `fist_part_stripped` and `second_part_stripped`

Makman2: \+ to avoid duplication with `lstrip().strip("\n")`, you could reassign `first_part` and `second_part`

```python
if not position_is_escaped(...):
    first_part = unescaped_rstrip(first_part)
    second_part = unescaped_rstrip(second_part)

return (first_part.lstrip().strip("\n"), second_part.lstrip().strip("\n"))
```

Makman2: \+ as you have already left-stripped, you can invoke `rstrip()` instead of plain `strip()` which should be slightly faster :+1:

coalib/parsing/LineParser.py
```diff
@@ -125,10 +127,19 @@ def __separate_by_first_occurrence(string,
...
-            string[delim_pos + (len(used_delim) if strip_delim else 0):].strip(
-                " \n"))
+        first_part = string[:delim_pos]
+        second_part = string[delim_pos + (
+            len(used_delim) if strip_delim else 0):]
+
+        first_part_stripped = unescaped_rstrip(first_part)
+        second_part_stripped = unescaped_rstrip(second_part)
+
+        if position_is_escaped(second_part, len(second_part) - 1):
```
Makman2: ah don't you need to check on `first_part_stripped` and `second_part_stripped`?

sils1297:
there has been a review, marking WIP, @SanketDG can you reiterate?

coalib/parsing/LineParser.py
SanketDG: Not really, ideally you want to stop the trimming once.

Makman2:
ack 667d232

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,66 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+        >>> from coalib.results.TextRange import TextRange
+        >>> Description = DocumentationComment.Description
+        >>> Parameter = DocumentationComment.Parameter
+        >>> python_default = DocstyleDefinition.load("python3", "default")
+        >>> parsed_doc = [Description(desc='\nDescription\n'),
+        ...               Parameter(name='age', desc='Age\n')]
+        >>> str(DocumentationComment.from_metadata(
+        ...                  parsed_doc, python_default,
+        ...                  python_default.markers[0], 4,
+        ...                  TextRange.from_values(0, 0, 0, 0)))
```
Makman2: indentation

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,66 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+        >>> from coalib.bearlib.languages.documentation.DocumentationComment \
+        ...     import DocumentationComment
+        >>> from coalib.bearlib.languages.documentation.DocstyleDefinition \
+        ...     import DocstyleDefinition
+        >>> from coalib.results.TextRange import TextRange
+        >>> Description = DocumentationComment.Description
+        >>> Parameter = DocumentationComment.Parameter
+        >>> python_default = DocstyleDefinition.load("python3", "default")
+        >>> parsed_doc = [Description(desc='\nDescription\n'),
+        ...               Parameter(name='age', desc='Age\n')]
```
Makman2: maybe you want add a space before `Age`? so the assembled documentation looks a bit nicer :)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -141,3 +141,66 @@ def _parse_documentation_with_symbols(self, param_identifiers,
...
+                                  docstyle_definition.metadata.param_end +
+                                  section_desc[0])
+
+            else:
+                assert isinstance(section, cls.ReturnValue)
+                assembled_doc += (docstyle_definition.metadata.return_sep +
+                                  section_desc[0])
+
+            for desc in section_desc[1:]:
+                assembled_doc += desc
```
Makman2: as you add everytime `section_desc[0]` as last element, you could remove it everywhere, changing this for to
```
for desc in section_desc:
```
and also use join:
`assembled_doc += ''.join(section_desc)`

you get also rid of the first if case :)

Makman2: ah and you need to change the `else` to an `elif x is ReturnValue` :+1:

tests/bearlib/languages/documentation/DocumentationCommentTest.py
Makman2: I think it should be inside `GeneralDocumentationCommentTest` as it's general language independent functionality^^

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -204,3 +204,22 @@ def from_metadata(cls, doccomment, docstyle_definition,
...
+    def assemble(self):
+        """
+        Assembles parsed documentation to the original documentation.
+
+        This function assembles the whole documentation comment, with the
+        given markers and indentation.
+        """
+        lines = self.documentation.splitlines(keepends=True)
+        assembled = ((self.indent if lines[0][0] == '\n' else '') +
+                     self.marker[0])
```
Makman2: not just `self.indent + self.marker[0]` as the extracted documentation starts after the first marker (which can't be `''`)?

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -204,3 +204,22 @@ def from_metadata(cls, doccomment, docstyle_definition,
...
+        if len(lines) == 1:
+            return assembled + lines[0] + self.marker[2]
```
Makman2: do you need this case actually? doesn't the join below handle this case too?

coalib/bearlib/languages/documentation/DocumentationComment.py
Makman2: ahh nvm understood, because the join already accounts for indentation

coalib/bearlib/languages/documentation/DocumentationComment.py
Makman2: nvm

Makman2: oh but
```
>>> "hello\n".splitlines(True)
['hello\n']
```
but the end marker has to be on the next line^^

Makman2:
ack 18d8f67

Makman2:
ack 89db4c0

Makman2:
ack 2ea9a23

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -36,7 +36,10 @@ def _extract_doc_comment_simple(content, line, column, markers):
     while line < len(content):
         pos = content[line].find(markers[2])
         if pos == -1:
-            doc_comment += content[line][align_column:]
+            if content[line][align_column:] == "":
+                doc_comment += "\n"
+            else:
+                doc_comment += content[line][align_column:]
```
Makman2: --> 
```
doc_comment += "\n" if conent[line][align_column:] == "" else content[line][align_column:]
```

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -1,8 +1,9 @@
-from collections import Iterable
+from collections import Iterable, namedtuple
 import os.path
 
 from coala_utils.decorators import (
     enforce_signature, generate_eq, generate_repr)
+from coala_utils.string_processing import unescape
```
Makman2: unused import
\+
the commit message: the last sentence does not fit any more :)

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -152,9 +172,15 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
             raise KeyError("Language {!r} is not defined for docstyle {!r}."
                            .format(language, docstyle))
 
+        metadata_settings = ("param_start", "param_end", "return_sep")
+
+        metadata = cls.Metadata(*[str(docstyle_settings.get(req_setting, ""))
+                                  for req_setting in metadata_settings])
```
Makman2: you don't need the list `[]`, unpacking generators is also possible :)
so just replace the `[]` with `()`

tests/bearlib/languages/documentation/DocstyleDefinitionTest.py
```diff
@@ -73,7 +92,13 @@ def test_load(self):
         self.assertEqual(result.docstyle, "default")
         self.assertEqual(result.markers, (('"""', '', '"""'),))
 
+        self.assertEqual(result.metadata, self.dummy_metadata)
+        self.assertEqual(result.metadata.param_start, ":param ")
+        self.assertEqual(result.metadata.param_end, ":")
+        self.assertEqual(result.metadata.return_sep, ":return:")
```
Makman2: I think the last three asserts are not needed, as if the tuple does equal the items do too

tests/bearlib/languages/documentation/DocstyleDefinitionTest.py
```diff
@@ -73,7 +92,13 @@ def test_load(self):
...
+
     def test_load_external_coalang(self):
+        empty_metadata = self.Metadata('', '', '')
```
Makman2: not for this PR, but maybe we can check for this case on docstyle load and use `None` for invalid / missing metadata definitions :+1:

Makman2:
ack 644b79f

Makman2:
ack 01b5cc7

coalib/bearlib/languages/documentation/DocumentationComment.py
SanketDG: no we talked about this, this decreases coverage

SanketDG: oh nvm you meant after doing that change

coalib/bearlib/languages/documentation/DocstyleDefinition.py
SanketDG: OH!

tests/bearlib/languages/documentation/DocstyleDefinitionTest.py
SanketDG: yeah that's just me trying to sho that's inside `metadata`

tests/bearlib/languages/documentation/DocumentationCommentTest.py
SanketDG: Um yes but I still would consider it to be a different class for the sake of modularity. Agreed that it is general, but I would put it more like a sub-class of general.

SanketDG:
reack 15f26cb e92e357 a080137 e009dbf f64071b f532601

Makman2:
ack dd82bbc

Makman2:
ack c2063ea

tests/bearlib/languages/documentation/DocumentationCommentTest.py
Makman2: I think I need to persist a bit on this, as when I read the class and test-function name, I think someone who has never seen this code doesn't understand what is actually tested^^

1. You test the function `from_metadata`, so I would call the test function `test_from_metadata`
2. `from_metadata` is language independent, so the test imo belongs to the general class, even though you use the python-default docstyle-definition.
3. It's some kind of "assembly" (assembling the parsed metadata), but it does not really refer to the function name. So the class should be called at least `FromMetadataTest` (or `MetadataAssemblyTest`)

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -197,3 +197,22 @@ def from_metadata(cls, doccomment, docstyle_definition,
...
+        given markers and indentation.
+        """
+        lines = self.documentation.splitlines(keepends=True)
+        assembled = self.indent + self.marker[0]
+        if len(lines) == 1:
+            assembled += lines[0]
+        else:
+            assembled += ''.join('\n' if line == '\n' and not self.marker[1]
+                                 else self.indent + self.marker[1] + line
+                                 for line in lines)
```
Makman2: got a corner case where it doesn't work:
Assume this doccomment (with markers `'/**', ' *', '*/`):
```
    /** line 1
     *  line2 */
```
the documentation string would be:
```
' line1\n  line2'
```
gets assembled to
```
    /**     * line1
     *  line2*/
```
please fix that + add a test case for that :+1:

Makman2: I just found about this function:
https://docs.python.org/3.4/library/textwrap.html#textwrap.indent

maybe you can use this one :)
so you just need to generate the stuff without indentation and make the indent special only for the start-marker and end-marker (if it's that easier, I don't know :3)

Makman2:
unack 9685c16 6b01f4b

coalib/bearlib/languages/documentation/DocumentationComment.py
SanketDG: It definitely doesn't make it easy. :3
The thing is, it is for python stuff.

I think a custom solution would work better to work across all languages and styles. :+1:

SanketDG:
reack 52d2f7b 198dab4 d659ed8 b82e7d6 8450cf3 5c08b6d 66d848b 1997323

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -64,6 +64,25 @@ def test_not_implemented(self):
         with self.assertRaises(NotImplementedError):
             not_implemented.parse()
 
+    def test_from_metadata(self):
+
```
Makman2: unneeded newline

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -64,6 +64,25 @@ def test_not_implemented(self):
...
+                       for doc in
+                       extract_documentation(data, "python", "default")]
+
+        docstyle_definition = DocstyleDefinition.load("python", "default")
+
+        assembled_docs = [DocumentationComment.from_metadata(
+                          doc[0], docstyle_definition, doc[1], doc[2], doc[3])
+                          for doc in parsed_docs]
+
+        self.assertEqual([str(doc) for doc in assembled_docs], original)
```
Makman2: With little modification you can also test for complete `DocumentationComment` object equality! :D
(As `DocumentationComment` uses `generate_eq`)

-->
```
original = list(extract_documentation(...))
parsed_docs = [(...)
             for doc in original]

# ...
self.assertEqual(assembled_docs, original)
```

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -195,3 +195,22 @@ def test_java_default(self):
                          desc='     the concatenated string\n')]]
 
         self.assertEqual(expected, parsed_docs)
+
+
+class DocumentationAssemblyTest(unittest.TestCase):
+
+    def test_python_assembly(self):
+
```
Makman2: unneeded newline

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -195,3 +195,22 @@ def test_java_default(self):
...
+        data = load_testdata("default.py")
+        docs = "".join(data)
+
+        for doc in extract_documentation(data, "python", "default"):
+            self.assertIn(doc.assemble(), docs)
+
+    def test_c_assembly(self):
+
```
Makman2: unneeded newline

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -135,8 +135,8 @@ def test_python_default(self):
                                    'Some more foobar-like text.\n')],
             [self.Description(desc='\nA nice and neat way of '
                                    'documenting code.\n'),
-             self.Parameter(name='radius', desc=' The explosion radius.\n')],
-            [self.Description(desc='\nA function that returns 55.\n')],
+             self.Parameter(name='radius', desc=' The explosion radius. ')],
+            [self.Description(desc='A function that returns 55.')],
```
Makman2: this doesn't belong to this commit :O

coalib/bearlib/languages/documentation/DocumentationComment.py
```diff
@@ -197,3 +197,20 @@ def from_metadata(cls, doccomment, docstyle_definition,
...
+    def assemble(self):
+        """
+        Assembles parsed documentation to the original documentation.
+
+        This function assembles the whole documentation comment, with the
+        given markers and indentation.
+        """
+        lines = self.documentation.splitlines(keepends=True)
+        assembled = self.indent + self.marker[0]
+        assembled += lines[0]
```
Makman2: what if `len(lines) == 0`?

tests/bearlib/languages/documentation/DocumentationCommentTest.py
Makman2: nvm ... concentration level drops ... :P

coalib/bearlib/languages/documentation/DocumentationComment.py
SanketDG: I will just return `''` for that

SanketDG: sorry `self.marker[0] + self.marker[2]`

Makman2:
ack f42bf20

Makman2:
ack d54201f

SanketDG:
reack 78ff315 dc35a0a 3a78aa9 edc67aa 31b0410 7de9aed e442dce f731ee4 5df5932 07135f5

SanketDG:
@rultor merge

PR #2461, opened 2016-07-19
WIP: Performance improvement

sils:
CC @hypothesist

### DIFF
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -15,7 +15,6 @@ import functools
 
 from pyprint.ConsolePrinter import ConsolePrinter
 
-from coalib.coala_main import run_coala
 from coalib.misc.Exceptions import get_exitcode
 from coalib.collecting.Collectors import filter_capabilities_by_languages
 from coalib.output.ConsoleInteraction import (
@@ -23,7 +22,6 @@ from coalib.output.ConsoleInteraction import (
     show_bears, show_language_bears_capabilities)
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.parsing.DefaultArgParser import default_arg_parser
-from coalib.settings.ConfigurationGathering import get_filtered_bears
 
 
 def main():
@@ -35,9 +33,12 @@ def main():
         args = default_arg_parser().parse_args()
 
         if args.show_bears:
+            from coalib.settings.ConfigurationGathering import (
+                get_filtered_bears)
             local_bears, global_bears = get_filtered_bears(
                 args.filter_by_language, log_printer)
 
+            from coalib.output.ConsoleInteraction import show_bears
             show_bears(local_bears,
                        global_bears,
                        args.show_description or args.show_details,
@@ -56,6 +57,11 @@ def main():
 
     except BaseException as exception:  # pylint: disable=broad-except
         return get_exitcode(exception, log_printer)
+
+    # Defer import so it's only done when needed
+    from coalib.coala_main import run_coala
+    from coalib.output.ConsoleInteraction import (
+        acquire_settings, nothing_done, print_results, print_section_beginning)
 
     partial_print_sec_beg = functools.partial(
         print_section_beginning,
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -2,6 +2,7 @@ import functools
 import os
 import pkg_resources
 import itertools
+from importlib import find_loader
 
 from pyprint.NullPrinter import NullPrinter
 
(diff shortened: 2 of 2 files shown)


coalib/collecting/Collectors.py
```diff
@@ -1,5 +1,7 @@
 import functools
 import os
+from importlib import find_loader
+
```
hypothesist: Why the blank line? The modules below are default modules as well

sils1297: this is a wip, don't judge codestyle please

coalib/coala.py
```diff
@@ -34,9 +30,11 @@ def main():
         args = default_arg_parser().parse_args()
 
         if args.show_bears:
+            from coalib.settings.ConfigurationGathering import get_filtered_bears
             local_bears, global_bears = get_filtered_bears(
                 args.filter_by_language, log_printer)
 
+            from coalib.output.ConsoleInteraction import show_bears
```
hypothesist: @sils1297 we can extend this to so many other things: for instance even `import functools` could be moved after the try except block since it is never used inside

hypothesist: But I guess that doesn't give a big performance gain (we should profile extensively)

sils1297: functools is probably not that relevent.

The coala imports are very evil because they again have lots and lots of imports and we're basically spending all the time on importing stuff

hypothesist:
@sils1297 so yeah, this looks pretty good. How much performance do we get by doing this?

sils1297:
the second commit doesn't work yet, the first one is a huge improve when using `coala -h` and `coala -v` from about .35 seconds to something like .19 seconds.

Makman2:
ping :)

coalib/coala.py
```diff
@@ -15,15 +15,13 @@
 
 from pyprint.ConsolePrinter import ConsolePrinter
 
-from coalib.coala_main import run_coala
 from coalib.misc.Exceptions import get_exitcode
 from coalib.collecting.Collectors import filter_capabilities_by_languages
 from coalib.output.ConsoleInteraction import (
     acquire_settings, nothing_done, print_results, print_section_beginning,
     show_bears, show_language_bears_capabilities)
```
sils1297: can't this import be deferred? At least acquire_settings isn't used and reimported later, can we import only what we need here?

coalib/coala.py
```diff
@@ -35,9 +33,12 @@ def main():
...
         if args.show_bears:
+            from coalib.settings.ConfigurationGathering import (
+                get_filtered_bears)
             local_bears, global_bears = get_filtered_bears(
                 args.filter_by_language, log_printer)
 
+            from coalib.output.ConsoleInteraction import show_bears
             show_bears(local_bears,
                        global_bears,
                        args.show_description or args.show_details,
```
AbdealiJK: A question about the code below, what is `show_capabilities` ? I can't seem to find it in DefaultArgParser

Don't you need to import `get_filtered_bears` there (In the elif) also ?

coalib/collecting/Collectors.py
```diff
@@ -284,13 +285,9 @@ def collect_registered_bears_dirs(entrypoint):
...
-    for ep in pkg_resources.iter_entry_points(entrypoint):
-        registered_package = None
-        try:
-            registered_package = ep.load()
-        except pkg_resources.DistributionNotFound:
-            continue
-        collected_dirs.append(os.path.abspath(
-            os.path.dirname(registered_package.__file__)))
-    return collected_dirs
+    eps = {find_loader("bears")
```
AbdealiJK: Are you assuming that the bear package will have a module named "bears" ?
That;s only valid in coala-bears right ?

AbdealiJK: Why loop over entrypoints and create `ep` is `ep` is not being used in the `find_loader("bears")` part ??
So, `eps` will just be a set of the same value repeated ?

coalib/coala.py
hypothesist: So you do `coala --show-capabilities Python` and it will list stuff like `Code Simplification, security, ...` and stuff. Basically a union of all `CAN_FIX`es of all python bears.

hypothesist: And welcome back to reviewing :)

coalib/collecting/Collectors.py
hypothesist: yeah, this can be improved more

marking as WIP for now

sils:
first commit was redone and put up for review on https://github.com/coala/coala/pull/2875

Mixih:
@sils didn't we implement proper caching since then? Is this pull still relevant?

sils:
@Mixih yeah it is relevant, mainly because we figured out that https://github.com/coala/coala/pull/2461/files#diff-e690c48863208d16808485e103164cd5L291 is a major performance killer. It's not necessarily a solution, I don't think it even solves it, but it shows the problem and a few attempts for someone who wants to pick this up.

PR #2463, opened 2016-07-20
Sils/usab

sils1297:


coalib/settings/SectionFilling.py
```diff
@@ -65,19 +65,13 @@ def fill_section(section, acquire_settings, log_printer, bears):
...
-                "get_non_optional_settings.".format(bear))
-        else:
-            needed = bear.get_non_optional_settings()
-            for key in needed:
-                if key in prel_needed_settings:
-                    prel_needed_settings[key].append(bear.name)
-                else:
-                    prel_needed_settings[key] = [needed[key][0],
-                                                 bear.name]
+        needed = bear.get_non_optional_settings()
```
sims1253: Even though coala "should" not fail here how a bout a try except to catch and communicate the fact that is did if it did?

sils1297: why? You don't do it if you expect a string right? You'd rather use enforce_signature. Which again is a peromance overhead and rather should use only or debugging.

sims1253:
ack 1fd3e1a
ack a5da95a

sims1253:
ack 	3afbd3d

sils1297:
reack
  1 pick fcc0c0e SectionFilling: Remove unneeded branch
  2 pick 2874988 require_setting: Remove unneeded check
  3 pick 83f94f5 require_setting: Print question in green

sils1297:
@rultor merge

PR #2498, opened 2016-07-21
fix codecov.yml

stevepeak:
The `threshold` should be read like `Allow the coverage to drop by ___%`.

I think you are looking to always hit 100%. So that is the target

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,5 +6,4 @@ coverage:
     project:
       default:
         enabled: true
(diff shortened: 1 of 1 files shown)


sils1297:
@Adrianzatreanu can you take this over?

Adrianzatreanu:
moved here https://github.com/coala-analyzer/coala/pull/2500

thanks a lot for the help :)

PR #2586, opened 2016-08-05
ConsoleInteraction: Sort questions by bear

hypothesist:
When multiple settings are missing for a bear, the questions
may be out of order, with other questions in between them. This
may be inconvinient. So sorting the questions by the bears
should group the settings.

Fixes https://github.com/coala-analyzer/coala/issues/2530

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -426,6 +426,29 @@ def print_affected_lines(console_printer, file_dict, section, sourcerange):
                         sourcerange)
 
 
+def join_names(values):
+    """
+    Produces a string by concatenating the items in ``values`` with
+    commas, except the last element, which is concatenated with an "and".
+
+    >>> join_names(["apples", "bananas", "oranges"])
+    'apples, bananas and oranges'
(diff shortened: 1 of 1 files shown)


coalib/output/ConsoleInteraction.py
```diff
@@ -426,6 +426,29 @@ def print_affected_lines(console_printer, file_dict, section, sourcerange):
                         sourcerange)
 
 
+def join_names(values):
+    """
+    Produces a string by concatenating the items in ``values`` with
+    commas and an "and" for the last element. Example:
```
Adrianzatreanu: this function looks good, but i dont really like the docstring.

maybe something like "Produces a string by concatenating the items in values with commas, besides the last element, which is concatenated with an ``and``".

hypothesist: `besides the last element` or `except the last element`?

Adrianzatreanu: except is better

Adrianzatreanu:
ack ef287a4

hypothesist:
@rultor merge

Issue #2075, opened 2016-04-26
Avoid multiple warnings for same malformed bear


AbdealiJK:
In coala -A if I have a malformed bear, it basically shows me the warning for every section. This is kinda irritating and confusing.

```
$ coala -A
[WARNING][09:04:21] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:21] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:21] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:21] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git/GitCommitBear.py. Probably the file is malformed or the module code raises an exception.
[WARNING][09:04:22] Unable to collect bears from /home/ajk/Documents/coala/coala-bears/bears/vcs/git
```

We should cache the list of malformed bears and not show the warning multiple times.

[...]

PR #2594, opened 2016-08-08
Fix repetition in bear collection

whonut:
Addresses #2075. 

I'm not sure if this is what @sils1297 has in-mind, but it does fix the issue of bears being collected multiple times.


whonut:
I forgot to run the tests...

Adrianzatreanu:
hey you have some partial coverage thingies (the ones with yellow) because the else branches are not being tested. to fix this, you need to test the cases in which the if condition is not met :)

whonut:
I've been trying to sort that since yesterday :) The one thing I can't do is find a `GlobalBear` for the `bear.kind()` conditional.

coalib/settings/SectionFilling.py
```diff
@@ -25,17 +26,38 @@ def fill_settings(sections, acquire_settings, log_printer):
...
-            [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
-            log_printer)
-        all_bears = copy.deepcopy(section_local_bears)
-        all_bears.extend(section_global_bears)
-        fill_section(section, acquire_settings, log_printer, all_bears)
+        bear_dirs_by_section[section_name] = section.bear_dirs()
+        bears_by_section[section_name] = list(section.get("bears", ""))
+
+    bears_to_collect = set(chain.from_iterable(bears_by_section.values()))
+    bear_dirs_to_collect = set(chain(*bear_dirs_by_section.values()))
```
sils1297: first you use chain.from_iterable and then chain with `*`, just use one of them consistenly shall we? ;)

whonut: I confess I did this for the sake of line length. Hobgoblin of Little Minds
and all that...

coalib/settings/SectionFilling.py
```diff
@@ -25,17 +26,38 @@ def fill_settings(sections, acquire_settings, log_printer):
...
+
+    all_local_bears, all_global_bears = collect_bears(
+        bear_dirs_to_collect,
```
sils1297: hm I'm not fully sure of this, we're joining the bear_dirs together here - this definitely changes the functionality because earlier I could have a coafile like:

```ini
[section1]
bear_dirs = ../something/
bear = SomeBear

[section2]
bear_dirs = ../else
bear = SomeBear
```

and it would be two different bears, one lying in something, the other in else. This is obviously not a very important corner case but we shouldn't break consistency. Maybe we should just get the bear dirs from the default section? I'm not so sure what to do here.

coalib/settings/SectionFilling.py
```diff
@@ -25,17 +26,38 @@ def fill_settings(sections, acquire_settings, log_printer):
...
+        section_local_bears = []
+        section_global_bears = []
+
+        for bear in all_bears:
+            if bear.name in bears_by_section[section_name]:
+                section_bears.append(bear)
+
+                if bear.kind() == BEAR_KIND.LOCAL:
+                    section_local_bears.append(bear)
+                elif bear.kind() == BEAR_KIND.GLOBAL:
```
sils1297: the coverage problem here is that the elif will always be true. Just make an `else` instead of `elif` as it can't be anything else, right?

coalib/settings/SectionFilling.py
whonut: Are you sure that's the case? I'm testing that exact setup and it seems to only run the bear that shows up first i.e. the one in `something`. If I switch the sections around it runs the one in `else`.

sils1297: Well it *should* run both bears for their sections respectively right?

whonut: It doesn't seem to be finding the bear in else. [Here](https://github.com/coala-analyzer/coala/files/411556/test-setup.zip) are the files/folders I'm using, if you want to give it a shot. Fully prepared to have done something stupid.

sils1297: no you havent, it's a conceptual problem, it cannot work if we don't do the
collection section wise :) so the question is do we really want this
anymore or do we want to change the functionality? (hint: I don't know :P
people always pretend I did know)

Makman2:
ping :)
@whonut do you need help? :D

whonut:
@Makman2 I'm not sure that I can proceed without some decisions being made about the corner case @sils1297 mentioned [above](https://github.com/coala-analyzer/coala/pull/2594#discussion_r74198516).

coalib/settings/SectionFilling.py
Makman2: if performance gets worse we should keep current behaviour^^
importing is one of our main performance impacts^^

AbdealiJK: I don't follow.
We should be only importing each bear once. A bear is not defined by only the class of the bear, it is defined by the whole path of the file too.
So, we should "cache" this somehow, so that the warning is not thrown and the import is not happening multiple times.

whonut: > A bear is not defined by only the class of the bear, it is defined by the whole path of the file too.
I hadn't thought of it this way. Good point. I'm a bit snowed under with other things at the moment, but I hope to get back to this.

AbdealiJK:
@whonut I'm not sure what discussions were made before. Could you please elaborate on what corner case @sils1297 mentioned (a link if possible, or a summary of discussions) ?

whonut:
@AbdealiJK Apologies for leaving this half-finished for so long. Life got in the way. I have a few days free at least now. [This](https://gitter.im/coala/coala?at=57ab6586f7f11a7936b318b5) is the only discussion I can find of it on Gitter with a quick search.

In summary (and as far as I can remember), my change doesn't allow bears with the same class name but defined in different files (e.g. `SomeBear` in `foo/SomeBear.py` and `SomeBear` in `bar/SomeBear.py`) to be treated separately. As far as I could tell at the time, the release version of coala didn't do this either. It makes some sense as a behaviour though, even if naming different bears identically is to be discouraged.

I hope that's helpful.

AbdealiJK:
I think you can ignore two bears with same name but different paths for now. The coala -b flag doesn't support that either to the best of my knowledge. It just chooses one of them (probably randomly)

So make an issue for bears with same name not working correctly in many places, as it would have to be fixed in many places not just this one. (show bears, mentioning bears in cli, mentioning bear as a dependency, etc)

whonut:
@AbdealiJK Will do. I just updated my branch by pulling from upstream, but I just realised that the commit title is overlong. D'oh, not sure what to do about that. Sorry.

I'll fix any CI issues now if I can.

AbdealiJK:
Hey, so we don't allow Merge commits like the one you pushed here into master normally. WHat you need to do is rebase on the current master - http://docs.coala.io/en/latest/Developers/Git_Basics.html?highlight=rebase#rebasing

whonut:
OK cool. I did that before, don't know why I forgot. Do I need to do anything before rebasing?

Also, I'm not entirely sure how to fix that failing test.

coalib/settings/SectionFilling.py
AbdealiJK: Please make a issue for this

coalib/settings/SectionFilling.py
```diff
@@ -25,17 +26,38 @@ def fill_settings(sections, acquire_settings, log_printer):
...
+
+    bears_to_collect = set(chain.from_iterable(bears_by_section.values()))
+    bear_dirs_to_collect = set(chain(*bear_dirs_by_section.values()))
+
+    all_local_bears, all_global_bears = collect_bears(
+        bear_dirs_to_collect,
+        bears_to_collect,
+        [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
+        log_printer)
+    all_bears = copy.deepcopy(all_local_bears)
```
AbdealiJK: Why a deepcopy ? Can't we just `chain` these ?

coalib/settings/SectionFilling.py
whonut: [#2783](https://github.com/coala/coala/issues/2783) 😄

coalib/settings/SectionFilling.py
whonut: Honestly I can't remember, `chain` should work.

whonut: It doesn't work properly with just `chain`, I have to call `list` on it which makes me think it's something to do with depleting the `chain` generator. That confuses me though, because I only refer to `all_bears` once.

whonut:
I'm not sure why CircleCI is failing, I can't make sense of it.

Adrianzatreanu:
when this kind of circleCI fails, it is because gitmate found problems. (see the 2 problems check here) where gitmate commented on your PR, please fix those and the tests will pass. a rebase is also needed :)

sils1297:
Uhm, usually you want the test change in the same commit as the functionality chagne so no commit breaks the tests.

sils1297:
Uhm @whonut a question about the change in general: if I have a coafile like this:

```ini
[python]
bears = KeywordBear

[C]
bears = KeywordBear
```

I assume after your change it'll ask me only once for the `keywords` setting for that bear? That would be wrong as I might want to have different keywords for the different sections.

So what we'd rather want to do is: first collect all bears - only once so if collection fails we don't have the warning multiple times - and then ask the user for every occurrence of every bear *but mentioning the section in the message*, like `Please enter a value for the setting "ci_keywords" (A list of keywords to search for (case insensitive).) needed by KeywordBear in the "python" section: `.

Makes sense?

tests/settings/SectionFillingTest.py
```diff
@@ -38,34 +38,37 @@ class SectionFillingTest(unittest.TestCase):
...
-        sections = {"test": self.section}
+        sections = {"test1": self.section1, "test2": self.section2}
         with simulate_console_inputs() as generator:
             fill_settings(sections,
                           acquire_settings,
                           self.log_printer)
             self.assertEqual(generator.last_input, -1)
 
-        self.section.append(Setting("bears", "SpaceConsistencyTestBear"))
+        self.section1.append(Setting("bears", "SpaceConsistencyTestBear"))
```
sils1297: can you put the same bear in section2 as well and simulate console inputs True, False?

sils1297:
Commit message: Instead of "Fix something" please describe the change, like "Collect bears only once" or so?

sils1297:
(also rebase needed while you're at it)

coalib/settings/SectionFilling.py
```diff
@@ -28,19 +28,39 @@ def fill_settings(sections, acquire_settings, log_printer):
...
-            bear_dirs,
-            bears,
-            [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
-            log_printer)
+        bear_dirs_by_section[section_name] = section.bear_dirs()
+        bears_by_section[section_name] = list(section.get("bears", ""))
+
+    bears_to_collect = set(chain.from_iterable(bears_by_section.values()))
+    bear_dirs_to_collect = set(
+        chain.from_iterable(bear_dirs_by_section.values()))
```
sils1297: if we go for this then I can't use different bear dirs for different sections, right? Then we should not check bear dirs from all sections but just from the default section.

coalib/settings/SectionFilling.py
```diff
@@ -28,19 +28,39 @@ def fill_settings(sections, acquire_settings, log_printer):
...
+        [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
+        log_printer)
+    all_bears = list(chain(all_local_bears, all_global_bears))
+
+    for section_name, section in sections.items():
+        section_local_bears = []
+        section_global_bears = []
+
+        for bear in all_bears:
+            if bear.name in bears_by_section[section_name]:
```
sils1297: hm this could fail if the bear.name is not the same as the filename! What do we do about this? Any alternative?

whonut: Sorry why is this? Does the `bears` setting use filenames, not class names?

sils1297: bear.name is the class name, bears are collected by filename! However the same file can hold multiple bears.

sils:
@whonut any news? Need help?

whonut:
Sorry, between uni and it being my birthday this weekend I haven’t had time
to sit down and sort this. Hopefully I can find time this week. Sorry to
drag this out.

sils:
Hey @Makman2 I think this PR was very promising and now is stale, might be something that might be covered by next gen bears? Or do it on top?

whonut:
Sorry all, I lack the time to get to grips with this now.

sils:
Maybe @whonut would be interested to pick this up again?

sils:
@whonut don't worry that's totally fine, just don't want your work wasted :P

Makman2:
@sils the core and bear collection are different things, also in the new core. But if you want I can pick this up as a next project^^ :)

whonut:
I would like to pick this up again if that's OK. If you'd rather I pick up a newcomer or difficulty low issue that's fine. I can only apologise again for inactivity, February was mad. If you're OK with me picking this up, I should be able to start tomorrow.

Makman2:
@whonut everything alright, no need for apologies :) And sure you can continue to work, "my next project" means I'm gonna start it... somewhen, so, in ~10 years :P

PR #2603, opened 2016-08-12
coalib/bearlib: Allow modifying deprecated_arg

Asnelchristian:


coalib/bearlib/__init__.py
```diff
@@ -16,6 +16,10 @@ def deprecate_settings(**depr_args):
      ... def run(new):
```
Adrianzatreanu: extra space after bracket in commit message.

also its For example: Having;
not For: Example having

coalib/bearlib/__init__.py
```diff
@@ -16,6 +16,10 @@ def deprecate_settings(**depr_args):
...
      ...     print(new)
```
Adrianzatreanu: also dont use first person in commit messages, use 3rd

coalib/bearlib/__init__.py
```diff
@@ -16,6 +16,10 @@ def deprecate_settings(**depr_args):
...
 
+     >>> @deprecate_settings(new=('old', lambda a: a + 'coala!'))
+     ... def func(new):
+     ...     print(new)
```
sils1297: please move this down to the example it belongs to

coalib/bearlib/__init__.py
```diff
@@ -41,16 +54,24 @@ def deprecate_settings(**depr_args):
     def _deprecate_decorator(func):
 
         def wrapping_function(*args, **kwargs):
-            for arg, deprecated_arg in depr_args.items():
+            for arg, _deprecated_arg in depr_args.items():
+                deprecated_arg, _func = (_deprecated_arg
+                                         if isinstance(_deprecated_arg, tuple)
+                                         else (_deprecated_arg, None))
```
sils1297: instead of `None` return `lambda x: x`, then you get rid of the if else below

coalib/bearlib/__init__.py
```diff
@@ -41,16 +54,24 @@ def deprecate_settings(**depr_args):
     def _deprecate_decorator(func):
 
         def wrapping_function(*args, **kwargs):
-            for arg, deprecated_arg in depr_args.items():
+            for arg, _deprecated_arg in depr_args.items():
```
sils1297: `_deprecated_arg` isn't actually the deprectated arg which is why your names are so confusing and clashing. It's rather a tuple or the arg right? Make a better name. Clashes usually suggest bad naming, underscores doesn't solve the real problem :)

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2603/commits/156adc5eb9cd6323bb58677584a9f6a22b2da466

sils1297:
@rultor merge

PR #2609, opened 2016-08-13
WIP: Udayan/newbears

Udayan12167:
And so it begins.


Adrianzatreanu:
thats how you mark it :D

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        :return: A list of lines of the file to iterate over.
+        """
+        return iter(self.lines)
+
+    def __hash__(self):
+        """
+        :return: Only hashes the filename.
+        """
+        return hash(self.filename)  # pragma: no cover
```
sils1297: you can test it by hashing two identical objects and asserting equality

coalib/processes/BearRunning.py
```diff
@@ -408,13 +397,27 @@ def run_local_bears(filename_queue,
     try:
         while True:
             filename = filename_queue.get(timeout=timeout)
+
+            if filename not in file_dict:
+                send_msg(message_queue,
+                         timeout,
+                         LOG_LEVEL.ERROR,
+                         "An internal error occurred.",
+                         Constants.THIS_IS_A_BUG)
```
sils1297: can you make a commit removing this whole if? This is an internal error and should never occurs we can just let it throw the exception :)

Udayan12167: But who will catch the exception?

sils1297: the main routine

meetmangukiya:
@udayan12167 I think @makman2 took this over so close?

Makman2:
Yeah think so. Let's keep the branch around for a while until the nextgen-core is completely implemented.

PR #2620, opened 2016-08-16
Implement function delete_lines

micrictor:
Implement a function delete_lines
https://github.com/coala-analyzer/coala/issues/2619

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -354,6 +354,13 @@ class Diff:
         linediff.delete = True
         self._changes[line_nr] = linediff
 
+    def delete_lines(self, line_nr_start, line_nr_end):
+        """
+        Delete lines in a specified range.
+        """
+        for line_nr in range(line_nr_start, line_nr_end):
+            self.delete_line(line_nr)
+
     def add_lines(self, line_nr_before, lines):
         """
         Adds lines after the given line number.
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -28,7 +28,16 @@ class DiffTest(unittest.TestCase):
     def test_delete_line(self):
         self.uut.delete_line(1)
         self.uut.delete_line(1)  # Double deletion possible without conflict
+        additions, deletions = self.uut.stats()
+        self.assertEqual(deletions, 1)
         self.assertRaises(ValueError, self.uut.delete_line, 0)
+
+    def test_delete_lines(self):
+        self.uut.delete_lines(1, 10)
+        self.uut.delete_lines(10, 20)
+        additions, deletions = self.uut.stats()
+        self.assertEqual(deletions, 19)
+        self.assertRaises(ValueError, self.uut.delete_lines, 0, 10)
 
     def test_change_line(self):
         self.assertEqual(len(self.uut), 0)


sils1297:
Hi,

thanks for your contribution!

I see you already made contact with GitMate. Essentially GitMate wants you to change the commits (which is why those comments are always reappearing, you're adding new commits and keeping the "wrong" ones).

So you can squash together your commits using `git rebase --interactive` and in the future you can simply fixup issues with the last commit using `git commit --amend` and then `git push --force` the new commit to replace the old one.

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

If you need further assistance, don't hesitate to ask, this is always a bit difficult :)

Cheers!

coalib/results/Diff.py
```diff
@@ -354,6 +354,16 @@ def delete_line(self, line_nr):
...
 
+    def delete_lines(self, line_nr_start, line_nr_end):
+        """
+        Delete lines in a range, exclusive.
+        """
+        for line_nr in range(line_nr_start, line_nr_end):
+            # Flatten delete_lines to reduce computations
+            linediff = self._get_change(line_nr)
+            linediff.delete = True
+            self._changes[line_nr] = linediff
```
sils1297: hey, one of the reasons why our checks on GitHub are failing is that this function is untested, I think a simple doctest (simlar to around line ~340) will be neat and simple!

coalib/results/Diff.py
```diff
@@ -354,6 +354,16 @@ def delete_line(self, line_nr):
         linediff.delete = True
         self._changes[line_nr] = linediff
 
+    def delete_lines(self, line_nr_start, line_nr_end):
+        """
+        Delete lines in a range, exclusive.
+        """
+        for line_nr in range(line_nr_start, line_nr_end):
+            # Flatten delete_lines to reduce computations
```
sils1297: so you're essentially trying to optimize away a function call at the expense of code duplication? This kind of thing usually only gives a very very limited performance improvement and increases the risk of having bugs/decreases maintainability.

sils1297:
Hey @micrictor I left you a quick review. General rule of thumb: if you disagree with the reviewer, start a discussion about it, that's fine. Some people seem to be scared to disagree - don't be. If it's about a general thing like codestyle/process, you'll rather want to discuss it on gitter or on an issue, we're open for improvement suggestions :)

micrictor:
@sils1297 Thanks for the support, this is my first time submitting code to a project I'm not in charge of, so there's a lot of little things I didn't do, either out of ignorance or absent-mindedness, and probably both.

One question, you mentioned doctests. How would I use them, given that delete_line, and therefore delete_lines, doesn't have a return value?

sils1297:
Hey,

great, so you already have some experience! We would be very interested in what you think we could do better.

The choice of doctests vs. unittests is yours, I don't have a strong opinion. Even in a doctest you could just get an instance, perform operations on it and invoke other functions that show what was changed.

tests/results/DiffTest.py
```diff
@@ -30,6 +30,11 @@ def test_delete_line(self):
         self.uut.delete_line(1)  # Double deletion possible without conflict
         self.assertRaises(ValueError, self.uut.delete_line, 0)
 
+    def test_delete_lines(self):
+        self.uut.delete_lines(1, 10)
+        self.uut.delete_lines(10, 20)
+        self.assertRaises(ValueError, self.uut.delete_lines, 0, 10)
```
sils1297: ahm so this will only raise a valueerror because you're trying to delete line 0, you're not really testing if th line deletion is actually happening - OTOH the other tests aren't really better. Maybe I wrote them... :P

Could you at least test the `len` of uut or something to assure that the deletions happen?

sils1297:
unack https://github.com/coala-analyzer/coala/pull/2620/commits/3aab8357d54b2bc7d2e34b9947ee9cb8a6696f60 , almost ready :)

sils1297:
ah for the commit message, I would add the `Diff:` tag to give some context in the first line/shortlog

micrictor:
I've added in better testing, both to my test for delete_lines and the existing test for delete_lines. The one downfall I can see is that if Diff.stats() is broken it would fail in these two tests first. Does that mean that I should just use len(self.uut) and assume that the lines are properly marked?

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2620/commits/ee79ec6da70eb88d8373d6f902d68e8fd15ad6f5 your solution is way better than what I had in mind!

sils1297:
So, your PR needs one last rebase it seems. [This page](http://coala.readthedocs.org/en/latest/Users/Tutorials/Git_Help.html) may help you to get started on this. (*Do not press the update branch button. It's evil.*)

Add the original coala repository as a remote:

```
git remote add upstream https://github.com/coala-analyzer/coala.git
```

Then go to your branch and rebase it onto our master:

```
git checkout <yourbranch>
git fetch upstream
git rebase --interactive upstream/master
```

Now push it up. You will need to do a force push because you change history here (do not do this on master branches :)):

```
git push --force
```

Ping us then so we can merge it ASAP.

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

Ping us on [our gitter channel](https://gitter.im/coala-analyzer/coala) if you need help!

sils1297:
Hey, for simplicify I'll cherry pick your commit into another PR and merge a few at once because there's a bit of stuff going on right now - so you don't need to do anything!

sils1297:
This was cherrypicked into https://github.com/coala-analyzer/coala/pull/2637 and will be merged there.

Thanks for your contribution! :)

PR #2630, opened 2016-08-16
Parser: Fix instances of "touples"

Daeke:
Replace every mention of "touples" with "tuple." This affects both
comments and named variables, but doesn't affect functionality.

Closes https://github.com/coala-analyzer/coala/issues/2412

### DIFF
--- a/coalib/parsing/CliParsing.py
+++ b/coalib/parsing/CliParsing.py
@@ -80,13 +80,13 @@ def parse_custom_settings(sections,
     :param line_parser:          The LineParser to use.
     """
     for setting_definition in custom_settings_list:
-        (_, key_touples, value, _) = line_parser.parse(setting_definition)
-        for key_touple in key_touples:
+        (_, key_tuples, value, _) = line_parser.parse(setting_definition)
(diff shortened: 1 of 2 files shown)


coalib/parsing/CliParsing.py
```diff
@@ -80,13 +80,13 @@ def parse_custom_settings(sections,
     :param line_parser:          The LineParser to use.
```
Adrianzatreanu: for the commit message: instead of ``shouldn't``, i think you should use ``doesn't``

sils1297:
Hey,

we use some bug prediction in coala which relies on the `Fixes` keyword in commit messages. To get good results from that we need to use `Closes` for normal issues instead of `Fixes` which should only be used for real bugs. (See also [the commit message docs](http://coala.readthedocs.io/en/latest/Getting_Involved/Writing_Good_Commits.html).)

Cheers!

sils1297:
ack https://github.com/coala-analyzer/coala/pull/2630/commits/9feb493d2e5dc27dc768dcbe027541685009903f

sils1297:
@rultor merge

PR #2642, opened 2016-08-19
Diff: Increment 2nd paramter of range in Diff.delete_lines

kmo1:
Closes https://github.com/coala-analyzer/coala/issues/2641

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -358,7 +358,7 @@ class Diff:
         """
         Delete lines in a specified range.
         """
-        for line_nr in range(line_nr_start, line_nr_end):
+        for line_nr in range(line_nr_start, line_nr_end + 1):
             self.delete_line(line_nr)
 
     def add_lines(self, line_nr_before, lines):
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -356,7 +356,7 @@ class Diff:
 
     def delete_lines(self, line_nr_start, line_nr_end):
         """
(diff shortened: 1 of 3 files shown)


Asnelchristian:
what about ``Diff: Increase `line_nr_end` in `delete_lines` `` as your short log

sils1297:
I think it's good after the commit message was changed :)

coalib/results/Diff.py
```diff
@@ -358,7 +358,7 @@ def delete_lines(self, line_nr_start, line_nr_end):
         """
         Delete lines in a specified range.
         """
-        for line_nr in range(line_nr_start, line_nr_end):
+        for line_nr in range(line_nr_start, line_nr_end + 1):
```
Makman2: please add some documentation that the given end-line is deleted inclusively :+1:

Makman2:
unack 28a4631

coalib/results/Diff.py
kmo1: @Makman2 - Should I just be updating the comments or is there documentation somewhere specific that should be updated?

Makman2: right above is the docstring, a little mention there would be good :)

Makman2:
ack 28a4631 e2ece4e

Makman2:
I thought of fixing the docstring in the previous commit, but in fact it's alright, as before it wasn't really well documented either :+1: :)

Makman2:
all that's missing now is a rebase, then we can merge

PR #2655, opened 2016-08-21
Diff/Result: Add filename property to Diff

micrictor:
Add a `filename` property to Diff, allowing a Results object to
be initialized with a collection of Diffs. Backwards compatable
with the previous method of initializing Diff and Result objects.

Addresses https://github.com/coala-analyzer/coala/issues/2194

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -12,24 +12,34 @@ class Diff:
     A Diff result represents a difference for one file.
     """
 
-    def __init__(self, file_list, filename, rename=False, delete=False):
+    def __init__(self, file_list, filename=None, rename=False, delete=False):
         """
         Creates an empty diff for the given file.
 
+        Test to see if the deprecation warning prints properly.
+
+        >>> diff = Diff([])
+        WARNING! Initializing Diff without filename deprecated
+
         :param file_list: The original (unmodified) file as a list of its
                           lines.
-        :param filename:  A string containing the name of the file
+        :param filename:  None or a string containing the name of the file
         :param rename:    False or str containing new name of file.
         :param delete:    True if file is set to be deleted.
         """
         self._changes = {}
         self._file = file_list
+
+        # FIXME: Use builtin logging when it becomes available
+        if(not filename):
+            print("WARNING! Initializing Diff without filename deprecated")
+
         self.filename = filename
         self.rename = rename
         self.delete = delete
 
     @classmethod
-    def from_string_arrays(cls, file_array_1, file_array_2, filename,
+    def from_string_arrays(cls, file_array_1, file_array_2, filename=None,
                            rename=False):
         """
         Creates a Diff object from two arrays containing strings.
@@ -39,7 +49,7 @@ class Diff:
 
         :param file_array_1: Original array
         :param file_array_2: Array to compare
-        :param filename:     Name of file
+        :param filename:     None or a string containing the name of the file
         :param rename:       False or str containing new name of file.
         """
         result = cls(file_array_1, filename, rename=rename)
@@ -94,7 +104,7 @@ class Diff:
                     type(file)(newvalue.splitlines(True)) +
                     file[fixit.range.end.line:])
 
-        return cls.from_string_arrays(file, new_file, filename)
+        return cls.from_string_arrays(file, new_file, filename or self.filename)
 
     def _get_change(self, line_nr, min_line=1):
         if not isinstance(line_nr, int):
@@ -224,7 +234,7 @@ class Diff:
         return list(diff.range()
                     for diff in self.split_diff(distance=0))
 
-    def split_diff(self, distance=1):
+    def split_diff(self, distance=1, filename=None):
         """
         Splits this diff into small pieces, such that several continuously
         altered lines are still together in one diff. All subdiffs will be
@@ -261,12 +271,13 @@ class Diff:
 
         :param distance: Number of unchanged lines that are allowed in between
                          two changed lines so they get yielded as one diff.
+        :param filename: None or a string containing the name of the file
         """
         if not self:
             return
 
         last_line = -1
-        this_diff = Diff(self._file, self.filename,
+        this_diff = Diff(self._file, filename or self.filename,
                          rename=self.rename, delete=self.delete)
         for line in sorted(self._changes.keys()):
             if line > last_line + distance + 1 and len(this_diff._changes) > 0:
@@ -282,7 +293,7 @@ class Diff:
         # always.
         yield this_diff
 
-    def range(self):
+    def range(self, filename=None):
         """
         Calculates a SourceRange spanning over the whole Diff. If something is
         added after the 0th line (i.e. before the first line) the first line
@@ -296,14 +307,15 @@ class Diff:
         >>> print(range.start.line)
         None
 
-        :return:                A SourceRange object.
+        :param filename: None or a string containing the name of the file
+        :return:         A SourceRange object.
         """
         if len(self._changes) == 0:
-            return SourceRange.from_values(self.filename)
+            return SourceRange.from_values(filename or self.filename)
 
         start = min(self._changes.keys())
         end = max(self._changes.keys())
-        return SourceRange.from_values(self.filename,
+        return SourceRange.from_values(filename or self.filename,
                                        start_line=max(1, start),
                                        end_line=max(1, end))
 
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -12,22 +12,25 @@ class Diff:
     A Diff result represents a difference for one file.
     """
 
-    def __init__(self, file_list, rename=False, delete=False):
+    def __init__(self, file_list, filename, rename=False, delete=False):
         """
         Creates an empty diff for the given file.
 
         :param file_list: The original (unmodified) file as a list of its
                           lines.
+        :param filename:  A string containing the name of the file
         :param rename:    False or str containing new name of file.
         :param delete:    True if file is set to be deleted.
         """
         self._changes = {}
         self._file = file_list
+        self.filename = filename
         self.rename = rename
         self.delete = delete
 
     @classmethod
-    def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
+    def from_string_arrays(cls, file_array_1, file_array_2, filename,
+                           rename=False):
         """
         Creates a Diff object from two arrays containing strings.
 
@@ -36,9 +39,10 @@ class Diff:
 
         :param file_array_1: Original array
         :param file_array_2: Array to compare
+        :param filename:     Name of file
         :param rename:       False or str containing new name of file.
         """
-        result = cls(file_array_1, rename=rename)
+        result = cls(file_array_1, filename, rename=rename)
 
         matcher = difflib.SequenceMatcher(None, file_array_1, file_array_2)
         # We use this because its faster (generator) and doesn't yield as much
@@ -68,12 +72,13 @@ class Diff:
         return result
 
     @classmethod
-    def from_clang_fixit(cls, fixit, file):
+    def from_clang_fixit(cls, fixit, file, filename=None):
         """
         Creates a Diff object from a given clang fixit and the file contents.
 
-        :param fixit: A cindex.Fixit object.
-        :param file:  A list of lines in the file to apply the fixit to.
+        :param fixit:    A cindex.Fixit object.
+        :param file:     A list of lines in the file to apply the fixit to.
+        :param filename: None or a string containing the file name
         :return:      The corresponding Diff object.
         """
         assert isinstance(file, (list, tuple))
@@ -89,7 +94,7 @@ class Diff:
                     type(file)(newvalue.splitlines(True)) +
                     file[fixit.range.end.line:])
 
-        return cls.from_string_arrays(file, new_file)
+        return cls.from_string_arrays(file, new_file, filename)
 
     def _get_change(self, line_nr, min_line=1):
         if not isinstance(line_nr, int):
@@ -209,15 +214,14 @@ class Diff:
         """
         return self.unified_diff
 
-    def affected_code(self, filename):
+    def affected_code(self):
         """
         Creates a list of SourceRange objects which point to the related code.
         Changes on continuous lines will be put into one SourceRange.
 
-        :param filename: The filename to associate the SourceRange's to.
-        :return:         A list of all related SourceRange objects.
+        :return:                A list of all related SourceRange objects.
         """
-        return list(diff.range(filename)
+        return list(diff.range()
                     for diff in self.split_diff(distance=0))
 
     def split_diff(self, distance=1):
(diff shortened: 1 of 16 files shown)


coalib/results/Diff.py
```diff
@@ -12,17 +12,20 @@ class Diff:
...
     """
 
-    def __init__(self, file_list, rename=False, delete=False):
+    def __init__(self, file_list, filename=None, rename=False, delete=False):
         """
         Creates an empty diff for the given file.
 
         :param file_list: The original (unmodified) file as a list of its
                           lines.
+
```
Makman2: redundant newline :)

coalib/results/Result.py
```diff
@@ -36,7 +36,7 @@ def __init__(self,
                  severity: int=RESULT_SEVERITY.NORMAL,
                  additional_info: str="",
                  debug_msg="",
-                 diffs: (dict, None)=None,
+                 diffs: (dict, set, list, None)=None,
```
Makman2: use instead `(dict, set, list, None)`--> `(Iterable, None)`
(`Iterable` lies inside `collections` :D)

tests/results/DiffTest.py
```diff
@@ -161,6 +166,14 @@ def test_addition_rename(self):
...
+        uut = Diff(self.file, filename="file")
+        other = Diff(self.file, filename="file")
+        self.assertEqual((other + uut).filename, "file")
+
+        other.filename = "elif"
+        self.assertRaises(ConflictError, other.__add__, uut)
+
     def test_from_string_arrays(self):
         a = ["q", "a", "b", "x", "c", "d"]
         b = ["a", "b", "y", "c", "d", "f"]
```
Makman2: As we maintain compatibility, could you move the modifications inside `Result` into another commit? :+1:

Makman2: and everything related :)

coalib/results/Result.py
```diff
@@ -63,6 +65,13 @@ def __init__(self,
         if severity not in RESULT_SEVERITY.reverse:
             raise ValueError("severity is not a valid RESULT_SEVERITY")
 
+        self.diffs = {}
+        if(diffs and type(diffs) != dict):
```
Makman2: use `isinstance` instead of `type(x) == sometype` :)

coalib/results/Diff.py
```diff
@@ -290,9 +297,19 @@ def range(self, filename):
         >>> print(range.start.line)
         None
 
-        :param filename: The filename to associate the SourceRange with.
+        >>> range = Diff([], filename="file").range()
```
sils1297: can we put this example first and after it a note that for backwards compatability you can still call it with a filename?

coalib/results/Diff.py
```diff
@@ -12,17 +12,20 @@ class Diff:
...
         :param file_list: The original (unmodified) file as a list of its
                           lines.
+
+        :param filename:  None or a string containing the name of the file
         :param rename:    False or str containing new name of file.
         :param delete:    True if file is set to be deleted.
         """
         self._changes = {}
         self._file = file_list
+        self.filename = filename
```
sils1297: if filename isn't given, can we print a warning? (Note that before @Uran198 s PR about logging isn't merged we don't use builtin logging yet, add a todo to transform it to logging later?)

sils1297: i.e. a deprecation warning

coalib/results/Result.py
```diff
@@ -63,6 +65,13 @@ def __init__(self,
...
+            for diff in diffs:
+                self.diffs[diff.filename] = diff
```
sils1297: use a generator expression here, much faster in terms of performance and usually simpler to understand

coalib/results/Result.py
```diff
@@ -63,6 +65,13 @@ def __init__(self,
...
+        else:
+            self.diffs = diffs
```
sils1297: actually I think we should set this to an empty dict instead of None if None is given since using Nones often involves special handling whereas the empty dict really is sane.

sils1297: but that's another commit

coalib/results/Diff.py
micrictor: How would I do this without breaking the unit testing? It seems most logical to output it at the warning debuglevel, but that makes all tests that don't use the new initializer toss an exception

tests/results/DiffTest.py
micrictor: By everything related, do you mean split this into two commits, one for Result and ResultTest, and the other for Diff and DiffTest?

coalib/results/Diff.py
```diff
@@ -276,23 +282,36 @@ def split_diff(self, distance=1):
...
         The range of an empty diff will only affect the filename:
 
+        >>> range = Diff([], filename="file").range()
+        >>> range.file is None
+        False
+        >>> print(range.start.line)
+        None
+
+        In the interests of compatibility, you can still call this with a
+        manually supplied filename as a parameter
```
Adrianzatreanu: end with period

coalib/results/Result.py
```diff
@@ -53,6 +53,8 @@ def __init__(self,
         :param diffs:           A dictionary with filenames as key and a
                                 sequence of ``Diff`` objects associated with
                                 them as values.
+                                May also be a collection of Diffs initialized
+                                with the ``filename`` property
```
Adrianzatreanu: end with period here too

tests/results/DiffTest.py
Makman2: right :+1:

coalib/results/Diff.py
Makman2: does logging a warning message break the tests?

micrictor: Using the Log Printer in conjunction with ConsolePrinter, yes

coalib/results/Diff.py
```diff
@@ -209,14 +211,18 @@ def __json__(self):
...
-    def affected_code(self, filename):
+    def affected_code(self, filename=None):
         """
         Creates a list of SourceRange objects which point to the related code.
         Changes on continuous lines will be put into one SourceRange.
 
         :param filename: The filename to associate the SourceRange's to.
         :return:         A list of all related SourceRange objects.
         """
+        assert filename or self.filename
```
AbdealiJK: A message like 
`assert filename or self.filename, "A filename is required create SourceRanges for"`
would go a long way to not confuse people later

coalib/results/Diff.py
```diff
@@ -276,23 +282,36 @@ def split_diff(self, distance=1):
...
+    def range(self, filename=None):
         """
         Calculates a SourceRange spanning over the whole Diff. If something is
         added after the 0th line (i.e. before the first line) the first line
         will be included in the SourceRange.
 
         The range of an empty diff will only affect the filename:
 
+        >>> range = Diff([], filename="file").range()
+        >>> range.file is None
```
AbdealiJK: Why not just do 
```
>>> print(range.file)
"file"
```
or
"""
>>> assert range.file == "file"
"""
Showing it is not None is awkward IMO

coalib/results/Result.py
```diff
@@ -36,7 +36,7 @@ def __init__(self,
                  severity: int=RESULT_SEVERITY.NORMAL,
                  additional_info: str="",
                  debug_msg="",
-                 diffs: (dict, None)=None,
+                 diffs: (Iterable, None)=None,
```
AbdealiJK: Don't you need to import Iterable anywhere ? (Same for the below use of Iterable)

coalib/results/Result.py
```diff
@@ -63,6 +65,13 @@ def __init__(self,
         if severity not in RESULT_SEVERITY.reverse:
             raise ValueError("severity is not a valid RESULT_SEVERITY")
 
+        self.diffs = {}
+        if(diffs and isinstance(diffs, dict)):
```
AbdealiJK: Redundant brackets:
`if diffs and isinstance(diffs, dict):`

Also, wouldnt `if isinstance(diffs, dict):` be sufficient ?

coalib/results/Result.py
```diff
@@ -63,6 +65,13 @@ def __init__(self,
...
+            for diff in diffs:
+                self.diffs[diff.filename] = diff
```
AbdealiJK: `self.diffs = {d.filename: d for d in diffs}` Simpler to read.
Also, this is if diffs is a dictionary ? Shouldnt it be when diffs is not a dictionary ?

coalib/results/Diff.py
micrictor: I was merely following the existing  doctest without really thinking about it. How I understand the comment on line 291 is that the filename will change, so the simplest way to test it is to check if there is, in fact, a file connected to the range.

coalib/results/Result.py
micrictor: Done, thanks

tests/results/DiffTest.py
micrictor: Done, thanks

coalib/results/Result.py
micrictor: Yes, thank you

coalib/results/Diff.py
sils1297: I think you don't even need the print

coalib/results/Diff.py
sils1297: @Makman2 @micrictor we have a PR pending switching to python builtin logging, CC @Uran198 , we could just do a print for now and give it a FIXME to change it to builtin logging later. That's what we're doing for other deprecation warnings as well.

sils1297:
Can you use Closes instead of Addresses in the second commit so github auto closes the issue?

sils1297:
see also https://github.com/coala-analyzer/coala/pull/2655/commits/3c5e4cad7f88b74f67e1e76930d305301d9c7e87#r77959590 and https://github.com/coala-analyzer/coala/pull/2655/commits/3c5e4cad7f88b74f67e1e76930d305301d9c7e87#r77960360

I think otherwise this is good

coalib/results/Diff.py
micrictor: So, just do a print and modify all the doctests that the print would make fail?

coalib/results/Diff.py
micrictor: Re-worked without the print, I had to give it an explicit filename, `/file`, to avoid it automatically prepending the working directory

sils1297:
Any idea what's wrong with the windows build? CC @Makman2

sils1297:
ah @micrictor maybe you want to go back to the previous test of the file thing because of windows :/

coalib/results/Diff.py
```diff
@@ -276,23 +284,39 @@ def split_diff(self, distance=1):
...
         will be included in the SourceRange.
 
         The range of an empty diff will only affect the filename:
 
-        >>> range = Diff([]).range("file")
-        >>> range.file is None
-        False
+        >>> range = Diff([], filename="/file").range()
+        >>> range.file
+        '/file'
```
sils1297: ok my fault, this isn't possible because it's a path and this test fails on windows :/

tests/results/DiffTest.py
```diff
@@ -60,31 +60,36 @@ def test_double_changes_with_same_diff(self):
             self.fail('We should not have a conflict on same diff!')
 
     def test_affected_code(self):
```
sils1297: I think this test should be split into a deprecated part and the rest so we can deprecate the old usage and then remove it?

coalib/results/Diff.py
sils1297: yes

coalib/results/Result.py
```diff
@@ -63,6 +66,12 @@ def __init__(self,
         if severity not in RESULT_SEVERITY.reverse:
             raise ValueError("severity is not a valid RESULT_SEVERITY")
 
+        self.diffs = {}
```
sils1297: this statement is unneeded I think

tests/results/DiffTest.py
micrictor: To the same end, I'm now restructuring it so that one commit has all of the changes needed to make deprecated code still work, such that when the time comes a simple git revert will get rid of it

coalib/results/Diff.py
```diff
@@ -12,24 +12,34 @@ class Diff:
...
     """
 
-    def __init__(self, file_list, filename, rename=False, delete=False):
+    def __init__(self, file_list, filename=None, rename=False, delete=False):
         """
         Creates an empty diff for the given file.
 
+        Test to see if the deprecation warning prints properly.
+
+        >>> diff = Diff([])
```
micrictor: I used a doctest here, as I think it's the cleanest way to catch and test the print.

sils1297:
shouldn't https://github.com/coala/coala/pull/2655/commits/c028a996ad1f9a4e8d8d4213f6935b46c29664ef be squashed with https://github.com/coala/coala/pull/2655/commits/9c7093fa968d840a2ade5ac61757b13248cf3e63 ? Otherwise the latter would break our codebase and that could e.g. interrupt a `git bisect run`

micrictor:
My idea was that having them split will allow for a simple git revert to be ran when we no longer want to support initializing Diff without the filename property. If using `git bisect run` is a needed functionality, I'll be happy to squash them together

sils1297:
Good idea actually, usually we don't want build breaking changes in any commits though.

coalib/results/Result.py
```diff
@@ -63,6 +64,11 @@ def __init__(self,
         if severity not in RESULT_SEVERITY.reverse:
             raise ValueError("severity is not a valid RESULT_SEVERITY")
 
+        if(diffs and not isinstance(diffs, dict)):
+            self.diffs = {d.filename: d for d in diffs}
+        else:
+            self.diffs = diffs
```
sils1297: the changes to this file actually belong to an own commit IMO whereas you would have the test changes (https://github.com/coala/coala/pull/2655/commits/8d1e072a5c6aa4cbf7ac21ecd2c77d9eed5c5c13) squashed into this (https://github.com/coala/coala/pull/2655/commits/7797f8816a1a72fd97ea1575eb6cc7a3048b37a5) again. The hunks in this file add new functionality, the others are just changing the use of the Diff.

sils1297:
Ok I know that squashing and splitting can be some work. I have three commits in mind:

- Add the filename parameter including backwards compatibility (just squash those commits into one)
- Pass the parameter all across the place
- Add the new functionality to the Result object that bears can just simply pass lists of diffs.

That would be cleaner and most importantly ease review I think so we find more potential bugs.

Makes sense? For the last two commits I'd put c028a99 and 9c7093f after the other two are squashed, then do a `git reset HEAD~2` and making the commits with `git add -p`.

sils1297:
the changes seem good on first sight though.

Btw.: If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

sils:
@micrictor should we fix it up for you?

micrictor:
Yes, sorry. It seems like you have a clearer vision for how the commits should be structured than I can wrap my head around at the moment. Maybe in a few months, when I have more free time, I'll come back to put a more complete effort into learning good commiting/python coding practices.

sils:
Closing this because of lots of conflicts, I'll make a note on the issue so this work doesn't get lost.

PR #2679, opened 2016-08-28
DistributionRequirement: Use distro module

SanketDG:
Some platform module functions are deprecated as of python version
3.5.
Source: https://bugs.python.org/issue1322

distro module is recommended as the alternative

Fixes https://github.com/coala-analyzer/coala/issues/2675

CC @AbdealiJK @Adrianzatreanu

### DIFF
--- a/coalib/bears/requirements/DistributionRequirement.py
+++ b/coalib/bears/requirements/DistributionRequirement.py
@@ -38,16 +38,18 @@ class DistributionRequirement(PackageRequirement):
                         'redhat': 'yum',
                         'arch': 'pacman'}
 
-        if (platform.linux_distribution()[0] in manager_dict.keys() and
-                manager_dict[platform.linux_distribution()[0]]
+        if platform.system() == 'Linux':  # pragma: no cover
+            import distro
+            distro_name = distro.linux_distribution()[0]
+        if (distro_name in manager_dict.keys() and manager_dict[distro_name]
                 in self.package.keys()):
-            manager = manager_dict[platform.linux_distribution()[0]]
+            manager = manager_dict[distro_name]
             return [manager.replace("_", "-"),
                     'install', self.package[manager]]
         else:
             package_possibilites = (
                 {package for package in self.package.values()})
-            print('The package could not be automatically installed on your '
-                  'operating system. Please try installing it manually. It'
+            print('The package could not be automatically installed on your'
+                  ' operating system. Please try installing it manually. It'
                   ' should look like this: ' + repr(package_possibilites))
             raise OSError
--- a/tests/bears/requirements/DistributionRequirementTest.py
+++ b/tests/bears/requirements/DistributionRequirementTest.py
@@ -1,23 +1,31 @@
 import platform
+import subprocess
 import unittest
 from unittest.mock import patch
+
+try:
+    import distro
+except subprocess.CalledProcessError:
+    pass
 
 from coalib.bears.requirements.DistributionRequirement import (
     DistributionRequirement)
 
 
+@unittest.skipIf(platform.system() == 'Windows',
+                 "This test cannot be run on Windows.")
 class DistributionRequirementTestCase(unittest.TestCase):
 
-    @patch('platform.linux_distribution', return_value=('Fedora',))
+    @patch('distro.linux_distribution', return_value=('Fedora',))
     def test_install_command_mock_fedora(self, call_mock):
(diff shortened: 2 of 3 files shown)


Adrianzatreanu:
apparently no module named distro. maybe you need to add it to requirements.txt or so, if its not default? or its from x import distro?

sils1297:
yeah you gotta add https://pypi.python.org/pypi/distro

SanketDG:
The appveyor tests are to fail, how would we get around that?

AbdealiJK:
Skip test on Windows?

SanketDG:
@AbdealiJK How would I go about doing that? Just add `not DistributionRequirement` on the execution command?

AbdealiJK:
Can't you simply do skipIf from unit test or pytest?

tests/bears/requirements/DistributionRequirementTest.py
```diff
@@ -8,16 +8,16 @@
```
Adrianzatreanu: do a skipif platform == windows :D?

Adrianzatreanu: helping you out: add this decorator to the class:

``@unittest.skipIf(platform.system() == 'Windows', "This test cannot be run on Windows.")``

and import platform anyway i guess ;d

Adrianzatreanu:
needs rebase too

tests/bears/requirements/DistributionRequirementTest.py
SanketDG: thanks :smile:

coalib/bears/requirements/DistributionRequirement.py
```diff
@@ -1,4 +1,9 @@
-import platform
+import subprocess
+
+try:
+    import distro
+except subprocess.CalledProcessError:  # pragma: no cover
+    pass
```
SanketDG: if I do this, the rest of the functionality would break with windows, need suggestions.

AbdealiJK: There are many ways:
 - Check if distro is in locals() or globals()
 - Use sys to check if a module named distro has been loaded
 - Create a flag which is a different value if this is loaded, and another value if it is not loaded
 - Use if conditions with platform.system()
 - Use a try catch with the rest of the distro code and suppress NameErrors because distro is not found

AbdealiJK: - Don't import it here, and only import it on demand when it's needed

coalib/bears/requirements/DistributionRequirement.py
```diff
@@ -38,16 +38,18 @@ def install_command(self):
                         'redhat': 'yum',
                         'arch': 'pacman'}
 
-        if (platform.linux_distribution()[0] in manager_dict.keys() and
-                manager_dict[platform.linux_distribution()[0]]
+        if platform.system() == 'Linux':  # pragma: no cover
+            import distro
+            distro_name = distro.linux_distribution()[0]
```
sils1297: so distro_name is undefined if this branch isn't taken?

sils1297: I think there should be an elif with a branch for mac and "anything unknown/windows", can you add brew support for mac while we're at it? (own commit of course :))

sils1297: see https://github.com/coala-analyzer/coala/issues/2678

tests/bears/requirements/DistributionRequirementTest.py
```diff
@@ -1,23 +1,31 @@
...
+    import distro
+except subprocess.CalledProcessError:
+    pass
+
 from coalib.bears.requirements.DistributionRequirement import (
     DistributionRequirement)
 
 
+@unittest.skipIf(platform.system() == 'Windows',
+                 "This test cannot be run on Windows.")
```
sils1297: it's not like it cannot be run but you mock linux specific things

aptrishu:
still on this @SanketDG  ?

AbdealiJK:
Moved to https://github.com/coala/coala/pull/2933

SanketDG:
Hey, sorry I have been insanely busy with other stuff.



> Moved to #2933

@AbdealiJK I dont see it there :O

AbdealiJK:
Was temporarily removed on @sils request. Will create a new PR tomorrow
with it.

PR #2706, opened 2016-09-01
Collectors: Improve dependency conflict message

sils1297:
Fixes https://github.com/coala-analyzer/coala/issues/2705

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-PyPrint>=0.2.4,==0.2.*
+PyPrint>=0.2.5,==0.2.*
 Pygments>=2.1.3
 setuptools>=19.2
 # Do *not* use 0.3 which is not backwards compatible to common clang versions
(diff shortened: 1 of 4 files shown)


sils1297:
@underyx review needed :)

requirements.txt
```diff
@@ -1,7 +1,7 @@
-PyPrint>=0.2.5,==0.2.*
-Pygments>=2.1.3
+PyPrint~=0.2.5
```
AbdealiJK: Why do I get the feeling that we're jumping from syntax to syntax in a loop every 2-3 weeks ...

AbdealiJK:
ack 4f1f2b5  43c5926 42cdfe4

requirements.txt
sils1297: because we're doing it for two projects and we switched to the complicated one because we thought it might help just to find out that it doesn't

sils1297:
reack 42cdfe4

sils1297:
@rultor merge

PR #2754, opened 2016-09-13
Fix ALL the things

fneu:

### DIFF
--- a/coalib/bears/requirements/NpmRequirement.py
+++ b/coalib/bears/requirements/NpmRequirement.py
@@ -49,7 +49,13 @@ class NpmRequirement(PackageRequirement):
 
         :param return: True if dependency is installed, false otherwise.
         """
-        cmd = ['npm', 'show', self.package]
-        if platform.system() == 'Windows':  # pragma: no cover
-            cmd = ['cmd', '/c'] + cmd
-        return not call_without_output(cmd)
+        for cmd in (['npm', 'list', self.package],
+                    ['npm', 'list', '-g', self.package]):
+
+            if platform.system() == 'Windows':  # pragma: no cover
+                cmd = ['cmd', '/c'] + cmd
+
+            if not call_without_output(cmd):
+                return True
+
+        return False
--- a/coalib/bears/requirements/GoRequirement.py
+++ b/coalib/bears/requirements/GoRequirement.py
@@ -49,4 +49,4 @@ class GoRequirement(PackageRequirement):
 
         :param return: True if dependency is installed, false otherwise.
         """
-        return not call_without_output(('go', 'doc', self.package))
+        return not call_without_output(('go', 'list', self.package))
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -343,6 +343,38 @@ def has_wildcard(pattern):
     return match is not None
 
 
+def _iglob(pattern):
+    dirname, basename = os.path.split(pattern)
+    if not has_wildcard(pattern):
+        for file in _absolute_flat_glob(pattern):
+            yield file
+        return
+
+    if basename == '**':
+        relative_glob_function = relative_recursive_glob
+    elif has_wildcard(basename):
+        relative_glob_function = relative_wildcard_glob
+    else:
+        relative_glob_function = relative_flat_glob
+
+    if not dirname:
+        for file in relative_glob_function(dirname, basename):
+            yield file
+        return
+
+    # Prevent an infinite recursion if a drive or UNC path contains
+    # wildcard characters (i.e. r'\\?\C:').
+    if dirname != pattern and has_wildcard(dirname):
+        dirs = iglob(dirname)
+    else:
+        dirs = [dirname]
+
+    for dirname in dirs:
+        for name in relative_glob_function(dirname, basename):
+            yield os.path.join(dirname, name)
+
+
+@yield_once
 def iglob(pattern):
     """
     Iterates all filesystem paths that get matched by the glob pattern.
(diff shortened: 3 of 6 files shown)


fneu:
please consider shell glob behaviour when reviewing the last commit:
```shell
fabian ~/testdir > echo **/**/**/**/**
dir dir/file dir/file dir/file dir/file
fabian ~/testdir > echo dir/**/**/**/**
dir/file
```

fneu:
ping @sils1297 @AbdealiJK

sils1297:
```
fabian ~/testdir > echo dir/**/**/**/**
dir/file
```

seems very sane, are we removing this behaviour? Please let's not do that :(

fneu:
no that's just current shell behaviour, bot of them. duplicate results on` '**/**/**'`, no idea wh.
I find it just interesting though, that this behaviour is shared between implementations.

with this commit we would never ever yield duplicate results, and only leave a trailing slash if the pattern itself contains one, matching only dirs. Otherwise all trailing slashes are stripped.

This way we are consistent with cpython globs in cases not containing `**/**`, and more sane in cases containing the pattern.

fneu:
@sils actually:

```shell
fabian ~/testdir > echo dir/**/**/**/**
dir/file
```

```python
fabian ~/testdir > python3
Python 3.5.1 (default, Dec 09 2015, 07:29:36) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from coalib.parsing.Globbing import glob as myglob
>>> from glob import glob as cpython_glob
>>> myglob("dir/**/**/**/**")
['dir', 'dir/file']
>>> cpython_glob("dir/**/**/**/**", recursive=True)
['dir/', 'dir/file', 'dir/file/', 'dir/file/', 'dir/file/']
```

we are keeping the dir right now since `**` matches 'nothing', looks right?

coalib/bears/requirements/NpmRequirement.py
```diff
@@ -49,7 +49,17 @@ def is_installed(self):
...
         if platform.system() == 'Windows':  # pragma: no cover
             cmd = ['cmd', '/c'] + cmd
-        return not call_without_output(cmd)
+        locally_installed = not call_without_output(cmd)
+
+        # global:
+        cmd = ['npm', 'list', '-g', self.package]
+        if platform.system() == 'Windows':  # pragma: no cover
+            cmd = ['cmd', '/c'] + cmd
+        globally_installed = not call_without_output(cmd)
```
Asnelchristian: this is just a repetition of the previous block right ( the only difference is `cmd`) maybe... defining a function that does is a good idea so that things are not repeated... :sweat_smile:

Asnelchristian:
ack https://github.com/coala-analyzer/coala/pull/2754/commits/ce394d3c7f2c21ca117914d17ed3a46c8f47b821

Asnelchristian:
ack https://github.com/coala-analyzer/coala/pull/2754/commits/99cd67abd9d042d7fbf8f14807cd53b321f94c63

Asnelchristian:
ack https://github.com/coala-analyzer/coala/pull/2754/commits/a0e1d0450ccb24c99fb5b79cfef7df1e3b666e0f

fneu:
reack https://github.com/coala-analyzer/coala/pull/2754/commits/20544c52a36583637ddf8d8b64ef87d45800953c

coalib/bears/requirements/NpmRequirement.py
fneu: actually even more importantly, we would have always performed both checks, which is unnecessary and kinda slow (those checks can take a second or two...)

now we return immediately if we find it. local check comes first because cib will default to installing locally.

fneu:
reack https://github.com/coala-analyzer/coala/pull/2754/commits/a0e1d0450ccb24c99fb5b79cfef7df1e3b666e0f

coalib/bears/requirements/NpmRequirement.py
```diff
@@ -49,7 +49,13 @@ def is_installed(self):
...
         """
-        cmd = ['npm', 'show', self.package]
-        if platform.system() == 'Windows':  # pragma: no cover
-            cmd = ['cmd', '/c'] + cmd
-        return not call_without_output(cmd)
+        for cmd in (['npm', 'list', self.package],
+                    ['npm', 'list', '-g', self.package]):
+
+            if platform.system() == 'Windows':  # pragma: no cover
+                cmd = ['cmd', '/c'] + cmd
```
sils1297: don't we have a function somewhere that does this windows stuff for you? Shell.py? CC @Makman2

fneu:
Idk it was done this way before.

It figured this can be done elsewhere, but figured this is probably not the 
scope of this change and I will likely change Requirements later anyway.


Am 15. September 2016 10:06:56 vorm. schrieb Lasse Schuirmann 
<notifications@github.com>:

coalib/bears/requirements/NpmRequirement.py
```diff
@@ -49,7 +49,13 @@ def is_installed(self):
 
         :param return: True if dependency is installed, false otherwise.
```
Adrianzatreanu: commit message: ``failes``-> ``fails``

fneu:
reack https://github.com/coala/coala/pull/2754/commits/15c346699cd225ccf981002887dd152b8ae4c1ec https://github.com/coala/coala/pull/2754/commits/24a3cc542a7777b84b6250235cf2dbab4ac18b0f https://github.com/coala/coala/pull/2754/commits/1a6eb98e942c863e7093cab2992b7f1ac6a7b583 https://github.com/coala/coala/pull/2754/commits/5ffb311b18b4e17c3bfc2dc7960a08c43c0857a5

fneu:
@rultor merge

sils1297:
@stevepeak the codecov state seems to be missing on multiple PRS lately

@underyx do we have coverage check via pytest enabled for coala and the bears? We could just disable codecov then I guess. Or not making it mandatory.

stevepeak:
@sils1297 we had a DNS outage for about 1 hour today. Are all the missed PR's today and maybe within that hour?

sils1297:
Guess so, what do we have to do?

stevepeak:
@sils1297 restart CI. Codecov did not get the reports. Thank you!

fneu:
@stevepeak yes I probably hit that time window. Re-pushed now, looks all good here, thank you.

fneu:
reack https://github.com/coala/coala/pull/2754/commits/b02142fd4859ad8f19065666d8675b50e50928f6

fneu:
@rultor merge

PR #2758, opened 2016-09-14
Tabs to spaces lookahead

jayvdb:
Determine the previous whitespace needed after enountering
non-whitespace or EOL, so that it is possible to determine
if a space immediately before the tab is followed by
whitespace and thus should be converted to a tab.

Fixes https://github.com/coala-analyzer/coala/issues/2756


sims1253:
ack 67e46f9

sils1297:
@stevepeak happening again here

jayvdb:
Maybe there is something special about me , or the way I am working?

jayvdb:
I have now enabled "Allow edits from maintainers. " in case that is the problem.

sils1297:
This doesn't seem to be urgent so I'll leave this open, I guess @stevepeak will look at it when there's daytime in the US again, he can look into the codecov logs and see what's wrong.

stevepeak:
@sils1297 so it looks like the status was attempted to be set. However, according to github api Travis CI failed for this commit and codecov did not set the failure because of this.

See here: https://api.github.com/repos/jayvdb/coala/commits/67e46f93a13584900d7f8a078534ea4ffa2f653f/status

I believe that if @jayvdb disables Travis and rebuilds a commit it will work.

Makman2:
commit message:
I would change to:
```
SpacingHelper: Lookahead tabs to spaces
```

Makman2:
unack 67e46f9

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -86,11 +86,28 @@ def replace_spaces_with_tabs(self, line: str):
         :param line: The string with spaces to replace.
         :return:     The converted string.
         """
+        def previous_whitespace():
+            # Find the previous real character, and remaining chars to fill tab
```
Makman2: why nesting functions :O what about putting this into class-level and prepending an underscore to the function name?

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -101,20 +118,23 @@ def replace_spaces_with_tabs(self, line: str):
                 currspaces += space_count
                 tabless_position += space_count
             else:
-                result += currspaces*" " + char
+                if currspaces:
```
Makman2: what if `currspaces` is zero?

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -101,20 +118,23 @@ def replace_spaces_with_tabs(self, line: str):
...
+                    if currspaces == 1:
+                        result += previous_char
+                    else:
+                        result += previous_whitespace()
```
Makman2: -->
```
result += previous_char if currspaces == 1 else previous_whitespace()
```

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -101,20 +118,23 @@ def replace_spaces_with_tabs(self, line: str):
...
+            previous_char = char
 
-                currspaces = 0
-
-        result += currspaces*" "
+        if currspaces:
+            if currspaces == 1:
+                result += previous_char
+            else:
+                char = None
```
Makman2: last time `char = None` occured outside of the if. Is this the right position?

Makman2: btw why do you set `char` to `None`? does this have any effect?

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -86,11 +86,28 @@ def replace_spaces_with_tabs(self, line: str):
...
+            # Find the previous real character, and remaining chars to fill tab
+            non_whitespace_position = tabless_position - currspaces
+            tab_fill = non_whitespace_position % self.tab_width
+
+            if tab_fill and tab_fill + currspaces >= self.tab_width:
+                whitespace = "\t"
+                remaining_spaces = currspaces - (self.tab_width - tab_fill)
+            else:
+                whitespace = ""
+                remaining_spaces = currspaces
```
Makman2: you could move a part away from the if:
```
remaining_spaces = currspaces
if ...:
    whitespace = '\t'
    remaining_spaces -= self.tab_width - tab_fill
else:
    whitespace = ''
```

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -86,11 +86,28 @@ def replace_spaces_with_tabs(self, line: str):
...
+
+            whitespace += "\t" * (remaining_spaces // self.tab_width)
+            whitespace += " " * (remaining_spaces % self.tab_width)
```
Makman2: maybe wanna append this in one go?
```
whitespace += ("\t" * (...)
               " " * (...))
``` 
but it's your choice, I think this is fine too :)

coalib/bearlib/spacing/SpacingHelper.py
```diff
@@ -86,11 +86,28 @@ def replace_spaces_with_tabs(self, line: str):
...
+            return whitespace
+
         currspaces = 0
         result = ""
         # Tracking the index of the string isnt enough because tabs are
         # spanning over multiple columns
         tabless_position = 0
+        previous_char = None
```
Makman2: no need to initialize to `None`, you can set variables the first time also in loops :)
But you use `previous_char` below in the if before it's set to a string. Maybe this should be
```
previous_char = ''
```
?

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: I'm flexible.  I dont know of any significant problems with nested funcs in Python 3 (they do have problems in Python 2).

But I dislike class level funcs unless it uses the `cls` for obvious reasons.  Is a module level private func ok?

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: ok

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: im flexible with this.

I would actually prefer the following, to avoid unnecessary string concatenations:

```python
    tab_count = remaining_spaces // self.tab_width
    if tab_count: whitespace += "\t" * tab_count

    space_count = remaining_spaces // self.tab_width
    if space_count: whitespace += "\t" * space_count
```

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: It needs to be initialised outside the loop, or in an loop `else` block, to handle the empty string.  If we assume that the `line` always has at least an EOL, that isnt necessary.

I've set it to `None` intentionally to ensure it breaks properly if it is used before there is a previous character.
It can never be `None` where it is currently used - if `currspaces` is > 0 , there has always been a previous character, so `previous_char` will always have a `str` value.

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: there is no previous whitespace to be added .... ?

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: that blows the line width ... ?

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: This can be removed - I was using `char` in an assert during testing,

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: You used single quotes here.  Is that a preference for this repo?  I see mostly double quotes.

coalib/bearlib/spacing/SpacingHelper.py
sils1297: you could use parentheses to break the line

coalib/bearlib/spacing/SpacingHelper.py
sils1297: we decided somewhere that single make more sense than double but we didn't write it down yet so you're free to use what you like most. At some point we'll probably write a bear that will bluntly replace all by single quotes except when it contains single quotes in the string.

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: Then how is it better than a normal `if` ?

coalib/bearlib/spacing/SpacingHelper.py
jayvdb: I prefer single.. Flake8 has quotes plugin that does this. Flake-quotes iirc.

coalib/bearlib/spacing/SpacingHelper.py
sils1297: it's two lines vs. 4. It's a matter of taste, really.

coalib/bearlib/spacing/SpacingHelper.py
sils1297: :+1: our bear would work for all coalang supported languages though because it's really simple to do. But we could use flake-quotes in the meantime. If you find it disturbing, file an issues. Otherwise we'll just wait for it to happen somewhen.

coalib/bearlib/spacing/SpacingHelper.py
Makman2: oh shit sry for missing your comments...
and yes module level is okay, though you can also use `@staticmethod` instead of `@classmethod` for class-level :+1:

coalib/bearlib/spacing/SpacingHelper.py
Makman2: nvm :3

coalib/bearlib/spacing/SpacingHelper.py
Makman2: no don't do that, an empty string concatenation should be equally cheap like an if^^

If you really want it make a performance test and see if it's worth it^^

coalib/bearlib/spacing/SpacingHelper.py
Makman2: you use it only for appending to a string, appending `None` raises an error, so there's imo no meaning to it setting it to `None` beforehand.

PR #2764, opened 2016-09-16
SpacingHelper: use string.expandtabs

jayvdb:
SpacingHelper.replace_tabs_with_spaces can be
implemented using string.expandtabs, which is
implemented in C for CPython.

SpacingHelper.yield_tab_lengths is now unused.


jayvdb:
The tests should also be removed, but I wanted to show they all passed first.

adtac:
@jayvdb In your commit message, shouldn't it be `str.expandtabs` instead of `string.expandtabs` since  the class name is `str`?

Also, please use double backticks for technical terms, variables, functions, etc.
Example:

    ``SpacingHelper.replace_tabs_with_spaces`` can be...

instead of

    SpacingHelper.replace_tabs_with_spaces can be...

sils1297:
also `use` -> `Use`. I think we can also remove `yield_tab_lengths`, at least it's not used in coala-bears and I doubt anyone outside coala uses it

jayvdb:
> In your commit message, shouldn't it be str.expandtabs instead of string.expandtabs since the class name is str?

Ugh, yes, it was removed from `string` in Python 3.

>  I think we can also remove yield_tab_lengths, at least it's not used in coala-bears and I doubt anyone outside coala uses it

Right, it was only created in Feb 2016 (2d8c38009), and it will not tested any longer.

jayvdb:
Created #2769, as the entire module/class needs a rewrite, or to be discarded.

jayvdb:
Note this is just a WIP as I have to run, and wanted to share how expandtabs can be used even for converting to tabs.

sils1297:
@jayvdb I prefixed your title with `wip:` so GitMate will mark it wip from now on on every push. If you want to get a review, remove the `wip:` prefix and *then* push so it'll mark it pending review again (or tell us of course).

jayvdb:
Thanks.

sils:
@jayvdb do you need assistance to continue?

jayvdb:
No. I've been occassionally testing it on real tabbed projects, with my pygments patch to help.
https://github.com/coala/docker-coala-base/issues/17 is slowing me down.

Issue #2742, opened 2016-09-09
Update the latest stable version in the README


adtac:
I think it's 0.8.1 now.

manthansharma:
I want to work on this issue

sils1297:
Hey,

I think you have done two newcomer issues and it seems you're well acquainted with the workflow! That's awesome :)

We recommend to not do newcomer issues anymore because they take us more time to file than to fix them - let other newcomers can have a few as well. The next level for you are the [low difficulty issues](http://tinyurl.com/coala-low) - you can now dive into whatever you want. Let us know if you need help or have any special interests - there's a lot of stuff going on and usually interests can be met :)

Cheers!

sils1297:
@adtac why do we have that badge if it's not auto updating? Can we just remove that?

adtac:
@sils1297 IMO we should mention the current version *somewhere* in the README. And I think a badge is a cool way for it to stand out. But I can't think of any way to auto-update it :(

adtac:
We could update it during the series of commits we do for releases - release notes, README, VERSION, then push to PyPI.

sils1297:
hm, maybe we can make a script for the release, shouldn't be too hard

mgracer48:
can i work on this?

sils1297:
Sure @mgracer48 !

mgracer48:
So whats the task basically i need to do (change the version numbers? ) (make a script to change it easily?)

sils1297:
The task here is just to change the version numbers indeed. If you want to
do something more challenging, add a release script somewhere ;) like
`.misc/release.py 0.8.1`, there's a script `.misc/adjust_version_number.py`
that you can reuse, it's already used in rultor, our merge bot; see
.rultor.yml.

PR #2771, opened 2016-09-17
README: Change version of 0.8.0 to 0.8.1

mgracer48:
The version is outdated so we need to update it

Closes https://github.com/coala/coala/issues/2742

Did not create the script yet so thats still a todo

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -229,7 +229,7 @@ License
 
 .. |Windows| image:: https://img.shields.io/badge/platform-Windows-brightgreen.svg
 .. |Linux| image:: https://img.shields.io/badge/platform-Linux-brightgreen.svg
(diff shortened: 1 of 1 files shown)


sils1297:
ack https://github.com/coala/coala/pull/2771/files/9dcbd2b17097694a0f43bd3bba42c3d2ebbcb81d but it seems that you didn't update your master before creatig your branch. I recommend pulling your master and then branching off.

Now you'll have to do another rebase: this way we verify that the commit still works on top of the new commits (sometimes a rebase/merge can break the codebase which is why we do it with WIP PRs) - second review will be shorter though because your change is already acknowledged.

sils1297:
rebased it for you onto https://github.com/coala/coala/pull/2772

Issue #2660, opened 2016-08-23
Make exception tracebacks default


adtac:
Instead of asking the user to run coala with `-L DEBUG`

Makman2:
Not `difficulty/newcomer`, as we don't want to execute coala always in debug-mode.

mgracer48:
Ill work on this
just more details please

sils1297:
CC @Uran198 does it make sense to address this issue before your PR is merged?

mgracer48:
@Makman2 then what do we want?

Makman2:
Change exception tracebacks and related information from DEBUG to ERROR :+1:

Uran198:
@sils1297 I think it does. There are a lot of places where the current logging will be exchanged for default one, so this does not really change that number, so not adding more work, while I like this feature and I think it might be useful.

sils1297:
Ok @mgracer48 take a look at LogPrinter, it has a method to log exceptions. The traceback is debug by default, we probably want to make it info or something like that at least. You'll probably have to change a few tests as well.

mgracer48:
Sorry i havent gotten to this u have been reading python books to improve my python and also havent had the time

mgracer48:
which test files?

Makman2:
the ones that will fail when you've done your changes :P

mgracer48:
```
E       AssertionError: '' != '[DEBUG][18:36:51] 4 r34l ch4ll3n63: 123 Ä[35 chars]n→ d'
E       + [DEBUG][18:36:51] 4 r34l ch4ll3n63: 123 ÄÖü ABc @€¥ §&% {[( ←↓→↑ ĦŊħ ß°^ \
E       + → d

tests/output/printers/LogPrinterTest.py:60: AssertionError

```

mgracer48:
Help?

Makman2:
@mgracer48 got a link to your commit / PR? :)

mgracer48:
https://github.com/coala/coala/pull/2782

mgracer48:
@Makman2

PR #2782, opened 2016-09-23
LogPrinter: Change log type DEBUG to log type info

mgracer48:
Instead of asking the user to run coala with -L DEBUG It's  better to have exception tracebacks default

Closes https://github.com/coala/coala/issues/2660

Warning: WIP
Have to fix this bug


Nosferatul:
Hello there! You should have a newline between shortlog _LogPrinter: Change log type DEBUG to log type info_ and the body _Instead of asking the user to run coala with -L DEBUG It's  better to have exception tracebacks default_ . Also you should check [https://coala.readthedocs.io/en/latest/Developers/Writing_Good_Commits.html](url)

Nosferatul:
Also you will need to do a rebase after you pass the checks!

tests/output/printers/LogPrinterTest.py
```diff
@@ -32,7 +32,7 @@ def test_logging(self):
         uut.log_message(self.log_message, end="")
         self.assertEqual(uut.printer.string, str(self.log_message))
 
-        uut = LogPrinter(StringPrinter(), log_level=LOG_LEVEL.DEBUG)
+        uut = LogPrinter(StringPrinter(), log_level=LOG_LEVEL.INFO)
```
Makman2: This is the cause of the failing test! :D
The `log_level` parameter sets the minimum log-level to print. `DEBUG` is the lowest one, `INFO` follows. So please don't change this line, then tests work fine :D

coalib/output/printers/LogPrinter.py
```diff
@@ -115,7 +115,7 @@ def log_exception(self,
 
         self.log(log_level, message, timestamp=timestamp, **kwargs)
```
Adrianzatreanu: are there 2 spaces between "It's" and "better"? make it one please.

also, it should be "It's"->"it's".

needs comma after the "-L DEBUG"

enclose "-L DEBUG" in double backticks as its a command.

end body with period.

PR #2784, opened 2016-09-24
Udayan/fileproxy

Udayan12167:


coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,45 @@
...
+class Fileproxy:
+    """
+    The file proxy is a wrapper object to wrap every file. The file proxy is an
+    object that contains properties about the file currently processed (such as
+    the filename). It contains:
+        * The filename, this is the starting point for every other attribute.
+        * The content of the file as a string.
+        * The content of the file, line-splitted.
+
+    Note: Other properties can be included later as and when required.
```
Adrianzatreanu: "as and when required"

what sense does that "as and" make?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,45 @@
...
+
+    The equality of the object is checked just based on it's filename.
```
Adrianzatreanu: it's -> its

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+
+    def __init__(self, filename):
+        """
+        Constructs a new fileproxy object.
+        :param filename: The name of the file to load.
+        """
+        self.filename = filename
+        with open(self.filename, "r", encoding="utf-8") as filehandle:
```
Adrianzatreanu: wouldnt omitting "r" make it "r" by default?

Adrianzatreanu:
ack a6fb83249d9eac979a4a6e9f328eeb46ef24fbea

coalib/files/Fileproxy.py
Udayan12167: as required or when required. But it should be both so as and when.

Adrianzatreanu: oh alright, it makes sense that you explain it, but i didnt get it just by reading it.

coalib/processes/BearRunning.py
```diff
@@ -408,13 +397,27 @@ def run_local_bears(filename_queue,
...
+                         timeout,
+                         LOG_LEVEL.ERROR,
+                         "An internal error occurred.",
+                         Constants.THIS_IS_A_BUG)
+                send_msg(message_queue,
+                         timeout,
+                         LOG_LEVEL.DEBUG,
+                         "The given file through the queue is not in the file "
+                         "dictionary.")
+                continue
```
sils1297: actually this shouldn't happen at all, can you remove that branch altogether? If the filename isn't in there the exception is the right thing to get thrown.

tests/files/FileproxyTest.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_str(self):
+        self.assertEqual(str(self.fileproxyobject),
+                         open(self.testfile).read())
+
+    def test_equal(self):
+        with open(self.testfile, 'a+') as file:
+            file.write("New line")
+            self.assertEqual(self.fileproxyobject, Fileproxy(self.testfile))
+            lines = file.readlines()
```
sils1297: unused variable

sils1297: oh sorry

sils1297: wouldn't it be easier to read the lines before writing the new line?

sils1297:
ack https://github.com/coala/coala/pull/2784/commits/2b9763f6dd536c150b9d3f469f0340d2a4b77625

sils1297:
ack https://github.com/coala/coala/pull/2784/commits/217b7c93277cb3557cd1e355188e184864dc5335 you're thinking ahead of me, thanks :)

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(self, filename):
+        """
+        Constructs a new fileproxy object.
+        :param filename: The name of the file to load.
+        """
+        self.filename = filename
+        with open(self.filename, encoding="utf-8") as filehandle:
+            self.content = filehandle.read()
+        self.lines = tuple(self.content.splitlines(True))
```
sils1297: can we make the lines and content readonly somehow? We had some discussions that if bears try to fiddle with the lines/content that should raise an exception because other bears within the process might get the changes as well and we don't want that. It's important that this one is immutable.

coalib/processes/BearRunning.py
```diff
@@ -404,7 +404,7 @@ def run_local_bears(filename_queue,
                                     local_result_dict,
                                     control_queue,
                                     filename,
-                                    file_dict[filename])
+                                    file_dict[filename].lines)
```
sils1297: you aren't passing the proxy here

coalib/processes/Processing.py
```diff
@@ -452,22 +453,22 @@ def yield_ignore_ranges(file_dict):
                                    start,
                                    1,
                                    line_number,
-                                   len(file[line_number-1])))
+                                   len(list(fileproxy)[line_number-1])))
```
sils1297: list has to iterate over it, why not lineproxy.lines?

coalib/processes/Processing.py
```diff
@@ -452,22 +453,22 @@ def yield_ignore_ranges(file_dict):
...
                 elif "ignore " in line:
-                    end_line = min(line_number + 1, len(file))
+                    end_line = min(line_number + 1, len(list(fileproxy)))
```
sils1297: fileproxy should have a `len` defined anyway

coalib/processes/Processing.py
sils1297: better: define a getindex for fileproxy

sils1297:
https://github.com/coala/coala/pull/2784/commits/21dee2f6f052baf941a01192debd648f5e899fa6 needs work

jayvdb:
I believe https://github.com/coala/coala-bears/issues/815 is waiting for this. cc @Flap-Py

Udayan12167:
reack a357bc9

Udayan12167:
reack aef1d2e

Udayan12167:
reack 142c0e4

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,60 @@
...
+        * The content of the file, line-splitted.
+
+    Note: Other properties can be included later as and when required.
+
+    The equality of the object is checked just based on its filename.
+    """
+
+    def __init__(self, filename):
+        """
+        Constructs a new fileproxy object.
```
sils: missing newline here

tests/files/FileproxyTest.py
```diff
@@ -0,0 +1,55 @@
...
+        self.fileproxyobject = Fileproxy(self.testfile)
+
+    def test_str(self):
+        self.assertEqual(str(self.fileproxyobject),
+                         open(self.testfile).read())
+
+    def test_equal(self):
+        with open(self.testfile, 'a+') as file:
+            lines = file.readlines()
+            file.write("New line")
```
sils: your test modifies the actual file? :/

sils:
yes, this is vitally important

tests/files/FileproxyTest.py
Udayan12167: I wanted to test the equivalence of the file is only on the basis of filename and not the contents.

sils: tests shouldn't modify persistent files on the filesystem. If they do, use tempfiles and clean them up properly on teardown.

sils:
@Udayan12167 ping?

Udayan12167:
reack 3c3afbf 4fad87e

coalib/processes/BearRunning.py
```diff
@@ -246,11 +247,11 @@ def run_global_bear(message_queue,
 
 def run_local_bears_on_file(message_queue,
                             timeout,
-                            file_dict,
```
jayvdb: To be backwards compatible, or at least break cleanly, `file_dict` here should be replaced with `file_contents`, and the function should check that `file_contents` is of the new expected type.

coalib/processes/BearRunning.py
```diff
@@ -160,9 +160,9 @@ def get_local_dependency_results(local_result_list, bear_instance):
 def run_local_bear(message_queue,
                    timeout,
                    local_result_list,
-                   file_dict,
```
jayvdb: To be backwards compatible, or at least break cleanly, `file_dict` here should be replaced with `file_contents`, and the function should check that `file_contents` is of the new expected type.

coalib/files/Fileproxy.py
jayvdb: I do not see the benefit of leaving a note like this in the API documentation.
Who is it helping?  API users?  Not likely.
Developers will do this, whether you give them permission or not.
When should this Note be removed?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,97 @@
...
+
+        :param filename: The name of the file to load.
+        """
+        self.filename = filename
+        with open(self.filename, encoding="utf-8") as filehandle:
+            self.__content = filehandle.read()
+        self.__lines = tuple(self.__content.splitlines(True))
```
jayvdb: Having `__lines` and `__content` 'doubles' memory usage compared to the existing implementation.
I strongly suggest that the proxy *attempts* to keep only one of `__lines` or `__content` in memory unless it appears that both are being used.

i.e.

1. `__init__` loads `__content` ; 
2. `__iter__` loads `__lines` and either discards `__content` or switches `__content` to a `weakref` so it can be garbage collected
3. `__str__` loads `__content` from `__lines` if necessary, and doesn't discard `__content`, as both are apparently being used.

Or, a better approach is to only load the file contents when `__iter__` or `__str__` are called, which should reduce startup time , thereby improving responsiveness that the user 'feels'.  However, that approach means the unicode decode handling also needs to be delayed.

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,97 @@
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq("filename")
+class Fileproxy:
+    """
+    The file proxy is a wrapper object to wrap every file. The file proxy is an
+    object that contains properties about the file currently processed (such as
+    the filename). The fileproxy object also protects your file from being
```
jayvdb: "protects your file" - eh?  what is being protected?

A bear can still `open()` the file and write to it.

A bear can still edit the contents of the fileproxy object using `fp[0][:] = 'abc'`.

coalib/files/Fileproxy.py
Udayan12167: So we actually did this to increase the speed of coala. As the fileproxy will be used by multiple bears hence it will have the content and the lines ready to go whenever a bear uses it. Avoiding to load it multiple times. This Fileproxy will also be available across sections and hence it will save time their too.

coalib/files/Fileproxy.py
Udayan12167: So by making the members start with `__` we make them protected. Also will that work even if it's not a list it's a tuple?

Udayan12167: Yes it can still edit the file using `open()`. What I wanted to say was it protects the file contents loaded into the fileproxy by having immutable objects and private class members so that other bears get the same content.

coalib/processes/BearRunning.py
Udayan12167: I don't gather. What new type?

coalib/files/Fileproxy.py
Udayan12167: Yes maybe we can generate the string version of the file from `__lines` though. Let me know what you think of it. Is the tradeoff on time joining all the lines together worth the memory we will save?

coalib/files/Fileproxy.py
jayvdb: > ... if it's not a list it's a tuple ...

`__lines` is not a tuple.  If the goal is to use immutable objects, please do that.

jayvdb: Also, as this is technical documentation, rather than "protects your file" , state something like " returns only immutable objects for bears to read only.".  Saying `immutable` is key for callers to understand what it does.

jayvdb: `str.splitlines` returns a mutable list.

coalib/processes/BearRunning.py
jayvdb: `file_contents` is going to have a different class than `file_dict` was ?

coalib/files/Fileproxy.py
sils: agree, I'd remove that from API docs

coalib/files/Fileproxy.py
Udayan12167: I do cast the list to tuple in the init method.

jayvdb: *nod* sorry, you are correct, I didnt notice that when I said `__lines is not a tuple.`.

tests/files/FileproxyTest.py
```diff
@@ -0,0 +1,57 @@
...
+        with open(self.testfile, 'w') as file:
+            file.writelines(lines)
+
+    def test_iter(self):
+        with open(self.testfile, 'r') as file:
+            self.assertEqual(tuple(self.fileproxyobject),
+                             tuple(file.readlines()))
+
+    def test_hashing(self):
+        lines = open(self.testfile).readlines()
```
sils: I dislike this way of testing, I'm not sure this test is even needed. If you think it is, make an own testcase object and make the file write back in a teardown method (read in setup)

Udayan12167:
reack 2afd019 6b7bd50

sils:
@Udayan12167 your tests fail...

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq("filename")
+class Fileproxy:
```
AbdealiJK: Opinionated: FileProxy seems more natural to me

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+    Old line
+    <BLANKLINE>
+
+    Get contents of the file as a string:
+
+    >>> fp.content
+    'Old line\\n'
+
+    Get all lines of the file as a list:
+    >>> fp.lines
```
AbdealiJK: Newline above this ?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+    ['Old line\\n']
+    """
+
+    def __init__(self, filename):
```
AbdealiJK: Few general comments here:
 - It makes sense to do the equality check with os.path.abspath(filename). Because ./test.py and ./abc/../test.py are actually the same ?
 - Also have a `absolute_filename` like thing for the absolute path to the file ?
 - Please mention that this does not be "closed" like a normal file pointer

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        Constructs a new fileproxy object.
+
+        :param filename: The name of the file to load.
+        """
+        self.filename = filename
+        with open(self.filename, encoding="utf-8") as filehandle:
+            self.__lines = tuple(filehandle.readlines())
```
AbdealiJK: Calculate __lines and __linecount lazily ! Why process them when not needed ?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __iter__(self):
+        """
+        :return: A list of lines of the file to iterate over.
+        """
+        return iter(self.__lines)
+
+    def __hash__(self):
+        """
+        :return: Only hashes the filename.
```
AbdealiJK: hash the absolute filename as thats what is unique

coalib/files/Fileproxy.py
AbdealiJK: What happens if the file doesnt exist ?
What about if the user does not have permission to read the file ?
This documentation needs `raises` ?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+    >>> fp.content
+    'Old line\\n'
+
+    Get all lines of the file as a list:
+    >>> fp.lines
+    ['Old line\\n']
+    """
+
+    def __init__(self, filename):
+        """
```
AbdealiJK: According to pythn constructors should not have documentation. Class should contain this info

coalib/processes/BearRunning.py
Udayan12167: But it doesn't affect backwards compatibility in any way. It's just not passing the full dictionary around when we need only one element of it.

coalib/settings/SectionFilling.py
```diff
@@ -3,7 +3,6 @@
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.collecting import Dependencies
 from coalib.collecting.Collectors import collect_bears
-from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
```
AbdealiJK: Don't we run PyUnusedCodeBear - https://github.com/coala/coala/blob/master/.coafile#L15 ?? 
How is this not being detected ?

coalib/output/ConsoleInteraction.py
```diff
@@ -172,7 +172,7 @@ def print_lines(console_printer,
                               color=FILE_LINES_COLOR,
                               end='')
 
-        line = file_dict[sourcerange.file][i - 1].rstrip("\n")
+        line = file_dict[sourcerange.file].lines[i - 1].rstrip("\n")
```
AbdealiJK: I think this is a fairly common usecase. Can FileProxy have a helper function to remove newlines ? That would make it more elegant to use.

So, essentially have a function `.lines(start=0, end=-1, with_newline=False)` This way we can even handle the splicing inside the function. There would be a similar function `.line(line_num, with_newline=False)` for a single line.

Maybe I am over thinking and over wrapping the File though ... opinions welcome.

coalib/processes/BearRunning.py
```diff
@@ -383,7 +383,7 @@ def run_local_bears(filename_queue,
     :param timeout:           The queue blocks at most timeout seconds for a
                               free slot to execute the put operation on. After
                               the timeout it returns queue Full exception.
-    :param file_dict:         Dictionary that contains contents of files.
+    :param file_dict:         Dictionary that contains fileproxy objects.
```
AbdealiJK: So, why is this a dictionary ? FileProxy already has the filename right ? Double info is bad.
Earlier when it was just list of lines it made sense,

coalib/processes/Processing.py
```diff
@@ -250,8 +251,7 @@ def get_file_dict(filename_list, log_printer):
     file_dict = {}
     for filename in filename_list:
         try:
-            with open(filename, "r", encoding="utf-8") as _file:
-                file_dict[filename] = tuple(_file.readlines())
+            file_dict[filename] = Fileproxy(filename)
```
AbdealiJK: Here's the issue. Why file_dict here ?
Just `fileproxy_set` or `file_set` is enough right ?

tests/output/ConsoleInteractionTest.py
```diff
@@ -49,6 +49,15 @@
 STR_PROJECT_WIDE = "Project wide:"
 
 
+class DummyFileProxy:
+
+    def __init__(self, lines):
+        self.lines = lines
```
AbdealiJK: I think a better test would be to have a `create_file_proxy` function which does 

```
def create_file_proxy(lines):
    fp = FileProxy('/dev/null')
    fp.__lines = lines
    return fp
```
as that will test file proxy rigorously in it's usage in all these files

AbdealiJK: And put that function in coalib/file/Fileproxy.py itself ? As it seems to be needed in multiple tests

coalib/processes/BearRunning.py
Udayan12167: Because that was planned as a later change. To make it a set.

tests/output/ConsoleInteractionTest.py
Udayan12167: The __lines won't be accesible. That is the purpose of using `__`. Maybe we can let fileproxy create two kinds of objects. One with a custom name and lines as input and one with just the absolute path of the file as input.

jayvdb:
For this to be useful for https://github.com/coala/coala-bears/issues/815 , it needs to preserve different EOLs.  I've done a quick check, and neither `readlines` nor `read()` preserve different EOLs if the file is opened as `open(self.filename, encoding="utf-8")`.

It needs to be opened in binary mode (without any encoding, ugly), or with the Python 3 [`newline=''`](https://docs.python.org/3.4/library/functions.html#open), and `readlines` works correctly in this mode however the EOLs need to be stripped from each line to be backwards compatible.

coalib/files/Fileproxy.py
jayvdb: have you done any performance testing on converting the returned list to a tuple?
As I have said previously, trying to prevent damage is not a good design.  If you want to write a proxy, do it in Cython.  Then you can really control access, and prevent damage.
A single underscore is sufficient.  PyLint and other tools warn about access to private attributes.  We should be using linters, rather than creating complexity in the code.

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+        Constructs a new fileproxy object.
+
+        :param filename: The name of the file to load.
+        """
+        self.filename = filename
+        with open(self.filename, encoding="utf-8") as filehandle:
```
jayvdb: See https://github.com/coala/coala/pull/2784#issuecomment-257217569

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
...
+        :return: A string of contents of the whole file.
+        """
+        return "".join(self.__lines)
+
+    @property
+    def lines(self):
+        """
+        :return: A string of contents of the whole file.
+        """
+        return list(self.__lines)
```
jayvdb: And now you convert it to a list again?  More inefficiency.  Bears shouldn't be playing with this returned value.  If they need to modify the lines, they should create their own collection of the lines, and alter it.

coalib/files/Fileproxy.py
jayvdb: If all filenames are relative to the same directory, there is no need for absolute filenames for uniqueness.
Can we rely on no bears doing a `os.chdir`?  If not, we need to convert all names to absolute somewhere - maybe in the constructor?

coalib/files/Fileproxy.py
```diff
@@ -0,0 +1,89 @@
+from coala_utils.decorators import generate_eq
```
jayvdb: I didnt see an answer to https://github.com/coala/coala/pull/2784#pullrequestreview-5991591 :

Why is this creating coalib/files/__init__.py ; those namespace files are only needed pre 3.4 ?

AbdealiJK: We currently create inits for all modules. This can be brought up in a new issue? If we're removing it here, let's remove it everywhere.
Example: https://github.com/coala/coala/blob/master/coalib/processes/__init__.py

coalib/files/Fileproxy.py
AbdealiJK: We do a os.chdir in the GitCommitBear iirc
Maybe we'll have to do it in other things too...
But either way : ./abc.txt and abc/../abc.txt still need a normpath or abspath even if base dir is same.

coalib/files/Fileproxy.py
AbdealiJK: Rather than doing readlines do a tuple(filehandler) as the file handler is also iterable
Also, yeah converting it to a list is redundant later.

tests/output/ConsoleInteractionTest.py
AbdealiJK: The __lines can be accessible to a static member of the class. Seems like we're talking about the same thing though ^.^

Udayan12167: Yup

thisch:
@Udayan12167 which issue does this PR solve?

palash25:
@Makman2  I think we can close this now.

jayvdb:
Link to cEP & PR pls

palash25:
This PR is closed in favour of https://github.com/coala/coala/pull/5485 which is a part of the cEP https://github.com/coala/cEPs/blob/master/cEP-0026.md

PR #2796, opened 2016-09-27
Remove newlines when adding settings

srisankethu:


Adrianzatreanu:
You need to fix the tests as well :)

Issue #2073, opened 2016-04-25
Linter: Remove prerequisite_check_command


Makman2:
It's nicer to manually override `check_prerequisites`, as it can be more convenient to use special checks the default linter does not implement. E.g. `check_prerequisites = check_java_module(...)`
This refactor won't use up more lines of code, as 
```
@linter(..., prerequisite_check_command=("exe", "arg1", "arg2", ...), ...)
```
would change to
```
@linter(...)
class MyLinter:
     # we can choose a different name :D
    check_prerequisites = custom_check_command("exe", "arg1", "arg2", ...)
```
This could also remove boilerplate-checking-code that is performed inside `@linter`.

CC @SanketDG

SanketDG:
Totally missed this! This makes sense, but we have to provide functions for checking of common language modules to avoid code duplications.

srisankethu:
Can I take this up?

sils1297:
sure!

PR #2804, opened 2016-10-01
Linter: Remove prerequisite_check_command

srisankethu:
Fixes https://github.com/coala/coala/issues/2073

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -5,6 +5,7 @@ from itertools import chain, compress
 import re
 import shutil
 from subprocess import check_call, CalledProcessError, DEVNULL
+import warnings
 from types import MappingProxyType
 
 from coalib.bears.LocalBear import LocalBear
@@ -196,6 +197,11 @@ def _create_linter(klass, options):
                         check_call(options['prerequisite_check_command'],
                                    stdout=DEVNULL,
                                    stderr=DEVNULL)
+                        warnings.simplefilter("default")
+                        warnings.warn("The bear {} uses "
+                                      "``prerequisite_check_command`` which "
+                                      "is deprecated.".format(cls.__name__),
+                                      PendingDeprecationWarning, 12)
                         return True
                     except (OSError, CalledProcessError):
                         return options['prerequisite_check_fail_message']


Makman2:
If you remove the current prerequisite_check parameters you need to change `coala-bears` too, bears are using this. + You need to provide the current functionality (you are about to remove) like described in the issue, using function assignment:
```python
# The name 'prerequisite_check' should be better...
check_prerequisites = prerequisite_check('executable2', 'arg1', 'arg2')

# Important: The 'old' prequisite check testing for plain executable-existance shall still be there!
```
Best is you put this (or this kind of function) into a different file (I think `bearlib.abstractions` is a good address, it's the same module `@linter` is in).

If you need help with that, ping me :)

sils1297:
Can we maybe just deprecate the option? Keep it around with a warn print
for a while

Makman2:
sure good idea :+1:

srisankethu:
@Makman2 So....are the changes good except [this](https://github.com/srisankethu/coala/commit/1c64c7b5f815a5c06bcf0445c95c265de4a654fc#commitcomment-19261396)?

Makman2:
you should print a deprecation warning instead of removing the functionality now :+1:

srisankethu:
should i raise a `DeprecationWarning`?

Makman2:
raising throws an exception and prevents code execution, just log something to console (using our logging facilities somehow if possible). :+1:

Makman2:
you can look at this one for reference :)
https://github.com/coala/coala/blob/e6d475de69087afa2dd2c1c4a59c319eb31e2d16/coalib/bearlib/__init__.py#L10

srisankethu:
@Makman2 I tried adding the deprecate_settings decorator. `@deprecate_settings(prerequisite_check = 'prerequisite_check_command')`. But I am getting a `KeyError: prerequisite_check`

Makman2:
naaah this is not working :D
`@deprecate_settings` is just for bear settings. I just wanted to point you to something similar using deprecation :3 You need to write your own deprecation code directly into `@linter`.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -543,13 +544,15 @@ def __repr__(self):
     return result_klass
 
 
+@deprecate_settings(prerequisite_check = 'prerequisite_check_command')
```
Adrianzatreanu: bump

Adrianzatreanu:
so where do i start? all tests fail, you have gitmate problems, but besides that, the change is needed and good :D waiting to fix these first, bumping the gitmate issue

srisankethu:
I will fix this soon

jayvdb:
@srisankethu are you still working on this?

srisankethu:
@jayvb I am busy with my college projects. I will see what I can do...

srisankethu:
From what I understand ``@linter`` executes the executable and returns ``Result`` objects which contain the output of the executable. If I need to give a depreciation warning here it is not possible. So it think it is preferable to write a depreciation warning in the Configuration gathering because it there are only a few(around 5) bears that use prerequisite_check_command.

Makman2:
you shall print a deprecation warning when the type is created. `ConfigurationGathering` btw has nothing to do with that, it is totally responsible for loading coafiles and giving coala the final setting set.

srisankethu:
In ``ConfigurationGathering`` we can get to know the local bears and global bears that are used. Since the bears that use ``prerequisite_check_command`` are around 5, we can give a depreciation warning using ``log_printer``. That was my initial idea. Then I realized that the user can  write his own bears. If I try importing and using ``LogPrinter`` object in ``Linter.py``, i am getting additional debug [messages](https://github.com/coala/coala/blob/d394b1035e01a4bbba0a89bc8d358e12ff3aab51/coalib/coala_main.py#L77).

JayVB:
I don't think you mean for me to be the included on this conversation...

________________________________

srisankethu:
@JayVB Sorry I do not know why that has happened? You weren't tagged.

Makman2:
no no we don't want to hardcode the bears into ConfigurationGathering and check if they are used. Rather yield a message directly where the bear gets created when it uses `prerequisite_check_command`^^

Makman2:
ah and please don't use our `LogPrinter`s any more, use the python built-in logging facilities via the `logging` module^^
so just do `import logging; logging.warn(...)`

JayVB:
Not sure why I am still being included in this thread... I'm pretty sure I have nothing to do with your project.  I don't even know what a "bear" is...

Sent from my iPhone

On Nov 5, 2016, at 8:59 AM, Mischa Krüger <notifications@github.com<mailto:notifications@github.com>> wrote:


ah and please don't use our LogPrinters any more, use the python built-in logging facilities via the logging module^^
so just do import logging; logging.warn(...)

sils:
@JayVB someone likely meant to include @jayvdb - you have to hit the unsusbscribe button at the "Notifications" area right on this page

srisankethu:
@Makman2 My initial idea is wrong. I understood that. I will use logging then!

JayVB:
Please remove me from this thread!

PLEASE!

________________________________

sils:
@JayVB we cannot remove you from the thread, you have to hit the unsubscribe button. We have no control over that.

JayVB:
UGH.  It will only let me do that if I log in, and I don't have my log-in information with me.  This is  pain.  You guys should start up a new thread without me.  This was not my fault.

________________________________

chauffer:
@JayVB my apologies for this, do you not see a `mute the thread` link in your email?

JayVB:
Yes, but it requires me to log into github to mute it.  I haven't used github for a while and my password is no longer stored by my laptop.  I'll try to fix this tonight when I get home and have time, but it has been annoying getting these updates about bears and your software issues every day for the last week or so.  A shame you can't take me off of it.  I regret having ever signed up for github years ago.

________________________________

chauffer:
@JayVB the github design does seem subpar for this, https://github.com/isaacs/github/issues/808 I've opened an issue and emailed support@github.com about it.

srisankethu:
@Makman2 Can you review?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -5,8 +5,8 @@
 import re
 import shutil
 from subprocess import check_call, CalledProcessError, DEVNULL
+import logging
 from types import MappingProxyType
-
```
Makman2: please don't delete this line, it parts system imports from coala imports

coalib/bearlib/abstractions/Linter.py
```diff
@@ -96,6 +96,9 @@ def _prepare_options(options):
         raise ValueError("Invalid `output_format` specified.")
 
     if options["prerequisite_check_command"]:
+        logging.warn("The bear which uses the following executable --{} "
+                     "uses ``prerequisite_check_command`` which is "
+                     "depreciated.".format(options["executable"]))
```
Makman2: it's deprecated, not depreciated :)
same in commit message

Makman2: and second: You should print the name of the bear directly instead of "The bear which uses the following executable -- ..."
(especially why do you have those `--`?)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -96,6 +96,9 @@ def _prepare_options(options):
...
         if "prerequisite_check_fail_message" in options:
             assert_right_type(options["prerequisite_check_fail_message"],
                               str,
```
Makman2: btw, when you deprecate the option, you should have an alternative :)
You can do another PR imo implementing the stuff mentioned in the issue, but please remove the "Fixes xxxx" reference, as you don't close the issue completely :)

coalib/bearlib/abstractions/Linter.py
srisankethu: I did not find a way to get the bear name so the just made it this way

srisankethu: :+1:

coalib/bearlib/abstractions/Linter.py
srisankethu: For now, we are just giving a deprecation warning https://github.com/coala/coala/pull/2804#issuecomment-251006959.

JayVB:
Ugh... I wish you didn't keep sending me these things.  I got eight e-mails with your thread last night.  It doesn't seem too hard to me for you guys to start a new thread without me.  But...

I will spend 10 minutes now trying to figure out how to unsubscribe myself from something you guys subscribed me to.

Annoyed,
  Jay


________________________________

JayVB:
Looks like I may have figured it out... hope you get all your coding issues resolved... mind those bears, they bite!

________________________________

coalib/bearlib/abstractions/Linter.py
```diff
@@ -192,6 +193,9 @@ def check_prerequisites(cls):
                          ""))
             else:
                 if options["prerequisite_check_command"]:
+                    logging.warn("The following bear {} uses "
```
sils: IMO remove "following"

sils:
:+1: that he figured that out, loosing discussion on PRs isn't nice for us as well

coalib/bearlib/abstractions/Linter.py
srisankethu: Done!

jayvdb:
As I understand it, this deprecation warning will be shown to users.  But it appears *normal* users cant do anything about it -- unless they become a developer and fix it, but that is not very friendly to commercial users, as the CI operations team dont appreciate user facing errors like this appearing unexpectedly.

srisankethu:
@jayvdb So what do you suggest?

Makman2:
@jayvdb problem is when devs execute coala to execute their stuff, they wanna see something like this. Using a debug message doesn't suffice, as you don't turn on debug logging level by default, so you would skip the message.
Users can turn off warnings if it's too annoying, I mean that's what warnings are for: They notify you that something happened, but it's not something crucial and operation *should* continue as expected.
To minimize the appearance of those messages, we should implement the new API (which does not exist yet...) as soon as possible.

sils:
@srisankethu didn't @jayvdb want you to use the warnings module?

srisankethu:
You have suggested [this](https://github.com/coala/cEPs/wiki/Deprecation-framework-(draft)), but ``prerequisite_check_command`` has not been deprecated yet. So, I think it is better to us ``PendingDeprecationWarning`` instead ``DeprecationWarning``.

Makman2:
Actually the warning should appear here: https://github.com/coala/coala/pull/2804/commits/c4937e5dbcf4283e3335be12a21a28f9b327b1c5#diff-6dee3b7ded9b28fd0db83b2e7d039b86R99

Makman2:
@srisankethu btw you are using `warnings` instead of `logging`^^

srisankethu:
@Makman2 Yeah I am using ``warnings`` -> https://github.com/coala/coala/pull/2804#issuecomment-260110433
I think there is no use of printing a warning when the prerequisites fail. What do you say? -> https://github.com/coala/coala/pull/2804#issuecomment-260124656

Makman2:
I think stuff gets complicated if we start to mix the warnings and logging module. I would just log a message with warning-severity.

srisankethu:
@Makman2  @jayvb can reason it. He has suggested on using warnings.

JayVB:
Why did you include me again in this thread!?!?!

Aaargh.  Are you trying to be rude?!?

Sent from my iPhone

On Nov 12, 2016, at 3:10 PM, Sri Sanketh Uppalapati <notifications@github.com<mailto:notifications@github.com>> wrote:


@Makman2<https://github.com/Makman2> @jayvb<https://github.com/jayvb> can reason it. He has suggested on using warnings.

Makman2:
Sorry again this is autocompletion I assume^^ @srisankethu please be careful when mentioning jayvdb, don't pick JayVB from autocompletion^^

Makman2:
@jayvdb but yes please reason it :D :P

srisankethu:
I am using my mobile. I am really sorry! This is weird.

JayVB:
Not sure what you mean by "please reason it"... there doesn't seem to be any "reason" to it... it's just a major nuisance to get tons of e-mails that I didn't ask for and that have nothing to do with me.
________________________________

Makman2:
@JayVB "please reason it" wasn't addressed to you^^ Please just hit the unsubscribe button, so you don't get any more mails, and we don't try to mention you. But sorry, it happens that people choose similar names on GitHub and they get addressed accidentally. If you don't use GitHub anyway or don't read its e-mails, turn off e-mail notifications inside the settings.

JayVB:
I'd rather you guys just keep me off of your threads.  I wish I could block you all... this is a major nuisance.  I did take myself off the thread, but you keep including me.


________________________________

sils:
@srisankethu please be more careful

AbdealiJK:
My understanding is that warning should be used when warning about some thing that can be suppressed by later code or other APIs. Also warnings are printed irrespective of the log level. Logging is just a log with various levels hence if a user only asks for fatal logs a Logging.warning() is hidden.

Logging seems to be the use case here.

srisankethu:
@AbdealiJK Earlier I used ``logging`` as @Makman2 suggested and now I am using ``warnings`` @jayvdb has suggested. I do not want to sway a swing here :) @jayvdb Can you explain your viewpoint so that we can get to a final decision? I would refer ``logging`` -> [When to use logging](https://docs.python.org/2/howto/logging.html#when-to-use-logging)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -196,6 +197,11 @@ def check_prerequisites(cls):
                         check_call(options["prerequisite_check_command"],
                                    stdout=DEVNULL,
                                    stderr=DEVNULL)
+                        warnings.simplefilter("default")
```
Makman2: I dislike the fact you need to turn on these warnings manually to see them (then we can also put them into the logging debug-channel btw, no need to utilize another module). The idea is actually that the developer can see what's wrong/not as it should be, and this should be fixed then as soon as possible. Especially for this kind of warning which is easy to fix for a developer I would still like to see a `logging.warn(...)`.
\+ users of coala are developers, though not explicitly for coala^^

jayvdb: coala is used in CI. putting lots of deprecation messages into other ppls CI logs is counter productive. They want clean logs.

Makman2: This doesn't lacks the fact that we need a warning that can be seen, otherwise nobody recognizes and as soon as we remove the deprecated feature, things are going to fail. Better a warning than an error long-term.

Makman2:
@srisankethu I would suggest you still rather use `logging` as we don't have a nice system for `warnings` yet and no-one would see the pending-deprecation-warning.
Anyway before you deprecate, you still need to write the prerequisite-functions mentioned in the issue, so one can do
```
check_prerequisites = simple_prerequisite_check("executable", "arg1")
```
Refer to bears how the check for prerequisites and develop functions accordingly.

We can't deprecate something if there isn't an alternative, otherwise the user has no choice to avoid the warning :)

@jayvdb I get your point on using `warnings`, but to get the PR done, let's merge it with `logging` (we currently handle deprecations with log-warnings currently), and open an issue to discuss the usage of `warnings` in general with respect to this usecase.

Nosferatul:
closing due to high inactivity

Issue #2798, opened 2016-09-28
Remove period from the settings print


Adrianzatreanu:
Right now if you execute SpaceConsistencyBear you will see this:

`Please enter a value for the setting "use_spaces" (True if spaces are to be used instead of tabs.) needed by SpaceConsistencyBear:`

i honestly think that the period afteR "tabs" is not fit there and should be removed.

bctaucsd:
Would love to give this a shot!

Adrianzatreanu:
Hi! Welcome to our community and thank you for your interest!

I have sent you an invitation to join our community. This is needed so we can coordinate tasks with you and make sure no two persons are working on the same thing. If you can't find the mail, you should be able to join us at http://github.com/coala-analyzer/!

We'd love to be able to help you when problems occur. Just join us on [our gitter channel](https://coala.io/chat) and tell us about them :)

raul1991:
assign it me @Adrianzatreanu

sils1297:
Unassigning you @raul1991 as you're already assigned to another newcomer issue - we only have a limited number of those available. After doing a newcomer issue, please advance to low! Newcomer issues are a hell lot of work for us.

kshitij10496:
Hello,
I would like to work on this.

sils1297:
Hi! Welcome to our community and thank you for your interest!

I have sent you an invitation to join our community. This is needed so we can coordinate tasks with you and make sure no two persons are working on the same thing. If you can't find the mail, you should be able to join us at http://github.com/coala-analyzer/!

We'd love to be able to help you when problems occur. Just join us on [our gitter channel](https://coala.io/chat) and tell us about them :)

kshitij10496:
@sils1297 Thanks for the invitation.
I'm raring to go :smile:

sils1297:
:+1: let us know if you need anything :)

nikssardana:
Hi! I am new to open source and would like to contribute to this community. I've locally made changes as required, but can't create a pull request. Please help!

sils1297:
Hi @nikssardana ! Welcome to our community and thank you for your interest!

I have sent you an invitation to join our community. This is needed so we can coordinate tasks with you and make sure no two persons are working on the same thing. If you can't find the mail, you should be able to join us at http://github.com/coala-analyzer/!

> Hint: if you're proud to be part of the team, you can make your membership public at https://github.com/orgs/coala/people

[...]

PR #2819, opened 2016-10-03
removed period from the settings print

nikssardana:
Hi! I just saw the beginner issue #2798 and applied this patch. Below is an image of the applied patch:

screenshot from 2016-10-03 14-07-39


sils1297:
Hey please ask for an assignment first, so situations like this don't happen. I have to close this PR because someone else is already assigned :/ Ping us on gitter if there aren't able to find another newcomer issue.

nikssardana:
Okay no problem!

Issue #2826, opened 2016-10-05
Use Circle CI trusty image


jayvdb:
In August 2016, trusty became the default image for new projects.
https://circleci.com/blog/ubuntu-14-04-trusty-is-the-default-os-for-new-projects/
Anyone creating their circleci project now will have a trusty image now.  Unless the main project build switches to trusty images, it is very likely that a new developer will hit red builds due to complex dependency differences that do not occur on the main project build.

jayvdb:
Note that the `coala` builds on Trusty fail on Julia Lint, same as https://github.com/coala/coala-bears/issues/859

PR #2827, opened 2016-10-05
Circle CI: Avoid staticfloat's julia PPA on trusty

jayvdb:
Fixes https://github.com/coala/coala/issues/2826
Fixes https://github.com/coala/coala-bears/issues/859

### DIFF
--- a/.misc/deps.sh
+++ b/.misc/deps.sh
@@ -9,8 +9,10 @@ case $CIRCLE_NODE_INDEX in
 esac
 
 # apt-get commands
-sudo add-apt-repository -y ppa:staticfloat/juliareleases
-sudo add-apt-repository -y ppa:staticfloat/julia-deps
+if [ "$CIRCLE_BUILD_IMAGE" = "ubuntu-12.04" ]; then
+  sudo add-apt-repository -y ppa:staticfloat/juliareleases
+  sudo add-apt-repository -y ppa:staticfloat/julia-deps
+fi
 sudo apt-get update
 deps="indent libclang1-3.4 r-base"
 deps_python_dbus="libdbus-glib-1-dev libdbus-1-dev"


jayvdb:
I am not sure what dependency from `ppa:staticfloat` julia is causing the problem on Trusty, so it is a bit difficulty to fill in the commit message with something meaningful, but I will improve it.  However, gotta run atm.

jayvdb:
Note this is sufficient to build on precise and trusty. But Im rebuilding on both without cache before getting too excited.

jayvdb:
Rebuilds without cache on precise and trusty are green.

sils1297:
ack https://github.com/coala/coala/pull/2827/commits/e1ec25773851184bf43ce3a309eb078f2a2629c5

jayvdb:
Note: I think I have another better solution coming.

sils1297:
@jayvdb any news on the better solution?

jayvdb:
I was hoping https://github.com/coala/coala-bears/issues/869 could be fixed before this `coala` fix, so they can both switch to trusty.
But that may take a few more days, so this can be merged if you want trusty `coala` builds to work now.
I haven't looked at the `coala` trusty build elapsed times, so switching the main repo builds to trusty may not be wise.

jayvdb:
I was hoping https://github.com/coala/coala-bears/issues/869 could be fixed before this `coala` fix, so they can both switch to trusty.
But that may take a few more days, so this can be merged if you want trusty `coala` builds to work now.
(Feel free to rebase it if you need this urgently; im limited avail. for next two days.)
I haven't looked at the `coala` trusty build elapsed times, so switching the main repo builds to trusty may not be wise.

sils:
argh the regex triggers because there's no space after the colon which is a common mistakes from newcomers.

jayvdb:
no worries; im revising it anyway.

jayvdb:
@sils , hmm.  hit the regex again.  This time, I think the regex that needs fixing.

sils:
Just rephrase it to `staticfloat ppa` instead of doing it with the colon or change the regex. Important: make a commit to the coafile *before* the commit that violates the current regex. GitMate checks every commit individually, not only the PR.

sils:
```diff
-shortlog_regex = ([^:]*|\S+: [A-Z0-9*].*)
+shortlog_regex = ([^:]*|[^:]+: [A-Z0-9*].*)
```

should do it. We need to do that on all repos though...

sils:
I added the coafile commit to https://github.com/coala/coala/pull/2873 to speed this up a bit since it seems important and easy

PR #2854, opened 2016-10-09
Include filename in warning

prafful002:


coalib/bears/Bear.py
```diff
@@ -249,8 +249,8 @@ def execute(self, *args, **kwargs):
             result = self.run_bear_from_section(args, kwargs)
             return [] if result is None else list(result)
         except:
-            self.warn("Bear {} failed to run. Take a look at debug messages"
-                      " (`-V`) for further information.".format(name))
+            self.warn("Bear {name} failed to run on file {filename}. Take a look at debug messages"
+                      " (`-V`) for further information.".format(name=name, filename=args[0]))
```
sils: args[0] isn't guaranteed to be the filename

sils:
Hey, you need to fix the gitmate issues and to do a rebase, here you have a tutorial for the latter:


Hi,

it looks like your PR is out of date and needs a rebase. [This page](http://coala.io/git) may help you to get started on this. (*Do not press the update branch button. It's evil.*)

Add the original coala repository as a remote:

```
git remote add upstream https://github.com/coala-analyzer/coala.git
```

(Be sure to use coala-bears on the coala-bears repository :))

Then go to your branch and rebase it onto our master:

```
git checkout <yourbranch>
git fetch upstream
git rebase --interactive upstream/master
```

Now push it up. You will need to do a force push because you change history here (do not do this on master branches :)):

```
git push --force
```

If you're just looking to fix an issue very quickly and not interested in contributing to coala long term, we can fix it up for you - just tell us :).

Ping us on [our gitter channel](https://coala.io/chat) if you need help!

coalib/bears/Bear.py
prafful002: Should I pass the filename as a separate parameter then ?

sils:
actually in some cases there won't be a filename, I think we might want to
generate this message in an overridable method or something? Take a look at
LocalBear and GlobalBear, I hope that helps understanding what's going on
here.

prafful002:
@sils so the filename is present only when the bear instance being run is that of a ``LocalBear`` and not a ``GlobalBear`` ?

adtac:
wip because of gitmate issues

sils:
@adtac could you not leave discussions unresolved and mark it WIP which means "doesn'tneed maintainer attention"? I think that isn'thelping.

@prafful002 yeah more or less, I think the message should be composed in the LocalBear or GlobalBear classes not in the Bear class.

adtac:
@sils yeah sorry about that

adtac:
This PR's commit history is really messed up. @prafful002 would you prefer to close this PR and create a new one freshly? I mean, you *can* fix up this PR, but that'd take a lot of time for both the reviewers and you (which I frankly don't think is worth when you can learn the proper way from the beginning.)

prafful002:
Ok will do.

Issue #2836, opened 2016-10-06
ConsoleInteraction: Typing Error in docstring


nikssardana:
The first line of the docstring for print_affected_files contains a typing error

![screenshot from 2016-10-06 13-59-38](https://cloud.githubusercontent.com/assets/14257973/19145634/72669e88-8bcd-11e6-9e6c-e4bec4eba41e.png)
:

The line should contain the word "affected" instead of "afected".

vreethesh:
I would like to get assigned to this issue.

deadpyxel:
I'm kinda new on github, does only one person get assigned to each issue? or there can be a multiple number of pull requests related to the same issue? In advance, good project, and keep up the good work.

Makman2:
@deadpyxel yes only one shall work on a single issue to prevent doing the same work (as one PR needs to be abandoned in that case :/ ). This issue is already assigned, but feel free to take another one.
If you want to start with something a bit more challenging, you can also pick a `difficulty/low` issue :)

deadpyxel:
@Makman2 Much thanks, and sorry for screwing up that PR, I wasn't really sure what to do since I never did contribute to other peoples projects, and used Github mostly for my own college stuff. Thanks for the tip.

Makman2:
hehe no problem at all, it was your PR :D
PRs don't harm the repository as long as we don't merge them :)

raul1991:
Duplicate of issue #2808

Makman2:
oh indeed :+1: 
closing the other in favor of this issue as this one is being worked on.

Issue #2847, opened 2016-10-07
PackageRequirement: Add `install_command()`


Makman2:
`install_command` is something that is not necessary bound to a `DistributionRequirement` only. `PackageRequirement` should have it too with a not-implemented version.

sils:
@Makman2 are you working on this already on #2843 ? Is @fneu aware of this? He's doing a refactoring in that area as well. Also please clarify on why this is needed and synch up with each other and @vreethesh .

fneu:
@Makman2 `install_command` might as well go, I removed it locally as cob won't need it and it's hard to do cross-distro.

sils:
@vreethesh plase stop working on this right now. Maybe pick it up later, when discussions are resolved but this stuff is likely to vanish.

vreethesh:
@sils OK

sils:
@Adrianzatreanu can this be continued in package_management or is it already resolved?

jayvdb:
This was completed in https://gitlab.com/coala/package_manager/merge_requests/21

PR #2857, opened 2016-10-11
#2836, #2847

vreethesh:
Fixes issue https://github.com/coala/coala/issues/2836

### DIFF
--- a/coalib/bears/requirements/PackageRequirement.py
+++ b/coalib/bears/requirements/PackageRequirement.py
@@ -37,6 +37,37 @@ class PackageRequirement:
         self.package = package
         self.version = version
 
+    def is_installed():
+        raise NotImplementedError
+
+    def install_command(self):
+        """
+        Creates the installation command for the instance of the class.
+
+        :param return: A string with the installation command. An empty string
+                       if the command could not be supplied.
+        """
+        manager_dict = {'Fedora': 'dnf',
+                        'Ubuntu': 'apt_get',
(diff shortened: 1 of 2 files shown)


sils:
Hey,

we use some bug prediction in coala which relies on the `Fixes` keyword in commit messages. To get good results from that we need to use `Closes` for normal issues instead of `Fixes` which should only be used for real bugs. (See also [the commit message docs](http://coala.io/commit).)

Cheers!

manu-chroma:
Hey @vreethesh, it appears you had opened [another PR](https://github.com/coala/coala/pull/2837) previously to solve this issue. 
Please close that one. 

Also, your commit message still uses ``fixes`` rather than ``closes``. https://github.com/coala/coala/pull/2857#issuecomment-252865399

manu-chroma:
Do away with the body of the commit specifying line number. Seems redundant.

vreethesh:
@manu-chroma Can be reviewed now. All changes you mentioned have been made.

adtac:
@vreethesh it's just `Closes https://github.com/coala/coala/issues/2836`, not `Closes issue https://github.com/coala/coala/issues/2836` if I'm not mistaken. (I generally hate being this pedantic, but the whole point of this Fixes/Closes thing is for a machine to parse so it'd be helpful if they follow a format.)

jayvdb:
Completed in https://github.com/coala/coala/pull/2869

PR #2858, opened 2016-10-11
FunctionMetadata: Store params that are deprecated

adtac:
This is useful in the generation of bear-docs to remove the
deprecated params from listing.

### DIFF
--- a/coalib/settings/FunctionMetadata.py
+++ b/coalib/settings/FunctionMetadata.py
@@ -17,7 +17,8 @@ class FunctionMetadata:
                  retval_desc: str="",
                  non_optional_params: (dict, None)=None,
                  optional_params: (dict, None)=None,
-                 omit: (set, tuple, list, frozenset)=frozenset()):
+                 omit: (set, tuple, list, frozenset)=frozenset(),
+                 deprecated_params: (set, tuple, list, frozenset)=frozenset()):
         """
         Creates the FunctionMetadata object.
 
@@ -34,6 +35,7 @@ class FunctionMetadata:
                                     the default value. To preserve the order,
                                     use OrderedDict.
         :param omit:                A set of parameters to omit.
+        :param deprecared_params:   A list of params that are deprecated.
         """
         if non_optional_params is None:
             non_optional_params = OrderedDict()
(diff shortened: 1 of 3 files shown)


coalib/settings/FunctionMetadata.py
```diff
@@ -101,6 +106,7 @@ def add_alias(self, original, alias):
         :param alias:     The name of the alias for the original.
         :raises KeyError: If the new setting doesn't exist in the metadata.
         """
+        self.deprecated_params.append(alias)
```
sils: you may want to rename this function into add_deprecated_setting or so

sils: also we don't want to have deprecated paras in the optional_params und non_optional_params list do we?

coalib/settings/FunctionMetadata.py
```diff
@@ -279,12 +285,14 @@ def merge(cls, *metadatas):
             (m.retval_desc for m in reversed(metadatas) if m.retval_desc), "")
         merged_non_optional_params = {}
         merged_optional_params = {}
+        merged_deprecated_params = []
```
sils: can we use sets?

coalib/settings/FunctionMetadata.py
```diff
@@ -279,12 +285,14 @@ def merge(cls, *metadatas):
...
 
         for metadata in metadatas:
             # Use the fields and not the properties to get also omitted
             # parameters.
             merged_non_optional_params.update(metadata._non_optional_params)
             merged_optional_params.update(metadata._optional_params)
+            merged_deprecated_params += metadata.deprecated_params
```
sils: `+=` might be undefined if one of them is a list and the other a set, I think update is better

sils:
ack https://github.com/coala/coala/pull/2858/commits/7a91f979533f572a1cb5223e25b0031a3f465cf3

sils:
@rultor merge

sils:
But @adtac can you please change the coala source code to show deprecated params properly when using `--show-bears` etc.

adtac:
@sils oh you merged it, alright I'll do that :+1:

PR #2870, opened 2016-10-17
WIP: Use builtin logging

sils:
CC @Uran198 and @AbdealiJK - tests will fail for now and not everything works yet but I think this is how it could look like.


Uran198:
I think we might also want to change how we are testing that the messages were logged. Right now we are substituting stdout, but I think a better way to do it, might be to substitute logger's handler with the one that will be writing into the stream that we might check later, or into a queue.

With these changes it might be easier to proceed with the complete switch.

sils:
CC @tulphoon this was one of my earlier attempts of what you're doing, I'm closing this now assuming it won't be needed anymore

Issue #2533, opened 2016-07-23
Add `noqa` to the keywords to start ignoring bears


adtac:
Some other linters use that so this will make coala backwards compatible

sils1297:
Nice idea, I like that - could be even simple to do.

sils1297:
that sounds like something simple, can we have it for this release?

RohanVB:
id like to try this one

hypothesist:
@RohanVB assigned you :+1:

yash-nisar:
Are you still working on this @RohanVB ?

RohanVB:
Yes, I am. Got delayed, sorry.

jayvdb:
the flake8 series of linters use this, but it must appear:

* in a comment at the end of the line to be ignored. This does **not** cover the 'next' line.
* in the *first* comment in the file.  This covers the entire file. (not all flake8 series linters support this)

pylint uses a different syntax.

What other linters use this?

RohanVB:
- > in a comment at the end of the line to be ignored. This does not cover the 'next' line.
Currently, this is what `noqa` does. Also, if you specify a bear, `noqa` would run only for that specific one. 

I see CoffeeLint uses this as : 
> You can also use # noqa as a shortcut for # coffeelint: disable-line

PR #2877, opened 2016-10-20
Add `noqa` to the keywords to start ignoring bears

RohanVB:
Previously 'ignore' was the only keyword being used
to ignore bears. 'noqa' performs the same function
as 'ignore'.This will make coala backwards compatible.

Closes https://github.com/coala/coala/issues/2533

### DIFF
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -437,13 +437,18 @@ def yield_ignore_ranges(file_dict):
         stop_ignoring = False
         for line_number, line in enumerate(file, start=1):
             # Before lowering all lines ever read, first look for the biggest
-            # common substring, case sensitive: I*gnor*e, start i*gnor*ing.
-            if 'gnor' in line:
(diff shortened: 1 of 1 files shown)


coalib/processes/Processing.py
```diff
@@ -461,6 +461,31 @@ def yield_ignore_ranges(file_dict):
                                line_number, 1,
                                end_line, len(file[end_line - 1])))
 
+            # Before lowering all lines ever read, first look for the biggest
+            # common substring, case sensitive: N*oqa, start n*oqa.
```
sils: hey, you copied the code above right? That's code duplication.That's usually not good because if we fund a bug in the code above it'll probably be forgotten to fix in your new code. Usually it's way easier to maintain if you don't duplicate chunks of code. Rather put them in a function, give it a variable or make a loop or something. I think we can be smart about it and make this without adding any lines of code by modifying the conditions

RohanVB: Alright. Will do. Is that why codecov was failing?

PR #2892, opened 2016-10-24
Globbing: fix copy-paste bug

bryongloden:
:bug: [hacktoberfest](https://hacktoberfest.digitalocean.com/)

Greetings,

`pattern` in `yield pattern` looks like a copy-paste defect. The original copy of `yield pattern` is on line [247](https://github.com/coala/coala/blob/master/coalib/parsing/Globbing.py#L247). Should the second instance of `yield pattern` on line [251](https://github.com/coala/coala/blob/master/coalib/parsing/Globbing.py#L251) say `yield dirname` instead?

Signed-off-by: Bryon Gloden, CISSP® cissp@bryongloden.com

### DIFF
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -248,7 +248,7 @@ def _absolute_flat_glob(pattern):
     else:
         # Patterns ending with a slash should match only directories
         if os.path.isdir(dirname):
-            yield pattern
+            yield dirname
     return
 
 


sils:
Hm, the commit message needs work for sure, I'll check if this is a bug indeed, though the test failure seems to indicate that this is probably wanted and corrected there.

FWIW we can do stuff like fixing up the commit message etc. if you're not interested in long term contributing, we're a bit strict there.

sils:
CC @fneu who wrote this code. The description of the PR should be in the commit message so that we can see what's going on, it's very helpful.

@fneu this looks suspicious indeed. [this line](https://github.com/coala/coala/blob/master/coalib/parsing/Globbing.py#L252) should be removed and the others should be summarized or corrected. I think @bryongloden s commit is right but the tests need adaption. Can you check this please?

fneu:
I am currently not at home, so it would be nice if somebody with an interpreter and debugger could correct me here if I am wrong.

At the top of the function, we say `dirname, basename = os.path.split(pattern)`, which leaves `basename` empty if and only if the pattern ends with a slash. 

The whole thing should evaluate like this:

```python
>>> import os
>>> pattern = "/some/dir/"
>>> dirname, basename = os.path.split(pattern)
>>> basename
''
>>> dirname
'/some/dir'
>>> pattern
'/some/dir/'
""
```

So `pattern` still contains the trailing slash while `dirname` doesn't. We yield pattern here and keep the slash to mirror the behavior of python's builtin `glob` module:

```python
>>> import glob
>>> glob.glob("dir/")
['dir/']
>>> glob.glob("dir")
['dir']
```

I'm confident that this is the correct behaviour. If the whole thing looks weird to others, we could maybe check the line above from `if os.path.isdir(dirname):` to `if os.path.isdir(pattern):`, the trailing slash shouldn't make a difference here.


-------

Concerning the return statement, be have a glorious battle between those two lines of the Zen of Python:

> simple is better than complex

and

> explicit is better than implicit

It is fact that this function does return None for some inputs (pattern does not describe an existing file or directory name) and thus the return statement is explicitly given here. We can aswell put a `return None` there or leave the line out completely, I don't care.

sils:
If that is true the condition should be refactored to `if (... and ...) or ...:` because essentially we only have one branch.

fneu:
codestyle is hard.

I like the more verbose version here and find it more readable but apparently it's confusing to some.

We can still argue afterwards for whether or not there should be a return afterwards.

I'm proposing to use this as a source of infinite newcomer issues ("remove the return statement"/"add a return statement for readability")

Anyway, I'm closing this particular pull request because the proposed change in behavior is not wanted.

PR #2893, opened 2016-10-24
DocumentationComment: fix invalid operations

bryongloden:
:bug: label: [hacktoberfest](https://hacktoberfest.digitalocean.com/)

Greetings,

It's always good to define string variables, at least to "", because if we try to retrieve its value before it gets assigned any actual (non-garbage) value, then it results in an exception such as a TypeError, e.g. line numbers `124` & `129` of [DocumentationComment.py](https://github.com/coala/coala/blob/master/coalib/bearlib/languages/documentation/DocumentationComment.py#L124).

Signed-off-by: Bryon Gloden, CISSP® cissp@bryongloden.com

### DIFF
--- a/coalib/bearlib/languages/documentation/DocumentationComment.py
+++ b/coalib/bearlib/languages/documentation/DocumentationComment.py
@@ -95,7 +95,8 @@ class DocumentationComment:
         parse_mode = self.Description
 
         cur_param = ""
-
+        retval_desc = ""
+        param_desc = ""
         desc = ""
         parsed = []
 


sils:
Hm, I'm not fully sure about this. You can easily see that in this state machine the variable will never be used before it's definition. But well - this is some code analysis that most tools cannot perform.

sils:
That being said I can also see the other POV...

sils:
As for the commit message: the body lines shouldn't be longer than 74 chars because they'll not be shown properly in git otherwise. The first line could be better, the GitMate fail here is because the first char after colon isn't upper case as required.

sils:
I asked for opinions on this on the chat, see :point_up: [October 24, 2016 11:44 PM](https://gitter.im/coala/coala?at=580e80a79aea834129ec055f)

Nosferatul:
@sils https://bytes.com/topic/python/answers/165539-no-variable-argument-declarations-necessary. I believe it's not fully necessary in our case. It was a problem if we used that variable without having a value.

Nosferatul:
Btw thank you for your concern @bryongloden  👍

sils:
@bryongloden are you aware that in any path through the data flow graph both variables get assigned first? I.e. there is no way that this variable is misused unless there is a bug. And according to the Zen of Python "Errors should never pass silently", i.e. the exception thrown in this case is exactly what we want to happen.

sils:
@bryongloden I've made up my mind. There should be no path (and I think you can mathematically prove that without too much effort), if there is there *should* be an exception because it shouldn't pass silently. Your patch would only make it pass silently.

What tool did you use? Most tools aren't that sophisticated. Thing is, such analysis of any turing complete language is mathematically an unsolvable problem.

sils:
Closing this then.

PR #2899, opened 2016-10-26
bearlib: Add aspects

adtac:


coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+
+    class NoDefault:
+        pass
+
+    @enforce_signature
+    def __init__(self,
+                 name='',
+                 description='',
+                 cast_type=str,
+                 default=NoDefault):
```
sils: docstring missing

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+
+class Aspect:
+    '''
+    This is the ``Aspect`` base class. Each aspect may have sub-aspects,
+    which are ``Aspect`` objects themselves. An aspect also contains
+    documentation, instantiated as an ``AspectDocumentation`` object.
+
+    There is a root ``Aspect`` object in which every other aspect lives.
+
+    >>> Aspects = Aspect("Aspects")
```
sils: root should be creatable without arguments IMO

adtac: Fixed both comments.

Udayan12167:
@adtac why is it not in separate files?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+documentation.
+"""
+
+from inspect import cleandoc
+
+from coala_utils.decorators import classproperty
+from coala_utils.decorators import enforce_signature
+
+
+def generate_consistency_check(*members):
```
Udayan12167: Should be a method in the coala_utils.Decorators.

Udayan12167:
This seems like a fairly complex enough module to have a cEP for it.

adtac:
@Udayan12167 it does: https://github.com/coala/cEPs/blob/master/cEP-0005.md

Also `from coalib.bearlib.aspects import Aspect, AspectDocumentation, AspectSetting` makes sense IMO, hence the single file. Plus most of it is just docs.

Udayan12167:
Also all testing is covered with Doctests?

sils:
apparently, after all coverage *is* full

adtac:
@Udayan12167 yep

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+
+    >>> Aspects = Aspect()
+
+    This is a placeholder aspect through which every other aspect can be
+    accessed.
+
+    To create new aspect, simply call the ``new_subaspect`` method with
+    the subaspect's name as a param:
+
+    >>> Aspects.new_subaspect("Redundancy")
```
Udayan12167: There is no point creating a subaspect without documentation right? Shouldn't that be enforced somehow.

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+    40
+
+    The user can use the qualifying name according to PEP 3155 for selecting
+    and unselecting aspects (see <https://www.python.org/dev/peps/pep-3155/>):
+
+    >>> Aspects.Redundancy.Clone.__qualname__
+    'Redundancy.Clone'
+    '''
+
+    def __init__(self, name="Aspects", parent=None):
```
sils: IMO the default name should be empty

sils: or `root` or something

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+        Returns the sub-aspect.
+
+        :param subaspect:
+            Name of the sub-aspect. This should have been instantiated
+            previously using the ``new_subaspect`` method.
+        :return:
+            An ``Aspect`` object corresponding to the sub-aspect. If the
+            sub-aspect is not found, ``None`` is returned.
+        """
+        if subaspect in self.__dict__["subaspects"]:
```
sils: you can just `self.subaspects` here, `__getattr__` is only called if `__getattribute__` fails

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+            Name of the aspect.
+        :param parent:
+            A reference to the parent object for this aspect. For first-level
+            aspects, this is the root aspect.
+        """
+        self.__name__ = name
+        self.docs = AspectDocumentation("", "", "", "")
+        self.subaspects = {}
+        self.parent = parent
+        self.__settings__ = []
```
sils: why do you name it like a builtin thing?

adtac: `_settings`?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+
+        current = self
+        quals = []
+
+        while current is not None and current.__name__ != "Aspects":
+            quals.append(current.__name__)
+            current = current.parent
```
sils: why not take `self.parent.__qualname__` and append `self.__name__`?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+        result = []
+
+        # Recursively go up and collect settings (the most local setting
+        # overrides previous values)
+        while current is not None:
+            for setting in current.__settings__:
+                if setting.name not in taken_settings:
+                    taken_settings.add(setting.name)
+                    result.append(setting)
+            current = current.parent
```
sils: why not simply go recursive, `self.parent.settings` and add `self`s settings on top?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+    ...     """, '...', '...', '...'
+    ... )
+
+    To create sub-aspects, simply call ``new_subaspect``:
+
+    >>> Aspects.Redundancy.new_subaspect("Clone")
+
+    Trying to access an uninitialized aspect will return ``None``:
+
+    >>> Aspects.UnknownAspect is None
```
sils: rather raise a NameError here right?

coalib/bearlib/aspects/__init__.py
sils: yep

coalib/bearlib/aspects/__init__.py
sils: I'd add settings here too so I can construct it right away with settings

coalib/bearlib/aspects/__init__.py
adtac: Because we want the closest settings to override higher level settings. IIUC with what you suggest, we would have multiple copies for the same settings (with possibly different values).

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+        :param description:
+            A description of what the setting is about and what it does.
+        :param cast_type:
+            The type to cast the user value into. For example, ``int`` will
+            convert and setting value as an integer.
+        :param default:
+            Default value for the setting.
+        """
+        self.name = name
+        self.description = description
```
sils: can we have exampe values or something? Maybe an explanation on what values this can have or so.

sils: I'd love to have more than just a description to be sure to get better qualitydocs for the settings

sils: also, do a cleandoc on strings like the description so I can make multiline strings with indentation easily :)

adtac: There's the example with `Root.Redundancy.Clone` - we create a setting called `max_clone_tokens`. Do you mean like a doctest for `AspectSetting`?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+        :param description:
+            A description of what the setting is about and what it does.
+        :param cast_type:
+            The type to cast the user value into. For example, ``int`` will
+            convert and setting value as an integer.
+        :param default:
+            Default value for the setting. (Optional)
+        :param example_values:
+            A tuple of example values that help the user understand what
+            kind of values this setting takes. (Optional)
```
sils: let's make this not optional please, we want high quality docs and that's not really an optional thing for aspects. Also put them in the right order

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+class AspectSetting:
+    """
+    This class contains the specification about a setting that should be given
+    for an aspect.
+    """
+
+    class NoDefault:
+        pass
+
+    @enforce_signature
```
sils: the decorator is useless if you don't have any annotations

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+    ...     )
+    ... )
+
+    This aspect is now accesible at ``Root.Redundancy``. Simply call
+    ``new_subaspect`` to keep creating new sub-aspects:
+
+    >>> Root.Redundancy.new_subaspect(
+    ...     "Clone",
+    ...     doc=AspectDocumentation(
+    ...         "This aspect detects clones in your source code...",
```
sils: an aspect doesn't detect anything! We have to be careful to demostrate a good style here because people will pick that up for a *hell lot of docs* for a *hell lot of aspects*

coalib/bearlib/aspects/__init__.py
sils: i.e. add the param to the consistency check as well

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+
+    >>> Root.Redundancy.Clone.settings = [
+    ...     AspectSetting("min_clone_token",
+    ...                   "The number of tokens that have to be "
+    ...                   "equal for it to be detected as a code clone",
+    ...                   int, default=20)
+    ... ]
+
+    ``settings`` can be defined for any aspect by creating a list of
+    ``AspectSetting`` objects with 3 params: the setting name,
```
sils: can you somehow like to the AspectSetting doc?

sils: because what you're describing here should go there

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+    fourth optional param - the default value for the setting.
+
+    All bears implementing a given aspect should implement all the settings
+    given and documented by it. Additional bear specific settings might be
+    added to allow further configuration.
+
+    You can also give the settings at object instantiation using the
+    ``settings`` param. Each setting can also contain a param called
+    ``example_values`` - these are helpful for the user to understand
+    what the setting does and what values it may take:
```
sils: same here, this doc doesn't belong here, rather link it

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+    ...                   int, default=40)
+    ... ]
+
+    >>> Root.Redundancy.Clone.settings[0].default
+    30
+    >>> Root.Redundancy.Clone.Logic.settings[0].default
+    40
+
+    The user can use the qualifying name according to PEP 3155 for selecting
+    and unselecting aspects (see <https://www.python.org/dev/peps/pep-3155/>):
```
sils: you should note that you're treating the name "Root" specially!

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+
+    class NoDefault:
+        pass
+
+    @enforce_signature
+    def __init__(self,
+                 name: str='',
+                 description: str='',
+                 cast_type: type=str,
```
sils: uhm this might also be a function, right? Is a function `type`? :/ I'd rather not typecheck this.

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+
+    ``settings`` can be defined for any aspect by creating a list of
+    ``AspectSetting`` objects with 5 params:
+        * the setting name
+        * a documentation string
+        * a cast type (in this case ``int`` will be applied to the given
+          setting)
+        * a tuple of example values that the setting might take
+        * (optional) the default value for the setting.
+    Please refer to the docs for the ``AspectSetting`` class to learn more.
```
sils: can't you make a sphinx link?

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+
+    >>> Root.Redundancy.Clone.settings = [
+    ...     AspectSetting("min_clone_token",
+    ...                   "The number of tokens that have to be "
+    ...                   "equal for it to be detected as a code clone",
+    ...                   int, example_values=(20, 40, 60), default=20)
+    ... ]
+
+    ``settings`` can be defined for any aspect by creating a list of
+    ``AspectSetting`` objects with 5 params:
```
sils: I'm pretty sure the builds will fail if you don't put proper newlines before and after the enumeration

coalib/bearlib/aspects/__init__.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+@generate_consistency_check('definition', 'cause', 'example',
+                            'example_language', 'importance_reason',
+                            'fix_suggestions')
+class AspectDocumentation:
+    """
+    This class contains documentation about an aspect described by the Aspect
+    class.
+    The documentation is consistent if all members are given:
```
sils1297: Shouldn't you have a newline her?

coalib/bearlib/aspects/__init__.py
sils1297: Actually you can use those for quickstart, maybe rename it to suggested values instead of example?

coalib/bearlib/aspects/__init__.py
adtac: Who is this @sils1297 ?! :P

sils:
ack 1e0bd145b675a551e087b1da9d928c5c779623fa

sils:
@rultor merge

Issue #2954, opened 2016-11-04
bearlib: Change `min_line_length` to `max_line_length`


Asnelchristian:
Check line ``184``

sils:
which file!?

badarsh2:
I found a usage of min_line_length in __init.py__

` ...         AspectSetting(
  ...             "min_line_length",
  ...             "Maximum length allowed for a line.",`

If it is so, can I fix it? 
P.S. I'm a newbie here :)

Makman2:
yeah seems this is the right one :)
@badarsh2 I've invited you to our org, be sure to check out our newcomer guides and feel free to talk to us at our gitter channel :)

As soon as you've accepted invitation, I'll assign you and you can start working :+1:

badarsh2:
Done!

PR #2960, opened 2016-11-05
Aspect: Change min_line_length to max_line_length

badarsh2:
Fixes a typo in **bearlib/aspects/__init__.py**
Line `183`: Changes `min_line_length` to `max_line_length`

Closes [https://github.com/coala/coala/issues/2954](https://github.com/coala/coala/issues/2954)


sils:
Hi, this is good!

It seems that your commit message doesn't contain enough information. Don't worry, most people don't know about good commit messages when they join us, including very experienced developers :) Standards vary from project to project. Especially it's missing a `Closes <full issue URL>` tag.

The most important part is the first line of your commit message - maintainers will use it to read through a large number of commits and thus it's important that it's short but most importantly concise. I'd heartily recommend reading <http://coala.io/commit>! In your case I'd prefix `Aspect:` so people know ths change is about the aspect class.

Cheers!

PR #2968, opened 2016-11-06
WIP: aspectclasses

userzimmermann:
As discussed with Lasse @ PyCon.DE in Munich:

Just an idea to define aspects in a nicer way using classes and ``@AspectClass.subaspect`` decorators, with some metaclass magic doing the job in the background.  Doc strings and included ``Root.Metadata...`` aspectclass definitions show how it works...

### Reviewers

@sils


userzimmermann:
Closing to reopen in a clean state

PR #3019, opened 2016-11-17
Sils/releasenotes

sils:


coalib/results/Diff.py
```diff
@@ -381,13 +381,52 @@ def add_lines(self, line_nr_before, lines):
...
         if linediff.change is not False and linediff.change[1] != replacement:
-            raise ConflictError("An already changed line cannot be changed.")
+            orig_diff = Diff.from_string_arrays(linediff.change[0],
+                                                linediff.change[1])
+            new_diff = Diff.from_string_arrays(linediff.change[0],
+                                               replacement)
+            try:
+                replacement = ''.join((orig_diff + new_diff).modified)
+            except:
+                raise ConflictError("The given changes aren't compatible.")
```
sils: then you'd also have to give file because otherwise it's meaningless, usually exceptions don't do that. I can't think of any *concrete* case where it'd be useful, i.e. the line is always known to the user when he gets to see this anyway/

RELEASE_NOTES.rst
```diff
@@ -1,3 +1,52 @@
+coala 0.9.0
+===========
+
+Command Line Interface Changes:
+
+- Users can press enter to dismiss a result by default.
+- A lot more strings, like ``roger`` or ``no way`` are allowed for boolean
```
fneu: maybe sort these by severity or something? This should move WAY to the bottom or nobody is taking the list serious from here on.

RELEASE_NOTES.rst
```diff
@@ -1,3 +1,52 @@
...
+  values. (https://github.com/coala/coala/commit/728b7b02da8ca8f91b91c67784872244c0820a77)
+- The section name is now displayed when asking the user for missing settings.
+- Minor performance improvements could be implemented.
```
fneu: sounds unimportant. Maybe "~~Minor~~ Some performance improvements could be ~~implemented~~ achieved"

coalib/results/Diff.py
fneu: I don't like the message much, it could explicitly mention a merge conflict or something, or be more verbose and tell you what actually went wrong. Are "given changes" and "compatible" obvious in this case?

coalib/results/Diff.py
```diff
@@ -381,13 +381,52 @@ def add_lines(self, line_nr_before, lines):
...
+        >>> diff.change_line(1,
+        ...                  'Hey there! Gorgeous.\n',
+        ...                  'Hey there! This is sad.\n')
+        >>> diff.modified
+        ['Hey there! This is sad.\n', "It's nice that we're here.\n"]
+
+        We can even merge changes within one line:
+
+        >>> diff.change_line(1,
+        ...                  'Hey there! Gorgeous.\n',
```
fneu: Giving the (now) WRONG first line as an argument looks so wrong and non-intuitive :( Looks like it will lead to bugs. Can I just pass some random stuff as the original text to get across what my change is?

I get that this is being done to make the diffs mergeable but it comes across really weird.

coalib/results/Diff.py
Makman2: But it's useful for the developer I think. Especially if the diff is bigger and because we now support in-line patching, this makes totally sense to include some additional information (and yes the filename should be there too then :) ). From a coala-user's standpoint we don't display the exception message anyway directly.

coalib/results/Diff.py
sils: actually *if* the lines have changed before and the new line is incompatible I could also pass: `Hey there! stupidshit.\n` and `Hello. :( stupidshit.\n` but I think if I do stupid shit before my change in the line the offset will be wrong.

I don't see this causing bugs, you're always supposed to give the original line. I guess we could fetch this from self.file automatically to be safe... wanna file an issue? That one should be totally easy

coalib/results/Diff.py
sils: actually this will never be shown to a user, if it gets shown it's with a traceback and that's a bug. So this should be a developers message. I'll try rephrasing anyway

RELEASE_NOTES.rst
```diff
@@ -1,3 +1,54 @@
...
+===========
+
+Command Line Interface Changes:
+
+- Users can press enter to dismiss a result by default.
+- Multiple patches within one line, even from different bears, can be
+  automatically merged by coala.
+- The section name is now displayed when asking the user for missing settings.
+- ``coala-ci`` shows results *and* patches by default now.
+- ``coala-dbus`` has been removed.
```
Adrianzatreanu: explain why at least :p ?

coalib/results/Diff.py
sils: #3024

coalib/results/Diff.py
```diff
@@ -381,13 +381,52 @@ def add_lines(self, line_nr_before, lines):
...
+        ...                  'Hello. :( Gorgeous.\n')
+        >>> diff.modified
+        ['Hello. :( This is sad.\n', "It's nice that we're here.\n"]
+
+        However, if we change something that has been changed before, we'll get
+        a conflict:
+
+
```
Makman2: obsolete newline

Makman2:
ack fba3820, 8b817e0

sils:
reack https://github.com/coala/coala/pull/3019/commits/083424e715c6984a6225026455f30dc9998bad9d

Makman2:
ack eded826

sils:
@rultor merge

Makman2:
@rultor merge

PR #3036, opened 2016-11-19
core: Add FileProxy

Makman2:


sils:
CC @jayvdb

sils:
@jayvdb our stand on this right now: merge this ASAP so we got something that's already better than what we have. This defines a nice API, we tried to leave out some simplifications so the user isn't tempted to do badly performant things, i.e. we removed the `__len__` wrapper etc.

We can change this and add caching and everything underneath later.

sils:
ack a9ccaea0508a346caa9af55caeea2fd9d814fa1b

sils:
unack a9ccaea as per discussion with @jayvdb we'll pause this.

@jayvdb we're working around this, you don't need to stress yourself too much.

Makman2:
Shall we merge this anyway? I kinda want to have it as an initial point for the future `FileProxy`^^

Makman2:
The "factory" term and approach do fit more into the picture. Keeping the branch alive for reference.

PR #3047, opened 2016-11-20
OpenEditorAction: Use `$EDITOR` if set

sils:


coalib/results/result_actions/OpenEditorAction.py
```diff
@@ -61,3 +62,6 @@ def apply(self, result, original_file_dict, file_diff_dict, editor: str):
                     rename=False if original_name == filename else filename)
 
         return file_diff_dict
+
+    if "EDITOR" in environ:
+        apply.__defaults__ = (environ['editor'],)
```
adtac: `EDITOR` not `editor` ;)

underyx: Wouldn't it be better to have a `getenv('EDITOR')` actually added as a default in the argument list? If I'm guessing what this is correctly (I don't even know!) this will change the last argument to optional at runtime, which is pretty confusing.

It would probably also make testing this easier.

adtac:
ack https://github.com/coala/coala/pull/3047/commits/febab039d9e0bf7aa640da2c3419ae0b0dd4bf79

adtac:
reack https://github.com/coala/coala/pull/3047/commits/0edf97a2d9458e93290086e2889054f402bbad65

sils:
@rultor merge

PR #3096, opened 2016-11-22
Sils/ignao

sils:


sils:
no idea what the dependency ci error is about, I disabled the service as it never found us anything and has no docs on if it needs configuration

jayvdb:
ack 41d92f125 a9c70f4d443d

jayvdb:
Re  	a7f4af4 , the old setting `format_str` is being removed.
Did it not work before ?
Is nobody using it?

sils:
HM, it was not documented that this existed, there night be people using
that

jayvdb:
ack a7f4af4021

sils:
@rultor merge

jayvdb:
@sils 

> HM, it was not documented that this existed, there night be people using that

coala-atom is using it, and this is one of the reasons it is broken.  A quick grep shows that `coala-vim` is also broken by this!

sils:
yeah we have a PR to atom already, the other plugins need attention as well indeed.

Issue #3149, opened 2016-12-04
Bear.get_non_optional_settings doesnt check dependencies


jayvdb:
`KeywordBear` depends on `AnnotationBear`, which has a required parameter `language`.

`KeywordBear.get_non_optional_settings()` doesnt include `language`.

```python
>>> bears.general.KeywordBear.KeywordBear.get_non_optional_settings()
OrderedDict()
>>> bears.general.KeywordBear.KeywordBear.get_metadata().non_optional_params
OrderedDict()
```

This appears to be one of the underlying causes of https://github.com/coala/coala-bears/issues/1012, and thus https://github.com/coala/coala/issues/3128

Perhaps `Bear.get_non_optional_settings` should have a parameter `recursive` which walks the dependency tree.

jayvdb:
I've added a workaround in https://github.com/coala/coala-quickstart/pull/40 .

> This appears to be one of the underlying causes of coala/coala-bears#1012, and thus #3128

Not quite correct; it only causes a decent error message, indicating that `language` isnt set, which at least helps people migrating to 0.9 to understand how to fix the problem.

jayvdb:
If `get_non_optional_settings` and `get_optional_settings` recurse, there needs to be a way to find out the non-recursive per-bear settings.

sils:
@Makman2 any progress? I heard you had a solution? I'm not too eager on merging the workaround from @jayvdb if we can have the full fix with close to equal the effort.

Makman2:
yes in fact we could use my `traverse_graph` function which is already fully tested and working on my branch. If you want we can already merge this part of the core, traverse each dependency and merge the parameters:
```
def merge_param(prev_bear, next_bear):
    pass #...

traverse_graph([self], lambda bear: bear.DEPS, merge_param)
```

sils:
we could just add dependencies in get_optional_settings recursively, that'd
be super easy and doesn't rely on that graph algorithm

Makman2:
oh yes that's for sure easier, let's do this 👍

jayvdb:
Oh look, this didnt make it into 0.9.1 ;-(

PR #3151, opened 2016-12-05
SectionFilling: Collect non-optional settings from dependencies

jayvdb:
AnnotationBear has a non-optional setting `language` which is not
being detected by fill_section when determining the non-optional
settings of KeywordBear.

The temporary fix is to look one level down the dependency tree,
and include any non-optional settings found.

Temporary workaround for #3149

### DIFF
--- a/coalib/settings/SectionFilling.py
+++ b/coalib/settings/SectionFilling.py
@@ -71,6 +71,10 @@ def fill_section(section, acquire_settings, log_printer, bears):
     prel_needed_settings = {}
     for bear in bears:
         needed = bear.get_non_optional_settings()
+        # FIXME: See https://github.com/coala/coala/issues/3149
+        for bear_dep in bear.BEAR_DEPS:
+            needed.update(bear_dep.get_non_optional_settings())
+
         for key in needed:
             if key in prel_needed_settings:
                 prel_needed_settings[key].append(bear.name)


jayvdb:
Probably should leave a FIXME in there ;-(

jayvdb:
This patch looks suitable for a hot fix.
`ClangCloneDetectionBear` depends on `ClangFunctionDifferenceBear`, which has no dependencies. So that is only a single level of dependencies.  Also, those deps do not have non-optional settings.
`git grep DEP` to dig into whether this hot fix is good enough until a better fix is ready to be merged.

coalib/settings/SectionFilling.py
```diff
@@ -71,6 +71,10 @@ def fill_section(section, acquire_settings, log_printer, bears):
     prel_needed_settings = {}
     for bear in bears:
         needed = bear.get_non_optional_settings()
+        # FIXME: See https://github.com/coala/coala/issues/3149
+        for bear_dep in bear.BEAR_DEPS:
+            needed.update(bear_dep.get_non_optional_settings())
```
sils: why not do that in get_non_optional_settings and get_optional_settings? Then it goes down recursively

jayvdb: Right.  This is a temporary fix, and should be unnecessary if https://github.com/coala/coala-bears/pull/1100 is an acceptable approach.

https://github.com/coala/coala/issues/3149 needs design.

Mixih:
marking WIP per the issue discussion.

jayvdb:
We have other solutions for this problem close to being ready.

PR #3161, opened 2016-12-05
Language: Change Language.extensions to tuples

jayvdb:
The language definitions for CSS, Java and JavaScript used
a single string value for the extensions attribute, as opposed to
tuples used elsewhere where multiple values may be needed,
including the aliases attribute.

Also add attribute extensions to the C# language definition.

cc @sils

### DIFF
--- a/coalib/bearlib/languages/definitions/CSS.py
+++ b/coalib/bearlib/languages/definitions/CSS.py
@@ -3,7 +3,7 @@ from coalib.bearlib.languages.Language import Language
 
 @Language
 class CSS:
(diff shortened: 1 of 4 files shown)


jayvdb:
The alternative is to check for strings everywhere.

sils:
ack https://github.com/coala/coala/pull/3161/commits/5b2df9e29e326c712980b194741e7a7aadca0f52

sils:
can you make the backport PR?

sils:
@rultor merge

PR #3162, opened 2016-12-06
[WIP] Import Pygments languages

jayvdb:
Main motivation is this may allow the `language` parameter of AnnotationBear to be optional, resolving https://github.com/coala/coala-bears/issues/1012


coalib/bearlib/languages/definitions/ABAP.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class ABAP:
+    extensions = '.abap', '.ABAP'
```
jayvdb: are extensions case-sensitive?

coalib/bearlib/languages/definitions/ActionScript3.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class ActionScript3:
```
jayvdb: Probably should be rolled into ActionScript using versions

coalib/bearlib/languages/definitions/Ada.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Ada:
+    extensions = '.adb', '.ads', '.ada'
```
jayvdb: Sort extensions, or add sensible ordering. S before B.

coalib/bearlib/languages/definitions/ApacheConf.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class ApacheConf:
+    extensions = '.htaccess', 'apache.conf', 'apache2.conf'
```
jayvdb: should we include filenames here? *..conf is going to conflict with many other formats.

coalib/bearlib/languages/definitions/Batchfile.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Batchfile:
+    extensions = '.bat', '.cmd'
```
jayvdb: These are different versions.

coalib/bearlib/languages/definitions/BlitzMax.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class BlitzMax:
```
jayvdb: How different from  BlitzBasic. Maybe pygments delegation can be used as part of heuristics for this.

coalib/bearlib/languages/definitions/COBOL.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class COBOL:
+    extensions = '.cob', '.COB', '.cpy', '.CPY'
```
jayvdb: case here, and the next few

coalib/bearlib/languages/definitions/C__.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class C__:
+    extensions = '.cpp', '.hpp', '.c++', '.h++', '.cc', '.hh', '.cxx', '.hxx', '.C', '.H', '.cp', '.CPP'
```
jayvdb: This class shouldnt have been created.

coalib/bearlib/languages/definitions/Mscgen.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Mscgen:
```
jayvdb: pretty sure this should have capital G, etx.

coalib/bearlib/languages/definitions/ObjectiveC__.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class ObjectiveC__:
```
jayvdb: needs a rename.

coalib/bearlib/languages/definitions/PHP.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class PHP:
+    extensions = '.php', '.inc', '.php3', '.php4', '.php5'
```
jayvdb: Interesting version problem here for sure.

coalib/bearlib/languages/definitions/Perl6.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Perl6:
```
jayvdb: same language?

coalib/bearlib/languages/definitions/PyPyLog.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class PyPyLog:
```
jayvdb: Need to see if this is a useful format for coala . Same with py tracebacks that follow.

coalib/bearlib/languages/definitions/ResourceBundle.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class ResourceBundle:
+    extensions = '.txt',
```
jayvdb: ugh. Not worth having this class for the problems a .txt will bring.

coalib/bearlib/languages/definitions/Rust.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Rust:
+    extensions = '.rs', '.rs.in'
```
jayvdb: .in here, is a different dialect

coalib/bearlib/languages/definitions/SquidConf.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class SquidConf:
```
jayvdb: c.f. apache conf, but much less useful.

coalib/bearlib/languages/definitions/XML.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class XML:
+    extensions = '.xml', '.xsl', '.rss', '.xslt', '.xsd', '.wsdl', '.wsf'
```
jayvdb: IMO this needs to be modelled properly, or drop the laundry list of extensions.

coalib/bearlib/languages/definitions/reg.py
```diff
@@ -0,0 +1,6 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class reg:
```
jayvdb: windows reg?

coalib/bearlib/languages/definitions/XML.py
sils: see http://api.coala.io/en/latest/coalib.bearlib.languages.html#module-coalib.bearlib.languages.Language we can derive XSL etc. from XML

coalib/bearlib/languages/definitions/ABAP.py
sils: good question, if pygments people thing they should maybe we should just go with it

coalib/bearlib/languages/definitions/ApacheConf.py
sils: IMO yes, my general opinion on those kinds of issues is they have probably thought longer on it and have more experience. This seems sensible to me.

coalib/bearlib/languages/definitions/ActionScript3.py
sils: yep, CC @userzimmermann one big omissioln in the Languages is that you cannot *yet* define different settings for different versions although the whole design is obviously built for it. We're not sure yet which syntax to use to make it intuitive.

coalib/bearlib/languages/definitions/ResourceBundle.py
sils: we should have something like

```python
@Language
class PlainText:
    extensions = '.txt'
```

as well IMO and it's not the definitions that should be "aware" of this but the algorithms that operate on them, they should deal with overlaps etc

coalib/bearlib/languages/definitions/ActionScript3.py
userzimmermann: @sils I already have an idea :)

PR #3178, opened 2016-12-07
Ud/sec inherit

Udayan12167:
### Checklist

- [ ] I have rebased properly. Please see [our tutorial on rebasing](http://coala.readthedocs.io/en/latest/Developers/Git_Basics.html#rebasing).
- [ ] I have gone through the [commit guidelines](http://coala.readthedocs.io/en/latest/Developers/Writing_Good_Commits.html) and I've followed them.
- [ ] I have followed the [guidelines for the commit shortlog](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#shortlog).
- [ ] I have followed the [guidelines for the commit body](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#commit-body).
- [ ] I have included the issue URL with the 'Fixes' keyword if the commit fixes a *real bug* and the 'Closes' keyword if it adds a feature or enhancement. For details, check the [issue reference section in our commit guidelines](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#issue-reference).
- [ ] I have [run all the tests](http://coala.readthedocs.io/en/latest/Developers/Executing_Tests.html) and they all pass. If any CI (below) is not green, please fix them. If GitMate issues appear you will have to amend your commits, pushing new commits on top is not sufficient!

### DIFF
--- a/coalib/coala_main.py
+++ b/coalib/coala_main.py
@@ -84,11 +84,11 @@ def run_coala(console_printer=None,
                                   VERSION))
 
         settings_hash = get_settings_hash(sections, targets)
-        flush_cache = bool(sections['default'].get('flush_cache', False) or
+        flush_cache = bool(sections['cli'].get('flush_cache', False) or
                            settings_changed(log_printer, settings_hash))
 
         cache = None
-        if not sections['default'].get('disable_caching', False):
+        if not sections['cli'].get('disable_caching', False):
             cache = FileCache(log_printer, os.getcwd(), flush_cache)
 
         for section_name, section in sections.items():
--- a/coalib/output/ConfWriter.py
+++ b/coalib/output/ConfWriter.py
@@ -17,7 +17,8 @@ class ConfWriter(ClosableObject):
                  key_delimiters=(',', ' '),
                  section_name_surroundings=MappingProxyType({'[': ']'}),
                  section_override_delimiters=('.',),
-                 unsavable_keys=('save',)):
+                 unsavable_keys=('save',),
+                 key_value_append_delimiters=('+=',)):
         ClosableObject.__init__(self)
 
         self.__file_name = file_name
@@ -26,6 +27,7 @@ class ConfWriter(ClosableObject):
 
         self.__file = open(self.__file_name, 'w')
         self.__key_value_delimiters = key_value_delimiters
+        self.__key_value_append_delimiters = key_value_append_delimiters
         self.__comment_separators = comment_separators
         self.__key_delimiters = key_delimiters
         self.__section_name_surroundings = section_name_surroundings
@@ -34,7 +36,8 @@ class ConfWriter(ClosableObject):
         self.__closed = False
 
         self.__key_delimiter = self.__key_delimiters[0]
-        self.__key_value_delimiter = self.__key_value_delimiters[0]
+        self.__key_value_delimiter = key_value_delimiters[0]
+        self.__key_value_append_delimiter = key_value_append_delimiters[0]
         (self.__section_name_surrounding_beg,
          self.__section_name_surrounding_end) = (
             tuple(self.__section_name_surroundings.items())[0])
@@ -70,11 +73,11 @@ class ConfWriter(ClosableObject):
                     keys.append(setting.key)
                 elif ((setting.key not in self.__unsavable_keys) or
                       (not setting.from_cli)):
-                    self.__write_key_val(keys, val)
+                    self.__write_key_val(keys, val, section.defaults)
                     keys = [setting.key]
                     val = str(setting)
         except StopIteration:
-            self.__write_key_val(keys, val)
+            self.__write_key_val(keys, val, section.defaults)
 
     def __write_section_name(self, name):
         assert not self.__closed
@@ -82,7 +85,7 @@ class ConfWriter(ClosableObject):
         self.__file.write(self.__section_name_surrounding_beg + name +
                           self.__section_name_surrounding_end + '\n')
 
-    def __write_key_val(self, keys, val):
+    def __write_key_val(self, keys, val, defaults):
         assert not self.__closed
 
         if keys == []:
@@ -101,8 +104,71 @@ class ConfWriter(ClosableObject):
                 for key in keys]
         val = escape(val, chain(['\\'], self.__comment_separators))
 
-        self.__file.write((self.__key_delimiter + ' ').join(keys) + ' ' +
-                          self.__key_value_delimiter + ' ' + val + '\n')
+        append_keys = []
+        other_keys = []
+        for key in keys:
+            if (defaults and
+                    (key in defaults and
+                     val.startswith(str(defaults[key])+','))):
+                append_keys.append(key)
+            else:
+                other_keys.append(key)
+        self.__write_keys_val_to_file(append_keys, other_keys, val, defaults)
+
+    def __write_keys_val_to_file(self, append_keys, other_keys, val, defaults):
+        """
+        This method helps in grouping keys with common appendable values and
+        writing all keys to the file.
+
+        :param append_keys: The keys that have to be written with an append
+                            delimiter.
+        :param other_keys:  The keys that have to be written with a normal
+                            delimiter.
+        :param val:         The value to be written for the keys.
+        :param defaults:    The defaults of the section the keys and value are
+                            to be written to.
+        """
+        if append_keys:
+            append_keys = sorted(append_keys,
+                                 key=lambda key: len(defaults[str(key)]),
+                                 reverse=True)
+            write_keys = []
+            def_val = None
+            for key in append_keys:
+                if str(defaults[key]) == def_val:
+                    write_keys.append(key)
+                    continue
+                else:
+                    if write_keys:
+                        append_val = self.__get_append_val(val, def_val)
+                        self.__write_value(
+                            write_keys,
+                            append_val,
+                            self.__key_value_append_delimiter)
+                    write_keys = [key]
+                    def_val = str(defaults[key])
+            append_val = self.__get_append_val(val, def_val)
+            self.__write_value(write_keys,
+                               append_val,
+                               self.__key_value_append_delimiter)
+
+        if other_keys:
+            self.__write_value(other_keys,
+                               val,
+                               self.__key_value_delimiter)
+
+    def __get_append_val(self, val, def_val):
+        def_val_list = def_val.split(self.__key_delimiter)
+        append_val = (self.__key_delimiter + ' ').join(
+            [v.strip() for v in val.split(
+                self.__key_delimiter)
+             if v not in def_val_list])
+        return append_val
+
+    def __write_value(self, keys, val, delimiter):
+        self.__file.write((self.__key_delimiter + ' ').join(keys) +
+                          ' ' + delimiter + ' ' +
+                          val + '\n')
 
     @staticmethod
     def is_comment(key):
--- a/coalib/parsing/CliParsing.py
+++ b/coalib/parsing/CliParsing.py
@@ -41,7 +41,7 @@ def parse_cli(arg_list=None,
     """
     arg_parser = default_arg_parser() if arg_parser is None else arg_parser
     origin += os.path.sep
-    sections = OrderedDict(default=Section('Default'))
+    sections = OrderedDict(cli=Section('cli'))
     line_parser = LineParser(key_value_delimiters,
                              comment_seperators,
                              key_delimiters,
@@ -64,6 +64,7 @@ def parse_cli(arg_list=None,
                                arg_key,
                                arg_value,
                                origin,
+                               section_name='cli',
                                from_cli=True)
 
     return sections
@@ -90,7 +91,7 @@ def parse_custom_settings(sections,
                                value=value,
                                origin=origin,
                                to_append=append,
-                               section_name=key_tuple[0],
+                               section_name=(key_tuple[0] or 'cli'),
                                from_cli=True)
 
 
--- a/coalib/parsing/CliParsing.py
+++ b/coalib/parsing/CliParsing.py
@@ -13,7 +13,8 @@ def parse_cli(arg_list=None,
               key_value_delimiters=('=', ':'),
               comment_seperators=(),
               key_delimiters=(',',),
-              section_override_delimiters=('.',)):
+              section_override_delimiters=('.',),
+              key_value_append_delimiters=('+=',)):
     """
     Parses the CLI arguments and creates sections out of it.
 
@@ -23,13 +24,17 @@ def parse_cli(arg_list=None,
     :param arg_parser:                  Instance of ArgParser that is used to
                                         parse none-setting arguments.
     :param key_value_delimiters:        Delimiters to separate key and value
-                                        in setting arguments.
+                                        in setting arguments where settings are
+                                        being defined.
     :param comment_seperators:          Allowed prefixes for comments.
     :param key_delimiters:              Delimiter to separate multiple keys of
                                         a setting argument.
     :param section_override_delimiters: The delimiter to delimit the section
                                         from the key name (e.g. the '.' in
                                         sect.key = value).
+    :param key_value_append_delimiters: Delimiters to separate key and value
+                                        in setting arguments where settings are
+                                        being appended.
     :return:                            A dictionary holding section names
                                         as keys and the sections themselves
                                         as value.
@@ -41,7 +46,8 @@ def parse_cli(arg_list=None,
                              comment_seperators,
                              key_delimiters,
                              {},
-                             section_override_delimiters)
+                             section_override_delimiters,
+                             key_value_append_delimiters)
 
     for arg_key, arg_value in sorted(
             vars(arg_parser.parse_args(arg_list)).items()):
@@ -76,12 +82,14 @@ def parse_custom_settings(sections,
     :param line_parser:          The LineParser to use.
     """
     for setting_definition in custom_settings_list:
-        (_, key_tuples, value, _) = line_parser.parse(setting_definition)
+        (_, key_tuples, value, append, _) = line_parser._parse(
+            setting_definition)
         for key_tuple in key_tuples:
             append_to_sections(sections,
                                key=key_tuple[1],
                                value=value,
                                origin=origin,
+                               to_append=append,
                                section_name=key_tuple[0],
                                from_cli=True)
 
--- a/coalib/parsing/LineParser.py
+++ b/coalib/parsing/LineParser.py
@@ -1,3 +1,5 @@
+import logging
+
 from coala_utils.string_processing.StringConverter import StringConverter
 from coala_utils.string_processing import (unescape, convert_to_raw,
                                            position_is_escaped,
@@ -11,7 +13,8 @@ class LineParser:
                  comment_separators=('#',),
                  key_delimiters=(',', ' '),
                  section_name_surroundings=None,
-                 section_override_delimiters=('.',)):
+                 section_override_delimiters=('.',),
+                 key_value_append_delimiters=('+=',)):
         """
         Creates a new line parser. Please note that no delimiter or separator
         may be an "o" or you may encounter undefined behaviour with the
@@ -31,12 +34,16 @@ class LineParser:
                                             puts the key into the section
                                             "section" despite of the current
                                             section.
+        :param key_value_append_delimiters: Delimiters to separate key and
+                                            value in setting arguments where
+                                            settings are being appended.
         """
         section_name_surroundings = (
             {'[': ']'} if section_name_surroundings is None
             else section_name_surroundings)
 
         self.key_value_delimiters = key_value_delimiters
+        self.key_value_append_delimiters = key_value_append_delimiters
         self.comment_separators = comment_separators
         self.key_delimiters = key_delimiters
         self.section_name_surroundings = section_name_surroundings
@@ -52,22 +59,48 @@ class LineParser:
         :return:     section_name (empty string if it's no section name),
                      [(section_override, key), ...], value, comment
         """
+        logging.warning('The parse method of LineParser is deprecated and will'
+                        ' be removed. Please use `_parse` which has a new '
+                        'return type, a tuple containing 5 values instead of '
+                        '4. Refer to the method documentation for further '
+                        'information.')
+        section_name, key_tuples, value, _, comment = self._parse(line)
+        return section_name, key_tuples, value, comment
+
+    def _parse(self, line):
+        """
+        Note that every value in the returned tuple *besides the value* is
+        unescaped. This is so since the value is meant to be put into a Setting
+        later thus the escapes may be needed there.
+
+        :param line: The line to parse.
+        :return:     section_name (empty string if it's no section name),
+                     [(section_override, key), ...], value, to_append (True if
+                     append delimiter is found else False), comment
+        """
         line, comment = self.__separate_by_first_occurrence(
             line,
             self.comment_separators)
         comment = unescape(comment)
         if line == '':
-            return '', [], '', comment
+            return '', [], '', False, comment
 
         section_name = unescape(self.__get_section_name(line))
         if section_name != '':
-            return section_name, [], '', comment
+            return section_name, [], '', False, comment
 
         # Escapes in value might be needed by the bears
-        keys, value = self.__extract_keys_and_value(line)
+        append = True
+        keys, value = self.__extract_keys_and_value(
+            line, self.key_value_append_delimiters)
+        if not value:
+            keys, value = self.__extract_keys_and_value(
+                line, self.key_value_delimiters, True)
+            append = False
 
         # Add all the delimiters that stored as tuples
         all_delimiters = self.key_value_delimiters
+        all_delimiters += self.key_value_append_delimiters
         all_delimiters += self.key_delimiters
         all_delimiters += self.comment_separators
         all_delimiters += self.section_override_delimiters
@@ -90,7 +123,7 @@ class LineParser:
                 True)
             key_tuples.append((unescape(section), unescape(key)))
 
-        return '', key_tuples, value, comment
+        return '', key_tuples, value, append, comment
 
     @staticmethod
     def __separate_by_first_occurrence(string,
@@ -148,12 +181,28 @@ class LineParser:
 
         return ''
 
-    def __extract_keys_and_value(self, line):
+    def __extract_keys_and_value(self,
+                                 line,
+                                 delimiters,
+                                 return_second_part_nonempty=False):
+        """
+        This method extracts the keys and values from the give string by
+        splitting them based on the delimiters provided.
+
+        :param line:                        The input string.
+        :param delimiters:                  A list of delimiters to split the
+                                            strings on.
+        :param return_second_part_nonempty: If no delimiter is found and this
+                                            is true the contents of the string
+                                            will be returned as value
+        :return:                            The parsed keys and values from a
+                                            line.
+        """
         key_part, value = self.__separate_by_first_occurrence(
             line,
-            self.key_value_delimiters,
+            delimiters,
             True,
-            True)
+            return_second_part_nonempty)
         keys = list(StringConverter(
             key_part,
             list_delimiters=self.key_delimiters).__iter__(
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -262,7 +262,9 @@ def get_config_directory(section):
 
     However if its origin is already a directory this will be preserved:
 
-    >>> section['files'].origin = os.path.abspath('/tmp/dir/')
+    >>> files = section['files']
+    >>> files.origin = os.path.abspath('/tmp/dir/')
+    >>> section.append(files)
     >>> os.makedirs(section['files'].origin, exist_ok=True)
     >>> get_config_directory(section) == section['files'].origin
     True
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -82,7 +82,7 @@ def save_sections(sections):
 
     :param sections: A section dict.
     """
-    default_section = sections['default']
+    default_section = sections['cli']
     try:
         if bool(default_section.get('save', 'false')):
             conf_writer = ConfWriter(
@@ -151,19 +151,17 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
     check_conflicts(cli_sections)
 
     if (
-            bool(cli_sections['default'].get('find_config', 'False')) and
-            str(cli_sections['default'].get('config')) == ''):
-        cli_sections['default'].add_or_create_setting(
-            Setting('config',
-                    re.escape(find_user_config(os.getcwd())),
-                    from_cli=True))
+            bool(cli_sections['cli'].get('find_config', 'False')) and
+            str(cli_sections['cli'].get('config')) == ''):
+        cli_sections['cli'].add_or_create_setting(
+            Setting('config', re.escape(find_user_config(os.getcwd()))))
 
     targets = []
     # We don't want to store targets argument back to file, thus remove it
-    for item in list(cli_sections['default'].contents.pop('targets', '')):
+    for item in list(cli_sections['cli'].contents.pop('targets', '')):
         targets.append(item.lower())
 
-    if bool(cli_sections['default'].get('no_config', 'False')):
+    if bool(cli_sections['cli'].get('no_config', 'False')):
         sections = cli_sections
     else:
         base_sections = load_config_file(Constants.system_coafile, log_printer)
@@ -176,10 +174,10 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
         user_config = str(user_sections['default'].get(
             'config', default_config))
         config = os.path.abspath(
-            str(cli_sections['default'].get('config', user_config)))
+            str(cli_sections['cli'].get('config', user_config)))
 
         try:
-            save = bool(cli_sections['default'].get('save', 'False'))
+            save = bool(cli_sections['cli'].get('save', 'False'))
         except ValueError:
             # A file is deposited for the save parameter, means we want to save
             # but to a specific file.
@@ -204,7 +202,10 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
                             'specify inheritance: the section \'all.python\' '
                             'will inherit all settings from \'all\'.')
 
-    str_log_level = str(sections['default'].get('log_level', '')).upper()
+    str_log_level = str(sections['cli'].get(
+        'log_level',
+        sections['default'].get('log_level', '')
+        if 'default' in sections else '')).upper()
     log_printer.log_level = LOG_LEVEL.str_dict.get(str_log_level,
                                                    LOG_LEVEL.INFO)
 
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -1,6 +1,7 @@
 import os
 import re
 import sys
+import logging
 
 from coalib.collecting.Collectors import (
     collect_all_bears_from_sections, filter_section_bears_by_languages)
@@ -153,7 +154,9 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
             bool(cli_sections['default'].get('find_config', 'False')) and
             str(cli_sections['default'].get('config')) == ''):
         cli_sections['default'].add_or_create_setting(
-            Setting('config', re.escape(find_user_config(os.getcwd()))))
+            Setting('config',
+                    re.escape(find_user_config(os.getcwd())),
+                    from_cli=True))
 
     targets = []
     # We don't want to store targets argument back to file, thus remove it
@@ -190,9 +193,16 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
 
         sections = merge_section_dicts(sections, cli_sections)
 
-    for section in sections:
-        if section != 'default':
-            sections[section].defaults = sections['default']
+    for name, section in list(sections.items()):
+        section.set_default_section(sections)
+        if name == 'default' and not all(
+                setting.from_cli for setting in section.contents.values()):
+            logging.warning('Implicit \'Default\' section inheritance is '
+                            'deprecated. It will be removed soon. To silence '
+                            'this warning remove settings in the \'Default\' '
+                            'section from your coafile. You can use dots to '
+                            'specify inheritance: the section \'all.python\' '
+                            'will inherit all settings from \'all\'.')
 
     str_log_level = str(sections['default'].get('log_level', '')).upper()
     log_printer.log_level = LOG_LEVEL.str_dict.get(str_log_level,
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -189,23 +189,34 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
 
         sections = merge_section_dicts(sections, coafile_sections)
 
+        if 'cli' in sections:
+            logging.warning('\'cli\' is an internally reserved section name. '
+                            'It may have been generated into your coafile '
+                            'while running coala with `--save`. The settings '
+                            'in that section will inherit implicitly to all '
+                            'sections as defaults just like CLI args do.'
+                            'Please change the name of that section in your '
+                            'coafile to avoid any unexpected behavior.')
+
         sections = merge_section_dicts(sections, cli_sections)
 
     for name, section in list(sections.items()):
         section.set_default_section(sections)
-        if name == 'default' and not all(
-                setting.from_cli for setting in section.contents.values()):
-            logging.warning('Implicit \'Default\' section inheritance is '
-                            'deprecated. It will be removed soon. To silence '
-                            'this warning remove settings in the \'Default\' '
-                            'section from your coafile. You can use dots to '
-                            'specify inheritance: the section \'all.python\' '
-                            'will inherit all settings from \'all\'.')
+        if name == 'default':
+            if section.contents:
+                logging.warning('Implicit \'Default\' section inheritance is '
+                                'deprecated. It will be removed soon. To '
+                                'silence this warning remove settings in the '
+                                '\'Default\' section from your coafile. You '
+                                'can use dots to specify inheritance: the '
+                                'section \'all.python\' will inherit all '
+                                'settings from \'all\'.')
+                sections['default'].update(sections['cli'])
+                sections['default'].name = 'cli'
+                sections['cli'] = sections['default']
+            del sections['default']
 
-    str_log_level = str(sections['cli'].get(
-        'log_level',
-        sections['default'].get('log_level', '')
-        if 'default' in sections else '')).upper()
+    str_log_level = str(sections['cli'].get('log_level', '')).upper()
     log_printer.log_level = LOG_LEVEL.str_dict.get(str_log_level,
                                                    LOG_LEVEL.INFO)
 
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -253,3 +253,34 @@ class Section:
         :param key: The key of the setting to be deleted
         """
         del self.contents[key]
+
+    def set_default_section(self, sections, section_name=None):
+        """
+        Find and set the defaults of a section from a dictionary of sections.
+        The defaults are found on the basis of '.' in section names:
+
+        >>> sections = {'all': Section('all')}
+        >>> section = Section('all.python')
+        >>> section.set_default_section(sections)
+        >>> section.defaults.name
+        'all'
+        >>> section = Section('all.python.syntax')
+        >>> section.set_default_section(sections)
+        >>> section.defaults.name
+        'all'
+
+        :param sections:     A dictionary of sections.
+        :param section_name: Optional section name argument to find the default
+                             section for. If not given then use member section
+                             name.
+        """
+        default_section = '.'.join((section_name or self.name).split('.')[:-1])
+
+        if default_section:
+            if default_section in sections:
+                self.defaults = sections[default_section]
+            else:
+                self.set_default_section(sections, default_section)
+        elif 'default' in sections and self.name.lower() != 'default':
+            # Implicit "default" section inheritance
+            self.defaults = sections['default']
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -5,6 +5,7 @@ from collections import OrderedDict
 
 from coalib.collecting.Collectors import collect_registered_bears_dirs
 from coala_utils.decorators import enforce_signature, generate_repr
+from coala_utils.string_processing import unescape
 from coalib.misc.DictUtilities import update_ordered_dict_key
 from coalib.settings.Setting import Setting, path_list
 from coalib.parsing.Globbing import glob_escape
@@ -155,7 +156,7 @@ class Section:
 
         if self.__contains__(key, ignore_defaults=True) and allow_appending:
             val = self[key]
-            val.value = str(val.value) + '\n' + setting.value
+            val.value = str(val._value) + '\n' + setting._value
             self.append(val, custom_key=key)
         else:
             self.append(setting, custom_key=key)
@@ -205,7 +206,7 @@ class Section:
         res = copy.deepcopy(self.contents.get(key, None))
         if res is not None:
             if res.to_append and self.defaults and res.key in self.defaults:
-                res.value = self.defaults[key].value + ', ' + res.value
+                res.value = self.defaults[key]._value + ', ' + res._value
                 res.to_append = False
                 return res
             res.to_append = False
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -15,7 +15,8 @@ def append_to_sections(sections,
                        value,
                        origin,
                        section_name=None,
-                       from_cli=False):
+                       from_cli=False,
+                       to_append=False):
     """
     Appends the given data as a Setting to a Section with the given name. If
     the Section does not exist before it will be created empty.
@@ -26,6 +27,8 @@ def append_to_sections(sections,
     :param origin:       The origin value of the setting to add.
     :param section_name: The name of the section to add to.
     :param from_cli:     Whether or not this data comes from the CLI.
+    :param to_append:    The boolean value if setting value needs to be
+                         appended to a setting in the defaults of a section.
     """
     if key == '' or value is None:
         return
@@ -36,14 +39,53 @@ def append_to_sections(sections,
     if not section_name.lower() in sections:
         sections[section_name.lower()] = Section(section_name)
 
-    sections[section_name.lower()].append(
-        Setting(key, str(value), origin, from_cli=from_cli))
+    sections[section_name.lower()].append(Setting(
+        key, str(value), origin, from_cli=from_cli, to_append=to_append))
 
 
 @generate_repr()
 class Section:
     """
     This class holds a set of settings.
+
+    To add settings and sections to a dictionary of sections we can use
+    ``append_to_sections``:
+
+    >>> sections = {}
+    >>> append_to_sections(sections,
+    ...                    'test1',
+    ...                    'val',
+    ...                    'origin',
+    ...                    section_name='all')
+    >>> 'all' in sections
+    True
+    >>> len(sections)
+    1
+    >>> str(sections)
+    "{'all': <Section object(contents=OrderedDict([('test1', ..."
+
+    We can also add settings that can be appended to other settings. Basically
+    it takes the default value of the setting which resides in the defaults of
+    the section and appends the value of the setting in the second and returns
+    the value of the setting:
+
+    >>> append_to_sections(sections,
+    ...                    'test1',
+    ...                    'val2',
+    ...                    'origin',
+    ...                    section_name='all.python',
+    ...                    to_append=True)
+
+    When the section has no defaults:
+
+    >>> str(sections['all.python']['test1'])
+    'val2'
+
+    After assigning defaults:
+
+    >>> sections['all.python'].set_default_section(sections)
+    >>> str(sections['all.python']['test1'])
+    'val, val2'
     """
 
     @staticmethod
@@ -114,6 +156,7 @@ class Section:
         if self.__contains__(key, ignore_defaults=True) and allow_appending:
             val = self[key]
             val.value = str(val.value) + '\n' + setting.value
+            self.append(val, custom_key=key)
         else:
             self.append(setting, custom_key=key)
 
@@ -159,8 +202,13 @@ class Section:
         if key == '':
             raise IndexError('Empty keys are invalid.')
 
-        res = self.contents.get(key, None)
+        res = copy.deepcopy(self.contents.get(key, None))
         if res is not None:
+            if res.to_append and self.defaults and res.key in self.defaults:
+                res.value = self.defaults[key].value + ', ' + res.value
+                res.to_append = False
+                return res
+            res.to_append = False
             return res
 
         if self.defaults is None or ignore_defaults:
@@ -169,7 +217,7 @@ class Section:
         return self.defaults[key]
 
     def __str__(self):
-        value_list = ', '.join(key + ' : ' + repr(str(self.contents[key]))
+        value_list = ', '.join(key + ' : ' + repr(str(self[key]))
                                for key in self.contents)
         return self.name + ' {' + value_list + '}'
 
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -330,6 +330,6 @@ class Section:
                 self.defaults = sections[default_section]
             else:
                 self.set_default_section(sections, default_section)
-        elif 'default' in sections and self.name.lower() != 'default':
-            # Implicit "default" section inheritance
-            self.defaults = sections['default']
+        elif 'cli' in sections and self.name.lower() != 'cli':
+            # CLI section is now default
+            self.defaults = sections['cli']
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -132,6 +132,8 @@ class Setting(StringConverter):
         if not isinstance(to_append, bool):
             raise TypeError('to_append needs to be a boolean value.')
 
+        self.to_append = to_append
+
         StringConverter.__init__(
             self,
             value,
@@ -142,7 +144,6 @@ class Setting(StringConverter):
         self.from_cli = from_cli
         self.key = key
         self.origin = str(origin)
-        self.to_append = to_append
 
     def __path__(self, origin=None, glob_escape_origin=False):
         """
@@ -216,6 +217,13 @@ class Setting(StringConverter):
         """
         return [Setting.__glob__(elem, self.origin) for elem in self]
 
+    def __iter__(self, remove_backslashes=True):
+        if self.to_append:
+            raise ValueError('Iteration on this object is invalid because the '
+                             'value is incomplete. Please access the value of '
+                             'the setting in a section to iterate through it.')
+        return StringConverter.__iter__(self, remove_backslashes)
+
     @property
     def key(self):
         return self._key
@@ -227,3 +235,11 @@ class Setting(StringConverter):
             raise ValueError('An empty key is not allowed for a setting.')
 
         self._key = newkey
+
+    @StringConverter.value.getter
+    def value(self):
+        if self.to_append:
+            raise ValueError('This property is invalid because the value is '
+                             'incomplete. Please access the value of the '
+                             'setting in a section to get the complete value.')
+        return self._value
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -86,7 +86,7 @@ def typed_ordered_dict(key_type, value_type, default):
         for key, value in OrderedDict(setting).items())
 
 
-@generate_repr('key', 'value', 'origin', 'from_cli')
+@generate_repr('key', 'value', 'origin', 'from_cli', 'to_append')
 class Setting(StringConverter):
     """
     A Setting consists mainly of a key and a value. It mainly offers many
@@ -100,7 +100,8 @@ class Setting(StringConverter):
                  strip_whitespaces=True,
                  list_delimiters=(',', ';'),
                  from_cli=False,
-                 remove_empty_iter_elements=True):
+                 remove_empty_iter_elements=True,
+                 to_append=False):
         """
         Initializes a new Setting,
 
@@ -121,9 +122,15 @@ class Setting(StringConverter):
                                            CliParser.
         :param remove_empty_iter_elements: Whether to remove empty elements in
                                            iterable values.
+        :param to_append:                  The boolean value if setting value
+                                           needs to be appended to a setting in
+                                           the defaults of a section.
         """
         if not isinstance(from_cli, bool):
             raise TypeError('from_cli needs to be a boolean value.')
+
+        if not isinstance(to_append, bool):
+            raise TypeError('to_append needs to be a boolean value.')
 
         StringConverter.__init__(
             self,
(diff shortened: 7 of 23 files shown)


coalib/parsing/LineParser.py
```diff
@@ -148,12 +155,19 @@ def __get_section_name(self, line):
 
         return ''
 
-    def __extract_keys_and_value(self, line):
-        key_part, value = self.__separate_by_first_occurrence(
-            line,
-            self.key_value_delimiters,
-            True,
-            True)
+    def __extract_keys_and_value(self, line, append_value=False):
```
sils: just give the delimiters as parameter, huh?

Mixih:
WIP because tests failing

coalib/parsing/LineParser.py
Udayan12167: Hmm  yeah cleaner.

sils:
first commit message has no verb

coalib/settings/ConfigurationGathering.py
```diff
@@ -132,6 +133,20 @@ def warn_config_absent(sections, argument, log_printer):
     return False
 
 
+def get_default_section(section, sections):
+    default_section = ('.'.join(section.split('.')[:-1])
+                       if unescaped_search_for('.', section) else None)
```
sils: nah this is buggy, if I have several `.`s and one of them is escaped this may fail

coalib/settings/ConfigurationGathering.py
```diff
@@ -132,6 +133,20 @@ def warn_config_absent(sections, argument, log_printer):
...
+
+    if default_section:
+        if default_section in sections:
+            return sections[default_section]
+        else:
+            sections[default_section] = Section(default_section)
+            return sections[default_section]
+
+    return sections['default']
```
sils: `'default'`?!?

coalib/settings/ConfigurationGathering.py
adtac: I think it's better if we just disallow `\.`s in section names. Certainly improves readability and I can't think of a use case where a section name would need a `.`

adtac: Disallow in names, allow in inheritance, obviously.

coalib/settings/ConfigurationGathering.py
```diff
@@ -132,6 +133,20 @@ def warn_config_absent(sections, argument, log_printer):
     return False
 
 
+def get_default_section(section, sections):
```
adtac: Missing docs, params, return

coalib/settings/ConfigurationGathering.py
```diff
@@ -132,6 +133,20 @@ def warn_config_absent(sections, argument, log_printer):
...
+    default_section = ('.'.join(section.split('.')[:-1])
+                       if unescaped_search_for('.', section) else None)
+
+    if default_section:
+        if default_section in sections:
```
adtac: I don't think there'd be a section named `""`. So we can safely remove the outer `if`?

adtac: Ah never mind

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +281,17 @@ def test_no_config(self):
...
+
+        with change_directory(test_dir):
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile'])
+            self.assertEqual(sections['all.python'].defaults, sections['all'])
+            self.assertEqual(sections['all.c']['config'],
+                             sections['default']['config'])
+            self.assertEqual(sections['java.test'].defaults, sections['java'])
```
adtac: Maybe test if `all.python[max_line_length] == 80`? After all that's the whole point of inheritence IIUC.

coalib/settings/ConfigurationGathering.py
sils: agree, just ignore backslashes, do a normal split :)

coalib/settings/ConfigurationGathering.py
```diff
@@ -146,6 +160,9 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
                         dict(str, Section), targets: list(str)). (Types
                         indicated after colon.)
     """
+    logging.warning('Implicit \'Default\' section inheritance is deprecated. '
+                    'It will be removed soon.')
```
sils: this should only be shown if there's a `default` section.

tests/coalaTest.py
```diff
@@ -47,22 +47,22 @@ def test_show_all_bears(self):
...
             # 6 bears plus 1 line holding the closing colour escape sequence
-            self.assertEqual(len(output.strip().splitlines()), 7)
+            self.assertEqual(len(output.strip().splitlines()), 8)
 
     def test_show_language_bears(self):
         with bear_test_module():
             retval, output = execute_coala(
                 coala.main, 'coala', '-B', '-l', 'java')
             self.assertEqual(retval, 0)
             # 2 bears plus 1 line holding the closing colour escape sequence
```
sils: you gotta update those comments, you can't just change the tests and expect me to believe this is right :(

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +281,19 @@ def test_no_config(self):
...
+            self.assertEqual(int(sections['all.python']['max_line_length']),
+                             80)
```
sils: add a test for double recursion please, what happens if I define a section `all.python.codestyle`? That should work

tests/settings/section_manager_test_files/inherit_coafile
```diff
@@ -0,0 +1,17 @@
...
+aspects = smell, redundancy.clone  # Inclusive
+
+[all.c]
+language = C
+files = **.(c|h)
+ignore_aspects = redundancy  # Exclusive
+
+[java.test]
+language = Java
+files = **.java
```
sils: missing newline before EOF

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,12 @@ def __get_section_name(self, line):
 
         return ''
 
-    def __extract_keys_and_value(self, line):
+    def __extract_keys_and_value(self, line, delimiters):
```
sils: add a doc comment here please, this method is getting more complicated, I'm not so enthusiastic about line 165 esp

coalib/settings/Section.py
```diff
@@ -14,6 +14,7 @@ def append_to_sections(sections,
                        key,
                        value,
                        origin,
+                       to_append=False,
```
sils: can you put the changes to the setting class to an own commit? Should be simple no?

sils:
you seem to have no test that actually tests the appending

tests/coalaJSONTest.py
```diff
@@ -53,7 +54,9 @@ def test_fail_acquire_settings(self):
 
     def test_show_all_bears(self):
         with bear_test_module():
-            retval, output = execute_coala(coala.main, 'coala', '--json', '-B')
+            retval, output = execute_coala(coala.main,
+                                           'coala', '--json', '-B',
+                                           stdout_only=True)
```
sils: changes to this file look unrelated, maybe put them in an own comit?

Udayan12167: It is related because the `default` section deprecation warnings were showing in the output and that was making the JSON invalid.

coalib/settings/ConfigurationGathering.py
```diff
@@ -192,9 +193,12 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
 
         sections = merge_section_dicts(sections, cli_sections)
 
-    for section in sections:
-        if section != 'default':
-            sections[section].defaults = sections['default']
+    for name, section in list(sections.items()):
+        if name != 'default':
+            section.set_default_section(sections)
```
sils: I'd move this outside the if even and just do `if name == 'default': logging.warning(...)`

coalib/settings/Section.py
```diff
@@ -253,3 +253,28 @@ def delete_setting(self, key):
...
+    def set_default_section(self, sections):
+        """
+        Find and set the defaults of a section from a dictionary of sections.
+        The defaults are found on the basis of '.' in section names:
+
+        >>> sections = {'all': Section('all')}
+        >>> section = Section('all.python')
+        >>> section.set_default_section(sections)
+        >>> section.defaults.name
+        'all'
```
sils: another testcase:

```
>>> sections = {'all': Section('all')}
>>> section = Section('all.python.syntax')
```

that'll fail!

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +281,23 @@ def test_no_config(self):
...
+        current_dir = os.path.abspath(os.path.dirname(__file__))
+        test_dir = os.path.join(current_dir, 'section_manager_test_files')
+
+        with change_directory(test_dir):
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile'])
+            self.maxDiff = None
+            print(sections['all.python'].defaults)
```
sils: debug print?

coalib/settings/Section.py
```diff
@@ -161,6 +164,13 @@ def __getitem__(self, item, ignore_defaults=False):
 
         res = self.contents.get(key, None)
         if res is not None:
+            if res.to_append:
```
sils: if ... and ...

coalib/settings/Section.py
```diff
@@ -161,6 +164,13 @@ def __getitem__(self, item, ignore_defaults=False):
...
+                if self.defaults:
+                    if res.key in self.defaults:
+                        default_val = self.defaults[res.key].value
+                        self.update_setting(
+                            res.key, new_value=default_val + ',' + res.value)
+                        res.to_append = False
```
sils: I really don't like this idea of creating a new setting when you get it. It feels like the setting might need a reference to the default setting or something that should be set upon setting the default section or so... I'm not fully sure but this doesn't feel clean

sils:
unack https://github.com/coala/coala/pull/3178/commits/3ec027203e5379fa456b57903ab96bd6e1686d47

sils:
unack https://github.com/coala/coala/pull/3178/commits/dcb1bc01d0c3cd8ac7da352658fe61bd8f39606b#r92730527 that's a major issue IMO

sils:
https://github.com/coala/coala/pull/3178/commits/2085dedd2fef74520ccf9a84492f8000c57e8b7f lgtm on first sight

sils:
maybe @userzimmermann has an idea about the architecture here

coalib/settings/Section.py
```diff
@@ -253,3 +253,37 @@ def delete_setting(self, key):
...
+
+        :param sections:     A dictionary of sections.
+        :param section_name: Optional section name argument to find the default
+                             section for. If not given then use member section
+                             name.
+        """
+        if section_name:
+            default_section = '.'.join(section_name.split('.')[:-1])
+        else:
+            default_section = '.'.join(self.name.split('.')[:-1])
```
sils: shorter:

    default_section = '.'.join((section_name or self.name).split('.')[:-1])

coalib/settings/Section.py
```diff
@@ -253,3 +253,37 @@ def delete_setting(self, key):
...
+
+        if default_section:
+            if default_section in sections:
+                self.defaults = sections[default_section]
+            else:
+                self.set_default_section(sections, default_section)
+        elif 'default' in sections:
```
sils: and self.name != 'default'

coalib/settings/Section.py
```diff
@@ -161,6 +164,12 @@ def __getitem__(self, item, ignore_defaults=False):
 
         res = self.contents.get(key, None)
         if res is not None:
+            if res.to_append and self.defaults:
```
sils: `and res.key in defaults`?

Udayan12167: Sorry forgot about the python evaluation.

sils:
unack https://github.com/coala/coala/pull/3178/commits/1fa2af6b48c8f5407aaa9ecb5e7ff96c845bf77e

sils:
@Udayan12167 can you rebase and please fix your tests?

sils:
@Udayan12167 see CI log, you have some problem with one of your doc comments

coalib/parsing/CliParsing.py
```diff
@@ -22,8 +23,12 @@ def parse_cli(arg_list=None,
                                         paths given as argument.
     :param arg_parser:                  Instance of ArgParser that is used to
                                         parse none-setting arguments.
-    :param key_value_delimiters:        Delimiters to separate key and value
-                                        in setting arguments.
+    :param key_value_define_delimiters: Delimiters to separate key and value
+                                        in setting arguments where settings are
+                                         being defined.
```
sils: indentation off, I think that's also causing the build fail

coalib/settings/ConfigurationGathering.py
```diff
@@ -192,9 +193,11 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
...
         sections = merge_section_dicts(sections, cli_sections)
 
-    for section in sections:
-        if section != 'default':
-            sections[section].defaults = sections['default']
+    for name, section in list(sections.items()):
+        section.set_default_section(sections)
+        if name == 'default':
+            logging.warning('Implicit \'Default\' section inheritance is '
+                            'deprecated. It will be removed soon.')
```
sils: I think there's no way the user can avoid this warning because ATM coala *always* creates a default section. I think we should change that.

coalib/settings/Section.py
sils: you forgot to resolve this

coalib/settings/Section.py
```diff
@@ -161,6 +164,12 @@ def __getitem__(self, item, ignore_defaults=False):
...
+                if res.key in self.defaults:
+                    return Setting(
+                        key=key,
+                        value=self.defaults[key].value + ',' + res.value,
+                        origin=res.origin)
```
sils: so there's an important case missing that should also be tested: how does writing back to file behave? I think it would write back the full concatenated setting.

Maybe we can build a mechanism that for every setting it checks if it `.startswith` it's default setting and the comma - and if so, strip that away and use `+=` automatically, then it'd even optimize it if I write coafiles that could profit from the `+=` operator but the user didn't use it.

sils: or alternatively get the setting from `.contents` to get the original one

coalib/parsing/CliParsing.py
```diff
@@ -22,8 +23,12 @@ def parse_cli(arg_list=None,
...
+    :param key_value_append_delimiters: Delimiters to separate key and value
+                                        in setting arguments where settings are
+                                        being defined.
```
sils: this is the same description as above, not so useful

coalib/parsing/CliParsing.py
sils: why change the description anyway

sils:
looks pretty good otherwise. Oh you have conflicts, probably something simple related to the PyCodestyle PR which touched a bunch of files codestyle wise.

coalib/settings/ConfigurationGathering.py
sils: at the very least we should check here if the section contains any important settings and allow the user to somehow get rid of this warning IMO. If this section isn't empty after your changes then your changes aren't complete yet are they?

tests/coalaTest.py
```diff
@@ -65,23 +65,25 @@ def test_show_all_bears(self):
         with bear_test_module():
             retval, output = execute_coala(coala.main, 'coala', '-B')
             self.assertEqual(retval, 0)
-            # 6 bears plus 1 line holding the closing colour escape sequence
-            self.assertEqual(len(output.strip().splitlines()), 7)
+            # 6 bears, 1 line holding the closing colour escape sequence, and
+            # 1 line containing a deprecation warning.
```
sils: can we rather do stdout only here too? (and below)

coalib/settings/ConfigurationGathering.py
sils: also the setting has an origin so you can even exclude settings from CLI and just look if any settings are from the coafile

sils:
https://github.com/coala/coala/pull/3178/commits/ef33c90c95e5e9935fa161a12d2822214f1e4f40 needs work

tests/settings/SectionTest.py
```diff
@@ -138,11 +140,25 @@ def test_append_to_sections(self):
...
 
+        append_to_sections(sections,
+                           'test1',
+                           'val2',
+                           'origin',
+                           to_append=True,
+                           section_name='section')
+        self.assertEqual(str(sections['section']['test1']), 'val2')
+        sections['section'].set_default_section(sections)
+        self.assertEqual(str(sections['section']['test1']), 'val,val2')
```
sils: could you add a doctest instead? Section and Setting both need better documentation *and* I think this would simplify the testsuite. Same for the Setting test. Just put it in the class documentation comment?

coalib/settings/Setting.py
```diff
@@ -135,6 +139,7 @@ def __init__(self,
         self.from_cli = from_cli
         self.key = key
         self.origin = str(origin)
+        self.to_append = to_append
```
sils: if `to_append` is `True` I think I shouldn't be able to retrieve the value of the setting since it's not complete

tests/settings/SectionTest.py
sils: and especially when adding tests we shouldn't mix them into existing testcases but make new ones, this will only lead to more crowded and less understandable test cases I fear

coalib/coala_main.py
```diff
@@ -86,11 +86,14 @@ def run_coala(console_printer=None,
...
-                           settings_changed(log_printer, settings_hash))
+        flush_cache = bool(
+            (sections['cli'] or sections['default']).get(
+                'flush_cache', False) or
+            settings_changed(log_printer, settings_hash))
 
         cache = None
-        if not sections['default'].get('disable_caching', False):
+        if not (sections['cli'] or sections['default']).get(
+                'disable_caching', False):
```
sils: uhm does this or statement really work? I mean if the cli section doesn't exist or is empty it will either raise an exception or it'll just take the default (False), this doesn't look too good to me. Maybe even make a function `get_cli_setting(key, type, default)`?

sils: actually why do you need to consider `default` here anyway?

coalib/settings/Section.py
```diff
@@ -161,6 +164,11 @@ def __getitem__(self, item, ignore_defaults=False):
 
         res = self.contents.get(key, None)
         if res is not None:
+            if res.to_append and self.defaults and (res.key in self.defaults):
+                return Setting(
+                    key=key,
+                    value=self.defaults[key].value + ',' + res.value,
+                    origin=res.origin)
```
sils: hm I mentioned earlier that we might want to override the `.value` property getter to raise an exception if to_append is true because the value is likely wrong, then this will have to be changed!

sils:
https://github.com/coala/coala/pull/3178/commits/70f41ec862f1b6da182a6b22ea8c0fae33f90ab0 needs work

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,19 @@ def __get_section_name(self, line):
...
+        :param delimiters: A list of delimiters to split the strings on.
+        """
         key_part, value = self.__separate_by_first_occurrence(
             line,
-            self.key_value_delimiters,
+            delimiters,
             True,
-            True)
+            # To return empty value in case of no append delimiter is found.
+            delimiters == self.key_value_define_delimiters)
```
sils: this is rather dirty I think and hurts readability, just make another bool parameter for this in the signature and explain it properly.

sils:
unack https://github.com/coala/coala/pull/3178/commits/4edc12e3be63378751d2eadd0fa3551caa9557be but very good and almost ready I think

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,35 @@ def __write_key_val(self, keys, val):
 
         # Add escape characters as appropriate
         keys = [escape(key, chain(['\\'],
-                                  self.__key_value_delimiters,
+                                  self.__key_value_define_delimiters,
```
sils: you forgot to add the append delimiters here too

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,35 @@ def __write_key_val(self, keys, val):
...
+                 if v not in def_val_list])
+            self.__file.write((self.__key_delimiter + ' ').join(append_keys) +
+                              ' ' + self.__key_value_append_delimiter + ' ' +
+                              append_val + '\n')
+            keys = [key for key in keys if key not in append_keys]
+
+        if keys:
+            self.__file.write((self.__key_delimiter + ' ').join(keys) + ' ' +
+                              self.__key_value_define_delimiter + ' ' + val +
+                              '\n')
```
sils: woha, this is too big to not be an own function of some kind.

sils: or simplified

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,35 @@ def __write_key_val(self, keys, val):
...
+        if append_keys:
+            def_val_list = str(section.defaults[append_keys[0]]).split(
+                self.__key_delimiter)
+            append_val = (self.__key_delimiter + ' ').join(
+                [v.strip() for v in val.split(self.__key_delimiter)
+                 if v not in def_val_list])
+            self.__file.write((self.__key_delimiter + ' ').join(append_keys) +
+                              ' ' + self.__key_value_append_delimiter + ' ' +
+                              append_val + '\n')
+            keys = [key for key in keys if key not in append_keys]
```
sils: I dislike this a bit, can't we just compile a second list in an else in 107 maybe? Or work with sets?

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,35 @@ def __write_key_val(self, keys, val):
...
                 for key in keys]
         val = escape(val, chain(['\\'], self.__comment_separators))
 
-        self.__file.write((self.__key_delimiter + ' ').join(keys) + ' ' +
-                          self.__key_value_delimiter + ' ' + val + '\n')
+        append_keys = []
+        for key in keys:
+            if (section.defaults and
+                    (key in section.defaults and
+                     (val.startswith(str(section.defaults[key])+',')))):
```
sils: so if I have this:

```ini
[defaults]
a = 4
b = 4,5
[defaults.new]
a = 4,5,6
b = 4,5,6
```

I think your algorithm would transform it to:

```ini
[defaults]
a = 4
b = 4,5
[defaults.new]
a, b += 5,6
```

maybe we should have like a generator expression to get the first append string and then check for which keys it's valid, that should be a good enough heuristic

sils:
unack https://github.com/coala/coala/pull/3178/commits/97b95321a9df5bcd3c775807fbfb370b8c28b16d , buggy but promising start

coalib/settings/ConfigurationGathering.py
```diff
@@ -82,7 +82,7 @@ def save_sections(sections):
 
     :param sections: A section dict.
     """
-    default_section = sections['default']
+    default_section = (sections['cli'] or sections['default'])
```
sils: I think this'll either fail (exception) or take the cli section

sils: as above

coalib/settings/ConfigurationGathering.py
```diff
@@ -199,7 +199,8 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
             logging.warning('Implicit \'Default\' section inheritance is '
                             'deprecated. It will be removed soon.')
 
-    str_log_level = str(sections['default'].get('log_level', '')).upper()
+    str_log_level = str(
+        (sections['cli'] or sections['default']).get('log_level', '')).upper()
```
sils: same as above

sils:
unack https://github.com/coala/coala/pull/3178/commits/97b95321a9df5bcd3c775807fbfb370b8c28b16d https://github.com/coala/coala/pull/3178/commits/351872a83bde5a27ed3b746d909330dbff9f3d4d#diff-8ecc903ba7ecebebacc61f64149f8eb7R85

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +282,40 @@ def test_no_config(self):
...
+        with self.assertLogs(logger, 'WARNING') as cm:
+            gather_configuration(lambda *args: True,
+                                 self.log_printer,
+                                 arg_list=[])
+
+        self.assertIn('WARNING', cm.output[0])
+
+        try:
+            with self.assertLogs(logger, 'WARNING') as cm:
+                load_configuration(['--no-config'], self.log_printer)
```
sils: are you trying to assert that there are no log messages here? this looks overly complicated

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +282,40 @@ def test_no_config(self):
...
+            self.assertEqual(sections['java.test'].defaults,
+                             sections['default'])
+            self.assertEqual(int(sections['all.python']['max_line_length']),
+                             80)
+            self.assertEqual(sections['all.python.codestyle'].defaults,
+                             sections['all.python'])
+            self.assertEqual(sections['all.java.codestyle'].defaults,
+                             sections['all'])
+
+    def test_default_section_deprecation_warning(self):
```
sils: you're testing this against the coafile of this repo, right? Put a comment! We'll change that eventually and people will be confused then

sils: took me a while...

coalib/settings/ConfigurationGathering.py
```diff
@@ -192,9 +195,14 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
...
 
-    for section in sections:
-        if section != 'default':
-            sections[section].defaults = sections['default']
+    for name, section in list(sections.items()):
+        section.set_default_section(sections)
+        if name == 'default' and not all(
+                setting.from_cli for setting in section.contents.values()):
+            logging.warning('Implicit \'Default\' section inheritance is '
+                            'deprecated. It will be removed soon. To silence '
```
sils: I'd add something like `You can use dots to specify inheritance: the section all.python will inherit all settings from all.` so people get to know the new thing right ahead for migration!

coalib/settings/Section.py
```diff
@@ -36,14 +39,50 @@ def append_to_sections(sections,
...
+    True
+    >>> len(sections)
+    1
+    >>> len(sections['default'].contents)
+    1
+
+    We can also add settings that can be appended to other settings. Basically
+    it takes the default value of the setting which resides in the defaults of
+    the section and appends the value of the setting in the second and returns
+    the value of the setting:
```
sils: missing newline here, this won't be rendered correctly

coalib/settings/Section.py
```diff
@@ -36,14 +39,50 @@ def append_to_sections(sections,
...
+
+    To add settings and sections to a dictionary of sections we can use
+    `append_to_sections`:
+
+    >>> sections = {}
+    >>> append_to_sections(sections,
+    ...                    'test1',
+    ...                    'val',
+    ...                    'origin',
+    ...                    section_name='default')
```
sils: why do you use default instead of `all` and `all.python`? This is deprecated after all

coalib/settings/Section.py
```diff
@@ -36,14 +39,50 @@ def append_to_sections(sections,
...
+    >>> 'default' in sections
+    True
+    >>> len(sections)
+    1
+    >>> len(sections['default'].contents)
+    1
```
sils: maybe just print the `repr` of the `sections` here and use `...` in the right places to make it well readable

coalib/settings/Setting.py
sils: no response to this @Udayan12167 ?

coalib/settings/Section.py
```diff
@@ -161,6 +200,11 @@ def __getitem__(self, item, ignore_defaults=False):
 
         res = self.contents.get(key, None)
         if res is not None:
+            if res.to_append and self.defaults and (res.key in self.defaults):
+                return Setting(
+                    key=key,
+                    value=self.defaults[key].value + ',' + res.value,
```
sils: let's have a space after the comma for beautification, shall we? :) Not that it's functionally relevant though.

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,26 @@ def __get_section_name(self, line):
...
+                                      the contents of the string will be
+                                      returned as value
+        """
         key_part, value = self.__separate_by_first_occurrence(
             line,
-            self.key_value_delimiters,
+            delimiters,
             True,
-            True)
+            # To return empty value in case of no append delimiter is found.
```
sils: this comment is redundant now

coalib/settings/Section.py
```diff
@@ -253,3 +253,34 @@ def delete_setting(self, key):
...
+                             name.
+        """
+        default_section = '.'.join((section_name or self.name).split('.')[:-1])
+
+        if default_section:
+            if default_section in sections:
+                self.defaults = sections[default_section]
+            else:
+                self.set_default_section(sections, default_section)
+        elif 'default' in sections and self.name != 'Default':
```
sils: rather lower the name, we can't rely on it always having an upper case D I think

tests/output/ConfWriterTest.py
```diff
@@ -25,7 +25,19 @@ class ConfWriterTest(unittest.TestCase):
                     '    key\\\\backslash = value\\\\backslash\n'
                     '    key\\,comma = value,comma\n'
                     '    key\\#hash = value\\#hash\n'
-                    '    key\\.dot = value.dot\n')
+                    '    key\\.dot = value.dot\n'
+                    '    a_default = val, val2\n')
+
+    example_file_2 = ('[defaults]\n'
```
sils: can we not enumerate them but give them proper names? append_example_file or so?

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,69 @@ def __write_key_val(self, keys, val):
...
+        for key in keys:
+            if (defaults and
+                    (key in defaults and
+                     (val.startswith(str(defaults[key])+',')))):
+                append_keys.append(key)
+            else:
+                other_keys.append(key)
+        self.__write_keys_val_to_file(append_keys, other_keys, val, defaults)
+
+    def __write_keys_val_to_file(self, append_keys, other_keys, val, defaults):
```
sils: some documentation here would be good IMO

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,69 @@ def __write_key_val(self, keys, val):
...
+
+    def __get_append_val(self, val, def_val):
+        def_val_list = def_val.split(self.__key_delimiter)
+        append_val = (self.__key_delimiter + ' ').join(
+            [v.strip() for v in val.split(
+                self.__key_delimiter)
+             if v not in def_val_list])
+        return append_val
+
+    def __write_to_file(self, keys, val, delimiter):
```
sils: I'd name it `__write_value` I think

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,69 @@ def __write_key_val(self, keys, val):
...
+    def __write_keys_val_to_file(self, append_keys, other_keys, val, defaults):
+        if append_keys:
+            append_keys = sorted(append_keys,
+                                 key=lambda key: len(defaults[str(key)]),
+                                 reverse=True)
+            append_keys_iter = append_keys.__iter__()
+            write_keys = []
+            def_val = None
+            try:
+                while True:
```
sils: the more I look at it the more I think that a simple for loop would do the job quite well, simplifying the code here a bit

coalib/settings/ConfigurationGathering.py
```diff
@@ -151,19 +151,17 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
...
-            bool(cli_sections['default'].get('find_config', 'False')) and
-            str(cli_sections['default'].get('config')) == ''):
-        cli_sections['default'].add_or_create_setting(
-            Setting('config',
-                    re.escape(find_user_config(os.getcwd())),
-                    from_cli=True))
+            bool(cli_sections['cli'].get('find_config', 'False')) and
+            str(cli_sections['cli'].get('config')) == ''):
+        cli_sections['cli'].add_or_create_setting(
+            Setting('config', re.escape(find_user_config(os.getcwd()))))
```
sils: I don't think we want to remove the `from_cli` thing, it's used in some places we should check that first, not part of this PR

sils: also the user can stil define a cli section in his coafile which is not from cli then right?

tests/misc/CachingTest.py
```diff
@@ -146,13 +146,11 @@ def test_caching_multi_results(self):
                 retval, output = execute_coala(
                    coala.main,
                    'coala',
-                   '-c', filename + '.coafile',
-                   '-f', filename + 'test.py')
+                   '-c', filename + '.coafile')
```
sils: why this change?

tests/parsing/CliParsingTest.py
```diff
@@ -33,23 +33,24 @@ def test_parse_cli(self):
...
                 ('default_key2', 'single_value'),
                 ('default_key3', 'first_value,second_value')},
             'section1': {
                 ('key1', 'value1,value2')},
             'section2': {
-                ('key2', 'only_this_value'),
-                ('key2a', 'k2a')}}
-        self.assertEqual(parsed_sections['default'].name, 'Default')
+                ('key2a', 'k2a'),
+                ('key2', 'only_this_value')}}
```
sils: why switching the order? Shouldn't matter

tests/parsing/CliParsingTest.py
```diff
@@ -33,23 +33,24 @@ def test_parse_cli(self):
...
+        self.assertEqual(parsed_sections['cli'].name, 'cli')
+        print(self.dict_from_sections(parsed_sections))
```
sils: debug print?

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -294,8 +297,6 @@ def test_section_inheritance(self):
                 self.log_printer,
                 arg_list=['-c', 'inherit_coafile'])
             self.assertEqual(sections['all.python'].defaults, sections['all'])
-            self.assertEqual(sections['all.c']['config'],
-                             sections['default']['config'])
```
sils: this is still needed to test the deprecation ight?

coalib/settings/Section.py
adtac: here and a couple of other places too

coalib/settings/Section.py
```diff
@@ -161,6 +200,11 @@ def __getitem__(self, item, ignore_defaults=False):
 
         res = self.contents.get(key, None)
         if res is not None:
+            if res.to_append and self.defaults and (res.key in self.defaults):
```
adtac: Are the brackets needed?

coalib/settings/Section.py
```diff
@@ -169,7 +213,7 @@ def __getitem__(self, item, ignore_defaults=False):
         return self.defaults[key]
 
     def __str__(self):
-        value_list = ', '.join(key + ' : ' + repr(str(self.contents[key]))
+        value_list = ', '.join(key + ' : ' + repr(str(self[key]))
```
adtac: tiny comment: no space before the colon ;) (hey lasse got to comment [this](https://github.com/coala/coala/pull/3178/commits/79852e06cc6ea03e24c068e0bcc7c7328a039c40#r95334456))

coalib/settings/Setting.py
```diff
@@ -114,6 +115,9 @@ def __init__(self,
                                            you want to specify a directory as
                                            origin be sure to end it with a
                                            directory separator.
+        :param to_append:                  The boolean value if setting value
+                                           needs to appended to a setting in
```
adtac: to *be* appended (I think this applies to the other param doc too)

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,26 @@ def __get_section_name(self, line):
...
+                                 return_value_nonempty=False):
+        """
+        This method extracts the keys and values from the give string by
+        splitting them based on the delimiters provided.
+        :param line:                  The input string.
+        :param delimiters:            A list of delimiters to split the
+                                      strings on.
+        :param return_value_nonempty: If no delimiter is found and this is true
+                                      the contents of the string will be
+                                      returned as value
```
adtac: you added the params, wanna add the return doc too? :P

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,26 @@ def __get_section_name(self, line):
...
         return ''
 
-    def __extract_keys_and_value(self, line):
+    def __extract_keys_and_value(self,
+                                 line,
+                                 delimiters,
+                                 return_value_nonempty=False):
+        """
+        This method extracts the keys and values from the give string by
+        splitting them based on the delimiters provided.
```
adtac: line break; things break if there isn't a line break I think

coalib/parsing/LineParser.py
```diff
@@ -148,12 +157,26 @@ def __get_section_name(self, line):
...
+        :param line:                  The input string.
+        :param delimiters:            A list of delimiters to split the
+                                      strings on.
+        :param return_value_nonempty: If no delimiter is found and this is true
```
adtac: why not call this `return_second_part_nonempty` like in `__separate_by_first_occurrence` - it's only passed to `__separate_by_first_occurrence`

coalib/output/ConfWriter.py
```diff
@@ -89,15 +92,69 @@ def __write_key_val(self, keys, val):
...
         val = escape(val, chain(['\\'], self.__comment_separators))
 
-        self.__file.write((self.__key_delimiter + ' ').join(keys) + ' ' +
-                          self.__key_value_delimiter + ' ' + val + '\n')
+        append_keys = []
+        other_keys = []
+        for key in keys:
+            if (defaults and
+                    (key in defaults and
+                     (val.startswith(str(defaults[key])+',')))):
```
adtac: redundant parantheses around `val.startswith(...)`

coalib/coala_main.py
```diff
@@ -86,11 +86,11 @@ def run_coala(console_printer=None,
                                   VERSION))
 
         settings_hash = get_settings_hash(sections, targets)
-        flush_cache = bool(sections['default'].get('flush_cache', False) or
+        flush_cache = bool(sections['cli'].get('flush_cache', False) or
                            settings_changed(log_printer, settings_hash))
 
         cache = None
-        if not sections['default'].get('disable_caching', False):
+        if not sections['cli'].get('disable_caching', False):
```
adtac: I can't disable caching from the coafile now? Sorry if I misunderstood this.

adtac: IIUC I can't disable caching from the coafile now?

coalib/settings/ConfigurationGathering.py
adtac: maybe `__cli`?

userzimmermann:
@Udayan12167 SORRY!!! Accidentally pushed my local `ud/SecInherit` checkout :( Please push yours again!

CC @sils

sils:
ack https://github.com/coala/coala/pull/3178/commits/2db98a13a95d136acfe709fd264ecadbdf9d883b

sils:
ack https://github.com/coala/coala/pull/3178/commits/4cddd1310abda7a31338899bb098abb4f81a1a7c

coalib/parsing/LineParser.py
sils: ping?

coalib/output/ConfWriter.py
sils: ping

Udayan12167:
reack af47d71

coalib/settings/Section.py
```diff
@@ -159,8 +202,13 @@ def __getitem__(self, item, ignore_defaults=False):
         if key == '':
             raise IndexError('Empty keys are invalid.')
 
-        res = self.contents.get(key, None)
+        res = copy.deepcopy(self.contents.get(key, None))
```
sils: you only need the copy inside the if, save performance!

sils: oh you *always* want to return the copy making them immutable? Then why not remove the setter for the setting

jayvdb:
@Udayan12167, is this still being targetted at 0.10?
If so, is 86656fb a mandatory part of this changeset?

The rest of it looks like new behaviour which is written in backwards compatible manner, and could be squeezed into 0.10.
86656fb is complex fiddling with existing code, and is the hard part to review to ensure that there are no surprise backwards incompatibilities hidden in there.

sils:
That's what the one week freeze is for right?

sils:
@jayvdb I think the commit is good. One thing is not backward compatible: if people use:

```ini
[default]
log_level = debug
```

that will no longer work I think. But this is an undocumented "feature" due to the design of our sections. People might know about it when running `coala -V --save` and get it generated into their coafile which now won't show logs anymore.

coalib/parsing/LineParser.py
```diff
@@ -7,7 +7,8 @@
 class LineParser:
 
     def __init__(self,
-                 key_value_delimiters=('=',),
```
jayvdb: this needs a deprecation.

coalib/parsing/LineParser.py
```diff
@@ -36,7 +37,8 @@ def __init__(self,
             {'[': ']'} if section_name_surroundings is None
             else section_name_surroundings)
 
-        self.key_value_delimiters = key_value_delimiters
```
jayvdb: `key_value_delimiters` should become a readonly `@property`.

coalib/parsing/LineParser.py
```diff
@@ -57,17 +59,24 @@ def parse(self, line):
             self.comment_separators)
         comment = unescape(comment)
         if line == '':
-            return '', [], '', comment
+            return '', [], '', False, comment
```
jayvdb: this is a breaking change, and the docstring isnt updated.

coalib/settings/Section.py
```diff
@@ -14,6 +15,7 @@ def append_to_sections(sections,
                        key,
                        value,
                        origin,
+                       to_append=False,
```
jayvdb: this is backwards incompatible.
It make it backwards compatible, move the new arg to the end of the args.

coalib/settings/Setting.py
```diff
@@ -97,6 +97,7 @@ def __init__(self,
                  key,
                  value,
                  origin='',
+                 to_append=False,
```
jayvdb: backwards incompatible; move to end of args

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -281,3 +283,39 @@ def test_no_config(self):
...
+    def test_default_section_deprecation_warning(self):
+        logger = logging.getLogger()
+
+        with self.assertLogs(logger, 'WARNING') as cm:
+            # This gathers the configuration from the '.coafile' of this repo.
+            gather_configuration(lambda *args: True,
+                                 self.log_printer,
+                                 arg_list=[])
+
+        self.assertIn('WARNING', cm.output[0])
```
Mixih: isn't it bad practice to use a cm object outside of the context manager?

Mixih:
ack 9607d9c11e44a205b2e0432a61ff61fd211d0744

coalib/output/ConfWriter.py
Mixih: ping

Mixih:
unack a2fc4f855967fe8382ac641420869f924e5b1fac
unack b9b23066f629d4930f32f015205f7b1ee987308e

coalib/coala_main.py
Mixih: ping!

Mixih:
ack 189cf0439c0557b1314f42ab46cb15150a3b1186

Mixih:
unack a2fc4f855967fe8382ac641420869f924e5b1fac
unack c9a627cbe8020a0b19d4bf3313266c9ca4993133

coalib/coala_main.py
adtac: Not the ideal way IMO.

Udayan12167:
@sils @jayvdb I read your conversation about the `cli` section and have a made a change which should satisfy @jayvdb's concerns. Lemme know if I have hit the right path.

coalib/parsing/LineParser.py
jayvdb: docstring not updated.
As a breaking change, if you think this should be safe to change without deprecation, the  function name should be preceeded with _ so that it is clear in the future that it may change without warning.

coalib/parsing/LineParser.py
jayvdb: Not done?

coalib/parsing/LineParser.py
jayvdb: Not done.

coalib/settings/Setting.py
```diff
@@ -121,10 +122,15 @@ def __init__(self,
...
         :param remove_empty_iter_elements: Whether to remove empty elements in
                                            iterable values.
+        :param to_append:                  The boolean value if setting value
+                                           needs to be appended to a setting in
+                                           the defaults of a section.
         """
         if not isinstance(from_cli, bool):
             raise TypeError('from_cli needs to be a boolean value.')
 
+        self.to_append = to_append
```
jayvdb: Check it is a bool, like above with from_cli..

tests/misc/CachingTest.py
jayvdb: Ping

coalib/output/ConfWriter.py
```diff
@@ -70,19 +73,19 @@ def write_section(self, section):
...
+            self.__write_key_val(keys, val, section.defaults)
 
     def __write_section_name(self, name):
         assert not self.__closed
 
         self.__file.write(self.__section_name_surrounding_beg + name +
                           self.__section_name_surrounding_end + '\n')
 
-    def __write_key_val(self, keys, val):
+    def __write_key_val(self, keys, val, defaults):
```
jayvdb: defaults here is unused. It looks like it isnt needed.

Mixih:
ack 148dc1791814336368d9d5dd90416d35b5368fea

Mixih:
ack 148dc17

tests/coalaJSONTest.py
```diff
@@ -35,7 +35,8 @@ def test_find_issues(self):
             retval, output = execute_coala(coala.main, 'coala', '--json',
                                            '-c', os.devnull,
                                            '-b', 'LineCountTestBear',
-                                           '-f', re.escape(filename))
+                                           '-f', re.escape(filename),
+                                           stdout_only=True)
```
jayvdb: You should capture and check stderr also.
That makes the tests more readable.

coalib/output/ConfWriter.py
Mixih: I see it's used in L110 though

Mixih:
ack a298a50

tests/settings/SettingTest.py
```diff
@@ -95,3 +95,12 @@ def test_inherited_conversions(self):
         self.assertEqual(str(self.uut), '22')
         self.assertEqual(int(self.uut), 22)
         self.assertRaises(ValueError, bool, self.uut)
+
+    def test_value_getter(self):
+        with self.assertRaises(ValueError):
```
Mixih: assertRaisesRegex to confirm the exception content?

tests/settings/SettingTest.py
```diff
@@ -95,3 +95,12 @@ def test_inherited_conversions(self):
...
+            self.uut = Setting('key', '22\n', '.', to_append=True)
+            self.uut.value
+
+        with self.assertRaises(ValueError):
```
Mixih: see above?

coalib/parsing/LineParser.py
sils: hm I'm not sure, our parsing stuff is pretty internal - is there really any chance of anyone outside this repo using this? Especially this low level utility.

coalib/output/ConfWriter.py
jayvdb: Whoops; missed that somehow.

jayvdb:
unack 9d5d5eda4e

jayvdb:
unack ef0c816dcd5

Udayan12167:
reack 8f0aae7 1d44cd9 1d54da5

userzimmermann:
@Udayan12167 @sils @Mixih @jayvdb @adtac When I checkout current `ud/SecInherit` branch and do:

```python
>>> from coalib.parsing.ConfParser import ConfParser
>>> ConfParser().parse('.coafile')
```

I get:

```
C:\Users\Zimmermann\Projects\coala\lib\coalib\parsing\ConfParser.py in parse(self, input_data, overwrite)
     54             self.__init_sections()
     55
---> 56         self.__parse_lines(lines, input_data)
     57
     58         return self.sections

C:\Users\Zimmermann\Projects\coala\lib\coalib\parsing\ConfParser.py in __parse_lines(self, lines, origin)
    125                                 # they fail to resolve this
    126                                 remove_empty_iter_elements=
--> 127                                 self.__remove_empty_iter_elements),
    128                                 # Stop ignoring
    129                         allow_appending=(keys == []))

C:\Users\Zimmermann\Projects\coala\lib\coalib\settings\Setting.py in __init__(self, key, value, origin, strip_whitespaces, list_delimiters, from_cli, remove_empty_iter_elements, to_append)
    140             strip_whitespaces=strip_whitespaces,
    141             list_delimiters=list_delimiters,
--> 142             remove_empty_iter_elements=remove_empty_iter_elements)
    143
    144         self.from_cli = from_cli

C:\Users\Zimmermann\Miniconda3\envs\coala\lib\site-packages\coala_utils\string_processing\StringConverter.py in __init__(self, value, strip_whitespaces, list_delimiters, dict_delimiter, remove_empty_iter_elements)
     34         self.__unescaped_list = None
     35         self.__dict = None
---> 36         self.value = value
     37
     38     def __str__(self):

C:\Users\Zimmermann\Miniconda3\envs\coala\lib\site-packages\coala_utils\string_processing\StringConverter.py in value(self, newval)
    148             self.__value = unescaped_strip(self.__value)
    149
--> 150         self.__prepare_list()
    151         self.__prepare_dict()
    152

C:\Users\Zimmermann\Miniconda3\envs\coala\lib\site-packages\coala_utils\string_processing\StringConverter.py in __prepare_list(self)
    102
    103     def __prepare_list(self):
--> 104         self.__escaped_list = self.__get_raw_list()
    105
    106         if self.__strip_whitespaces:

C:\Users\Zimmermann\Miniconda3\envs\coala\lib\site-packages\coala_utils\string_processing\StringConverter.py in __get_raw_list(self)
     98
     99         return list(unescaped_split(pattern,
--> 100                                     self.value,
    101                                     use_regex=True))
    102

C:\Users\Zimmermann\Projects\coala\lib\coalib\settings\Setting.py in value(self)
    243                              'incomplete. Please access the value of the '
    244                              'setting in a section to get the complete value.')
--> 245         return self._value

AttributeError: 'Setting' object has no attribute '_value'
```

What's going on here? Why does CI work? Anyone else encountering that?
Is this real? ;)

jayvdb:
I dont see that on Linux.
@userzimmermann, What version of coala_utils do you have installed?

coalib/parsing/LineParser.py
jayvdb: Resolved; thanks!

tests/settings/SettingTest.py
```diff
@@ -14,9 +14,12 @@ class SettingTest(unittest.TestCase):
     def test_construction(self):
         self.assertRaises(ValueError, Setting, '', 2, 2)
         self.assertRaises(TypeError, Setting, '', '', '', from_cli=5)
+        self.assertRaisesRegex(TypeError, 'to_append.*',
```
jayvdb: `.*` is redundant

tests/settings/SettingTest.py
```diff
@@ -14,9 +14,12 @@ class SettingTest(unittest.TestCase):
...
+                               Setting, '', '', '', to_append=10)
 
     def test_path(self):
-        self.uut = Setting('key', ' 22\n', '.' + os.path.sep, True)
+        self.uut = Setting(
+            'key', ' 22\n', '.' + os.path.sep, strip_whitespaces=True)
```
jayvdb: These `strip_whitespaces=` are not necessary any longer.

I have no objection to them being changed, as it is more readable, but it isnt a required change for this changeset.

tests/settings/SettingTest.py
```diff
@@ -90,7 +93,17 @@ def test_typed_ordered_dict(self):
...
 
     def test_inherited_conversions(self):
-        self.uut = Setting('key', ' 22\n', '.', True)
+        self.uut = Setting('key', ' 22\n', '.', strip_whitespaces=True)
         self.assertEqual(str(self.uut), '22')
         self.assertEqual(int(self.uut), 22)
         self.assertRaises(ValueError, bool, self.uut)
+
+    def test_value_getter(self):
+        with self.assertRaisesRegex(ValueError, 'This property is invalid.*'):
```
jayvdb: `.*`

tests/settings/SettingTest.py
```diff
@@ -90,7 +93,17 @@ def test_typed_ordered_dict(self):
...
+            self.uut = Setting('key', '22\n', '.', to_append=True)
+            self.uut.value
+
+        with self.assertRaisesRegex(ValueError,
+                                    'Iteration on this object is invalid.*'):
```
jayvdb: `.*`

jayvdb:
A huge improvement, and I feel it is close to being finished, if it isnt already.
However there are still questions on this PR which should be answered, but have not been answered.

userzimmermann:
@jayvdb Stupid me ;) My outdated `coala_utils` were in fact the problem. Thanks :)

Udayan12167:
@jayvdb So except the Settings object being made unmodifiable there is no comment that I have left unresolved or not answered. If there is any specific one you are looking at can you please drop a ping. About the settings being made unmodifiable I am going to look at the `attrs` library but that is certainly going to take time and isn't absolutely necessary right now. Only your ack remains.

coalib/settings/Section.py
jayvdb: ping.

tests/coalaJSONTest.py
```diff
@@ -56,7 +57,9 @@ def test_fail_acquire_settings(self):
 
     def test_show_all_bears(self):
         with bear_test_module():
-            retval, output = execute_coala(coala.main, 'coala', '--json', '-B')
+            retval, output = execute_coala(coala.main,
+                                           'coala', '--json', '-B',
+                                           stdout_only=True)
```
jayvdb: When I remove this, the stderr stream contains:
`Implicit 'Default' section inheritance is deprecated. It will be removed soon. To silence this warning remove settings in the 'Default' section from your coafile. You can use dots to specify inheritance: the section 'all.python' will inherit all settings from 'all'.`

This is caused by the repos `.coafile`, which still has the deprecated `[Default]` section.
I think a better solution is to add command line argument `-I`, so that this and future changes to `.coafile` dont break this test case.
This is especially true because here we are using `bear_test_module`, which provides a different set of bears than are used by the repo `.coafile` , so there really should be lots of errors if the repo `.coafile` is not ignored.

tests/coalaTest.py
```diff
@@ -51,35 +51,36 @@ def test_python_version_34(self):
...
-                                       '-S', 'default.enabled=false')
+                                       '-S', 'cli.enabled=false')
         self.assertEqual(retval, 2)
         self.assertIn('Nothing to do.', output)
 
     def test_show_all_bears(self):
         with bear_test_module():
-            retval, output = execute_coala(coala.main, 'coala', '-B')
+            retval, output = execute_coala(
+                coala.main, 'coala', '-B', stdout_only=True)
```
jayvdb: again I think `-I` should be used instead of silencing stderr.
stderr shouldnt normally have any output here, and this will be masking bugs in the future.

tests/coalaTest.py
```diff
@@ -51,35 +51,36 @@ def test_python_version_34(self):
...
             self.assertEqual(retval, 0)
-            # 6 bears plus 1 line holding the closing colour escape sequence
+            # 6 bears and 1 line holding the closing colour escape sequence.
             self.assertEqual(len(output.strip().splitlines()), 7)
 
     def test_show_language_bears(self):
         with bear_test_module():
             retval, output = execute_coala(
-                coala.main, 'coala', '-B', '-l', 'java')
+                coala.main, 'coala', '-B', '-l', 'java', stdout_only=True)
```
jayvdb: `-I` again

tests/coalaTest.py
```diff
@@ -51,35 +51,36 @@ def test_python_version_34(self):
...
             self.assertEqual(retval, 0)
-            # 2 bears plus 1 line holding the closing colour escape sequence
+            # 2 bears and 1 line holding the closing colour escape sequence.
             self.assertEqual(len(output.splitlines()), 3)
 
     def test_show_capabilities_with_supported_language(self):
         with bear_test_module():
             retval, output = execute_coala(
-                coala.main, 'coala', '-p', 'R')
+                coala.main, 'coala', '-p', 'R', stdout_only=True)
```
jayvdb: `-I` again

coalib/settings/Section.py
jayvdb: ping

tests/misc/CachingTest.py
jayvdb: ping

jayvdb:
https://github.com/coala/coala-bears/pull/1348 is merged; it might fix your circle problems.

tests/misc/CachingTest.py
jayvdb: As discussed, this is a regression.
previously `coala -f abc.py -c .coafile` would check `abc.py` using bears in section `LineCounting `, but after this PR it would not.
This presumably only occurred if `files` was empty in `LineCounting` and `default`.  Still need to verify that.
This is possible a small regression and undesirable to support, even for the short term.
More analysis needed, to be very confident what exactly is being de-supported (It should be relnoted), and that it is unlikely anyone needed this.

jayvdb: As discussed on gitter, this would also apply to `-m`, which means [this PR](https://github.com/processing/processing/pull/4781/files) would fail after 0.10 docker is released

jayvdb: This is still here?
Is it necessary?

coalib/settings/ConfigurationGathering.py
```diff
@@ -209,6 +209,13 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
     log_printer.log_level = LOG_LEVEL.str_dict.get(str_log_level,
                                                    LOG_LEVEL.INFO)
 
+    if sections['cli'].contents and 'default' in sections:
+        logging.warning('Inheritance of CLI arguments to \'Default\' section '
+                        'has been deprecated. Please modify the \'Default\' '
+                        'section by using the --settings argument.')
```
sils: I don't feel this message is good, as a user I'm not supposed to use normal cli args anymore? I'm not supposed to use the default section anymore right? And if I do that my cli settings won't be stored, is that correct?

Udayan12167: What I wanted to say was that from further on the CLI args won't affect the whole coafile. That cli would be executed as an isolated section in itself.

Mixih:
postponed

tests/misc/CachingTest.py
jayvdb: Is `coala -f filename` (without `-b`) legal?

If so, then this change here shouldnt be necessary.

If not, have you created a test to check it emits an error message?

sils:
@Udayan12167 sorry we got a few test conflicts, I will certainly have time to review this today I think, can you try resolving them soon?

tests/misc/CachingTest.py
Udayan12167: @jayvdb I fixed it in https://github.com/coala/coala/pull/3178/commits/f1c1b30e12a1b543cf631e964ab469003694c525. It was needed here because it needed to be in the default section. In the later commit I merge them for proper deprecation and hence fixed it back. `coala -f filename` is needed in case a section doesn't have a files argument.

coalib/settings/ConfigurationGathering.py
```diff
@@ -262,7 +262,9 @@ def get_config_directory(section):
 
     However if its origin is already a directory this will be preserved:
 
-    >>> section['files'].origin = os.path.abspath('/tmp/dir/')
+    >>> files = section['files']
+    >>> files.origin = os.path.abspath('/tmp/dir/')
+    >>> section.append(files)
```
sils: how's this related to this commit? (https://github.com/coala/coala/pull/3178/commits/2df2936b7f6448c568b6252304c1c6e90d65742e)

sils:
ack https://github.com/coala/coala/pull/3178/commits/43a40fcfac4ea057c8c183253d375a2870306df1#diff-c9588a55c89d73e1bd3f6e2061e739d6L91

tests/settings/SettingTest.py
sils: agreed, I'd personally be willing to let this slip through to get this merged. Big PRs should be avoided before they happen but it's too late for that now and we gotta get this finished.

sils:
ack https://github.com/coala/coala/pull/3178/commits/2df2936b7f6448c568b6252304c1c6e90d65742e#r98327938

sils:
ack https://github.com/coala/coala/pull/3178/commits/f9b64fca069cb3bdf5a59980f3828f23228721fc

sils:
ack https://github.com/coala/coala/pull/3178/commits/20f612e63e98d525d27ffe03999e3647effa6e75 this is so super gorgeous

coalib/settings/ConfigurationGathering.py
Udayan12167: https://github.com/coala/coala/pull/3178/commits/2df2936b7f6448c568b6252304c1c6e90d65742e#diff-41b4f7bdbd660f8c693c7d0304602d79R205 . Because of this here it doesn't allow changing the settings within a Section without replacing them.

Udayan12167:
reack 944718c

Udayan12167:
reack 931483f

Udayan12167:
reack 3314a1c

Udayan12167:
reack b0fd0d0

coalib/coala_main.py
sils: IMO ok and doesn't need deprecation

coalib/settings/ConfigurationGathering.py
sils: so why do you remove it again? We still want the correct deprecation warning don't we?

sils: and doesn't it still make sense here? They're from the cli after all.

sils:
unack erhttps://github.com/coala/coala/pull/3178/commits/4120645e422e86049b208b31d8436148ae702f85#diff-8ecc903ba7ecebebacc61f64149f8eb7R193 let's improve message as pointed out on gitt

sils:
ack https://github.com/coala/coala/pull/3178/commits/02ad795a735d1ae38a8b06cba069546c53575266

sils:
ack https://github.com/coala/coala/pull/3178/commits/ded977f150a784b8a5ac5f8a1540c422d9981b77

sils:
ack https://github.com/coala/coala/pull/3178/commits/8cf84d683abcdaa8c8e7d770024f1a77435bef1f

sils:
ack https://github.com/coala/coala/pull/3178/commits/0acecba5fa0ee65fd3929d5a6a39dc220157a921

sils:
ack https://github.com/coala/coala/pull/3178/commits/74bd3b546e0f1074c64737c14afa05f99d09a5cf

Udayan12167:
@rultor merge

Issue #3191, opened 2016-12-09
Allow Travis-CI builds


jayvdb:
We've had some problems with Circle CI, so we need an alternative ATM.

Mixih:
Also look into drone.io adtac seem to recommend it.

PR #3206, opened 2016-12-10
Basic Travis CI precise build script

jayvdb:
Uses ppa:staticfloat/juliareleases for Julia.
Uses Travis apt addon for R and clang.

Closes https://github.com/coala/coala/issues/3191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,46 @@
+language: python
+python:
+  - 3.4
+  - 3.5
+  - 3.6-dev
+
+addons:
+  apt:
+    packages:
+      - r-base
+      - clang-3.4
+    sources:
+      - ubuntu-toolchain-r-test
+      - llvm-toolchain-precise
+
+before_install:
+  - sudo add-apt-repository -y ppa:staticfloat/juliareleases
+  - sudo apt-get update
+  - sudo apt-get install julia
+  # https://github.com/coala/coala/issues/3183
+  # https://github.com/coala/coala/issues/3184
+  - pip install -U py==1.4.29 pip
+  - cat test-requirements.txt docs-requirements.txt >> requirements.txt
+  - sed -i '/^-r/d' requirements.txt
+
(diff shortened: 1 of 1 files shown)


jayvdb:
Green build at https://travis-ci.org/jayvdb/coala/builds/182846769

jayvdb:
Gah.  Another "Merge" commit https://github.com/coala/coala/commit/ad3bfc023086ac8e4

maybe we need to solve https://github.com/coala/coala-bears/issues/1037 first.

.travis.yml
```diff
@@ -0,0 +1,45 @@
+language: python
+python:
+  - 3.4
+  - 3.5
```
sils: can you add 3.6 while we're at it?

jayvdb: [no](https://travis-ci.org/jayvdb/coala/builds/182831634).  well, `3.6` hasnt been added to Travis yet. maybe they have it available under a different label, but I havent looked.  we can add it soon, I am sure.

Mixih: Is that going to slow the the testing process even further though?

jayvdb: oi! The Travis builds are faster than Circle CI :P
But, no, Travis runs the jobs in parallel, so adding more jobs doesnt impact elapsed time until there are more than 5 concurrent jobs, and I have seen up to 15 concurrent OSS jobs in Travis off-peak times.

sils: use `3.6-dev`

sils: not sure, you may need to give it as a string: https://docs.travis-ci.com/user/languages/python/#Default-Python-Version

Mixih: It's working now in the latest build (I think)

sils:
ack https://github.com/coala/coala/pull/3206/commits/527c7a46eae46594142ec67959ef93e70932e711

Mixih:
ack 527c7a4

Issue #3212, opened 2016-12-10
Removal of autoapply in run_coala causes difficulty when using coala in a serverside checking application


Mixih:
The removal of auto-apply parameter from run_coala means that coala still auto applies applies it's actions when run from run_coala. We need to have a way for the user to disable auto-patch application so coala can be used to show code errors, not fix them. (This almost broke gitmate)

sils:
marking critical importance because this affects gitmate and thus also us!

sils:
CC @fneu @seblat or anyone really, if you have time, fix this please

Mixih:
decided to add `autoapply` parameter. default it to true, when false, don't apply action.
per gitter https://gitter.im/coala/coala?at=584c5c47c29531ac5d480430

Mixih:
I advise we patch coala-incremental for now as it's critical and testing will be a mess. It's appearently a tiny workaround there that will be easy to implement. https://gitter.im/coala/coala?at=584c59ccaeb49008046ff5d0

Mixih:
https://github.com/coala/coala/pull/3216 closes. Due to a failure on my part, I failed to catch the multiple commit issues before merging. *sigh*

PR #3216, opened 2016-12-10
https://github.com/coala/coala/issues/3212

sils:
https://github.com/coala/coala/issues/3212

### DIFF
--- a/coalib/coala_main.py
+++ b/coalib/coala_main.py
@@ -24,6 +24,7 @@ def run_coala(console_printer=None,
               acquire_settings=fail_acquire_settings,
               print_section_beginning=do_nothing,
               nothing_done=do_nothing,
+              autoapply=True,
               force_show_patch=False,
               arg_parser=None,
               arg_list=None):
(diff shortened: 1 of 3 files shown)


Mixih:
Not exactly sure how we're going to test this as all run_coala options are tests as parts of coala-modes. Do we need a new mode, or should we write a test for this file?

coalib/coala_main.py
```diff
@@ -91,7 +94,9 @@ def run_coala(console_printer=None,
             if not section.is_enabled(targets):
                 continue
 
-            if force_show_patch:
+            if not autoapply:
+                section['default_actions'] = ''
+            elif force_show_patch:
                 section['default_actions'] = '*: ShowPatchAction'
```
Mixih: I think you need to make it clear in the doc string that `autoapply` and `force_show_patch` cannot be used together

coalib/coala_main.py
```diff
@@ -49,6 +50,9 @@ def run_coala(console_printer=None,
     :param nothing_done:            A callback that will be called with only a
                                     log printer that shall indicate that
                                     nothing was done.
+    :param autoapply:               Set this to false to not autoapply any
+                                    actions. If you set this to `True`,
```
Mixih: I think you mean `False`

Mixih:
ack 06a772f

Mixih:
ack cb1c7d

sils:
@rultor merge

Issue #2649, opened 2016-08-20
Diff: `add_line` function


Makman2:
it's already possible to add several lines at once, to delete several ones and to delete a single one.

`add_line` should be implemented too.

Nosferatul:
Hello, I wanna work on this!

sils1297:
Hey - nice that you came back :)! You know how it goes, if you need help, we'll be there for you :)

Nosferatul:
Sure, thank you!

Nosferatul:
What differences should be between 'add_line' and 'add_lines' ?

Makman2:
`add_line` for just adding a single line without the need for creating a list^^

Nosferatul:
I don't know where is created a list :(. `linediff = self._get_change(line_nr_before, min_line=0)` create a new object of type DiffLine if doesn't exist an object with key = line_nr_before in list __changes_. I believe that `linediff.add_after` is just a variable from DiffLine.

sils1297:
ahm @Makman2 I'm not so sure if we don't just want to remove add_line and delete_line altogether, passing a list with one element is quite trivial

Nosferatul:
If we want to remove delete_line, it will change delete_lines also.

Makman2:
I wouldn't remove `delete_line` (after some thinking), as it's a bit redundant to do:
```
delete_lines(3, 3)
```
this is a bit nicer to write and to read:
```
delete_line(3)
```
and to be consequent it's maybe good to have also `add_line`? not sure any more...

ashwini0529:
Is anyone working on this issue?

sils:
@Nosferatul ?

Nosferatul:
sorry, you can take it @ashwini0529 .

ashwini0529:
@Nosferatul Alright. I'll start working on it after 30th Nov.

ashwini0529:
@sils Can you please guide me like what should exactly be done and brief me about the issue?

Makman2:
@ashwini0529 please don't ping sils for such generic questions, other people can answer too. Especially those who filed the issue ;)

For what's to be done: Please read the first and uppermost comment at the top

ashwini0529:
Hey. Can you explain me how the `add_line` functionality would work? And which files do I need to change?

Nosferatul:
check Diff.py. `add_line` needs to add a single line, we have already the `add_lines`

ashwini0529:
[...]

PR #3218, opened 2016-12-11
Diff: Implement function add_line

ashwini0529:
Add a function, add_line, to add a line after a line number
Fixes https://github.com/coala/coala/issues/2649

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -381,6 +381,16 @@ class Diff:
         linediff.add_after = lines
         self._changes[line_nr_before] = linediff
 
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the addition.
+                               Use 0 to insert line before everything.
+        :param line:           Line to add.
+        """
+        return self.add_lines(line_nr_before, [line])
+
     def change_line(self, line_nr, original_line, replacement):
         r"""
         Changes the given line with the given line number. The replacement will
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -17,6 +17,11 @@ class DiffTest(unittest.TestCase):
         self.uut.add_lines(0, ['t'])
         self.uut.add_lines(0, [])
 
+    def test_add_line(self):
+        self.uut.add_line(0, 't')
+        self.assertRaises(ConflictError, self.uut.add_line, 0, 't')
+        self.assertEqual(self.uut.modified, ['t', '1', '2', '3', '4'])
+
     def test_double_addition(self):
         self.uut.add_lines(0, ['t'])
 


Nosferatul:
Needs test

Nosferatul:
Also, `Fixes` -> `Closes`

ashwini0529:
@Nosferatul Can you please guide me regarding tests?

Nosferatul:
Check DiffTest.py, it should be similar with `test_add_lines` method

ashwini0529:
@Nosferatul Is this fine?
Added test for `add_line`

Nosferatul:
check what gitmate-bot doesn't like. The test it's not enough, I'm not sure why, check travis

ashwini0529:
@Nosferatul The logs gives the following error : 
selection_027

I'm not sure what's wrong in my commit.

Nosferatul:
first resolve the gitmate-bot problem and push again to see if travis has the same error

ashwini0529:
@Nosferatul Already resolved `gitmate-bot` issue.

ashwini0529:
I guess `gitmate-bot` won't mind now. 😅 I already solved that gitmate-bot code fix.

tests/results/DiffTest.py
```diff
@@ -17,6 +17,11 @@ def test_add_lines(self):
         self.uut.add_lines(0, ['t'])
         self.uut.add_lines(0, [])
 
+    def test_add_line(self):
```
aptrishu: You'll have put that under assert statement, other wise you see you're getting conflicterror. 
Dosomething like
`self.assertRaises(ConflictError, self.uut.add_line, 0, 't')`
after adding the first line.

ashwini0529: Oh So that was the way the test was supposed to be written.
Thanks :+1: 
I'll try it right now.

ashwini0529: Hey, I added that into test. And it returned the error again. :/
Did I write something wrong?

ashwini0529: Passed successfully

ashwini0529:
I was wondering if some changes were needed?

coalib/results/Diff.py
```diff
@@ -381,6 +381,19 @@ def add_lines(self, line_nr_before, lines):
         linediff.add_after = lines
         self._changes[line_nr_before] = linediff
 
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the additions.
```
aptrishu: `additions` ? there is going to be a single addition right?
also in the conflicterror it would say `Cannot add "lines"` but there is a single line.

coalib/results/Diff.py
```diff
@@ -381,6 +381,19 @@ def add_lines(self, line_nr_before, lines):
...
+                               Use 0 for insert lines before everything.
```
aptrishu: `Use 0 for insert "lines"` -> `Use 0 to insert the line`

ashwini0529:
@aptrishu done!

coalib/results/Diff.py
ashwini0529: yeah! I wrote complete function for `add_line` , but @Makman2 suggested to use the `add_lines` function. 
I've modified the ConflictError's to `line(s)`.
That will cover both the cases.

coalib/results/Diff.py
```diff
@@ -375,12 +375,25 @@ def add_lines(self, line_nr_before, lines):
...
 
         linediff.add_after = lines
         self._changes[line_nr_before] = linediff
 
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the addition.
+                               Use 0 to insert lines before everything.
```
Nosferatul: `lines` -> `line`

coalib/results/Diff.py
```diff
@@ -375,12 +375,25 @@ def add_lines(self, line_nr_before, lines):
 
         linediff = self._get_change(line_nr_before, min_line=0)
         if linediff.add_after is not False:
-            raise ConflictError('Cannot add lines after the given line since '
-                                'there are already lines.')
+            raise ConflictError('Cannot add line(s) after the given line since '
+                                'there are already line(s).')
```
Makman2: I think you don't need to change this, as this is the generalizing plural^^

coalib/results/Diff.py
```diff
@@ -375,12 +375,25 @@ def add_lines(self, line_nr_before, lines):
...
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the addition.
+                               Use 0 to insert line before everything.
+        :param line:           Line to add
+        """
+        if line == '':
+            return  # No action
```
Makman2: no need for this, `add_lines` does this check by itself^^

Makman2:
unack c70a4b1

coalib/results/Diff.py
ashwini0529: So what should be done here? 
Should it remain as it is or should I change this to something else?

ashwini0529: I think using `line(s)` would work in both the cases. When a user gets a ConflictError on adding a single line, it is understood that user has added a single line.

Even if `add_lines` is used, and a single line in a `list` is passed, then also it will treat it as a multiple lines.
So using a generic statement would be fine.

Makman2: even when using `add_line`, the message makes sense when it stays like before :)

ashwini0529: So should I let it be as it is? i.e `lines` ?

ashwini0529: @aptrishu told to look for single line ConflictError as well.
I think we should go with `line(s)` . This will make things generic.
Otherwise it would require another `condition check` to see if the array contains only one element and to assert different `ConflictError` message for that which doesn't makes sense.

Makman2: what I mean is that the sentence, though it's written with the plural form of `line` -> `lines`, `lines` is meant generically. It's like a general condtition that holds:
"You can't generally add lines (whether these are 1 or many), when there are already lines (1 or many)"

I know this is slightly nitpicking, though I'm not a fan of making changes that aren't necessary for the commit^^ And in this case I believe the grammar is totally okay, whether a single or multiple lines are meant.

ashwini0529: So should I change this again to `lines` ?

Makman2: yes please keep the old version :)

ashwini0529: Just gimme a minute.

ashwini0529: @Makman2 Done!

coalib/results/Diff.py
```diff
@@ -381,6 +381,17 @@ def add_lines(self, line_nr_before, lines):
...
 
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the addition.
+                               Use 0 to insert line before everything.
+        :param line:           Line to add
+        """
+
```
Makman2: redundant newline ;)

coalib/results/Diff.py
```diff
@@ -381,6 +381,17 @@ def add_lines(self, line_nr_before, lines):
         linediff.add_after = lines
         self._changes[line_nr_before] = linediff
 
+    def add_line(self, line_nr_before, line):
+        """
+        Adds line after the given line number.
+
+        :param line_nr_before: Line number of the line before the addition.
+                               Use 0 to insert line before everything.
+        :param line:           Line to add
```
ashwini0529: Fixed.

tests/results/DiffTest.py
```diff
@@ -17,6 +17,10 @@ def test_add_lines(self):
...
+        self.uut.add_line(0, 't')
+        self.assertRaises(ConflictError, self.uut.add_line, 0, 't')
```
Makman2: could you make a test whether successfully a line was added? Meaning to check the diff stats / changed version.

ashwini0529: I didn't get you. What extra test needs to be added?

Makman2: a test like
```
self.assertEqual(self.uut.changed, ['t', '2', '3', '4'])
```

ashwini0529:
@aptrishu Do I need to change tests or not?

aptrishu:
You need to add tests. right now your tests only checks for errors. It doesn't check if the line is actually added there or not.

tests/results/DiffTest.py
ashwini0529: There's nothing as `uut.changed` :/ 
What should be done here?

Nosferatul: try `change_line`

ashwini0529: I did change that to `change_line`.
Ended up getting this error.
https://travis-ci.org/coala/coala/jobs/183574360#L1001

ashwini0529: So something like `self.assertEqual(self.uut.modified, ['t', '2', '3', '4'])`

Makman2: my example was wrong, you add a line instead of changing the first one^^

ashwini0529: `self.assertRaises(ConflictError, self.uut.add_line, 1, 't')` This?

Makman2: no:
```
self.assertEqual(self.uut.modified, ['t', '1', '2', '3', '4'])
```
so the test output is correct, the test is wrong^^

ashwini0529: Done!

Makman2:
ack 2913570

Makman2:
@ashwini0529 you need to rebase :)

ashwini0529:
@Makman2 Rebased! :)
Hope this is good to go! 
Looking forward to work on other issues! :)

aptrishu:
ack e44b314

aptrishu:
@rultor merge

Makman2:
yes congrats! pick a new one that you like :+1: :)

Issue #3310, opened 2016-12-18
Python 3.3 version check does not work


jayvdb:
It is possible to install coala on any Python 3.3, except perhaps Python 3.3.0

The logic in https://github.com/coala/coala/blob/569d894/coalib/__init__.py#L25 is incorrect.

In Python 3.3
```python
>>> from coalib import assert_supported_version
>>> assert_supported_version()
>>> import sys
>>> print('true') if not sys.version_info > (3, 3) else print('false')
false
>>> print('true') if not sys.version_info[0:2] > (3, 3) else print('false')
true
```

jayvdb:
Appropriate that this issue is numbered 3310 ;-)

PR #3312, opened 2016-12-18
assert_supported_version: Reject Python 3.3.x

jayvdb:
The logic in 805d8a0c to prevent installation on Python 3.3
was incorrect as Version 3.3.1 is greater than Version 3.3,
so only 3.3.0 was rejected by the current logic.

Fixes https://github.com/coala/coala/issues/3310

### DIFF
--- a/coalib/__init__.py
+++ b/coalib/__init__.py
@@ -21,7 +21,7 @@ VERSION = get_version()
 __version__ = VERSION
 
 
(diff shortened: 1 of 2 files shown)


sils:
ack https://github.com/coala/coala/pull/3312/commits/7f4edb8954156652beba359047d983a28ae62089

sils:
@rultor merge

PR #3328, opened 2016-12-18
setup.py: Move Python version assert after imports

jayvdb:
Avoids pycodestyle import errors.

All of the imports from coalib do not depend on large amounts of coalib.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,5 @@
 #!/usr/bin/env python3
 
-# Start ignoring PyImportSortBear as imports below may yield syntax errors
-from coalib import assert_supported_version, VERSION, get_version
-
-assert_supported_version()
-# Stop ignoring
-
 import datetime
 import locale
 import platform
@@ -14,14 +8,19 @@ from os import getenv
 from subprocess import call
 
 import setuptools.command.build_py
-from coalib.misc.BuildManPage import BuildManPage
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
+
+from coalib import VERSION, assert_supported_version, get_version
+from coalib.misc.BuildManPage import BuildManPage
 
 try:
     locale.getlocale()
 except (ValueError, UnicodeError):
     locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
+
+
+assert_supported_version()
 
 
 class BuildPyCommand(setuptools.command.build_py.build_py):
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 python:
+  - 2.7
+  - 3.3
   - 3.4
   - 3.5
   - 3.6-dev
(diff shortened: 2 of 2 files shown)


sils:
@jayvdb the big problem is python 2 because of the syntax differences!

sils:
I think this would break it.

sils:
if you add a test that works with python 2 that'd be actually really great, marking WIP for now

jayvdb:
Sigh.  This is all patching around the wrong approach.  setup.py should not import the module it is trying to check dependencies for.

But even with that design, you could easily check that my origin patch df119eb worked on Python 2.7.

With 1886221 we make Travis slower to include Python 2.7.

fwiw, currently this is what happens on Python 2.6, so I havent added Python 2.6 to Travis as it would make the build script even more ugly just to prove my point.

```
Collecting coala_utils~=0.5.1 (from -r requirements.txt (line 2))
  Downloading coala_utils-0.5.1.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-UJTZGj/coala-utils/setup.py", line 17
        maintainers={'Adrian Zatreanu', 'Alexandros Dimos', 'Lasse Schuirmann'},
                                      ^
    SyntaxError: invalid syntax
```

We do want to break unsupported versions of Python, but not like that.

The most appropriate way to do that in `setup.py`  is checking needed features, without importing the package that has not yet been installed.
e.g.
```
import sys
try:
    import selectors
except:
    print('....')
    sys.exit(4)
import os
```

That passes static analysis tools based on pyflakes, and is explicit about what Python system library features are *used*.
An explicit `sys.version_info` check can appear afterwards, if necessary.

The most appropriate way to do that with packaging is by not distributing `py2` wheels (done), and then shipping py34,py35,py36 wheels only, or using Environment Markers in the wheel so that the wheel installer knows it isnt compatible with Python 3.3 or lower (not yet done).

Mixih:
@jayvdb @sils Is the discussion here resolved?

sils:
yes, this is good to go from my side since his tests obviously prove that his solution is indeed sufficient and will ensure it stays so in the future. Without the tests I'd be afraid of the past problems coming up again.

Mixih:
So the python2 and 3.3 stuff in travis is just to check depreciated versions really are depreciated right? Is this a permanent change?

setup.py
```diff
@@ -14,16 +8,21 @@
 from subprocess import call
 
 import setuptools.command.build_py
-from coalib.misc.BuildManPage import BuildManPage
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
+from coalib import assert_supported_version, VERSION, get_version
```
Mixih: should we sort these alphabetically? `assert_supported_version, VERSION, get_version`

jayvdb:
I believe that the python2 and 3.3 stuff in travis is permanent, to ensure there is no regression.

setup.py
jayvdb: done

sils:
Yes, we had exactly this regression in the past multiple times

Mixih:
ack d960d12

jayvdb:
@rultor merge

PR #3378, opened 2016-12-23
TravisCI: Error on failed tests

jayvdb:
d960d125 moved .misc/tests.sh into an if statement,
followed by other commands.  This meant that if tests.sh
failed the exit status would be lost.


Mixih:
ack fd5428b

Mixih:
@rultor merge

PR #3380, opened 2016-12-23
TravisCI: Add OSX job

jayvdb:


jayvdb:
See 57754019 for creation of previous OSX CI.

Mixih:
Please rebase.

Mixih:
Huh, looks like diff.py isn't covered now :(

jayvdb:
And if I [remove the env var](https://github.com/jayvdb/coala/commit/0ff23c5ac2bf7627c968b107304f07b7fab7d5e3), Diff [doesnt have coverage](https://travis-ci.org/jayvdb/coala/jobs/187441373#L1139) because:
> SKIP [1] tests/results/DiffTest.py:194: dlsym(RTLD_DEFAULT, clang_annotateTokens): symbol not found. Please ensure that your python bindings are compatible with your libclang.so version.

It is actually finding a different clang which is also installed, which is why the message isnt "image not found".

sils:
@jayvdb you got conflicts

jayvdb:
I got lots of self-conflicts.
will rebase and merge when they are ack'ed.

.travis.yml
```diff
@@ -18,16 +27,36 @@ addons:
...
 
 before_install:
+  - if [[ $TRAVIS_OS_NAME == osx ]]; then TRAVIS_PYTHON_VERSION=3.5; fi
   - export TRAVIS_PYTHON_VERSION_MAJOR=${TRAVIS_PYTHON_VERSION%.[0-9]}
   - export TRAVIS_PYTHON_VERSION_MINOR=${TRAVIS_PYTHON_VERSION#[0-9].}
-  - sudo add-apt-repository -y ppa:staticfloat/juliareleases
-  - sudo apt-get update
-  - sudo apt-get install julia
+
+  - >
```
Mixih: Just wondering why you have these `>` even though travis's examples don't....

Mixih:
I'm not quite sure what the rationale behind this was as @sils mentioned that OSX CI mostly outputs the same stuff as linux CI. However, I'm acking on the assumption that this was intended to introduce official OSX support.

Mixih:
ack cf5d2e3

.travis.yml
```diff
@@ -6,6 +6,15 @@ python:
...
 
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode6.4
+      language: generic
+      python: 3.5
+      env:
+        - DYLD_LIBRARY_PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/
+
```
nemaniarjun: Just add the directories you want to cache as done in the following file. 

https://github.com/NIT-dgp/UIP/blob/master/.travis.yml

Mixih:
unack cf5d2e3

Mixih:
I strongly recommend caching implementation as @nemaniarjun described because of the time increase.

.travis.yml
```diff
@@ -16,13 +28,26 @@ addons:
...
+      sudo ln -sf $(which python3) $(which python)
+      sudo ln -sf $(which pip3) $(which pip 2>/dev/null || echo /usr/bin/pip)
+    fi
+
   # https://github.com/coala/coala/issues/3183
   # Travis automatically installs the `requirements.txt` in "install" stage
   - cp requirements.txt requirements.orig
   - cat test-requirements.txt docs-requirements.txt >> requirements.txt
-  - sed -i '/^-r/d' requirements.txt
+  - sed '/^-r/d' requirements.txt > requirements.new && mv requirements.new requirements.txt
```
sils: what's the point of this moving instead of just doing it in place? I think a comment would be nice

sils:
I think if we have a comment there (or revert the behaviour back to using in place) we're good to go here

.travis.yml
jayvdb: It isnt supported by OSX.
I had a comment "sed -i isnt supported by OSX", but it seemed a bit silly because the line isnt using -i , and there are multiple cases of this, and anyone who changes it will get a broken build anyway.

sils: http://stackoverflow.com/a/22084103/3212182 `sed -i.bak 's/foo/bar/' filename` is supposed to work with both (and creates a backup file which isn't harmful) and better readable. When I see this line I immediately begin to look for hours on "why did he do that" and I think having that sed with backup file expression will really be better

sils:
marking WIP as the rest looks good, if you disagree with https://github.com/coala/coala/pull/3380#discussion_r94311997 respond and mark pending review again

.travis.yml
jayvdb: Then you have everyone wondering what `-i.bak` means. It is shorter. That is all.
Explicit is better than implicit.

Mixih:
ack d2d67fa

Mixih:
looks like we are going with the .bak syntax

jayvdb:
@rultor merge

Issue #3155, opened 2016-12-05
source lines are not available as a --format placeholder


jayvdb:
Current `--format` placeholders are: origin, file, line, end_line, column, end_column, severity, severity_str, message .

It would be useful to have the source lines available as a placeholder, as that can be used to specify a custom format that is more similar to the current default UI.

sils:
not too sure this is smart, currently the main assumption is that the format string is a one liner, this'll definitely introduce multiple lines and it's likely badly parsable. At least it shouldn't be active by default IMO

jayvdb:
Ya, I've no intention for it to be active by default. That would be a breaking change.
Also we could use a new line marker instead of `\n`, like pygments uses.

Abhisek-:
If we render everything in on line without using newline, then it will become too confusing to understand as what is wrong with source file. Not sure about the pygments thing

PR #3386, opened 2016-12-24
Add source_lines as --format placeholder

Abhisek-:
ConsoleInteraction.py: source_lines is added
for --format placeholder to know about affected
lines of a file.
    
Closes https://github.com/coala/coala/issues/3155

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -292,6 +292,7 @@ def print_diffs_info(diffs, printer):
 def print_results_formatted(log_printer,
                             section,
                             result_list,
+                            file_dict,
                             *args):
     """
     Prints results through the format string from the format setting done by
@@ -329,6 +330,8 @@ def print_results_formatted(log_printer,
 
             for range in result.affected_code:
                 format_args['affected_code'] = range
+                format_args['source_lines'] = range.affected_source(file_dict)
+
                 print(format_str.format(file=range.start.file,
                                         line=range.start.line,
                                         end_line=range.end.line,
--- a/coalib/results/SourceRange.py
+++ b/coalib/results/SourceRange.py
@@ -120,6 +120,47 @@ class SourceRange(TextRange):
                                        tr.end.line,
                                        tr.end.column)
 
+    @enforce_signature
+    def affected_source(self, file_dict: dict):
+        r"""
+        Tells which lines are affected in a specified file within a given range.
+
+        >>> from os.path import abspath
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=2)
+        >>> sr.affected_source({
+        ...     abspath('file_name'): ('def fun():\n', '    x = 2  \n')
+        ... })
+        ('    x = 2  \n',)
+
+        If more than one line is affected.
+
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=3)
+        >>> sr.affected_source({
+        ...     abspath('file_name'): ('def fun():\n',
+        ...                            '    x = 2  \n', '    print(x)  \n')
+        ... })
+        ('    x = 2  \n', '    print(x)  \n')
+
+        If the file indicated at the source range is not in the `file_dict` or
+        the lines are not given, this will return `None`:
+
+        >>> sr = SourceRange.from_values('file_name_not_present',
+        ...     start_line=2, end_line=2)
+        >>> sr.affected_source({abspath('file_name'):
+        ...     ('def fun():\n', '    x = 2  \n')})
+
+        :param file_dict:
+            It is a dictionary where the file names are the keys and
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            A tuple of affected lines in the specified file.
+            If the file is not affected or the file is not present in
+            ``file_dict`` return ``None``.
+        """
+        if self.start.file in file_dict and self.start.line and self.end.line:
+            # line number starts from 1, index starts from 0
+            return file_dict[self.start.file][self.start.line - 1:self.end.line]
+
     def __json__(self, use_relpath=False):
         _dict = get_public_members(self)
         if use_relpath:
--- a/tests/output/ConsoleInteractionTest.py
+++ b/tests/output/ConsoleInteractionTest.py
@@ -802,7 +802,7 @@ class PrintFormattedResultsTest(unittest.TestCase):
             print_results_formatted(self.logger,
                                     self.section,
                                     [Result('1', '2', affected_code)],
-                                    None,
+                                    {},
                                     None)
             self.assertRegex(stdout.getvalue(), expected_string)
 
@@ -836,3 +836,16 @@ class PrintFormattedResultsTest(unittest.TestCase):
                                 None,
                                 None,
                                 None)
+
+    def test_source_lines(self):
+        self.section.append(Setting(key='format', value='{source_lines}'))
+        affected_code = (SourceRange.from_values('file', 2, end_line=2),)
+        with retrieve_stdout() as stdout:
+            print_results_formatted(
+                self.logger,
+                self.section,
+                [Result('SpaceConsistencyBear', message='msg',
+                        affected_code=affected_code)],
+                {abspath('file'): ('def fun():\n', '    pass  \n')})
+            self.assertEqual(stdout.getvalue(),
+                             "('    pass  \\n',)\n")
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -100,7 +100,7 @@ To run coala without user interaction, run the `coala --non-interactive`,
              '"Message: {message}"; possible placeholders: '
              'id, origin, file, line, end_line, column, end_column, '
              'severity, severity_str, message, message_base, '
-             'message_arguments, affected_code')
+             'message_arguments, affected_code, source_lines')
 
     config_group = arg_parser.add_argument_group('Configuration')
 


sims1253:
Maybe add an ConsoleInteraction identifier to the beginning of your commit message for reference

coalib/output/ConsoleInteraction.py
```diff
@@ -312,16 +314,30 @@ def print_results_formatted(log_printer,
...
+
+                if args[0] and (
+                        range.start.line and range.end.line):
+                    source_lines = []
+                    file = range.start.file
+                    idx = int(range.start.line)
+
+                    while idx <= int(range.end.line):
+                        source_lines.append(args[0][file][idx - 1])
+                        idx += 1
```
sils: this is long and complicated, try to simplify it. if that's not possible, make a function, we probably want this to be a member function of the object anyway

sils:
you gotta add the new param to the help in DefaultArgParser as well for users to see it's available

Abhisek-:
@sils I've added a new param to the help in `DefaultArgParser` in the current commit. Is there something wrong with that ?About other things I'll change as suggested.

sils:
sounds good if there was no misunderstanding, just push it so we can see it when you're done

coalib/output/ConsoleInteraction.py
```diff
@@ -312,16 +314,30 @@ def print_results_formatted(log_printer,
...
             for range in result.affected_code:
+                source_lines = None
+
+                if args[0] and (
+                        range.start.line and range.end.line):
+                    source_lines = []
+                    file = range.start.file
+                    idx = int(range.start.line)
+
+                    while idx <= int(range.end.line):
```
Abhisek-: Oh..I forgot about `extend`. I though of first one before, but it couldn't be possible because of [this](https://github.com/coala/coala/issues/3398). thanks :+1:

coalib/results/Result.py
```diff
@@ -264,6 +264,39 @@ def location_repr(self):
...
 
+    @enforce_signature
+    def get_affected_source_lines(self,
+                                  file_name: str,
+                                  start_line: (int, None) = None,
+                                  end_line: (int, None) = None):
+        """
+        Tells which are the lines affected in the file within a specified
+        range. If both start_line and end_line are None return all the
+        affected lines of the specified file.
```
sils: can we have a doctest here?

sils:
its #3394

sils:
@Abhisek- the other PR was merged, can you rebase? A few things changed a bit, I think they'll make your stuff easier possibly.

sils:
It may even be better to make a new PR since I imagine your code will change as well, you will know best what to do

Abhisek-:
sorry for the delay.

sils:
no issue

coalib/output/ConsoleInteraction.py
```diff
@@ -322,12 +323,20 @@ def print_results_formatted(log_printer,
...
                 format_args['affected_code'] = range
-                print(format_str.format(file=range.start.file,
-                                        line=range.start.line,
-                                        end_line=range.end.line,
+                file_name = range.start.file
+                start_line = range.start.line
+                end_line = range.end.line
+                source_lines = result.get_affected_source_lines(file_name,
+                                                                start_line,
+                                                                end_line)
```
sils: can't you please add this to the `format_args` dict? That's what it's for. Then you don't need to give the args explicitly below

coalib/results/Result.py
```diff
@@ -264,6 +264,55 @@ def location_repr(self):
         return ', '.join(repr(relpath(range_path))
                          for range_path in sorted(range_paths))
 
+    @enforce_signature
+    def get_affected_source_lines(self,
```
sils: actually I think this function should reside in the SourceRange class instead! Then you also don't need to pass filename, start and end line and you'll just use `range.affected_source`, maybe even make it a property.

The way of doctesting you do here is great :)

Abhisek-: `get_affected_source_lines` uses the contents of a file. `SourceRange` class only has the information about line numbers and file name and not about contents of a file, so I can not make that as a property.

Probably I will move that function to `SourceRange` class and pass file content information to the file instead of start, end and filename.

coalib/output/ConsoleInteraction.py
```diff
@@ -322,6 +323,8 @@ def print_results_formatted(log_printer,
 
             for range in result.affected_code:
                 format_args['affected_code'] = range
+                format_args['source_lines'] = range.affected_source(args[0])
```
sils: args[0] isn't well readable, just add the file_dict arg above and use it directly

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,47 @@ def expand(self, file_contents):
...
+        range.
+
+        >>> from os.path import abspath
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=2)
+        >>> sr.affected_source({abspath('file_name'):
+        ...     ('def fun():\n', '    x = 2  \n')})
+        ['    x = 2  \n']
+
+        If there is no file to search for the contents then it will
+        produce None as the result.
```
sils: enclose none with backticks

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,47 @@ def expand(self, file_contents):
                                        tr.end.line,
                                        tr.end.column)
 
+    @enforce_signature
+    def affected_source(self, file_dict: (dict, None)):
```
sils: why allow None anyway and then have special handling for that?

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,47 @@ def expand(self, file_contents):
...
+            List of lines that are affected in the specified file.
+            If the file is not affected return None
+        """
+        if file_dict:
+            source_lines = []
+            file_contents = file_dict[self.start.file]
+            if self.start.line and self.end.line:
+                # line number starts from 1, index starts from 0
+                for line_number in range(self.start.line - 1, self.end.line):
+                    source_lines.append(file_contents[line_number])
```
sils: how about you `yield file_contents[line_number]`? (Google python yield if you're unsure how that works)

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,47 @@ def expand(self, file_contents):
...
+        ...     ('def fun():\n', '    x = 2  \n')})
+        ['    x = 2  \n']
+
+        If there is no file to search for the contents then it will
+        produce None as the result.
+
+        >>> sr.affected_source(None)
+        >>>
+
+        If more than one line is infected.
```
Abhisek-: affected

coalib/results/SourceRange.py
Abhisek-: Because there are some test cases whose `file_dict`(`args[0]`) is None. maybe I should handle the `None` case inside `ConsoleInteraction.py`. But I thought handling `None` inside `affected_source` is more meaningful. What do you say ?

coalib/results/SourceRange.py
Abhisek-: `yield file_contents[line_number]` generates one result upon one call. For `range.start.line=3` and `range.end.line=5` we have to call the generator `3` times to get  the affected source lines. That means again a for loop in `ConsoleInteraction.py`. Is it ok ? I was printing the `list` of the affected lines. May be you are telling to print the *each* source line on its *own* line. Should we print each source line on its own line ? Is it something else you are saying ?

Abhisek-: If we print *each* source line on its *own* line, then a format like `"{line}: {end_line} {source_lines}"` will end up in 

    3: 5 third line
    3: 5 fourth line
    3: 5 fifth line

coalib/results/SourceRange.py
sils: you'd invoke it like `list(range.affected_source`

tests/output/ConsoleInteractionTest.py
```diff
@@ -849,3 +849,16 @@ def test_empty_list(self):
                                 None,
                                 None,
                                 None)
+
+    def test_formatted_string(self):
```
Mixih: I think you should rename this test to test_source_lines to make it more clear it is testing the functionality you added.

tests/output/ConsoleInteractionTest.py
```diff
@@ -849,3 +849,16 @@ def test_empty_list(self):
...
+        affected_code = (SourceRange.from_values('file', 2, end_line=2),)
+        with retrieve_stdout() as stdout:
+            print_results_formatted(
+                self.logger,
+                self.section,
+                [Result('SpaceConsistencyBear', message='msg',
+                        affected_code=affected_code)],
+                {abspath('file'): ('def fun():\n', '    pass  \n')})
+            self.assertRegex(stdout.getvalue(),
+                             "['    pass  \\n']")
```
Mixih: Shouldn't you also be asserting it printed out the right source lines numbers too?

tests/results/SourceRangeTest.py
```diff
@@ -81,6 +81,23 @@ def test_invalid_comparison(self):
...
+        file_dict = {abspath('file_name'): ('def fun():\n', '    x = 2  \n')}
+        sr1 = SourceRange.from_values('file_name', start_line=2, end_line=2)
+
+        self.assertEqual(list(sr1.affected_source(file_dict)),
+                         ['    x = 2  \n'])
+
+        sr2 = SourceRange.from_values('file_name')
+        self.assertEqual(list(sr2.affected_source(file_dict)), [])
+
+        file_dict1 = {abspath('file_name'): ('def fun():\n', '    x = 2  \n',
```
Mixih: why not overwrite file_dict?

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,38 @@ def expand(self, file_contents):
...
+        ['    x = 2  \n', '    print(x)  \n']
+
+        :param file_dict:
+            It is a dictionary where the file names are the keys and
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected return ``None``
+        """
+        file_contents = file_dict[self.start.file]
```
adtac: What if I give a file_dict that doesn't contain the filename? Uncaught exception.

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,38 @@ def expand(self, file_contents):
...
 
+    @enforce_signature
+    def affected_source(self, file_dict: dict):
+        r"""
+        Make an iterator that tells which are the lines affected in the
+        file within a specified range.
+
+        >>> from os.path import abspath
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=2)
+        >>> list(sr.affected_source({abspath('file_name'):
```
adtac: Wait, how does this pass the tests? The `SourceRange` has `file_name` while the `file_dict` has the absolute path? Shouldn't the lack of a `try` - `except` raise an exception? Sorry if I'm missing something here.

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,38 @@ def expand(self, file_contents):
...
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected return ``None``
+        """
+        file_contents = file_dict[self.start.file]
+        if self.start.line and self.end.line:
+            # line number starts from 1, index starts from 0
+            for line_number in range(self.start.line - 1, self.end.line):
+                yield file_contents[line_number]
```
adtac: Just a `return file_contents[self.start.line-1:self.end.line]` will do ;)

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,38 @@ def expand(self, file_contents):
...
+
+        :param file_dict:
+            It is a dictionary where the file names are the keys and
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected return ``None``
+        """
+        file_contents = file_dict[self.start.file]
+        if self.start.line and self.end.line:
```
adtac: I think it's possible to have a `SourceRange` with `self.end=None` - I think it's used for single line results. Not sure, maybe someone more familiar with bears can chip in.

tests/results/SourceRangeTest.py
Abhisek-: A variable name should be used for only one purpose. Isn't it ?

tests/output/ConsoleInteractionTest.py
Abhisek-: I think `source_lines` is only for the lines of the file. for line numbers there is `line` and `end_line`. Or Am I missing something ?

coalib/results/SourceRange.py
Abhisek-: When somebody passes the name of the file to `SourceRange.from_values`, it automatically converts the name of the file to absolute path. see `coalib.results.SourcePosition.py` constructor.

coalib/results/SourceRange.py
Abhisek-: for single line results `self.start` and `self.end` value are same.

tests/output/ConsoleInteractionTest.py
Mixih: oh right, I misread, sorry :/

Udayan12167:
The filename should be in the shortlog of the commit message

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,43 @@ def expand(self, file_contents):
...
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected return ``None``
+        """
+        try:
+            file_contents = file_dict[self.start.file]
+            if self.start.line and self.end.line:
+                # line number starts from 1, index starts from 0
+                return file_contents[self.start.line - 1: self.end.line]
+        except KeyError:
```
adtac: Uh, wouldn't a simple `if self.start.file in file_dict` be better than a try-except? try-excepts are computationally costlier and we should try to be as performant as possible (plus a `if` is more readable anyway).

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,43 @@ def expand(self, file_contents):
...
+            raise KeyError('{} is not present in {}'.format(
```
adtac: And you're raising an Exception here that you aren't catching in ConsoleInteraction. IMO a simply returning a `None` makes sense here.

adtac:
unack https://github.com/coala/coala/pull/3386/commits/560841308893c357dbc3174ba0c3e7c5e872d4c2

Abhisek-:
@Udayan12167 more than one file has been changed. A function has been added in `SourceRange.py` and the changes in `ConsoleInteraction.py` seems little but main. what file name should I include in the shortlog ?

coalib/results/SourceRange.py
Abhisek-: yeah, that makes sense.

Abhisek-:
If I add the file name to the shortlog, it exceeds the limit of 50 characters. I already have added the file name to the body of the commit message

Mixih:
Imo commit is fine as is

coalib/output/ConsoleInteraction.py
```diff
@@ -309,6 +309,8 @@ def print_results_formatted(log_printer,
                       'column:{column}:end_line:{end_line}:end_column:'
                       '{end_column}:severity:{severity}:severity_str:'
                       '{severity_str}:message:{message}')
+    file_dict = args[0]
```
sils: https://github.com/coala/coala/pull/3386#discussion_r94294547 you gotta add this to the params, this isn't well readable.

coalib/output/ConsoleInteraction.py
```diff
@@ -338,6 +340,13 @@ def print_results_formatted(log_printer,
 
             for range in result.affected_code:
                 format_args['affected_code'] = range
+
+                if file_dict:
```
sils: coala will always give a file dict, IMO this if is unneeded

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,39 @@ def expand(self, file_contents):
...
+        ...     ('def fun():\n', '    x = 2  \n')})
+        ('    x = 2  \n',)
+
+        If more than one line is affected.
+
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=3)
+        >>> sr.affected_source({abspath('file_name'):
+        ...     ('def fun():\n', '    x = 2  \n', '    print(x)  \n')})
+        ('    x = 2  \n', '    print(x)  \n')
+
```
sils: I'd add a sentence and example here:

    If the file indicated at the source range is not in the `file_dict` or the lines are not given, this will return `None`:

    >>> ...

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,39 @@ def expand(self, file_contents):
...
+            It is a dictionary where the file names are the keys and
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected or the file is not present in
+            ``file_dict`` return ``None``.
+        """
+        if self.start.file in file_dict:
+            file_contents = file_dict[self.start.file]
+            if self.start.line and self.end.line:
```
sils: this if can be joined with the previous: `if self.start.file in file_dict and self.start.line and self.end.line:`

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,39 @@ def expand(self, file_contents):
...
+        :param file_dict:
+            It is a dictionary where the file names are the keys and
+            the contents of the files are the values(which is of type tuple).
+        :return:
+            An iterator that returns affected lines in the specified file.
+            If the file is not affected or the file is not present in
+            ``file_dict`` return ``None``.
+        """
+        if self.start.file in file_dict:
+            file_contents = file_dict[self.start.file]
```
sils: unneeded variable, just use `file_dict[...][...]` right at the return IMO

tests/output/ConsoleInteractionTest.py
sils: are we sure we want them to be printed as a list and not the list joined with a newline?

tests/results/SourceRangeTest.py
```diff
@@ -81,6 +81,27 @@ def test_invalid_comparison(self):
         with self.assertRaises(TypeError):
             SourceRange(self.result_fileB_noline, self.result_fileB_line2) < 1
 
+    def test_affected_source(self):
+        file_dict = {abspath('file_name'): ('def fun():\n', '    x = 2  \n')}
+        sr1 = SourceRange.from_values('file_name', start_line=2, end_line=2)
+
+        self.assertEqual(sr1.affected_source(file_dict),
+                         ('    x = 2  \n',))
```
sils: don't repeat your doctests, that's just unneeded maintenance and code overhead. Remove all tests that are covered by your doctests.

coalib/output/ConsoleInteraction.py
Abhisek-: there are unit tests that pass None value as file_dict(see `tests/output/PrintFormattedResultsTest.test_default_format()`. I think that `if` is needed, isn't it ?

coalib/results/SourceRange.py
aptrishu: In that case do we need to handle the case with try catch in code and return None?

aptrishu: nvm we are already checking it.

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,45 @@ def expand(self, file_contents):
                                        tr.end.line,
                                        tr.end.column)
 
+    @enforce_signature
+    def affected_source(self, file_dict: dict):
+        r"""
+        Make an iterator that tells which are the lines affected in the
```
sils: I would have loved to just merge this but you're not making an iterator here but returning a tuple and that's a vital difference

tests/output/ConsoleInteractionTest.py
```diff
@@ -849,3 +849,16 @@ def test_empty_list(self):
...
+        self.section.append(Setting(key='format', value='{source_lines}'))
+        affected_code = (SourceRange.from_values('file', 2, end_line=2),)
+        with retrieve_stdout() as stdout:
+            print_results_formatted(
+                self.logger,
+                self.section,
+                [Result('SpaceConsistencyBear', message='msg',
+                        affected_code=affected_code)],
+                {abspath('file'): ('def fun():\n', '    pass  \n')})
+            self.assertRegex(stdout.getvalue(),
```
sils: why regex and not equality?

coalib/results/SourceRange.py
Abhisek-: Oh after making the suggested changes, forgot to check the docs. really stupid.

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,44 @@ def expand(self, file_contents):
...
 
+    @enforce_signature
+    def affected_source(self, file_dict: dict):
+        r"""
+        Tells which lines are affected in a specified file within a given range.
+
+        >>> from os.path import abspath
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=2)
+        >>> sr.affected_source({abspath('file_name'):
+        ...     ('def fun():\n', '    x = 2  \n')})
```
arush0311: perhaps you could format it better? Its a little bit hard to read.

coalib/results/SourceRange.py
```diff
@@ -120,6 +120,44 @@ def expand(self, file_contents):
...
+        ('    x = 2  \n',)
+
+        If more than one line is affected.
+
+        >>> sr = SourceRange.from_values('file_name', start_line=2, end_line=3)
+        >>> sr.affected_source({abspath('file_name'):
+        ...     ('def fun():\n', '    x = 2  \n', '    print(x)  \n')})
```
arush0311: same here

coalib/results/SourceRange.py
sils: yeah I'd go for

```
asdasdasdasd({
    'key': ...
})
```

coalib/output/ConsoleInteraction.py
sils: that test assumes that this function doesn't need a file dict which is no longer true so the test should be corrected

coalib/results/SourceRange.py
adtac: @sils don't we close the opening braces/parens in the same line as the last element? Or has that been relaxed?

sils:
we're using PEP8 ATM and everything that is allowed according to that ;)

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net
http://coala.io/ - http://viperdev.io/ - http://gitmate.com/

On 17 January 2017 at 12:16, Adhityaa Chandrasekar <notifications@github.com

coalib/results/SourceRange.py
Abhisek-: I think we can use both, when we use vertical alignment then we close the brace in the same line as the last element and when we use hanging indent then we close after the last line

Abhisek-:
I just need to rebase or some other things that I need to do.

coalib/output/ConsoleInteraction.py
jayvdb: ping @Abhisek-

yukiisbored:
@Abhisek- Hello, the shortlog of a commit message is the first line of the commit message.

sils:
ack 23dedd70f598867b663590bef65c867a3043aadf

sils:
great code!

sils:
(I'll probably cherry pick it later with other PRs.)

sils:

PR #3404, opened 2016-12-27
assertRaisesRegex used in LinterTest.py

rhemon:
Replaced assertRaises with assertRaisesRegex in LinterTest.py file

I am not confident about the strings used, incase I have done it wrong, please tell me, I will fix it. 

Task: https://codein.withgoogle.com/dashboard/task-instances/4861422196490240/

### DIFF
--- a/tests/bearlib/abstractions/LinterTest.py
+++ b/tests/bearlib/abstractions/LinterTest.py
@@ -29,6 +29,8 @@ def get_testfile_name(name):
 
 class LinterComponentTest(unittest.TestCase):
 
+    PARAM_TYPERERROR_RE = '[a-z_]+ must be an instance'
+
     # Using `object` instead of an empty class results in inheritance problems
     # inside the linter decorator.
     class EmptyTestLinter:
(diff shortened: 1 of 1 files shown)


jayvdb:
Great to see you got the tests to pass.

The current code for `LinterTest.py` has a few cases like:

```python
    def test_decorator_invalid_parameters(self):
        with self.assertRaises(ValueError) as cm:
            linter('some-executable', invalid_arg=88,
                   ABC=2000)(self.EmptyTestLinter)
        self.assertEqual(
            str(cm.exception),
            "Invalid keyword arguments provided: 'ABC', 'invalid_arg'")
```

Especially notice the last line there.

Your commit has changes like:
```diff
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaisesRegex(ValueError,
+                                    'Invalid keyword arguments provided') as cm:
```

The parameter that you are adding (`'Invalid keyword arguments provided'`) is correct.
However, it is not as good as the existing equality comparison below it, which ensures the exception message is:

    `"Invalid keyword arguments provided: 'ABC', 'invalid_arg'"`

So, for these cases, with an existing exception message check (using `cm`), you have two options:

1.  Don't convert them to `assertRaisesRegex`, as they are already a good unit test, with a precise exception message check.

2.  Convert them to `assertRaisesRegex`, and the new argument needs to be an exact match of the exception message text, which means you can then delete the `self.assertEqual` line as redundant.

The second option is of course preferred, but the first option is not bad.

Once you have finished the code changes, use `git commit --amend`, and then use a similar commit message to the example in the GCI task. i.e. end with "`Related to <url>`"

rhemon:
So I went with the first option, since it was easier. so now the ones that didnt use `cm` are using `assertRaisesRegex`. For those do I have to do any change in the arguments. Do those have to be completely same as the message? @jayvdb

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -144,18 +144,20 @@ def test_decorator_invalid_states(self):
...
         uut = linter('some-executable')(self.ManualProcessingTestLinter)
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaisesRegex(NotImplementedError, ''):
             uut.create_arguments('filename', 'content', None)
 
     def test_decorator_invalid_parameter_types(self):
         # Provide some invalid severity maps.
-        with self.assertRaises(TypeError):
+        with self.assertRaisesRegex(TypeError,
+                                    'severity_map must be an instance'):
```
jayvdb: this should be a constant, stored in the class, with an ALL_CAPS_RE variable name, like the example.

jayvdb: "[a-z_]+ must be an instance" would be a good regex, to match all of these `TypeError`

rhemon: @jayvdb ok did the changes, please have a look

jayvdb: I dont see any new constants in your code.

jayvdb:
Need to rebase

jayvdb:
ack 23c023b

jayvdb:
@rultor merge

PR #3444, opened 2016-12-31
test-requirements: Put pytest before its plugins

jayvdb:
pytest dependencies are higher and more important
than those listed by its plugins.

Closes https://github.com/coala/coala/issues/3184

PR #3448, opened 2017-01-01
CI: Use coala-bears extra 'alldeps'

jayvdb:
Before we can remove the python dependencies from
the coala-bears install requirements, the other coala
repositories need to switch to using the 'alldeps'
extra.

Related to https://github.com/coala/coala-bears/issues/1000

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ script:
       bash .misc/tests.sh
       python setup.py bdist_wheel
       pip install ./dist/coala-*.whl
-      pip install coala-bears --pre -U
+      pip install coala-bears[alldeps] --pre -U
       # https://github.com/coala/coala-bears/issues/1037
       if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
         python3 -m nltk.downloader punkt maxent_treebank_pos_tagger averaged_perceptron_tagger
--- a/circle.yml
+++ b/circle.yml
@@ -15,7 +15,7 @@ test:
         parallel: true
     - python setup.py install:
         parallel: true
-    - pip install coala-bears --pre -U:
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -43,7 +43,7 @@ script:
       bash .misc/tests.sh
       python setup.py bdist_wheel
       pip install ./dist/coala-*.whl
-      pip install coala-bears --pre -U
+      pip install coala-bears[alldeps] --pre
```
sils: why the removal of `-U`? IMO we should either consistently use the newest (with -U) or spec a version (`==0.10.0dev.....`)

jayvdb: putting a version here makes no difference at all.  `coala-bears` isnt installed, so the `-U` has no effect wrt that.

 `-U` upgrades the deps to newer versions.  Not using `-U` will mean we get a `VersionConflict` if pip should have installed a newer version than was already installed, indicating that our requirements between the two packages are slightly problematic.   Problems like https://github.com/coala/coala/pull/3444 will be found sooner (if it happens in real requirements, not test-requirements).

We really need https://github.com/coala/coala-bears/issues/1113 , or that can be built into cib.

(and we should enable PySafetyBear)

sils:
ack https://github.com/coala/coala/pull/3448/commits/a53c80c58a986b7260a06e13c5d2d2617d608647

sils:
@jayvdb sorry, just merging another PR, I figured you doing a rebase will take mere seconds while the newer people will be annoyed for a longer time.

jayvdb:
(If ack the commit and note it needs rebasing, I can take care of it after that)

sils:
bdec071 ack

sils:
@rultor merge

Issue #3323, opened 2016-12-18
@linter decorator causes ERROR on Windows


userzimmermann:
@sils See https://github.com/coala/coala-bears/issues/1172 and https://github.com/coala/coala/issues/3322 respectively.

sils:
CC @Makman2

Makman2:
phew... will investigate soon^^

userzimmermann:
@Makman2 Any progress? Just got the same problem with `MarkdownBear`. Feeling lost on Windows... ;)

userzimmermann:
Found some useful info here:

http://stackoverflow.com/questions/26025878/what-is-being-pickled-when-i-call-multiprocessing-process

userzimmermann:
It's all about the pickling when using `multiprocessing`, which doesn't happen on Linux.

Doesn't seem like we can get those `@linter` classes pickled... :(

Mixih:
I think this is done.... Reopen if it isn't

PR #3522, opened 2017-01-10
Redesign @linter for Windows compatibility

userzimmermann:
@sils @Makman2 

IMO the only way to solve https://github.com/coala/coala/issues/3323 :

* Define `LinterBase` and `LinterMeta` globally in `Linter` module and not dynamically in `_create_linter` function, because dynamically created classes can't be pickled
* Set `klass.options = options` to make them available as `cls.options` and `self.options` in `LinterBase` methods instead of accessing `options` through the method closures as done in the dynamic class creation approach
* Explicitly derive all `@linter` bears from `LinterBase`:

```python
@linter(...)
class SomeLinterBear(LinterBase):
    ...
```

The above looks ugly, but that can maybe further redesigned with some other way for setting linter options than with `@linter` decorator.

My changes don't break `@linter` bears that don't explicitly derive from `LinterBase`. They will just still be not usable under Windows.

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -572,7 +572,8 @@ def _create_linter(klass, options):
     # Mixin the linter into the user-defined interface, otherwise
     # `create_arguments` and other methods would be overridden by the
     # default version.
-    result_klass = type(klass.__name__, (klass, LinterBase), {})
+    result_klass = type(klass.__name__, (klass, LinterBase), {
+        '__module__': klass.__module__})
     result_klass.__doc__ = klass.__doc__ or ''
     return result_klass
 


Makman2:
From discussion in gitter:
> I dislike to change usage via a base class as this is repeating yourself^^

\+ This is an API change,  which especially adds redundancy^^

arush0311:
Perhaps the internals of linter can be changed so that LinterBase is used instead of dynamically creating a LinterBase class. This way both linter decorator as well as bears deriving from LinterBase can work.

userzimmermann:
@arush0311 That's already provided by my solution.

CC @Makman2

userzimmermann:
@Makman2 Found a way to keep the existing `@linter` only style: Making `_create_linter` function copy over all methods from `LinterBase` to the decorated class and defining an `__init__` that creates a `self.base = LocalBear(...)` and a `__getattr__` that delegates to the `LocalBear` base.

It might be a bit hacky but it works without changing any bear implementation!

Only problem is that `@linter` bear classes are not derived from `LocalBear` anymore. So the according check in `coalib.processes.BearRunning.run_local_bear` function fails. I commented that out for now. We could add another mechanism to specify a class as being a bear class without that base class lookup...

userzimmermann:
@sils @Makman2 @arush0311 Forget all the above changes and discussions. Almost all ;) Pickling is still the actual problem. But the pickling fail was only introduced by the `.__module__` name of the `@linter` decorated class. So just set it to the same as the original `.__module__` name and everything is fine :)

userzimmermann:
Sorry for the trouble ;D But I think we all could learn something ;)

userzimmermann:
ack https://github.com/coala/coala/pull/3522/commits/926506685e94f379752457b3f9787496b2cf6e44

Makman2:
unack 9265066

Makman2:
commit message ->
```
Linter: Add __module__ attribute

Needed for Windows compatability to support pickling of Linter classes.
```

Makman2:
ack 48ef412

Makman2:
@rultor merge

PR #3523, opened 2017-01-11
WIP: Makman2/nextgen3

Makman2:
Part 1 -> The complete new core. **Integration not yet done**

### DIFF
--- a/coalib/core/Core.py
+++ b/coalib/core/Core.py
@@ -7,6 +7,11 @@ from coalib.core.DependencyTracker import DependencyTracker
 from coalib.core.Graphs import traverse_graph
 
 
+# FIXME Improve performance by using a `CoalaRunContext` class. This avoids
+# FIXME   passing always parameters around like `result_callback`,
+# FIXME   `dependency_tracker`, `event_loop`, `running_tasks` and `executor`.
+
+
 def group(iterable, key=lambda x: x):
     """
     Groups elements (out-of-order) together in the given iterable.
--- a/coalib/core/FileBear.py
+++ b/coalib/core/FileBear.py
@@ -1,0 +1,42 @@
+from coalib.core.Bear import Bear
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+class FileBear(Bear):
+    """
+    This bear base class parallelizes tasks for each file given.
+    """
+
+    def __init__(self, section, file_dict):
+        """
+        :param section:
+            The section object where bear settings are contained. A section
+            passed here is considered to be immutable.
+        :param file_dict:
+            A dictionary containing filenames to process as keys and their
+            contents (line-split with trailing return characters) as values.
+        """
+        Bear.__init__(self, section, file_dict)
+
+        self._kwargs = self.get_metadata().create_params_from_section(section)
+
+    def execute_task(self, args, kwargs):
+        # To optimize performance and memory usage, we use kwargs from this
+        # class instance instead of passing the same settings thousand times in
+        # the tasks themselves.
+        return Bear.execute_task(self, args, self._kwargs)
+
+    @classmethod
+    def get_metadata(cls):
+        """
+        :return:
+            Metadata for the ``analyze`` function extracted from its signature.
+            Excludes parameters ``self``, ``filename`` and ``file``.
+        """
+        return FunctionMetadata.from_function(
+            cls.analyze,
+            omit={'self', 'filename', 'file'})
+
+    def generate_tasks(self):
+        return (((filename, file), {})
+                for filename, file in self.file_dict.items())
--- a/coalib/core/ProjectBear.py
+++ b/coalib/core/ProjectBear.py
@@ -1,0 +1,42 @@
+from coalib.core.Bear import Bear
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+class ProjectBear(Bear):
+    """
+    This bear base class does not parallelize tasks at all, it runs on the
+    whole file base provided.
+    """
+
+    def __init__(self, section, file_dict):
+        """
+        :param section:
+            The section object where bear settings are contained. A section
+            passed here is considered to be immutable.
+        :param file_dict:
+            A dictionary containing filenames to process as keys and their
+            contents (line-split with trailing return characters) as values.
+        """
+        Bear.__init__(self, section, file_dict)
+
+        self._kwargs = self.get_metadata().create_params_from_section(section)
+
+    def execute_task(self, args, kwargs):
+        # To optimize performance a bit and memory usage, we use args and
+        # kwargs from this class instance, instead of passing them via the
+        # task.
+        return Bear.execute_task(self, (self.file_dict,), self._kwargs)
+
+    @classmethod
+    def get_metadata(cls):
+        """
+        :return:
+            Metadata for the ``analyze`` function extracted from its signature.
+            Excludes parameters ``self`` and ``files``.
+        """
+        return FunctionMetadata.from_function(
+            cls.analyze,
+            omit={'self', 'files'})
+
+    def generate_tasks(self):
+        return (tuple(), {}),
--- a/tests/core/FileBearTest.py
+++ b/tests/core/FileBearTest.py
@@ -1,0 +1,117 @@
+import unittest
+
+from coalib.core.FileBear import FileBear
+from coalib.core.Core import run
+from coalib.settings.Section import Section
+
+
+class TestFileBear(FileBear):
+
+    def analyze(self, filename, file):
+        yield filename
+
+
+class TestFileBearWithParameters(FileBear):
+
+    def analyze(self, filename, file, results_per_file: int=1):
+        for i in range(results_per_file):
+            yield filename + str(i)
+
+
+class FileBearTest(unittest.TestCase):
+
+    def assertResultsEqual(self, bear_type, expected,
+                           section=None, file_dict=None):
+        """
+        Asserts whether the expected results do match the output of the bear.
+
+        Asserts for the results out-of-order.
+
+        :param bear_type:
+            The bear class to check.
+        :param expected:
+            A sequence of expected results.
+        :param section:
+            A section for the bear to use. By default uses a new section with
+            name ``test-section``.
+        :param file_dict:
+            A file-dictionary for the bear to use. By default uses an empty
+            dictionary.
+        """
+        if section is None:
+            section = Section('test-section')
+        if file_dict is None:
+            file_dict = {}
+
+        uut = bear_type(section, file_dict)
+
+        results = []
+
+        def on_result(result):
+            results.append(result)
+
+        run({uut}, on_result)
+
+        self.assertEqual(sorted(expected), sorted(results))
+
+    def test_bear_without_parameters(self):
+        self.assertResultsEqual(
+            TestFileBear,
+            file_dict={},
+            expected=[])
+        self.assertResultsEqual(
+            TestFileBear,
+            file_dict={'fileX': []},
+            expected=['fileX'])
+        self.assertResultsEqual(
+            TestFileBear,
+            file_dict={'fileX': [], 'fileY': []},
+            expected=['fileX', 'fileY'],)
+        self.assertResultsEqual(
+            TestFileBear,
+            file_dict={'fileX': [], 'fileY': [], 'fileZ': []},
+            expected=['fileX', 'fileY', 'fileZ'])
+
+    def test_bear_with_parameters_but_keep_defaults(self):
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            file_dict={},
+            expected=[])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            file_dict={'fileX': []},
+            expected=['fileX0'])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            file_dict={'fileX': [], 'fileY': []},
+            expected=['fileX0', 'fileY0'])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            file_dict={'fileX': [], 'fileY': [], 'fileZ': []},
+            expected=['fileX0', 'fileY0', 'fileZ0'])
+
+    def test_bear_with_parameters(self):
+        section = Section('test-section')
+        section['results_per_file'] = '2'
+
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            section=section,
+            file_dict={},
+            expected=[])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            section=section,
+            file_dict={'fileX': []},
+            expected=['fileX0', 'fileX1'])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            section=section,
+            file_dict={'fileX': [], 'fileY': []},
+            expected=['fileX0', 'fileX1', 'fileY0', 'fileY1'])
+        self.assertResultsEqual(
+            TestFileBearWithParameters,
+            section=section,
+            file_dict={'fileX': [], 'fileY': [], 'fileZ': []},
+            expected=['fileX0', 'fileX1', 'fileY0', 'fileY1', 'fileZ0',
+                      'fileZ1'])
--- a/tests/core/ProjectBearTest.py
+++ b/tests/core/ProjectBearTest.py
@@ -1,0 +1,117 @@
+import unittest
+
+from coalib.core.ProjectBear import ProjectBear
+from coalib.core.Core import run
+from coalib.settings.Section import Section
+
+
+class TestProjectBear(ProjectBear):
+
+    def analyze(self, files):
+        yield '\n'.join(filename + ':' + str(files[filename])
+                        for filename in sorted(files))
+
+
+class TestProjectBearWithParameters(ProjectBear):
+
+    def analyze(self, files, prefix: str='!!!'):
+        yield '\n'.join(prefix + filename + ':' + str(files[filename])
+                        for filename in sorted(files))
+
+
+class ProjectBearTest(unittest.TestCase):
+
+    def assertResultsEqual(self, bear_type, expected,
+                           section=None, file_dict=None):
+        """
+        Asserts whether the expected results do match the output of the bear.
+
+        Asserts for the results out-of-order.
+
+        :param bear_type:
+            The bear class to check.
+        :param expected:
+            A sequence of expected results.
+        :param section:
+            A section for the bear to use. By default uses a new section with
+            name ``test-section``.
+        :param file_dict:
+            A file-dictionary for the bear to use. By default uses an empty
+            dictionary.
+        """
+        if section is None:
+            section = Section('test-section')
+        if file_dict is None:
+            file_dict = {}
+
+        uut = bear_type(section, file_dict)
+
+        results = []
+
+        def on_result(result):
+            results.append(result)
+
+        run({uut}, on_result)
+
+        self.assertEqual(sorted(expected), sorted(results))
+
+    def test_bear_without_parameters(self):
+        self.assertResultsEqual(
+            TestProjectBear,
+            file_dict={},
+            expected=[''])
+        self.assertResultsEqual(
+            TestProjectBear,
+            file_dict={'fileX': []},
+            expected=['fileX:[]'])
+        self.assertResultsEqual(
+            TestProjectBear,
+            file_dict={'fileX': [], 'fileY': ['hello']},
+            expected=["fileX:[]\nfileY:['hello']"],)
+        self.assertResultsEqual(
+            TestProjectBear,
+            file_dict={'fileX': [], 'fileY': ['hello'], 'fileZ': ['x\ny']},
+            expected=["fileX:[]\nfileY:['hello']\nfileZ:['x\\ny']"])
+
+    def test_bear_with_parameters_but_keep_defaults(self):
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            file_dict={},
+            expected=[''])
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            file_dict={'fileX': []},
+            expected=['!!!fileX:[]'])
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            file_dict={'fileX': [], 'fileY': ['hello']},
+            expected=["!!!fileX:[]\n!!!fileY:['hello']"], )
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            file_dict={'fileX': [], 'fileY': ['hello'], 'fileZ': ['x\ny']},
+            expected=["!!!fileX:[]\n!!!fileY:['hello']\n!!!fileZ:['x\\ny']"])
+
+    def test_bear_with_parameters(self):
+        section = Section('test-section')
+        section['prefix'] = '???'
+
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            section=section,
+            file_dict={},
+            expected=[''])
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            section=section,
+            file_dict={'fileX': []},
+            expected=['???fileX:[]'])
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            section=section,
+            file_dict={'fileX': [], 'fileY': ['hello']},
+            expected=["???fileX:[]\n???fileY:['hello']"], )
+        self.assertResultsEqual(
+            TestProjectBearWithParameters,
+            section=section,
+            file_dict={'fileX': [], 'fileY': ['hello'], 'fileZ': ['x\ny']},
+            expected=["???fileX:[]\n???fileY:['hello']\n???fileZ:['x\\ny']"])


Makman2:
Everything is ready except the first commit:
```
core: Add Bear v2 
```
(you can already see my fixups, I forgot to write tests for the new bear + some concept changes...)

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,330 @@
...
+    >>> tracker.resolve(object3)
+    {<object object at ...>}
+    """
+
+    def __init__(self):
+        self._dependency_dict = {}
+
+    def get_dependants(self, dependency):
+        """
+        Returns all dependants for the given dependency.
```
aptrishu: better to say get immediate dependants.

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+        :param bear:
+            The bear instance to add results for.
+        :param dependency_results:
+            The results to add.
+        """
+        bear_type = type(bear)
+
+        if bear_type not in self._dependency_results:
+            current_results = []
+            self._dependency_results[bear_type] = current_results
```
aptrishu: This statement should be outside the if else, after  `current_results += dependency_results` .
IMO dict is not getting updated properly.

coalib/core/Graphs.py
```diff
@@ -0,0 +1,67 @@
...
+        A callable that is run on each edge during traversing. Takes in two
+        parameters, the previous- and next-node which form an edge. The default
+        is an empty function.
+    :raises CircularDependencyError:
+        Raised when the graph is cyclic.
+    """
+    path = set()
+    visited_nodes = set()
+
+    def visit(node):
```
aptrishu: you are adding and removing from path two times..
why didn't you just do path.add(node) at the start of visit() method and path.remove(node) at the end of visit() method, and after that remove path.add() and path.remove() from other places ?

aptrishu: same goes with visited_nodes.add()

Makman2: works, good catch :D :100:

aptrishu:
well, sorry for not suggesting everything at once.
but this would be a better code for your graphs.py (performs less checking)

>def visit(node):
    if node not in visited_nodes:
        visited_nodes.add(node)
        path.add(node)
        for subnode in get_successive_nodes(node):
            if subnode not in visited_nodes:
                visit(subnode)
            elif subnode in path:
                raise CircularDependencyError(subnode)
    path.remove(node)

for node in start_nodes:
        visit(node)

aptrishu:
Well sorry for not suggesting everything at once but I think this would be a better code for graphs.py (performs less checking) (a little bit more change)  

`def visit(node):
    if node not in visited_nodes:
        visited_nodes.add(node)
        path.add(node)
        for subnode in get_successive_nodes(node):
            if subnode not in visited_nodes:
                visit(subnode)
            elif subnode in path:
                raise CircularDependencyError(subnode)
    path.remove(node)

for node in start_nodes:
        visit(node)`

coalib/core/Graphs.py
```diff
@@ -0,0 +1,62 @@
+from coalib.core.CircularDependencyError import CircularDependencyError
+
+
+def traverse_graph(start_nodes, get_successive_nodes,
+                   run_on_edge=lambda prev, nxt: None):
+    """
+    Traverses all edges of a directed acyclic graph once. Detects cyclic graphs
```
aptrishu: it is confusing, cyclic graph in acyclic graph not possible.  
Plus we are expecting more than one connected graphs and No, we can't ensure it is traversing every edge exactly once. so, maybe change it to something like - 
`Traverses the 'graph/s'. Detects cycle by raising..`

Makman2: okay shall I rephrase?

> No, we can't ensure it is traversing every edge exactly once.

why? algorithm does that :O

aptrishu: Actually I misinterpreted that statement. and yes rest of the statement needs the change.

tests/core/CoreTest.py
```diff
@@ -0,0 +1,600 @@
...
+
+
+class MultiResultBear(TestBearBase):
+
+    def analyze(self, bear, section_name, file_dict):
+        yield 1
+        yield 2
+
+
+# This bear runs a certain number of tasks dependencing on the number of
```
aptrishu: Typo - dependencing -> depending

Makman2: omg was that me? :O

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+    ... 'Code Simplification'}
+    >>> CAN_FIX = {'Syntax', ...}
+
+    Specifying something to ``CAN_FIX`` makes it obvious that it can be
+    detected too, so it may be omitted:
+
+    >>> class SomeBear(Bear):
+    ...     CAN_DETECT = {'Syntax', 'Security'}
+    ...     CAN_FIX = {'Redundancy'}
+    >>> list(sorted(SomeBear.can_detect))
```
sils: `list` not needed, sorted gives a list

sils: since sorted is O(nlogn) or so anyway there's no benefit in making it an iterator

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+
+    @classproperty
+    def maintainers_emails(cls):
+        """
+        :return:
+            A set containing ``MAINTAINERS_EMAILS`` if specified, else takes
+            ``AUTHORS_EMAILS`` by default.
+        """
+        return (cls.AUTHORS_EMAILS if cls.MAINTAINERS_EMAILS == set()
+                else cls.MAINTAINERS)
```
sils: you mean `cls.MAINTAINERS_EMAILS` right?

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+        """
+        This method has to determine which settings are needed by this bear.
+        The user will be prompted for needed settings that are not available
+        in the settings file so don't include settings where a default value
+        would do.
+
+        :return: A dictionary of needed settings as keys and a tuple of help
+                 text and annotation as values
+        """
+        return cls.get_metadata().non_optional_params
```
sils: not ok we need to get the non optonal params of the dependency here too!

coalib/core/Bear.py
Makman2: seems like, copied that from the old bear without modification^^

sils: see #3566

Makman2:
bump 🎉

coalib/core/Core.py
```diff
@@ -145,6 +149,91 @@ def finish_task(bear,
...
+    #       |       |                                         V
+    #       ---------------------------------------------> BearType3
+    #
+    # We need to traverse this graph and instantiate dependency bears
+    # accordingly, one per section.
+
+    # Group bears by sections and file-dictionaries. These will serve as
+    # entry-points for the dependency-instantiation-graph.
+    grouping = groupby(bears, key=lambda bear: (bear.section, bear.file_dict))
+    for (section, file_dict), bears_per_section in grouping:
```
Makman2: I've actually discovered a serious flaw: `groupby` does not work as expected:

> The operation of groupby() is similar to the uniq filter in Unix. It generates a break or new group every time the value of the key function changes (which is why it is usually necessary to have sorted the data using the same key function). That behavior differs from SQL’s GROUP BY which aggregates common elements regardless of their input order.

Means we need to sort the bears before using the section and file-dicts as sort-keys.
\+ I need to add a test-case where 3 bears are added, 2 of them using `section1`, one of them using `section2`. And they are added out-of-order: `section1Bear1, section2Bear, section1Bear2`.

The functionality could be also moved into an own little function^^

Makman2: This should work:
```python
def group(iterable, key=None):
    if key is None:
        return groupby(sorted(iter))
    else:
        return groupby(sorted(iter, key=key), key=key)
```

aptrishu: yeah docs clearly mention that.

coalib/core/Core.py
```diff
@@ -0,0 +1,264 @@
...
+    #       |       |                                         |
+    #       |       |                                         V
+    #       ---------------------------------------------> BearType3
+    #
+    # We need to traverse this graph and instantiate dependency bears
+    # accordingly, one per section.
+
+    # Group bears by sections and file-dictionaries. These will serve as
+    # entry-points for the dependency-instantiation-graph.
+    grouping = groupby(bears, key=lambda bear: (bear.section, bear.file_dict))
```
aptrishu: Since we need to sort the data first, so it'd be a better efficient way to do the grouping with dict. like,
`def group_by(l, key=lambda..):
    d = defaultdict(list)
    for item in l:
        d[key(item)].append(item)
    return d.items()`

Makman2:
Grouping problem addressed. New function `group` supports out-of-order-processing + unhashable keys (needed because file-dicts are used as keys).

Makman2:
Integrated af800e05b3f250d4bad3b9a6134984fac4413682 and 0e68e1367a882beea472ff819eca1e0939abd270 into the new `core.Bear`.
33293e9bbcf22d51135550f0dd5f5acacb3078e8 does not apply, as `execute` does not exist any more like this.

coalib/core/Bear.py
aptrishu: lists are mutable. so when we are updating current_results later in L256, `self._dependency_results[bear_type]` is also getting updated with that.

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+        :param dependency_results:
+            The results to add.
+        """
+        bear_type = type(bear)
+
+        if bear_type not in self._dependency_results:
+            current_results = []
+            self._dependency_results[bear_type] = current_results
+        else:
+            current_results = self._dependency_results[bear_type]
```
aptrishu: We are handling `dependency_results` only once in the current design. So It seems better ?

coalib/core/Core.py
```diff
@@ -0,0 +1,175 @@
...
+        running_tasks[bear].remove(task)
+        cleanup_bear(bear, result_callback, running_tasks, event_loop,
+                     executor)
+
+    if results is not None:
+        for result in results:
+            try:
+                # FIXME Long operations on the result-callback could block the
+                # FIXME   scheduler significantly. It should be possible to
+                # FIXME   schedule new Python Threads on the given event_loop
```
Makman2: Actually I'm not sure how result-callbacks are handled, but it could be that they are executed sequentially (which I currently assume). Offloading them into Python threads makes them at least executing "in parallel" (not real-time, but Python switches those threads regularly).

coalib/core/Core.py
```diff
@@ -0,0 +1,175 @@
...
+                # FIXME   and process the callback there.
+                result_callback(result)
```
Makman2: exactly :+1:

coalib/core/Core.py
```diff
@@ -145,6 +149,91 @@ def finish_task(bear,
...
+    #       V       V                  |                      |
+    #  BearType1  BearType2            -----------------------|
+    #       |       |                                         |
+    #       |       |                                         V
+    #       ---------------------------------------------> BearType3
+    #
+    # We need to traverse this graph and instantiate dependency bears
+    # accordingly, one per section.
+
+    # Group bears by sections and file-dictionaries. These will serve as
```
Makman2: A section does not contain the actual file-dict, it just contains the `files`-setting which contains a (unevaluated) glob. Actually some section settings will be treated specially in our current design (like `bears`, `files`, `ignore`, `default_actions` etc), as some of them are bound to the CLI, and some do control how bears shall be run outside of the scope of the bear itself.
This will get more separated with aspects, though this is still a way to go ;)

Makman2: The main idea behind this is also to support virtual files (useful for editor plugins), so the user defines how files look like and their names inside the file-dict.

coalib/core/Core.py
```diff
@@ -145,6 +149,91 @@ def finish_task(bear,
...
+        # Now traverse each edge of the graph, and instantiate a new dependency
+        # bear if not already instantiated. For the entry point bears, we hack
+        # in identity-mappings because those are already instances. Also map
+        # the types of the instantiated bears to those instances, as if the
+        # user already supplied an instance of a dependency, we reuse it
+        # accordingly.
+        type_to_instance_map = {}
+        for bear in bears_per_section:
+            type_to_instance_map[bear] = bear
+            type_to_instance_map[type(bear)] = bear
```
Makman2: yes both required to build up a proper dependency graph.

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,129 @@
...
+    >>> tracker = DependencyTracker()
+    >>> tracker.add(object2, object1)
+
+    This would define that ``object1`` is dependent from ``object2``.
+
+    If you define that ``object2`` has its dependency duty fulfilled, you can
+    resolve it:
+
+    >>> resolved = tracker.resolve(object2)
+    >>> resolved  # +ELLIPSIS
```
Makman2: Ah yeah forgot to remove it again... :+1:

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,129 @@
...
+    def add(self, dependency, dependant):
+        """
+        Add a bear-dependency to another bear manually.
+
+        This function does not check for circular dependencies.
+
+        >>> tracker = DependencyTracker()
+        >>> tracker.add(0, 1)
+        >>> tracker.add(0, 2)
+        >>> tracker.get_dependants(0)
```
Makman2: Oh yeah damn... trying to reorder/squash commits somehow...

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+        """
+        Override JSON export of ``Bear`` class.
+        """
+        # Those members get duplicated if they aren't excluded because they
+        # exist also as fields.
+        excluded_members = {'can_detect', 'maintainers', 'maintainers_emails'}
+
+        # json cannot serialize properties, so drop them
+        data = {
+            key: value
```
Makman2: we can't^^ We need to implement `str` on those which don't have one / we want to change so they look better^^

coalib/core/Bear.py
```diff
@@ -0,0 +1,469 @@
...
+            'desc': metadata.desc,
+            'non_optional_params': {param: non_optional_params[param][0]
+                                    for param in non_optional_params},
+            'optional_params': {param: optional_params[param][0]
+                                for param in optional_params}}
+
+        return data
+
+    @staticmethod
+    def setup_dependencies():
```
Makman2: nope, the base bear has no dependencies, so the setup step does not do anything :)

coalib/core/Bear.py
Makman2: I've reduced code size by doing this now:
```
        if bear_type not in self._dependency_results:
            self._dependency_results[bear_type] = list(dependency_results)
        else:
            self._dependency_results[bear_type] += dependency_results
```

coalib/core/Core.py
```diff
@@ -0,0 +1,175 @@
...
+            except Exception as ex:
+                # FIXME Try to display only the relevant traceback of the
+                # FIXME result handler if error occurred there, not the
+                # FIXME complete event-loop traceback.
+                logging.error(
+                    'An exception was thrown during result-handling.',
+                    exc_info=ex)
+
+
+def run(bears, result_callback):
```
Makman2: what do you want specifically being improved?

coalib/core/DependencyTracker.py
```diff
@@ -181,6 +181,24 @@ def dependencies(self):
         """
         return set(self._dependency_dict.keys())
 
+    def __iter__(self):
```
Makman2: doctests should suffice^^

tests/core/CoreTest.py
```diff
@@ -57,12 +57,278 @@ def generate_tasks(self):
...
+
+        dependency_tracker, bears_to_schedule = initialize_dependencies(
+            {bear_b, bear_c})
+
+        self.assertEqual(dependency_tracker.dependants, {bear_c})
+        dependencies = dependency_tracker.dependencies
+        self.assertEqual(len(dependencies), 1)
+        dependency = dependencies.pop()
+        self.assertIsInstance(dependency, BearB)
+        self.assertIsNot(dependency, bear_b)
```
Makman2: the idea is that if the section is different, a new bear get instantiated. Imo this is okay^^

coalib/core/Core.py
```diff
@@ -42,14 +47,32 @@ def cleanup_bear(bear,
...
+
         del running_tasks[bear]
 
     if not running_tasks:
+        # Check the DependencyTracker additionally for remaining
+        # dependencies.
+        resolved = dependency_tracker.all_dependencies_resolved
+        if not resolved:  # pragma: no cover
+            logging.warning(
+                'Core finished with run, but it seems some dependencies '
```
Makman2: You think this is okay?
```
Core finished with run, but it seems some dependencies were unresolved: {}. Ignoring them, but this is a bug, please report it to the developers.
```

coalib/core/Core.py
```diff
@@ -71,44 +96,54 @@ def schedule_bears(bears,
...
-
-        if not tasks:
-            # We need to recheck our runtime if something is left to
-            # process, as when no tasks were offloaded the event-loop could
-            # hang up otherwise.
-            cleanup_bear(bear, result_callback, running_tasks, event_loop,
-                         executor)
+        if dependency_tracker.get_dependencies(bear):  # pragma: no cover
+            logging.warning(
+                'Dependencies for {!r} not yet resolved, holding back. This '
```
Makman2: Appending `Please report this to the developers.` suffices?

coalib/core/Bear.py
SanketDG: so atleast `pass` wouldn't hurt :smile:

tests/core/DependencyTrackerTest.py
```diff
@@ -0,0 +1,41 @@
...
+        uut.add(6, 0)
+
+        self.assertEqual(uut.resolve(0), {1, 2, 3})
+        # Dependants already resolved.
+        self.assertEqual(uut.resolve(0), set())
+
+        # Though 0 had a dependency, it was still forcefully resolved.
+        self.assertEqual(uut.resolve(6), set())
+
+        # Let's re-add a dependency for 0.
```
SanketDG: dependency -> dependant?

coalib/core/Bear.py
Makman2: no need for `pass`, when providing a docstring `pass` can be omitted^^

coalib/core/Core.py
```diff
@@ -0,0 +1,175 @@
...
+    except NotImplementedError:  # pragma: no cover
+        # cpu_count is not implemented for some CPU architectures/OSes
+        return 1
+
+
+def cleanup_bear(bear,
+                 result_callback,
+                 running_tasks,
+                 event_loop,
+                 executor):
```
SanketDG: why so many params? are you trying to have the same signature across all functions?

tests/core/CoreTest.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self.assertEqual(len(comparable_real), len(expected))
+        for result in expected:
+            self.assertIn(result, comparable_real)
+            comparable_real.remove(result)
+
+    def test_run_simple(self):
+        # Test single bear without dependencies.
+        section = Section('test-section')
+        filedict = {}
```
SanketDG: these are being used throughout the rest of the code, maybe use global variables to inline them?

Also, in some they are inlined, while in others you use variables.

coalib/core/Bear.py
```diff
@@ -0,0 +1,501 @@
...
+                                  'runnable bear.')
+
+    def generate_tasks(self):
+        """
+        This method is responsible for providing the job arguments ``analyze``
+        gets called with.
+
+        :return:
+            An iterable containing the positional and keyword arguments
+            organized in pairs: ``(args-tuple, kwargs-dict)``
```
Makman2: what kind of doctests? This just returns a sequence of tuples the overrider defines + it's currently `NotImplemented` :)

tests/core/BearTest.py
```diff
@@ -0,0 +1,262 @@
...
+            str(cm.exception),
+            'The bear BearWithPrerequisitesOverride does not fulfill all '
+            'requirements. This is on purpose due to running inside a test.')
+
+    def test_get_metadata(self):
+        uut = BearWithAnalysis.get_metadata()
+        self.assertEqual(uut.non_optional_params,
+                         {'x': ('First value.', int),
+                          'y': ('Second value.', int)})
+        print(uut.optional_params)
```
Makman2: oh damn thx :+1:

coalib/core/Bear.py
```diff
@@ -0,0 +1,501 @@
...
+    @classproperty
+    def maintainers_emails(cls):
+        """
+        :return:
+            A set containing ``MAINTAINERS_EMAILS`` if specified, else takes
+            ``AUTHORS_EMAILS`` by default.
+        """
+        return (cls.AUTHORS_EMAILS
+                if cls.MAINTAINERS_EMAILS == set() else
+                cls.MAINTAINERS_EMAILS)
```
Makman2: otherwise you can't set them easily via `AUTHORS_EMAILS = ["a@b.com"]`.

Also we need to maintain compatibility

Makman2: I think this is something we should reconsider afterwards, the current/old bear does the same

coalib/core/DependencyTracker.py
```diff
@@ -151,6 +153,20 @@ def append_to_dependencies(prev, nxt):
...
+        Returns a set of all registered dependants.
+
+        >>> tracker = DependencyTracker()
+        >>> tracker.add(0, 1)
+        >>> tracker.add(0, 2)
+        >>> tracker.add(1, 3)
+        >>> tracker.dependants
+        {1, 2, 3}
+        """
+        return set(chain.from_iterable(self._dependency_dict.values()))
```
SanketDG: why use from_iterable?

tests/core/CoreTest.py
Makman2: Sometimes you need two different sections, sometimes not. I wasn't sure how to unify this, so I made variables for them everywhere (seems I missed some).

So shall I use variables consistently or shall I put those sections and file-dicts into `setUp()`?

coalib/core/CircularDependencyError.py
```diff
@@ -0,0 +1,19 @@
...
+
+        :param node:
+            The node that was encountered twice and closes the dependency
+            circle.
+        """
+        message = 'Circular dependency detected.'
+        if node is not None:
+            message += ' {0!r} -> ... -> {0!r}'.format(node)
+
+        Exception.__init__(self, message)
```
Makman2: we use in coala currently the `__init__` method, there was a discussion about to switch to `super`. Imo that's a different topic for now^^

coalib/core/Graphs.py
```diff
@@ -0,0 +1,59 @@
+from coalib.core.CircularDependencyError import CircularDependencyError
+
+
+def traverse_graph(start_nodes, get_successive_nodes,
```
Makman2: it's actually useful, see https://github.com/coala/coala/pull/3523/commits/5b3b87c96f262769946d1e3617170c915323b49e

Each edge I traverse in the dependency graph is a dependency relation I feed the `DependencyTracker` with.

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    A ``DependencyTracker`` allows to register and manage dependencies between
+    objects.
+
+    This class uses a directed graph to track relations.
+
+    Add a dependency relation between two objects:
+
+    >>> object1 = object()
+    >>> object2 = object()
```
Makman2: `object` shall demonstrate that you can use arbitrary objects + a new object is unique

Makman2: I like having `object` here, but if you want I can change to integers^^

coalib/core/DependencyTracker.py
Makman2: because I chain from iterable :) This would also be possible:
```
chain(*self._dependency_dict.values())
```

coalib/core/Bear.py
SanketDG: I was thinking of something like creating a new bear and implementing on the method, but it looks like its more suitable for the tests

tests/core/CoreTest.py
SanketDG: setUp() looks better for uniformity, but feel free to use anything.

coalib/core/CircularDependencyError.py
SanketDG: you would still use `__init__` though, just `super().__init__(message)`

but if this is what is always used in coala, then that's okay.

coalib/core/DependencyTracker.py
SanketDG: let's keep it then

coalib/core/Bear.py
Makman2: a plain implementation of `generate_tasks` is meaningless without the rest of the class^^ a doctest is too much imo^^

coalib/core/DependencyTracker.py
SanketDG: uh nvm

coalib/core/Core.py
Makman2: good point I don't need them yet in this commit. :+1:

coalib/core/Bear.py
SanketDG: yes true, ignore

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+        {}
+        >>> dependency_bear = Bear(section, file_dict)
+        >>> bear.add_dependency_results(dependency_bear, [1, 2, 3])
+        >>> bear.dependency_results
+        {<class 'coalib.core.Bear.Bear'>: [1, 2, 3]}
+
+        :param bear:
+            The bear instance to add results for.
+        :param dependency_results:
+            The results to add.
```
satwikkansal: Maybe mention `An iterable of results to add.` because later in line 251, doing `list(dependency_result)` will lead to an `TypeError` or unwanted results if a non-iterable or string is passed.

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+        """
+        filename = join(cls.data_dir, filename)
+        if exists(filename):
+            return filename
+
+        logging.info('{}: Downloading {} into {!r}.'
+                     .format(cls.name, url, filename))
+
+        response = requests.get(url, stream=True, timeout=20)
+        with open(filename, 'wb') as file:
```
satwikkansal: Shouldn't we check for the status code of the response here?

tests/core/BearTest.py
```diff
@@ -0,0 +1,261 @@
...
+                          'y': ('Second value.', int)})
+        self.assertEqual(
+            uut.optional_params,
+            {'z': ("Third value. (Optional, defaults to '33'.)", int, 33)})
+
+    def test_get_non_optional_settings(self):
+        self.assertEqual(BearWithAnalysis.get_non_optional_settings(),
+                         {'x': ('First value.', int),
+                          'y': ('Second value.', int)})
+
```
satwikkansal: Is it a good idea to test for circular dependency chains and confirm that nothing is returned?

coalib/core/Bear.py
Makman2: you are right, though this function will be removed (see https://github.com/coala/coala/pull/3711) as it's too primitive and using `dependency_dictionary` directly is easier :)

coalib/core/Bear.py
Makman2: 1. Not sure if `requests.get` will thren throw an exception
2. This is the current code of the old bear, just copied it. If you have improve suggestions, file an issue please :)

tests/core/BearTest.py
Makman2: This function will never return then. It's simpler on purpose, because:
1. less code
2. less complicated
3. it's actually hard to create a circular dependency, you really have to create one purposely

See the old/current bear where we do this (nearly) the same

coalib/core/Bear.py
adhikasp: For note this will **not** throw exception, [source](http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes).

Makman2: This was implemented for the old bear on master and adapted in https://github.com/coala/coala/pull/3711 :+1: This function now properly throws an exception.

sils:
please don't submit this for review. Submit your small PRs, keep this out of the queue, it's wasting peoples time.

Mixih:
after some discussion, we've decided to move next gen to 0.12 earliest

thisch:
@Makman2 what are your plans in this PR? https://github.com/coala/coala/pull/3523#issuecomment-271765839 mentions that it contains Bear v2, which it does not unfortunately. Can you please update the description of this PR?

Makman2:
@thisch this was the original PR containing all new core code. I've made several smaller PRs, and rebased this one each time a merge did happen. As only the `FileBear` and `ProjectBear` are left to merge, there's no need any more for this PR.

Please don't delete the branch yet.

PR #3526, opened 2017-01-11
Bear: Implement default `check_all_deps`

arush0311:
This commit adds a default implementation of check_all_deps
for checking all requirements associated with that bear, it
also calls `Bear.check_prerequisite` for custom requirements
handling by the bear developer.

Closes https://github.com/coala/coala/issues/3219

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -212,13 +212,14 @@ def _create_linter(klass, options):
             return options['executable']
 
         @classmethod
-        def check_prerequisites(cls):
+        def check_all_deps(cls):
             """
             Checks whether the linter-tool the bear uses is operational.
 
             :return:
                 True if operational, otherwise a string containing more info.
             """
+
             if shutil.which(cls.get_executable()) is None:
                 return (repr(cls.get_executable()) + ' is not installed.' +
                         (' ' + options['executable_check_fail_info']
@@ -233,7 +234,7 @@ def _create_linter(klass, options):
                         return True
                     except (OSError, CalledProcessError):
                         return options['prerequisite_check_fail_message']
-                return True
+                return super().check_all_deps()
 
         @classmethod
         def _get_create_arguments_metadata(cls):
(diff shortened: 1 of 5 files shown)


coalib/bears/Bear.py
```diff
@@ -316,9 +316,9 @@ def setup_dependencies():
         """
 
     @classmethod
-    def check_prerequisites(cls):
+    def check_all_deps(cls):
```
rhemon: I am not sure but isn't this kind of risky? since there are other bears in coala-bears which are implementing this method with check_prequisities and later tests call that. this requires to have those changes in coala-bears as well, right?

arush0311: actually no, check_all_deps later calls check_prequisites and bears can still use that :D

rhemon:
oh ok. cool

coalib/bearlib/abstractions/Linter.py
```diff
@@ -212,13 +212,14 @@ def get_executable():
...
         @classmethod
-        def check_prerequisites(cls):
+        def check_all_deps(cls):
             """
             Checks whether the linter-tool the bear uses is operational.
 
             :return:
                 True if operational, otherwise a string containing more info.
             """
+
```
jayvdb: please remove the extra blank line here.

jayvdb:
As far as I can see, currently this PR doesnt actually do anything very differently.  It appears to rename `check_prequisites` to `check_all_deps` which then invokes `check_prequisites` , which will have the effect of the logic currently in `check_prequisites` not being overridden by the bears that implement `check_prequisites` .
If that is all it is doing, it would be simpler to require that the bears implementing `check_prequisites` also call `super().check_prequisites()`

arush0311:
@sils is this pr needed ^?

jayvdb:
Blocked by https://github.com/coala/coala/issues/3712

jayvdb:
Also blocked by blocked by https://github.com/coala/coala/issues/4064 because the `is_installed` checks of some Requirement classes were very broken .

arush0311:
@jayvdb it doesnot matter is as isinstalled is already called by check_prerequisites method I think :(

arush0311:
@jayvdb 
> it would be simpler to require that the bears implementing check_prequisites also call super().check_prerequisites()

Also super().check_prerequisites() wont work in case of @linter as super() wont resolve to anything

https://github.com/coala/coala/pull/3526#issuecomment-272793295

jayvdb:
https://github.com/coala/coala/pull/4134 is merged; the bears need updating so they dont fail this new `check_all_deps`. Now it is blocked on https://github.com/coala/coala-bears/issues/1650

jayvdb:
Here is what this looks like so far : https://travis-ci.org/jayvdb/coala/builds/226002010

Need more of https://github.com/coala/coala-bears/issues/1650 fixed ;-)

jayvdb:
So the blocker is currently https://gitlab.com/coala/package_manager/issues/121

PR #3527, opened 2017-01-11
Sils/simply

sils:


Nosferatul:
ack 8890d0e

sils:
reack https://github.com/coala/coala/pull/3527/commits/8890d0e0db64571fe569e6ca9664b02f8949c8be

sims1253:
ack 1b976e5cf9aa13c03ffdf082abb8fa0643e2e7a3

coalib/output/ConsoleInteraction.py
```diff
@@ -201,23 +214,18 @@ def print_lines(console_printer,
 
         printed_chars = 0
         if i == sourcerange.start.line and sourcerange.start.column:
-            console_printer.print(highlight_text(
-                no_color, line[:sourcerange.start.column - 1], lexer), end='')
+            print_normal(line[:sourcerange.start.column - 1])
```
sims1253: BackgroundSourceRangeStyle was not in the old call but in the new one.
So maybe this also is a print_normal?

sils: yes it is, what's wrong?

sils:
we're getting a coverage issue on the CI because the code previously was only covered because one of the lines was doubly run... I'm going to remove that commit changing this and let the rest be like this

coalib/output/ConsoleInteraction.py
sims1253: must have skipped a line. Thought you used the other method... Fine then :)

PR #3534, opened 2017-01-12
Linter: Use shutil.which for real executable path

userzimmermann:
@sils @Makman2 

`@linter(executable='...')` causes problems under Windows when executables are `.cmd` or `.bat` files, which is default for node.js tools for example. So I improved `LinterBase.get_executable()` to return the real full executable path with ` shutil.which(options['executable'])`

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -207,9 +207,9 @@ def _create_linter(klass, options):
             Returns the executable of this class.
 
             :return:
-                The executable name.
+                The full path to the executable file.
             """
-            return options['executable']
+            return shutil.which(options['executable'])
 
         @classmethod
         def check_prerequisites(cls):


jayvdb:
I think this needs an issue which walks through the problem you are encountering.
If the same problem can not occur on unix, the fix should not be in unix code paths.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -207,9 +207,9 @@ def get_executable():
             Returns the executable of this class.
 
             :return:
-                The executable name.
+                The full path to the executable file.
             """
-            return options['executable']
+            return shutil.which(options['executable'])
```
Makman2: maybe just assign `shutil.which(executable)` to `options['executable']`, below at `def linter(...)`?

Makman2:
@jayvdb it's still good to work with the full path though I think^^

Makman2:
let's rebase @userzimmermann :)

userzimmermann:
@Makman2 done :)

coalib/bearlib/abstractions/Linter.py
Makman2: can we maybe cache the value? For 1000+ files this could be a bit slow^^

sils:
Simple lru cache

coalib/bearlib/abstractions/Linter.py
arush0311: sorry for the noob question, but why not just this https://github.com/coala/coala/pull/3534#discussion_r95806398 as you suggested instead going for cache?

Makman2: @userzimmermann did not change and I thought he had a reason for it. But assigning the full path is definitely better, you're right^^ @arush0311 :+1:

rhemon:
This is related to Issue https://github.com/coala/coala/issues/3654

userzimmermann:
Obsoleted by #3679

Issue #3537, opened 2017-01-12
deprecate_settings: Allow multiple deprecations


pratyushprakash:
`deprecate_settings` currently allows only one deprecation per setting.

PR #3546, opened 2017-01-13
deprecate_settings: Allow multiple deprecations

pratyushprakash:
Add support for multiple deprecations of the
same setting.

Closes https://github.com/coala/coala/issues/3537

### DIFF
--- a/coalib/bearlib/__init__.py
+++ b/coalib/bearlib/__init__.py
@@ -9,6 +9,9 @@ import logging
 from coalib.settings.FunctionMetadata import FunctionMetadata
 
 
+def _do_nothing(x): return x
+
+
 def deprecate_settings(**depr_args):
     """
     The purpose of this decorator is to allow passing old settings names to
@@ -64,15 +67,29 @@ def deprecate_settings(**depr_args):
     >>> run(old='Hello!', new='Hello!')
     Hello!
 
+    Multiple deprecations can be provided for the same setting. The modifier
+    function for each deprecated setting can be given as a value in a dict
+    where the deprecated setting is the key. A default modifier may be
+    specified at the end of the deprecated settings tuple.
+
+    >>> @deprecate_settings(new=({'old': lambda x: x + ' coala!'},
+    ...                          'older',
+    ...                           lambda x: x + '!' ))
+    ... def run(new):
+    ...     print(new)
+    >>> run(old='Hi')
+    WARNING:root:The setting `old` is deprecated. Please use `new` instead.
+    Hi coala!
+    >>> run(older='Hi')
+    WARNING:root:The setting `older` is deprecated. Please use `new` instead.
+    Hi!
+
     The metadata for coala has been adjusted as well:
 
     >>> list(run.__metadata__.non_optional_params.keys())
     ['new']
     >>> list(run.__metadata__.optional_params.keys())
-    ['old']
-
-    You cannot deprecate an already deprecated setting. Don't try. It will
-    introduce non-deterministic errors in your program.
+    ['old', 'older']
 
     :param depr_args: A dictionary of settings as keys and their deprecated
                       names as values.
@@ -82,11 +99,9 @@ def deprecate_settings(**depr_args):
         logged_deprecated_args = set()
 
         def wrapping_function(*args, **kwargs):
-            for arg, depr_arg_and_modifier in depr_args.items():
-                deprecated_arg, _func = (
-                    depr_arg_and_modifier
-                    if isinstance(depr_arg_and_modifier, tuple)
-                    else (depr_arg_and_modifier, lambda x: x))
+            for arg, depr_value in wrapping_function.__metadata__.depr_values:
+                deprecated_arg = depr_value[0]
+                _func = depr_value[1]
                 if deprecated_arg in kwargs:
                     if deprecated_arg not in logged_deprecated_args:
                         logging.warning(
@@ -105,11 +120,32 @@ def deprecate_settings(**depr_args):
             return func(*args, **kwargs)
 
         new_metadata = FunctionMetadata.from_function(func)
-        for arg, depr_arg_and_modifier in depr_args.items():
-            deprecated_arg = (depr_arg_and_modifier[0]
-                              if isinstance(depr_arg_and_modifier, tuple)
-                              else depr_arg_and_modifier)
-            new_metadata.add_deprecated_param(arg, deprecated_arg)
+        new_metadata.depr_values = []
+        for arg, depr_values in depr_args.items():
+            if not isinstance(depr_values, tuple):
+                depr_values = (depr_values,)
+
+            if callable(depr_values[-1]):
+                deprecated_args = depr_values[:-1]
+                default_modifier = depr_values[-1]
+            else:
+                deprecated_args = depr_values
+                default_modifier = _do_nothing
+
+            for depr_value in deprecated_args:
+                if isinstance(depr_value, dict):
+                    deprecated_arg = list(depr_value.keys())[0]
+                    modifier = depr_value[deprecated_arg]
+                    new_metadata.depr_values.append((arg,
+                                                     (deprecated_arg,
+                                                      modifier)))
+                else:
+                    deprecated_arg = depr_value
+                    new_metadata.depr_values.append((arg,
+                                                     (deprecated_arg,
+                                                      default_modifier)))
+                new_metadata.add_deprecated_param(arg, deprecated_arg)
+
         wrapping_function.__metadata__ = new_metadata
 
         return wrapping_function


namanyadav12:
I think you should add tests to show that your changes work.

pratyushprakash:
I have added tests.

namanyadav12:
Oh. Didn't know that.
Thanks for the info.

coalib/bearlib/__init__.py
```diff
@@ -82,34 +96,49 @@ def _deprecate_decorator(func):
...
+                if not isinstance(depr_args_and_modifier, tuple):
+                    depr_args_and_modifier = (depr_args_and_modifier,)
+
+                if callable(depr_args_and_modifier[-1]):
+                    deprecated_args = depr_args_and_modifier[:-1]
+                    _func = depr_args_and_modifier[-1]
+                else:
+                    deprecated_args = depr_args_and_modifier
+
+                    def _func(x): return x
```
satwikkansal: I'm wondering what this line does?

pratyushprakash: We need a function to modify the deprecated arg.

pratyushprakash: @satwikkansal `depr_arg_value = _func.__call__(kwargs[deprecated_arg])`

coalib/bearlib/__init__.py
```diff
@@ -64,12 +64,26 @@ def deprecate_settings(**depr_args):
     >>> run(old='Hello!', new='Hello!')
     Hello!
 
+    Multiple deprecations can be provided for the same setting. Note that
+    the modifier function if required must always be given as the last element
+    of the deprecated settings tuple.
+
+    >>> @deprecate_settings(new=('old','older'))
```
sils: uhm what if I have two different conversion functions?

Asnelchristian:
this looks good to me

coalib/bearlib/__init__.py
pratyushprakash: How would you like to handle this?

sils: maybe alllow a dict so for every old value you can give a conversion function or just None for the `lambda x:x`

coalib/bearlib/__init__.py
```diff
@@ -64,12 +64,27 @@ def deprecate_settings(**depr_args):
     >>> run(old='Hello!', new='Hello!')
     Hello!
 
+    Multiple deprecations can be provided for the same setting. The modifier
+    function for each deprecated setting can be given as a value in  a dict
+    where the deprecated setting is the key. A default modifier may be
+    specified at the end of the deprecated settings tuple.
+
+    >>> @deprecate_settings(new=({'old': lambda x: x + ' coala!'},'older'))
```
jayvdb: one space needed before `'older'`

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,57 @@ def _deprecate_decorator(func):
...
+                if not isinstance(depr_args_and_modifier, tuple):
+                    depr_args_and_modifier = (depr_args_and_modifier,)
+
+                if callable(depr_args_and_modifier[-1]):
+                    deprecated_args = depr_args_and_modifier[:-1]
+                    _func = depr_args_and_modifier[-1]
+                else:
+                    deprecated_args = depr_args_and_modifier
+
+                    def _func(x): return x
```
jayvdb: this should be a module level (private) function.  It does not need to be re-evaluated every time the decorator is run with a non-callable.

coalib/bearlib/__init__.py
```diff
@@ -82,34 +100,56 @@ def _deprecate_decorator(func):
...
+                        if deprecated_arg not in logged_deprecated_args:
+                            logging.warning(
+                                'The setting `{}` is deprecated. Please use'
+                                ' `{}` instead.'.format(deprecated_arg, arg))
+                            logged_deprecated_args.add(deprecated_arg)
+                        depr_arg_value = modifier.__call__(
+                            kwargs[deprecated_arg])
+                        if arg in kwargs and depr_arg_value != kwargs[arg]:
+                            logging.warning(
+                                'The value of `{}` and `{}` are conflicting.'
```
jayvdb: there is no test coverage of the 'are conflicting' message.  Could you please add it as a doctest in the docstring

coalib/bearlib/__init__.py
```diff
@@ -82,34 +100,56 @@ def _deprecate_decorator(func):
         logged_deprecated_args = set()
 
         def wrapping_function(*args, **kwargs):
-            for arg, depr_arg_and_modifier in depr_args.items():
-                deprecated_arg, _func = (
```
jayvdb: I dont see why you need to modify this line.

`deprecated_arg` will contain a dict, but that can be fixed later.

coalib/bearlib/__init__.py
```diff
@@ -82,34 +100,56 @@ def _deprecate_decorator(func):
...
-                        logged_deprecated_args.add(deprecated_arg)
-                    depr_arg_value = _func.__call__(kwargs[deprecated_arg])
-                    if arg in kwargs and depr_arg_value != kwargs[arg]:
-                        logging.warning(
-                            'The value of `{}` and `{}` are conflicting.'
-                            ' `{}` will be used instead.'.format(
-                                  deprecated_arg, arg, arg))
+            for arg, depr_args_and_modifier in depr_args.items():
+                if not isinstance(depr_args_and_modifier, tuple):
+                    depr_args_and_modifier = (depr_args_and_modifier,)
```
jayvdb: I find it confusing to modify the `key` or `value` variables pulled out of a dict.  When I read `for arg, depr_args_and_modifier in depr_args.items()` at the start of a loop, I expect `depr_args_and_modifier` will always be the same as what occurred in the dict.

coalib/bearlib/__init__.py
pratyushprakash: actually `deprecated_arg` will contain a tuple

coalib/bearlib/__init__.py
pratyushprakash: L57 already covers this

coalib/bearlib/__init__.py
pratyushprakash: Instead of handling the case of one deprecated setting and multiple deprecated settings individually, `depr_args_and_modifier = (depr_args_and_modifier,)` basically just converts the single deprecated setting into a tuple and handles it in the same way.

coalib/bearlib/__init__.py
jayvdb: :+1:

coalib/bearlib/__init__.py
pratyushprakash: Also each deprecated setting can have its own modifier.

pratyushprakash:
@jayvdb Could you take a look at this?

coalib/bearlib/__init__.py
sils: @jayvdb is right. You're getting an arg_tuple here just to rename it to depr_args_and_modifier again and never use arg_tuple. That's unneeded complexity.

pratyushprakash: This was done on request of @jayvdb. He did not want me changing the value pulled out of a dict.

pratyushprakash:
@jayvdb I still don't know what has to be changed. Could you please elaborate your review? Sorry for the trouble but I'm getting conflicting reviews

coalib/bearlib/__init__.py
jayvdb: For this case, you modify a for loop variable to be a `tuple`, and then unpack that variable, and don't use the for loop variable again.  That is unnecessary complexity.

I believe it can be changed to

```python
    if (isinstance(depr_args_and_modifier, tuple) and
            callable(depr_args_and_modifier[-1])):
        deprecated_args = depr_args_and_modifier[:-1]
        _func = depr_args_and_modifier[-1]
    else:
        deprecated_args = depr_args_and_modifier
        _func = _do_nothing
```

Cleaner, simpler.

pratyushprakash:
I don't know why the circle build always fails.

coalib/bearlib/__init__.py
```diff
@@ -64,12 +67,27 @@ def deprecate_settings(**depr_args):
...
     The metadata for coala has been adjusted as well:
 
     >>> list(run.__metadata__.non_optional_params.keys())
     ['new']
     >>> list(run.__metadata__.optional_params.keys())
-    ['old']
+    ['old', 'older']
 
     You cannot deprecate an already deprecated setting. Don't try. It will
     introduce non-deterministic errors in your program.
```
sils: delete this note it's meant for the case you're fixing

pratyushprakash: @sils Doesn't it still hold. The keyword of the keyword argument to the deprecate_settings decorator must always be the in use setting?

pratyushprakash: I think the note is warning against the use of `@deprecate_settings(s=(s1,s2),s1=s2)` . Correct me if I am wrong.

sils: it should at least be rephrased or removeed IMO

pratyushprakash:
@sils Done

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,57 @@ def _deprecate_decorator(func):
...
+
             return func(*args, **kwargs)
 
         new_metadata = FunctionMetadata.from_function(func)
-        for arg, depr_arg_and_modifier in depr_args.items():
-            deprecated_arg = (depr_arg_and_modifier[0]
-                              if isinstance(depr_arg_and_modifier, tuple)
-                              else depr_arg_and_modifier)
-            new_metadata.add_deprecated_param(arg, deprecated_arg)
+        for arg, depr_args_and_modifier in depr_args.items():
```
jayvdb: this section of code should be the only code which looks at `depr_args`.
It should save whatever is needed into `__metadata__`, possibly with new fields added.

This chunk of code is run once -- **everything** *possible* should be done here to prepare the decoration, so that each invocation of `wrapping_function` is as fast as possible.

Rename `arg, depr_args_and_modifier` to `arg, depr_value` as its contents & datatype is now not known here

coalib/bearlib/__init__.py
```diff
@@ -82,34 +100,57 @@ def _deprecate_decorator(func):
...
-                            'The setting `{}` is deprecated. Please use `{}` '
-                            'instead.'.format(deprecated_arg, arg))
-                        logged_deprecated_args.add(deprecated_arg)
-                    depr_arg_value = _func.__call__(kwargs[deprecated_arg])
-                    if arg in kwargs and depr_arg_value != kwargs[arg]:
-                        logging.warning(
-                            'The value of `{}` and `{}` are conflicting.'
-                            ' `{}` will be used instead.'.format(
-                                  deprecated_arg, arg, arg))
+            for arg, depr_args_and_modifier in depr_args.items():
```
pratyushprakash: @jayvdb I don't understand. The previous implementation also iterated through `depr_args`. Also I don't think it's possible to use the iterator provided by `FunctionMetadata` as `wrapping_function.__metadata__.deprecated_params` would only contain the deprecated settings and not the settings that replaced them.

JayVB:
Why am I being added to this thread?!?

nemaniarjun:
@JayVB Someone might have typed your handle instead of @jayvdb

coalib/bearlib/__init__.py
```diff
@@ -64,15 +67,27 @@ def deprecate_settings(**depr_args):
...
+    Hi!
+
     The metadata for coala has been adjusted as well:
 
     >>> list(run.__metadata__.non_optional_params.keys())
     ['new']
     >>> list(run.__metadata__.optional_params.keys())
-    ['old']
-
-    You cannot deprecate an already deprecated setting. Don't try. It will
```
pratyushprakash: @sils requested that this be removed.

jayvdb: c.f. https://github.com/coala/coala/pull/3546#discussion_r99503527

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,52 @@ def _deprecate_decorator(func):
...
+                else:
+                    deprecated_arg = depr_value
+                    modifier = (
+                        wrapping_function.__metadata__.default_modifier)
                 if deprecated_arg in kwargs:
                     if deprecated_arg not in logged_deprecated_args:
                         logging.warning(
-                            'The setting `{}` is deprecated. Please use `{}` '
-                            'instead.'.format(deprecated_arg, arg))
+                            'The setting `{}` is deprecated. Please use'
```
jayvdb: why is this modified?

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,52 @@ def _deprecate_decorator(func):
...
+                            ' `{}` instead.'.format(deprecated_arg, arg))
                         logged_deprecated_args.add(deprecated_arg)
-                    depr_arg_value = _func.__call__(kwargs[deprecated_arg])
+                    depr_arg_value = modifier.__call__(
+                        kwargs[deprecated_arg])
                     if arg in kwargs and depr_arg_value != kwargs[arg]:
                         logging.warning(
                             'The value of `{}` and `{}` are conflicting.'
                             ' `{}` will be used instead.'.format(
-                                  deprecated_arg, arg, arg))
```
jayvdb: this doesnt need changing

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,52 @@ def _deprecate_decorator(func):
...
+                                deprecated_arg, arg, arg))
                     else:
                         kwargs[arg] = depr_arg_value
                     del kwargs[deprecated_arg]
+
```
jayvdb: while I do prefer a blank line here, I'd much prefer that this patch is shorter so it is easier to see what is being changed.

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,52 @@ def _deprecate_decorator(func):
...
                 if deprecated_arg in kwargs:
                     if deprecated_arg not in logged_deprecated_args:
                         logging.warning(
-                            'The setting `{}` is deprecated. Please use `{}` '
-                            'instead.'.format(deprecated_arg, arg))
+                            'The setting `{}` is deprecated. Please use'
+                            ' `{}` instead.'.format(deprecated_arg, arg))
                         logged_deprecated_args.add(deprecated_arg)
-                    depr_arg_value = _func.__call__(kwargs[deprecated_arg])
+                    depr_arg_value = modifier.__call__(
```
jayvdb: can you continue using the variable name `_func`, so this change is smaller and more clear.

coalib/bearlib/__init__.py
```diff
@@ -82,34 +97,52 @@ def _deprecate_decorator(func):
         logged_deprecated_args = set()
 
         def wrapping_function(*args, **kwargs):
-            for arg, depr_arg_and_modifier in depr_args.items():
-                deprecated_arg, _func = (
-                    depr_arg_and_modifier
-                    if isinstance(depr_arg_and_modifier, tuple)
-                    else (depr_arg_and_modifier, lambda x: x))
+            for arg, depr_value in wrapping_function.__metadata__.depr_values:
+                if isinstance(depr_value, dict):
```
jayvdb: I think this logic can also be removed into the creation of ` __metadata__` , probably by adding a new optional argument to `add_deprecated_param`

coalib/bearlib/__init__.py
```diff
@@ -82,18 +97,17 @@ def _deprecate_decorator(func):
...
+                deprecated_arg = depr_value[0]
+                _func = depr_value[1]
                 if deprecated_arg in kwargs:
                     if deprecated_arg not in logged_deprecated_args:
                         logging.warning(
                             'The setting `{}` is deprecated. Please use `{}` '
                             'instead.'.format(deprecated_arg, arg))
                         logged_deprecated_args.add(deprecated_arg)
-                    depr_arg_value = _func.__call__(kwargs[deprecated_arg])
+                    depr_arg_value = _func.__call__(
```
jayvdb: unnecessary modification

pratyushprakash:
@jayvdb Done

pratyushprakash:
@jayvdb is this ready?

jayvdb:
I do not see a test case for a "A default modifier may be specified at the end of the deprecated settings tuple."

pratyushprakash:
@jayvdb I modified the doctest.

jayvdb:
ack 3cfe7a9

jayvdb:
@rultor merge

Issue #3529, opened 2017-01-11
files setting can not contain non-unicode files


rhemon:
```
(venv) MBPs-MacBook-Pro:coala-bears rhemon$ coala -f ./test-img/img.png 
Executing section Default...
[WARNING][18:23:39] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
Executing section python...
[WARNING][18:23:40] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
Executing section autopep8...
[WARNING][18:23:40] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
Executing section linelength...
[WARNING][18:23:43] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
Executing section CI...
Executing section commit...
[WARNING][18:23:44] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
Executing section rst...
Executing section PythonPackageInit...
[WARNING][18:23:45] Failed to read file '/Users/mbp13inch/Documents/Workspace/coala-bears/test-img/img.png'. It seems to contain non-unicode characters. Leaving it out.
```
Same result if I use jpg files.

pratyushprakash:
We don't have support for image files

Makman2:
We don't have generally support for files that are not utf8-encoded. This is going to come with the `FileProxy` we plan for bears. Currently we use the file-dict, which is loaded before a bear-run, so we can't do anything about it for now.

See https://github.com/coala/coala/issues/3530

jayvdb:
The PEP for FileProxy has the same problem. My PR of that PEP went some way towards removing that limitation.
However all that is orthogonal to this problem.  This is a global external linter bear with no dependencies. There is no reason for coala to be opening these files. Any opening and reading of these files by coala is performance problem, as the external linter will be doing the same.

Makman2:
The PEP is a bit outdated^^

[...]

PR #3560, opened 2017-01-15
Allow Bears to use raw files

yukiisbored:
Processing allows raw files if the Bear state that it uses raw
files. Bear that uses binary files can't be mixed with Bear that uses
text files. If this is enabled, bears are incharge of doing the file
handling (opening it, closing it, reading it, etc).

Closes https://github.com/coala/coala/issues/3529

### Checklist

### DIFF
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -242,14 +242,16 @@ def print_result(results,
     return retval or len(results) > 0, patched_results
 
 
-def get_file_dict(filename_list, log_printer):
+def get_file_dict(filename_list, log_printer, allow_raw_files=False):
     """
     Reads all files into a dictionary.
 
-    :param filename_list: List of names of paths to files to get contents of.
-    :param log_printer:   The logger which logs errors.
-    :return:              Reads the content of each file into a dictionary
-                          with filenames as keys.
+    :param filename_list:   List of names of paths to files to get contents of.
+    :param log_printer:     The logger which logs errors.
+    :param allow_raw_files: Allow the usage of raw files (non text files),
+                            disabled by default
+    :return:                Reads the content of each file into a dictionary
+                            with filenames as keys.
     """
     file_dict = {}
     for filename in filename_list:
@@ -257,6 +259,9 @@ def get_file_dict(filename_list, log_printer):
             with open(filename, 'r', encoding='utf-8') as _file:
                 file_dict[filename] = tuple(_file.readlines())
         except UnicodeDecodeError:
+            if allow_raw_files:
+                file_dict[filename] = None
+                continue
             log_printer.warn("Failed to read file '{}'. It seems to contain "
                              'non-unicode characters. Leaving it '
                              'out.'.format(filename))
@@ -339,7 +344,8 @@ def instantiate_processes(section,
                           cache,
                           log_printer,
                           console_printer,
-                          debug=False):
+                          debug=False,
+                          use_raw_files=False):
     """
     Instantiate the number of processes that will run bears which will be
     responsible for running bears in a multiprocessing environment.
@@ -351,6 +357,7 @@ def instantiate_processes(section,
     :param cache:            An instance of ``misc.Caching.FileCache`` to use as
                              a file cache buffer.
     :param log_printer:      The log printer to warn to.
+    :param use_raw_files:    Allow the usage of raw files (non text files)
     :param console_printer:  Object to print messages on the console.
     :param debug:            Bypass multiprocessing and activate debug mode
                              for bears, not catching any exceptions on running
@@ -369,7 +376,8 @@ def instantiate_processes(section,
     # This stores all matched files irrespective of whether coala is run
     # only on changed files or not. Global bears require all the files
     complete_filename_list = filename_list
-    complete_file_dict = get_file_dict(complete_filename_list, log_printer)
+    complete_file_dict = get_file_dict(complete_filename_list, log_printer,
+                                       use_raw_files)
 
     if debug:
         from . import DebugProcessing as processing
@@ -398,7 +406,8 @@ def instantiate_processes(section,
     # run only for the changed files if caching is enabled.
 
     # Start tracking all the files
-    if cache and loaded_valid_local_bears_count == loaded_local_bears_count:
+    if cache and (loaded_valid_local_bears_count == loaded_local_bears_count
+                  and not use_raw_files):
         cache.track_files(set(complete_filename_list))
         changed_files = cache.get_uncached_files(
             set(filename_list)) if cache else filename_list
@@ -464,6 +473,11 @@ def yield_ignore_ranges(file_dict):
         start = None
         bears = []
         stop_ignoring = False
+
+        # Do not process raw files
+        if file is None:
+            continue
+
         for line_number, line in enumerate(file, start=1):
             # Before lowering all lines ever read, first look for the biggest
             # common substring, case sensitive: I*gnor*e, start i*gnor*ing,
@@ -722,6 +736,22 @@ def execute_section(section,
         except IndexError:
             running_processes = get_cpu_count()
 
+    bears = global_bear_list + local_bear_list
+    use_raw_files = set(bear.USE_RAW_FILES for bear in bears)
+
+    if len(use_raw_files) > 1:
+        log_printer.err("Bears that uses raw files can't be mixed with "
+                        'Bears that uses text files. Please move the following '
+                        'bears to their own section: ' +
+                        ', '.join(bear.name for bear in bears
+                                  if not bear.USE_RAW_FILES))
+        return ((), {}, {}, {})
+
+    # use_raw_files is expected to be only one object.
+    # The if statement is to ensure this doesn't fail when
+    # it's running on an empty run
+    use_raw_files = use_raw_files.pop() if len(use_raw_files) > 0 else False
+
     processes, arg_dict = instantiate_processes(section,
                                                 local_bear_list,
                                                 global_bear_list,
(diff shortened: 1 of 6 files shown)


sils:
hey, I don't understand why we need this. Bears can use raw files right now and just ignore whatever coala gives them and that works reasonably well. Also the case of mixing bears that use raw files with native ones works. Your change will make that *not work* and users won't like that. Also it adds more code and more stuff I need to add to my bear and I don't see any benefit. Is there one that I miss?

jayvdb:
@yukiisbored , please rebase this.

jayvdb:
> Bears can use raw files right now and just ignore whatever coala gives them and that works reasonably well.

@sils, They can only do that if `files` has dummy values, and then all the 'modified files' functionality is broken.

> Also it adds more code and more stuff I need to add to my bear and I don't see any benefit.

No, this provides a default value, which only needs to be modified if the bear is happy to accept raw files.

This is blocking three binary bears:

- [ ] https://github.com/coala/coala-bears/pull/1288 - ImageDimensionBear
- [ ] https://github.com/coala/coala-bears/pull/1302 - ImageCompressionBear
- [ ] https://github.com/coala/coala-bears/pull/1543 - MP3 Bear

I need the first two for WMID.

coalib/processes/Processing.py
```diff
@@ -365,8 +370,18 @@ def instantiate_processes(section,
     # only on changed files or not. Global bears require all the files
     complete_filename_list = filename_list
 
+    # Checks if this section uses raw files or not
+    bears = chain(local_bear_list, global_bear_list)
+    use_raw_files = set(bear.USE_RAW_FILES for bear in bears)
+    if len(use_raw_files) > 1:
+        log_printer.err("Bears that uses raw files can't be mixed with "
+                        'Bears that uses text files. Skipping this '
+                        'section.')
```
sils: can we make them mixable? If they're mixed just load a raw and a normal file dict into memory or so?

jayvdb: That introduces lots of edge cases, and so far we have no sensible use case for that.  All raw bears we have in the pipeline will be loading files that cant be loaded as text file.

tests/processes/ProcessingTest.py
```diff
@@ -494,47 +529,6 @@ def test_yield_ignore_ranges(self):
         self.assertEqual(source_range.end.line, 1)
         self.assertEqual(source_range.end.column, 14)
 
-    def test_loaded_bears_with_error_result(self):
```
yukiisbored: Whoops, accidentally removed this during rebase.

tests/processes/ProcessingTest.py
```diff
@@ -36,8 +37,7 @@
 p=subprocess.Popen([sys.executable,
                   "-c",
                   "import time; time.sleep(0.1)"]);
-pgid = p.pid if platform.system() == "Windows" else os.getpgid(p.pid);
-print(p.pid, pgid)
+pgid = p.pid if platform.system() == "Windows" else os.getpgid(p.pid);print(p.pid, pgid)
```
yukiisbored: okay, I have no idea how this happened.

tests/processes/ProcessingTest.py
```diff
@@ -155,6 +157,63 @@ def test_empty_run(self):
...
+                         ", message='test msg', aspect=NoneType\\) at "
+                         '0x[0-9a-fA-F]+>'.format(hex(local_result.id)))
+        self.assertRegex(repr(global_result),
+                         "<Result object\\(id={}, origin='GlobalTestRawBear', "
+                         'affected_code=\\(.*start=.*file=.*section_executor_'
+                         'test_files.*line=None.*end=.*\\), severity=NORMAL, c'
+                         "onfidence=100, message='test message', "
+                         'aspect=NoneType\\'
+                         ') at 0x[0-9a-fA-F]+>'.format(hex(global_result.id)))
+
```
yukiisbored: nu-uh.

coalib/processes/Processing.py
```diff
@@ -385,8 +396,16 @@ def instantiate_processes(section,
     # the whole project is accessible to every bear. However, local bears are
     # run only for the changed files if caching is enabled.
 
+    if len(use_raw_files) > 1:
+        log_printer.err("Bears that uses raw files can't be mixed with "
+                        'Bears that uses text files. Skipping this '
+                        'section.')
+        mixed = True
+    use_raw_files = use_raw_files.pop() if len(use_raw_files) != 0 else False
```
Mixih: Isn't this going to always make use_raw_file False since the set {True, False} will pop False first because of current hash ordering? The pop also makes me a bit uncomfortable since it spec'ed as non-deterministic in the python docs.

coalib/processes/Processing.py
```diff
@@ -385,8 +396,16 @@ def instantiate_processes(section,
     # the whole project is accessible to every bear. However, local bears are
     # run only for the changed files if caching is enabled.
 
+    if len(use_raw_files) > 1:
+        log_printer.err("Bears that uses raw files can't be mixed with "
+                        'Bears that uses text files. Skipping this '
+                        'section.')
```
sils: I think it's crucial here to propose a solution to the user. E.g. "Move the following bears into an own section: ..."

The user doesn't care if a bear uses raw files or not and he also doesn't know so this message needs to be very specific about how to solve this IMO

coalib/processes/Processing.py
yukiisbored: Actually, it expects it to be only one object inside the Set. If there's more than one, this section will be skipped anyway since it's mixing bears that uses raw files and the ones that doesn't.

the `if len(use_raw_files) != 0` is added because there are some tests that executes this function without giving a bear.

coalib/processes/Processing.py
```diff
@@ -337,9 +342,10 @@ def instantiate_processes(section,
                           global_bear_list,
                           job_count,
                           cache,
-                          log_printer,
```
jayvdb: why are you removing this?

coalib/processes/Processing.py
```diff
@@ -337,9 +342,10 @@ def instantiate_processes(section,
...
+                          use_raw_files,
```
jayvdb: Never add parameters in the middle.  That is a breaking change, which causes errors like the following.  Move it to the end, and add a default value.

```
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/coalib/coala_main.py", line 122, in run_coala
    debug=debug or args and args.debug)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/coalib/processes/Processing.py", line 763, in execute_section
    debug=debug)
TypeError: instantiate_processes() got multiple values for argument 'console_printer'
```

coalib/processes/Processing.py
yukiisbored: Whoops, accidentally deleted it during rebase.

tests/processes/section_executor_test_files/ProcessingGlobalTestRawFileBear.py
```diff
@@ -0,0 +1,13 @@
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.results.Result import Result
+
+
+class ProcessingGlobalTestRawFileBear(GlobalBear):  # pragma: no cover
```
jayvdb: why no cover?
Add a comment above explaining it.
I am trying very hard (https://github.com/coala/coala/pull/4339) to remove "pragma: no cover", so only very specific exceptions are allowed.

tests/processes/section_executor_test_files/ProcessingLocalTestRawFileBear.py
```diff
@@ -0,0 +1,15 @@
+import time
+
+from coalib.bears.LocalBear import LocalBear
+from coalib.results.Result import Result
+
+
+class ProcessingLocalTestRawFileBear(LocalBear):  # pragma: no cover
```
jayvdb: per other comment.

tests/processes/ProcessingTest.py
```diff
@@ -155,6 +158,66 @@ def test_empty_run(self):
...
+                                  None,
+                                  self.log_printer,
+                                  console_printer=self.console_printer)
+        self.assertTrue(results[0])
+
+        local_results = self.result_queue.get(timeout=0)
+        global_results = self.result_queue.get(timeout=0)
+        self.assertTrue(self.result_queue.empty())
+
+        self.assertEqual(len(local_results), 1)
```
yukiisbored: Should I put it into another commit? Because I'm using the same way as the other one.

tests/processes/ProcessingTest.py
```diff
@@ -329,6 +387,25 @@ def test_get_file_dict_non_existent_file(self):
                        'an unknown error.'),
                       self.log_printer.log_queue.get().message)
 
+    def test_get_file_dict_allow_raw_file(self):
+        file_dict = get_file_dict([self.unreadable_path], self.log_printer,
+                                  True)
+        self.assertNotEqual(file_dict, {})
+        self.assertEqual(file_dict[self.unreadable_path], None)
+        self.assertNotIn(("Failed to read file '{}'. It seems to contain "
```
jayvdb: An idea that I just had over at https://github.com/coala/coala/pull/4270#discussion_r121594930 , ...

It would be more useful to assert that there are no warnings (or errors) in the message queue.

And as part of that, instead of `self.log_printer.log_queue.get().message`, fetch all messages into an array, so that if another message is emitted somewhere in new code before the target message, it doesnt cause this test to silently fail.

jayvdb:
ack 095aeb3

jayvdb:
@rultor merge

PR #3581, opened 2017-01-17
RequirementsTest.py: Add tests to check versions

### LINKED ISSUE #3511: `py.test` fails due to version conflict of packages
Enhance `py.test` to first install the appropriate packages before starting the tests. 
Can prevent confusions.

Shade5:
Added tests to check versions with requirements.txt

Closes #3511

### Checklist

### DIFF
--- a/tests/RequirementsTest.py
+++ b/tests/RequirementsTest.py
@@ -1,0 +1,67 @@
+import os
+import sys
+import unittest
+
+from coalib.misc.Shell import run_shell_command
+from packaging import version
+
+
+class coalaaCITest(unittest.TestCase):
+
+    def test_find_no_issues(self):
+        with open(os.path.dirname(__file__) + '/../requirements.txt') as f:
+            content = f.readlines()
+        reqiredpac = []
+        req_pac_ver = []
+
+        for x in content:
+            if '=' in x:
+                reqiredpac.append(x[:x.find('=')-1])
+                req_pac_ver.append(x[x.find('=')+1:-1])
+
+        with open(os.path.dirname(__file__) + '/../test-requirements.txt') as f:
+            content = f.readlines()
+        for x in content:
+            if '=' in x:
+                reqiredpac.append(x[:x.find('=')-1])
+                req_pac_ver.append(x[x.find('=')+1:-1])
+
+        output = run_shell_command([sys.executable, '-m', 'pip', 'freeze'])
+        split_list = output[0].split('\n')
+        package = []
+        temporary = []
+        for i in split_list:
+            package.append(i[:i.find('=')])
+            temporary.append(i[i.find('==')+2:])
+
+        package = [x.lower() for x in package]
+        reqiredpac = [x.lower() for x in reqiredpac]
+        package = [x.replace('-', '_') for x in package]
+        reqiredpac = [x.replace('-', '_') for x in reqiredpac]
+
+        cur_version = []
+        for k in temporary:
+            while(k.count('.') > 1):
+                k = k[:k.rfind('.')]
+            cur_version.append(k)
+
+        req_version = []
+        for k in req_pac_ver:
+            while(k.count('.') > 1):
+                k = k[:k.rfind('.')]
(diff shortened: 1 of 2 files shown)


Shade5:
Requirement.txt says that `colorlog  2.10.0` is necessary, but even travis has only `colorlog 2.7`

vijeth-aradhya:
Correct me if I'm wrong, shouldn't it be `Need 2.10.0 Installed: 2.7`? because `ver[pac.index(k)]` has `pip freeze` output right (that is what is installed)?

Shade5:
@vijeth-aradhya You were right. I messed It up. I've fixed it now. Also `wheel` and 'libclang-py3` don't show up in `pip-freeze', so I've excluded their tests.

Shade5:
The funny thing is 2.10 is newer than 2.7, although 2.10 < 2.7

tests/RequirementsTest.py
```diff
@@ -0,0 +1,64 @@
...
+        temp = []
+        for i in split_list:
+            package.append(i[:i.find('=')])
+            temp.append(i[i.find('==')+2:])
+
+        package = [x.lower() for x in package]
+        reqiredpac = [x.lower() for x in reqiredpac]
+        package = [x.replace('-', '_') for x in package]
+        reqiredpac = [x.replace('-', '_') for x in reqiredpac]
+
```
HitRam: This may cause a bug later I guess. I think this may be useful http://stackoverflow.com/questions/11887762/compare-version-strings

tests/RequirementsTest.py
```diff
@@ -0,0 +1,64 @@
...
+        for k in temp:
+            while(k.count('.') > 1):
+                k = k[:k.rfind('.')]
+            version.append(k)
+
+        req_version = []
+        for k in req_pac_ver:
+            while(k.count('.') > 1):
+                k = k[:k.rfind('.')]
+            req_version.append(k)
```
HitRam: This may introduce a bug later I guess. 
I think this might be useful http://stackoverflow.com/questions/11887762/compare-version-strings

Shade5:
@HitRam Thanks, that helped. I have used short forms for variables as the complicated list accesses become very long (more than 80 char) and become very ugly.
`x and y` are temporary variables and are used only in the loop, so it doesn't matter.
About the possible bug you mentioned, I don't think it is possible as it takes only the first two numbers no matter how long the versions are.

Shade5:
Travis does not have the `packaging` package, should I write a test for comparing versions?

HitRam:
Version 1.9.1 > 1.9.0 and according to your code, it would be mark as installed. So, I am not sure if this would matter or not?

Shade5:
@HitRam a newer package will be fine right? I don't think we should enforce a specific verision

tests/RequirementsTest.py
```diff
@@ -0,0 +1,69 @@
...
+
+        for k in reqiredpac:
+            if 'whe' not in k:
+                self.assertIn(k, package)
+
+        for k in reqiredpac:
+            if 'whe' not in k:
+                print(cur_version[package.index(k)]+'  ' +
+                      req_version[reqiredpac.index(k)])
```
sils: debug print?

Shade5: I'll fix that, but can the packages module be added to travis? It'll makes the version comparison simpler.

jayvdb:
Author is going to try creating a pytest plugin.

tests/RequirementsTest.py
```diff
@@ -0,0 +1,67 @@
...
+import unittest
+
+from coalib.misc.Shell import run_shell_command
+from packaging import version
+
+
+class coalaaCITest(unittest.TestCase):
+
+    def test_find_no_issues(self):
+        with open(os.path.dirname(__file__) + '/../requirements.txt') as f:
```
thisch: Use os.path.join to join the directories of the path, sth like

os.path.join(os.path.dirname(__file__), os.pardir, 'requirements.txt')

Issue #3367, opened 2016-12-22
setup.py install: Install a "commit-sha" field into coalib


Makman2:
As self-installs from a git clone have `dev9999999...` version, we should add a commit sha into the module, like `coalib.__sha__` or `coalib.__commit__` or whatever

This makes bisecting problems easier.

I would also add to the commit string whether the branch was dirty on install, or not, so it looks like
`ab87df1-dirty` or `ab87df1*`

Techievena:
@Makman2 see this [discussion](https://github.com/Homebrew/brew/issues/1738)

Makman2:
@Techievena where's the relevant information?

Techievena:
https://github.com/Homebrew/brew/issues/1738

Makman2:
Your issue is about installing from commit hash, this issue is about placing a commit-sha-variable into our `coalib` module or some other suited module. I'll rephrase the title a bit^^

arush0311:
Can I take this?

Makman2:
@arush0311 how do you want to call the field? :)

arush0311:
There is already a Version in coalib directory, we can add a `__commit__` file when setup.py runs?

arush0311:
@Makman2 but it should only be applicable for install right? ( not develop and not for pip installs ), perhaps will have to read more about setup.py

Makman2:
also for develop and especially for pip installs we should have a `__commit__` :)
you should look at how we set `__version__` :+1:

do you want to support also dirty git branches? I think it's good to have, if someone submits a bug report and you see something like `8afbc21*` then you know that he has untracked modifications made to the repo^^

arush0311:
So the __commit__ will change as soon as there are changes in the repo (in case of develop), perhaps git hooks can help (but they will not be persistent).
And how do we set version, I see there is a version file but who updates that?

Makman2:
no git hooks :) the `__commit__` variable will be generated by `setup.py install` or similar targets. If it detects during installation that it's inside a git repo (or we just require it), it will create a `__commit__` field with the commit sha and checks whether it's a dirty branch or not and modifies/generates `__commit__` accordingly.

[adjust_version_number.py](https://github.com/coala/coala/blob/master/.misc/adjust_version_number.py) is run inside [release.sh](https://github.com/coala/coala/blob/master/.misc/release.sh) which creates the `VERSION` file :+1:

arush0311:
[...]

PR #3595, opened 2017-01-19
coalib: Create `get_commit`

arush0311:
This commit introduces `__commit__` variable containing
the sha hash of the current commit or the commit of `HEAD`
at the time of installation

Closes https://github.com/coala/coala/issues/3367

### Checklist

### Reviewers
@Makman2

### DIFF
--- a/.misc/check_setuptools.py
+++ b/.misc/check_setuptools.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import subprocess
 import sys
+import re
 
 import setuptools
 
@@ -9,7 +10,8 @@ import setuptools
 def get_setuptools_version():
     with open('requirements.txt') as f:
         for line in f:
-            if line.startswith('setuptools'):
+            # Ensure that line matched is setuptools
+            if re.match(r'^setuptools(>|<|~|=|$)', line):
                 line = line.rstrip()
                 if '>=' not in line:
                     raise ValueError('%s doesnt use ">="' % line)
--- a/coalib/__init__.py
+++ b/coalib/__init__.py
@@ -7,17 +7,21 @@ formatting and settings are also included in coalib.
 
 import sys
 from os.path import join, dirname
+from pkg_resources import get_distribution, DistributionNotFound
 
 
 VERSION_FILE = join(dirname(__file__), 'VERSION')
 
 
-def get_version():
+def get_public_version():
     with open(VERSION_FILE, 'r') as ver:
         return ver.readline().strip()
 
 
-VERSION = get_version()
+try:
+    VERSION = get_distribution('coala').version
+except DistributionNotFound:
+    VERSION = get_public_version()
 __version__ = VERSION
 
 
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,9 @@ import setuptools.command.build_py
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
-from coalib import VERSION, assert_supported_version, get_version
+from setuptools_scm.version import get_local_node_and_date
+
+from coalib import assert_supported_version, get_public_version
 from coalib.misc.BuildManPage import BuildManPage
 
 try:
@@ -21,6 +23,8 @@ except (ValueError, UnicodeError):
 
 
 assert_supported_version()
+
+VERSION = get_public_version()
 
 
 class BuildPyCommand(setuptools.command.build_py.build_py):
@@ -72,6 +76,17 @@ with open('README.rst') as readme:
     long_description = readme.read()
 
 
+def _get_public_version_string(scm_version):
+    return VERSION
+
+
+def _get_local_version_string(scm_version):
+    if 'dev' in VERSION:
+        return get_local_node_and_date(scm_version)
+    else:
+        return ''
+
+
 if __name__ == '__main__':
     if platform.system() != 'Windows':
         data_files = [('.', ['coala.1'])]
@@ -79,7 +94,10 @@ if __name__ == '__main__':
         data_files = [('.', [])]
 
     setup(name='coala',
-          version=VERSION,
+          use_scm_version={
+              'version_scheme': _get_public_version_string,
+              'local_scheme': _get_local_version_string
+          },
           description='Linting and Fixing Code for All Languages',
           author='The coala developers',
           author_email='coala.analyzer@gmail.com',
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,5 @@ libclang-py3~=3.4.0
 Pygments~=2.1
 PyPrint~=0.2.6
 requests~=2.12
+setuptools-scm~=1.15.0
 setuptools>=17.0


setup.py
```diff
@@ -62,6 +63,9 @@ def run(self):
               '-b {}'.format(current_version)])
         VERSION = get_version()
 
+with open(COMMIT_FILE, 'w') as cf:
+    cf.write(get_current_commit() + '\n')
+
```
arush0311: Travis python 2.7 complains is we use `print(get_current_commit, file=cf)` as print is not a function

coalib/__init__.py
```diff
@@ -7,10 +7,46 @@
...
+COMMIT_FILE = join(dirname(__file__), 'COMMIT')
+
+
+def get_current_commit():
+    '''
+    :return:    string containing the current commit sha if coala is
+                initialized as a git repo otherwise an empty string
+    '''
+    try:
+        commit = check_output(['git', 'rev-parse', 'HEAD'])
```
jayvdb: Use `sarge` instead of using `check_output`, as `sarge` provides a sane and safe cross platform interface.

coalib/__init__.py
```diff
@@ -7,10 +7,46 @@
...
 from os.path import join, dirname
+from subprocess import check_output, CalledProcessError
 
 
 VERSION_FILE = join(dirname(__file__), 'VERSION')
 
+COMMIT_FILE = join(dirname(__file__), 'COMMIT')
+
+
+def get_current_commit():
```
jayvdb: this needs to be run in the coalib directory, otherwise it will be wrong.
It is more versatile if the caller specifies the directory that it should be run in, as an argument to this function.

coalib/__init__.py
```diff
@@ -7,10 +7,46 @@
...
+    '''
+    :return:    string containing the current commit sha if coala is
+                initialized as a git repo otherwise an empty string
+    '''
+    try:
```
jayvdb: Before trying to run a subprocess, which is a heavy call, check there is a `.git` directory, which is a quick check.

coalib/__init__.py
```diff
@@ -7,10 +7,46 @@
...
+        commit = ''
+    return commit
+
+
+def get_commit(commit_file):
+    '''
+    :return:    string containing the current commit sha if coala is a git repo
+                otherwise commit sha at the time of installation.
+    '''
+    commit = get_current_commit()
```
jayvdb: specify the coalib directory here.

coalib/__init__.py
```diff
@@ -7,10 +7,46 @@
 
 import sys
 from os.path import join, dirname
+from subprocess import check_output, CalledProcessError
 
 
 VERSION_FILE = join(dirname(__file__), 'VERSION')
```
jayvdb: Please save the `dirname` result, so it can be used multiple times.

coalib/__init__.py
arush0311: Actually, the .git is not present coalib, its only present in coala superdirectory, which is not present after the installation, so we cant check for .git folder

coalib/__init__.py
arush0311: Why exactly do we need sarge?
IMO sarge donot provide any benifit over check_output in this case
And it also introduces a new dependency. 
Also can we use an external dependency in setup.py?

coalib/__init__.py
```diff
@@ -6,10 +6,61 @@
...
+
+        status = check_output(['git', 'status', '--porcelain'])
+        status = status.decode('utf-8').strip()
+        if status != '':
+            commit += '*'
+    except (OSError, CalledProcessError) as e:
+        commit = ''
+
+    chdir(old_dir)
+    return commit
```
pratyushprakash: Maybe you could use `change_directory` from `coala_utils` here to shorten this up  a bit.

arush0311: actually we may not be able to use external dependencies in setup.py, and ithis function is directly being called by setup.py

arush0311:
@Makman2 can you review?

Udayan12167:
Put the tests in coalaTest

coalib/__init__.py
Makman2: > actually we may not be able to use external dependencies in setup.py

why not?

coalib/__init__.py
```diff
@@ -6,10 +6,61 @@
 
 
 import sys
+
```
Makman2: obsolete newline

coalib/__init__.py
arush0311: because one of the purpose of setup.py is to install dependencies, and these functions are run before that!

Makman2: ah obviously, nvm then :3

arush0311:
@Udayan12167 Actually, should I do that? the tests do have their own setup and teardown, which donot at all apply to other tests.

Udayan12167:
@arush0311 if you put the methods in the misc directory then you can create a separate file to test them. This way I feel it's difficult to find the tests and could make the test directory cluttered.

arush0311:
@Udayan12167 I understand your point, I will move the tests to coalaTest

Udayan12167:
No I think going the misc directory way is better. Make a `gitHelper.py` in that directory and a parallel `gitHelperTest.py`.

arush0311:
Is it good though? that way we have to keep track of one more file which donot contain external dep and is python2 friendly.

setup.py
```diff
@@ -13,6 +13,8 @@
 
 from coalib import VERSION, assert_supported_version, get_version
 from coalib.misc.BuildManPage import BuildManPage
+from coalib import get_current_commit, COMMIT_FILE, COALIB_DIR
```
jayvdb: these imports should be near the other imports from the same module.
one way to do that is to move this line up one line.

coalib/__init__.py
```diff
@@ -6,10 +6,60 @@
...
+COALIB_DIR = join(dirname(__file__))
+
+VERSION_FILE = join(COALIB_DIR, 'VERSION')
+
+COMMIT_FILE = join(COALIB_DIR, 'COMMIT')
+
+
+def get_current_commit(git_dir):
+    '''
+    :param git_dir:     string containing the directory to check the
```
jayvdb: 'git_dir' is quite ambiguous.  It could be the directory where `git` is located.

I think this should be more like `git_repo_dir`.

coalib/__init__.py
```diff
@@ -6,10 +6,60 @@
...
+                        current commit in.
+    :return:            string containing the current commit sha if `git_dir`
+                        is initialized as a git repo otherwise an empty string.
+    '''
+    old_dir = getcwd()
+    chdir(git_dir)
```
jayvdb: if we are already in the git repo dir (or any subdirectory), this `chdir` and the one below are not necessary.

coalib/__init__.py
```diff
@@ -6,10 +6,60 @@
...
-VERSION_FILE = join(dirname(__file__), 'VERSION')
+COALIB_DIR = join(dirname(__file__))
+
+VERSION_FILE = join(COALIB_DIR, 'VERSION')
+
+COMMIT_FILE = join(COALIB_DIR, 'COMMIT')
+
+
+def get_current_commit(git_dir):
+    '''
```
jayvdb: coala uses `"""` for docstring start and end.

jayvdb: coala uses `"""` for docstring start and end.

coalib/__init__.py
```diff
@@ -6,10 +6,60 @@
...
+    old_dir = getcwd()
+    chdir(git_dir)
+
+    try:
+        commit = check_output(['git', 'rev-parse', 'HEAD'])
+        commit = commit.decode('utf-8').strip()
+
+        status = check_output(['git', 'status', '--porcelain'])
+        status = status.decode('utf-8').strip()
+        if status != '':
```
jayvdb: `if status`

coalib/__init__.py
```diff
@@ -21,6 +71,10 @@ def get_version():
 __version__ = VERSION
 
 
+COMMIT = get_commit(COALIB_DIR, COMMIT_FILE)
```
jayvdb: this is a large startup overhead for the properly installed package, which is not a git repo.
The overhead for the properly installed package should be as **small** as possible.  No `chdir`; no `subprocess`.
The overhead for the development package, in a git repo, can be large, because it is not the production version.

arush0311:
@Makman2 please review :D

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+            commit += '*'
+    except (OSError, CalledProcessError) as e:
+        commit = ''
+
+    if(cwd_changed):
+        chdir(old_dir)
+    return commit
+
+
+def get_commit(git_repo_dir, commit_file):
```
SanketDG: why not just use the `COMMIT_FILE` variable?

arush0311: makes it easier to test.

setup.py
```diff
@@ -5,15 +5,21 @@
...
+import setuptools.command.install
+
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
 from coalib import VERSION, assert_supported_version, get_version
+from coalib import get_current_commit, COMMIT_FILE, COALIB_DIR
 from coalib.misc.BuildManPage import BuildManPage
 
+
```
Makman2: unrelated change --> separate commit :)

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+                            of installation.
+    :return:                string containing the current commit sha if
+                            `git_repo_dir` is a git repo otherwise commit sha
+                            in commit_file.
+    """
+    try:
+        with open(commit_file) as cf:
+            commit = cf.readline().strip()
+    # Not using FileNotFoundError, as python2 CI starts complaining
+    except (OSError, IOError):
```
Makman2: ah damn... we need to file an issue to get `FileNotFoundError` there after merge :)

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+
+def get_commit(git_repo_dir, commit_file):
+    """
+    :param git_repo_dir:    string containing the directory to check the
+                            current commit in.
+    :param commit_file:     filename containing commit sha of HEAD at the time
+                            of installation.
+    :return:                string containing the current commit sha if
+                            `git_repo_dir` is a git repo otherwise commit sha
+                            in commit_file.
```
Makman2: You should start with upper-case letters + Adjust alignment ;)
would recommend you to use this format btw:
```
:param x:
    4-space indented description
:return:
    And again.
```

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+COMMIT_FILE = join(COALIB_DIR, 'COMMIT')
+
+
+def get_current_commit(git_repo_dir):
+    """
+    :param git_repo_dir:    string containing the directory to check the
+                            current commit in.
+    :return:                string containing the current commit sha if
+                            `git_repo_dir`is initialized as a git repo
+                            otherwise an empty string.
```
Makman2: Start sentences with upper-case + Alignment (see below for recommendation on how to format docs)

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+
+COMMIT_FILE = join(COALIB_DIR, 'COMMIT')
+
+
+def get_current_commit(git_repo_dir):
+    """
+    :param git_repo_dir:    string containing the directory to check the
+                            current commit in.
+    :return:                string containing the current commit sha if
+                            `git_repo_dir`is initialized as a git repo
```
Makman2: missing space before `is`

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+                            otherwise an empty string.
+    """
+    old_dir = getcwd()
+    cwd_changed = False
+
+    if(old_dir != git_repo_dir):
```
Makman2: unnecessary braces

coalib/__init__.py
Makman2: I think it makes sense to inline the `COMMIT_FILE` and `COALIB_DIR` variables.

Do you know the `unittest.patch` decorator? :)
It allows you to patch fields in a different module, e.g.
```python
@unittest.patch('coalib.COMMIT_FILE', '/home/xyz')
def test_get_commit(self):
    ...
```

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+                            in commit_file.
+    """
+    try:
+        with open(commit_file) as cf:
+            commit = cf.readline().strip()
+    # Not using FileNotFoundError, as python2 CI starts complaining
+    except (OSError, IOError):
+        commit = ''
+
+    if(commit == 'develop'):
```
Makman2: maybe return the git sha when there's just no `COMMIT` file present?

setup.py
arush0311: What issue should I refer :D ? imo it seems very trivial

coalib/__init__.py
arush0311: as @jayvdb said it may increase production overhead. No commit file means the repo was not installed from a git repo, so there will be no git sha..

coalib/__init__.py
arush0311: you mean make them one variable? or remove them from parameters?

coalib/__init__.py
arush0311: noted :D. as for the case most of the description started with `string`. should I capitalize that?(or perhaps use `str`?)

coalib/__init__.py
arush0311: https://github.com/coala/coala/pull/3595#discussion_r96984596

coalib/__init__.py
arush0311: https://github.com/coala/coala/pull/3595#discussion_r98343215

setup.py
Makman2: no issue, just a new commit

I know this is trivial, but for the sake of making commits as atomic as possible and ease review (even a bit) we should do it :)

coalib/__init__.py
Makman2: If you use `str` you can should enclose it in double backticks then, as you refer to the actual type. So capitalize or wrap `str` in double backticks (for sentences that begin with "string")

coalib/__init__.py
Makman2: hmm... imo people install from pip or from repository. If they use e.g. a tar package or similar, the commit file should be included there as well (as it's a packed release normally). I can't imagine what @jayvdb wanted to point out...

arush0311: Sorry the commit file will always be there even if someone donot install from git repo. Exception handling is there just in case

arush0311: should probably change that behaviour though

.gitignore
```diff
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+COMMIT
```
arush0311: @Makman2 Is this okay? while doing a release one commit file will have to be manually placed

coalib/__init__.py
Makman2: ah okay thx for referring to that comment, keep it then :)

coalib/__init__.py
Makman2: I think easiest is to have no `COMMIT` file unless an installation (not develop-installation!). If there's no `COMMIT` file, try to get the git hash, if that does not work, return `''` :+1:

@arush0311 what do you think? :)

.gitignore
Makman2: I thought you generate it during install?

arush0311: Yes but that wont happen during installation of released version, as there is no git directory

coalib/__init__.py
arush0311: @makman2 during installation of released version there will be no commit file, unless its manually placed before release. In that case the overhead of release version will be more.

setup.py
```diff
@@ -23,11 +27,28 @@
...
+    """
+    Creates COMMIT file with `HEAD` sha if `COALIB_DIR` is a git repo.
+    """
+    current_commit = get_current_commit(COALIB_DIR)
+    if(current_commit != ''):
+        with open(os.path.join(COALIB_DIR, COMMIT_FILE), 'w') as cf:
+            cf.write(get_current_commit(COALIB_DIR) + '\n')
+
+
+class SdistCommand(setuptools.command.sdist.sdist):
```
arush0311: sure gitmate :stuck_out_tongue:

arush0311:
@Mixih please review if you have time..

sils:
ack https://github.com/coala/coala/pull/3595/commits/e0295d2c668c7b60a0c262ce6bbd5a894f851a9a

sils:
are we sure we wanna do IO on every import of our init py? Performance IS a problem with coala and IMO this isn't important enough to make it worse. On installation? Good. On every run? Not so sure.

arush0311:
what about we remove the commit variable and only use get_commit function ?



  <https://mailtrack.io/> Sent with Mailtrack
<https://mailtrack.io/install?source=signature&lang=en&referral=arushgyl@gmail.com&idSignature=22>

arush0311:
@sils @Mixih ^

sils:
sounds good

sils:
sounds good

nemaniarjun:
ack b2324f0fdd076d76b5362237b967994e01406053

nemaniarjun:
@jayvdb there are some changes requested by you, can you look into them once? :D

Mixih:
ping @jayvdb your opinion please.

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+    :return:
+        String containing the current commit sha if `git_repo_dir` is a
+        git repo otherwise commit sha in commit_file.
+    """
+    try:
+        with open(commit_file) as cf:
+            commit = cf.readline().strip()
+    # Not using FileNotFoundError, as python2 CI starts complaining
+    except (OSError, IOError):
+        commit = ''
```
jayvdb: `commit = None`.  Do not use the empty string unless it is actually an empty string that you need.

coalib/__init__.py
```diff
@@ -6,10 +6,69 @@
...
+
+    try:
+        commit = check_output(['git', 'rev-parse', 'HEAD'])
+        commit = commit.decode('utf-8').strip()
+
+        status = check_output(['git', 'status', '--porcelain'])
+        status = status.decode('utf-8').strip()
+        if status:
+            commit += '*'
+    except (OSError, CalledProcessError) as e:
```
jayvdb: why catch these?
When these exceptions occur, this method is *not* returning a commit representation, which means it has failed, so it should be throwing an exception.

setup.py
```diff
@@ -52,6 +75,14 @@ def run(self):
         sys.exit(errOne or errTwo)
 
 
+class DevelopCommand(setuptools.command.develop.develop):
+
+    def run(self):
+        with open(os.path.join(COALIB_DIR, COMMIT_FILE), 'w') as cf:
+            cf.write('develop\n')
```
jayvdb: what is the benefit of writing 'develop' into the file.  Isn't that the same as no commit file?
The COMMIT file is mandatory because it is listed in `package_data`?

Maybe after running `setuptools.command.develop.develop.run(..)`, remove the `COMMIT` file again...?

setup.py
```diff
@@ -23,11 +28,29 @@
 assert_supported_version()
 
 
+def create_commit_file():
+    """
+    Creates COMMIT file with `HEAD` sha if `COALIB_DIR` is a git repo.
+    """
+    current_commit = get_current_commit(COALIB_DIR)
+    if(current_commit != ''):
```
jayvdb:
```diff
-if(current_commit != '')
+if (current_commit != '')
```

but ... wtf...

If you can not get a commit identifier, it is probably because the user is installing from a tarball.  You must write a COMMIT file so that the package doesnt try to run `git` at startup every time the package is loaded.

coalib/__init__.py
arush0311: yeah its not needed now, previously it was needed when we were using `__commit__` variable

SanketDG:
unack b2324f0

Mixih:
I think we've drifted a bit from the original goal. Can we have a discussion on the issue please?

setup.py
```diff
@@ -72,14 +76,28 @@ def run(self):
...
 
+def _get_public_version_string(scm_version):
+    return VERSION
+
+
+def _get_local_version_string(scm_version):
+    if 'dev' in VERSION:
+        return get_local_node_and_date(scm_version)
+    else:
+        return ''
```
shreyans800755: Can be done in a single line like `return get_local_node_and_date(scm_version) if 'dev' in VERSION else ''`

.misc/check_setuptools.py
```diff
@@ -2,14 +2,16 @@
...
 
 import setuptools
 
 
 def get_setuptools_version():
     with open('requirements.txt') as f:
         for line in f:
-            if line.startswith('setuptools'):
+            # Ensure that line matched is setuptools
+            if re.match(r'^setuptools(>|<|~|=|$)', line):
```
shreyans800755: `re.compile` is better option since we need to match it multiple times and regex remains same ever time.

Makman2:
@arush0311 why closing? 😞

arush0311:
Hey! I am sorry.. I don't think this pr is relevant anymore and i thought it might be wasting time of reviewers and I keep getting spammed by gitmate

Makman2:
I think it's still good to have :) However if you don't want to continue, it's fine :)

Issue #1862, opened 2016-03-11
Auto dump log on errors


AbdealiJK:
This is something I think would be very useful.

The feature request is that if coala fails, we auto dump a `.coala-log` or `coala-log.txt` file in the current directory where it's being run from. Why ?

There have been many instances where I find that coala failed due to an unusual error. Specifically when I'm running coala and using the editor/auto apply patch in various places. I get the "Sorry, something went wrong, this is a bug" error but I am unable to reproduce it later :/

It's frustrating, and we should have a "dump all error messages if it bugged out" because the issues aren't always reproducible.

Other notes: ros, pip, and npm have this.

sils1297:
Neat, sounds interesting and not particularly hard to do either, we have the exception catching in one routine somewhere where this could be done, right?

AbdealiJK:
I think the python `logging` module may have this inbuilt and probably does it nicely

sils1297:
blocked by https://github.com/coala-analyzer/coala/issues/975

sils:
No longer blocked, we do use builtin logging now

Makman2:
What about logging coala always instead of only on errors (which is a bit complicated, isn't it?)

nemaniarjun:
No, Its not. @Makman2 This can be easily implemented. Assign me if still needed.

nemaniarjun:
A bit of design needed, we need to write all the logs in a log file, in the place where coala is run, is that it?

Makman2:
Yes would say so :+1: :)

nemaniarjun:
Nearly done with this.
Do we want the file to be colored for console viewing or not? ( I think not, just confirming :P )

Makman2:
No by default no colors.

nemaniarjun:
Will send a PR as soon as the current one is merged. https://github.com/coala/coala/pull/3494

jayvdb:
@nemaniarjun , what is the status of this. https://github.com/coala/coala/pull/3596 was closed, and a new PR was going to be opened ... ;-)

nemaniarjun:
@jayvdb I was a bit busy with exams, etc. You can expect it to be completed by the end of this week.
Current PR : https://github.com/coala/coala/pull/3740

jayvdb:
No worries; I have cross linked the PR with the issue so it doesnt get lost.

nemaniarjun:
Creating a new PR. It is very rough as I am very rusty and this issue is / was complicated.
[...]

PR #3596, opened 2017-01-19
WIP : Logging: Add File Logging

nemaniarjun:
Running `coala` in all modes except `json`
will output a log file in the folder it has been run
named `coala-log` with all log messages.

Closes https://github.com/coala/coala/issues/1862

### Checklist

- [ ] I have rebased properly. Please see [our tutorial on rebasing](http://coala.io/git#rebasing).
- [ ] I have gone through the [commit guidelines](http://coala.io/commit) and I've followed them.
- [ ] I have followed the [guidelines for the commit shortlog](http://coala.io/commit#shortlog).
- [ ] I have followed the [guidelines for the commit body](http://coala.io/commit#commit-body).
- [ ] I have included the issue URL with the 'Fixes' keyword if the commit fixes a *real bug* and the 'Closes' keyword if it adds a feature or enhancement. For details, check the [issue reference section in our commit guidelines](http://coala.io/commit#issue-reference).
- [ ] I have [run all the tests](http://api.coala.io/en/latest/Developers/Executing_Tests.html) and they all pass. If any CI (below) is not green, please fix them. If GitMate issues appear you will have to amend your commits, pushing new commits on top is not sufficient!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+**/coala-log
--- a/coalib/coala_delete_orig.py
+++ b/coalib/coala_delete_orig.py
@@ -29,7 +29,7 @@ def start_delete_orig(section: Section=None):
     not_deleted = 0
     for ofile in orig_files:
         logging.info('Deleting old backup file... '
-                         + os.path.relpath(ofile))
+                     + os.path.relpath(ofile))
         try:
             os.remove(ofile)
         except OSError as oserror:
@@ -39,9 +39,9 @@ def start_delete_orig(section: Section=None):
 
     if not_deleted:
         logging.warn(str(not_deleted) + ' .orig backup files could not be'
-                         ' deleted, possibly because you lack the permission'
-                         ' to do so. coala may not be able to create'
-                         ' backup files when patches are applied.')
+                     ' deleted, possibly because you lack the permission'
+                     ' to do so. coala may not be able to create'
+                     ' backup files when patches are applied.')
     return 0
 
 
--- a/coalib/coala_delete_orig.py
+++ b/coalib/coala_delete_orig.py
@@ -13,7 +13,10 @@ from coalib.parsing.Globbing import glob_escape
 
 def main(section: Section=None):
     configure_logging()
+    return start_delete_orig(section)
 
+
+def start_delete_orig(section: Section=None):
     start_path = get_config_directory(section)
 
     if start_path is None:
--- a/coalib/coala_delete_orig.py
+++ b/coalib/coala_delete_orig.py
@@ -1,4 +1,5 @@
 import os
+import logging
 
 from pyprint.ConsolePrinter import ConsolePrinter
 
@@ -10,12 +11,10 @@ from coalib.settings.Section import Section
 from coalib.parsing.Globbing import glob_escape
 
 
-def main(log_printer=None, section: Section=None):
+def main(section: Section=None):
     configure_logging()
 
     start_path = get_config_directory(section)
-    log_printer = (LogPrinter(ConsolePrinter()) if log_printer is None
-                   else log_printer)
 
     if start_path is None:
         return 255
@@ -26,17 +25,17 @@ def main(log_printer=None, section: Section=None):
 
     not_deleted = 0
     for ofile in orig_files:
-        log_printer.info('Deleting old backup file... '
+        logging.info('Deleting old backup file... '
                          + os.path.relpath(ofile))
         try:
             os.remove(ofile)
         except OSError as oserror:
             not_deleted += 1
-            log_printer.warn("Couldn't delete {}. {}".format(
+            logging.warn("Couldn't delete {}. {}".format(
                 os.path.relpath(ofile), oserror.strerror))
 
     if not_deleted:
-        log_printer.warn(str(not_deleted) + ' .orig backup files could not be'
+        logging.warn(str(not_deleted) + ' .orig backup files could not be'
                          ' deleted, possibly because you lack the permission'
                          ' to do so. coala may not be able to create'
                          ' backup files when patches are applied.')
--- a/coalib/misc/Caching.py
+++ b/coalib/misc/Caching.py
@@ -14,8 +14,15 @@ class FileCache:
 
     >>> from pyprint.NullPrinter import NullPrinter
     >>> from coalib.output.printers.LogPrinter import LogPrinter
+    >>> from coalib.output.Logging import configure_logging
     >>> import logging
     >>> import copy, time
+    >>> configure_logging()
+    20
+
+    configure_logging should always be run before creating a
+    LogPrinter instance.
+
     >>> log_printer = LogPrinter()
     >>> log_printer.log_level = logging.CRITICAL
 
--- a/coalib/output/Logging.py
+++ b/coalib/output/Logging.py
@@ -14,18 +14,18 @@ def configure_logging():
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
-            'colored': {
+            'console-handler': {
                 'class': 'logging.StreamHandler',
-                'formatter': 'colored',
+                'formatter': 'color',
                 'stream': sys.stderr
             }
         },
         'root': {
             'level': 'DEBUG',
-            'handlers': ['colored']
+            'handlers': ['stream-handler']
         },
         'formatters': {
-            'colored': {
+            'color': {
                 '()': 'colorlog.ColoredFormatter',
                 'format': '%(log_color)s[%(levelname)s]%(reset)s[%(asctime)s]'
                           ' %(message)s',
@@ -51,7 +51,7 @@ def configure_json_logging():
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
-            'json': {
+            'json-handler': {
                 'class': 'logging.StreamHandler',
                 'formatter': 'json',
                 'stream': stream
@@ -59,7 +59,7 @@ def configure_json_logging():
         },
         'root': {
             'level': 'DEBUG',
-            'handlers': ['json']
+            'handlers': ['json-handler']
         },
         'formatters': {
             'json': {
--- a/coalib/output/Logging.py
+++ b/coalib/output/Logging.py
@@ -4,25 +4,50 @@ import io
 import logging
 import logging.config
 
+from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 
-def configure_logging():
+
+def configure_logging(new_log_level=logging.INFO, incremental=False):
     """
     Configures the logging with hard coded dictionary.
+    When changing the log_level, the incremental param should be True.
+    json-handler are created so that they exist when
+    the Log Level is changed via `configure_logging`
+
+    :param new_log_level: The new log level to set for json and console
+                          handlers.
+    :param incremental: `False` if setting up new loggers, `True` if only
+                        changing the log level.
+
+    :return: The new_log_level is returned.
     """
     import sys
 
     logging.config.dictConfig({
         'version': 1,
+        'incremental': incremental,
         'handlers': {
             'console-handler': {
                 'class': 'logging.StreamHandler',
                 'formatter': 'color',
-                'stream': sys.stderr
+                'stream': sys.stderr,
+                'level': LOG_LEVEL.reverse.get(new_log_level)
+            },
+            'file-handler': {
+                'class': 'logging.FileHandler',
+                'formatter': 'no-color',
+                'filename': 'coala-log',
+                'mode': 'w',
+                'level': 'DEBUG'
+            },
+            'json-handler': {
+                'class': 'logging.NullHandler',
+                'level': LOG_LEVEL.reverse.get(new_log_level)
             }
         },
         'root': {
             'level': 'DEBUG',
-            'handlers': ['stream-handler']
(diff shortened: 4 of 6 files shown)


nemaniarjun:
CC @Makman2

.gitignore
```diff
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+coala-log
```
Makman2: miep

coalib/output/Logging.py
```diff
@@ -15,18 +15,24 @@ def configure_logging():
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
-            'colored': {
+            'console-logger': {
```
Makman2: I think renaming the handler should go into an own commit^^

coalib/output/Logging.py
```diff
@@ -15,18 +15,24 @@ def configure_logging():
...
                 'class': 'logging.StreamHandler',
-                'formatter': 'colored',
+                'formatter': 'color',
```
Makman2: why changing this setting suddenly?

coalib/output/Logging.py
```diff
@@ -15,18 +15,24 @@ def configure_logging():
...
                 'stream': sys.stderr
+            },
+            'file-logger': {
+                'class': 'logging.FileHandler',
+                'formatter': 'no-color',
+                'filename': 'coala-log',
+                'mode': 'w+'
```
Makman2: I would truncate, otherwise the log will grow quite big^^

Makman2: Also can you turn on debug log messages? That's the whole purpose of the log file actually ;)

coalib/output/Logging.py
```diff
@@ -37,6 +43,12 @@ def configure_logging():
                     'INFO': 'blue',
                     'DEBUG': 'green'
                 }
+            },
+            'no-color': {
+                '()': 'logging.Formatter',
+                'format': '[%(levelname)s][%(asctime)s]'
+                          ' %(message)s',
```
Makman2: doesn't fit into a single line? :)

coalib/output/Logging.py
nemaniarjun: Just for the sake of it. Color and no-color, instead of colored and non-colored. 
This setting doesn't have any effect outside this function. So it doesn't matter much.

Makman2: then -> own commit ;)

coalib/output/Logging.py
nemaniarjun: What do you mean by truncate? 
> `w+` Opens a file for both writing and reading. Overwrites the existing file if the file exists. If the file does not exist, creates a new file for reading and writing.

coalib/output/Logging.py
nemaniarjun: One commit for renaming both handler and formatter seems fair

coalib/output/Logging.py
Makman2: truncating a file means that you remove its contents on open.
seems `w+` does truncate ;) but do you need `+`?

nemaniarjun: I am having a bit of difficulty doing this. I can't understand how the arg `-L ` works.

Makman2: ok then let's do this later :+1:

nemaniarjun: No I dont think so :)

nemaniarjun: I am trying to understand this :')

nemaniarjun:
This is with respect to why the build failed.

Got a problem, in coalaDeleteOrigTest under `test_nonexistent_coafile` we are mocking `os.getcwd()` to be `None`. This create a problem when I try to create a log file. (since that would need an actual path to be made.)
Should I change this to point to a empty folder? (or maybe create and delete one for this test?)


another fix : configure_logging after the above test ends. i.e. on L22 instead of L14 in https://github.com/coala/coala/blob/master/coalib/coala_delete_orig.py#L14

coalib/output/Logging.py
nemaniarjun: Nearly done. Requires a certain amount of testing, if you have any corner cases that need to checked do comment :)

nemaniarjun:
@Makman2 according to your solution I tried the first way, but that wont work as it will require changing some lines in `ConfigurationGathering` something i am unwilling to do.

Also 3rd Fix will not work (It will just try to config logging again)

Thought of a 4th Fix: change the name of the current `main` function to something else and test that.
while configuring logging in a new main function.

Makman2:
@nemaniarjun you are talking about using a temp-directory for tests, right?

> but that wont work as it will require changing some lines in ConfigurationGathering something i am unwilling to do.

what exactly needs change?

> Also 3rd Fix will not work (It will just try to config logging again)

what 3rd fix?

> Thought of a 4th Fix: change the name of the current main function to something else and test that.
while configuring logging in a new main function.

okay don't exactly understand this, just try ;)

nemaniarjun:
I tried to import `args` from `coalib/coala.py` and use it to check if the mode is json in `Logging.py` but I get a ImportError

Makman2:
So if I understand correctly, json-logging-configuration overrides the configuration that contains the file-handler?

nemaniarjun:
Nope! When run in json mode, first I override with json-logging-config, but after that when log_printer is used it overwrites the json-logging with file-logging

Makman2:
Ah yeah sorry confused the order :P :P

Imo there are two possibilities:
- We merge all logging configuration functions into a single one, so we can handle them correctly together
- We modularize json-logging, normal stdout logging and file logging (first two are already modularized), and calling one of the functions does not override other configurations.

I prefer 2nd if possible (so we have then `enable_json_logging`, `enable_file_logging`, `enable_std_logging` or so, together with a `reset_logging` function). @nemaniarjun is it possible to patch the logging configuration once it's set?

nemaniarjun:
Found [this](https://docs.python.org/3/library/logging.config.html#incremental-configuration)

Makman2:
Looks like the right direction, though this worries me:

> Thus, when the incremental key of a configuration dict is present and is True, the system will completely ignore any formatters and filters entries, and process only the level settings in the handlers entries, and the level and propagate settings in the loggers and root entries.

Do we get problems with this behaviour?

nemaniarjun:
I think this is good. And I can figure this out. It might cause a problem but got an intuition that it will not be a big one.

Makman2:
@nemaniarjun cool :)

nemaniarjun:
@Makman2 This works :D 
Can you however checkout this branch and try it once on your computer ?
Thanks :)

nemaniarjun:
Don't know why the test failed, some error in gitmate?

.gitignore
```diff
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+**/coala-log
```
Makman2: you don't need `**` :)

coalib/coala_delete_orig.py
```diff
@@ -13,7 +13,10 @@
 
 def main(section: Section=None):
     configure_logging()
+    return start_delete_orig(section)
 
+
+def start_delete_orig(section: Section=None):
```
Makman2: good change, but separate commit :+1:

coalib/coala_delete_orig.py
```diff
@@ -26,7 +29,7 @@ def main(section: Section=None):
     not_deleted = 0
     for ofile in orig_files:
         logging.info('Deleting old backup file... '
-                         + os.path.relpath(ofile))
+                     + os.path.relpath(ofile))
```
Makman2: actually I'm not sure about this one...
1. definitely separate commit (also for the change below)
2. could you read up in PEP8 again how this shall be handled? The problem is we seem to allow both styles, and next time some guy changes it back^^ If PEP8 has no opinion on this case, I would rather tell you to revert this change, sorry :/

coalib/misc/Caching.py
```diff
@@ -14,8 +14,11 @@ class FileCache:
 
     >>> from pyprint.NullPrinter import NullPrinter
     >>> from coalib.output.printers.LogPrinter import LogPrinter
+    >>> from coalib.output.Logging import configure_logging
     >>> import logging
     >>> import copy, time
+    >>> configure_logging()
+    20
```
Makman2: what is `20`? :O

coalib/misc/Caching.py
```diff
@@ -14,8 +14,11 @@ class FileCache:
...
     >>> log_printer = LogPrinter()
     >>> log_printer.log_level = logging.CRITICAL
```
Makman2: I think you can omit those then

Makman2: Nvm

Makman2: Maybe separate commit?

Makman2: We definitely need to explain why we `configure_logging` and then use a `LogPrinter`, this is definitely confusing for unaware people^^

coalib/output/Logging.py
```diff
@@ -4,25 +4,40 @@
...
+
+def configure_logging(new_log_level=logging.INFO, incremental=False):
     """
     Configures the logging with hard coded dictionary.
     """
     import sys
 
     logging.config.dictConfig({
         'version': 1,
+        'incremental': incremental,
```
Makman2: Don't we want to have this set to `True`?

coalib/output/Logging.py
```diff
@@ -4,25 +4,40 @@
...
                 'formatter': 'color',
-                'stream': sys.stderr
+                'stream': sys.stderr,
+                'level': LOG_LEVEL.reverse.get(new_log_level)
+            },
+            'file-handler': {
+                'class': 'logging.FileHandler',
+                'formatter': 'no-color',
+                'filename': 'coala-log',
+                'mode': 'w+',
```
Makman2: shouldn't this be just `w`?

coalib/output/printers/LogPrinter.py
```diff
@@ -121,22 +122,26 @@ def __init__(self,
...
+        if not printer:
+            self.log_level = log_level
 
     @property
     def log_level(self):
         """
         Returns current log_level used in logger.
         """
-        return self.logger.getEffectiveLevel()
+        # return self.logger.getEffectiveLevel()
```
Makman2: delete commented out code ;)

coalib/output/printers/LogPrinter.py
```diff
@@ -121,22 +122,26 @@ def __init__(self,
...
+        return self.level
 
     @log_level.setter
     def log_level(self, log_level):
         """
         Sets log_level for logger.
         """
-        self.logger.setLevel(log_level)
+        # self.logger.setLevel(log_level)
+        # print(self.printer)
```
Makman2: commented out code

coalib/output/Logging.py
nemaniarjun: Not always, thats why taking a new argument, which by default is False.

coalib/coala_delete_orig.py
nemaniarjun: I didnt make this change, it was made by PystyleCodeBear

.gitignore
nemaniarjun: Sometimes there are extra logs created in tests folder, to ignore these too I used the `**` :D

coalib/misc/Caching.py
nemaniarjun: Should I put that comment here or in `LogPrinter.py`?

coalib/output/Logging.py
nemaniarjun: When `configure_logging` is called for the first time, the increment should be `False`, when using it to update the levels we set it to `True`, thus this change.

coalib/output/Logging.py
nemaniarjun: Yeah forgot to change that!

coalib/coala_delete_orig.py
```diff
@@ -26,17 +25,17 @@ def main(log_printer=None, section: Section=None):
 
     not_deleted = 0
     for ofile in orig_files:
-        log_printer.info('Deleting old backup file... '
+        logging.info('Deleting old backup file... '
                          + os.path.relpath(ofile))
```
nemaniarjun: @Makman2 This is still weird.

CC @sils because Gitmate issue.

coalib/coala_delete_orig.py
Makman2: okay, separate commit then :)

coalib/misc/Caching.py
Makman2: here before `log_printer = LogPrinter()`

coalib/output/Logging.py
Makman2: hm this sounds complicated. When do we update levels?

coalib/misc/Caching.py
nemaniarjun: Okay, also added a comment to `LogPrinter`

coalib/output/Logging.py
nemaniarjun: We use levels a lot in testing and doctests.
I dont know exactly when we do this.
But basically when we configure_logging for the first time we set increment to false.
When we only want to change the level we set it to True.

nemaniarjun:
Creating a new PR, too many comments and gitmate is still irritating about that change :(

Issue #3587, opened 2017-01-17
Improve coafile/KeywordBear shortlog regex


Makman2:
I think the regex should be `[^:]*|[^:]*[^: ]: [A-Z0-9*].*`
The current one would also allow messages like
s:: description
and not allow
t: description (it has to match two letters currently)

Also the redundant braces around the whole expression can be removed^^

Makman2:
oh damn I forgot to put the regex into double quotes to make it better readable...

But actually your commit doesn't fix the issue @affanv14 , you now removed the `[^:]*|` part which allows to match commit messages without tag. \+ The braces are still there^^

jayvdb:
We should test the modified regex against last 100 entries in git log

affanv14:
Damn.I thought that the tag was mandatory.

@Makman2 putting the regex in double quotes would include the double quotes in the regex while parsing the coafile.
Wont the braces around the regex make it more readable instead?

Makman2:
the double quotes were just for getting properly rendered here in the issue, so you can read it better ;)

Actually I'm always confused when I see the braces, because they are redundant^^

> We should test the modified regex against last 100 entries in git log

Indeed. We should create a `coafileTest` inside our test directory for such things regarding our coafile^^.

Updated the top description of the issue to add a test.

sils:
okayyyy seems @affanv14 was a bit overenthusiastic about the exactly one colon thing: commit messages without a colon should be allowed!

affanv14:
@sils yeah.This one just exploded!

jayvdb:
> Also the redundant braces around the whole expression can be removed^^

IMO (and this is personal taste only), `|` without round brackets is effectively a syntax error because it *must* be `(..  |..  )` if the regex was to be embedded into another string. i.e. it is not a valid regex fragment.
 
They are redundant, and most implementation implicitly add the round brackets, but that is a feature which should only be used for really simple example regex like `a|b`.
The explicit round brackets helps people understand the regex.  They know there will be alternation somewhere in it, as there isnt any modifiers immediately after the `(`.

I fully appreciate that the round brackets are redundant, and wont -1 the patch as a result. So I rant here. ;-)

jayvdb:
> t: description (it has to match two letters currently)

@Makman2 , why do we want to allow a one letter prefix before the `:`?  Do you have a valid use for that?  If not, I suggest we wait until we encounter a need for it.
`.ci:` is the shortest valid use that I can think of.

[...]

PR #3599, opened 2017-01-20
coafileTest: check last 100 commits

affanv14:
Fix the shortlog regex and add test to
check the regex against the last 100 commits

Fixes https://github.com/coala/coala/issues/3587

### Checklist

### DIFF
--- a/tests/coafileTest.py
+++ b/tests/coafileTest.py
@@ -1,0 +1,20 @@
+import unittest
+import re
+from coalib.settings.ConfigurationGathering import load_config_file
+
+
+class coafileTest(unittest.TestCase):
+
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.commits_filepath = 'tests/commits.txt'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        regex_string = self.sections['commit']['shortlog_regex'].value
+        shortlog_regex = re.compile(regex_string)
+        git_log = open(self.commits_filepath, 'r').read()
+        for message in git_log.splitlines():
+            self.assertIsNotNone(shortlog_regex.fullmatch(message),
+                                 msg='{!r} does not match the '
+                                 'given regex {}'.format(message, regex_string))
--- a/tests/commits.txt
+++ b/tests/commits.txt
@@ -1,0 +1,100 @@
+coafile: Add test for shortlog regex
+Writing_Good_Commits.rst: Edit Commit Messages
+Writing_Good_Commits.rst: Reword line
+Bear.py: Include filename on which the bear failed
+DefaultArgParser.py: Change order of logging
+Instructions to add CI to fork
+AppveyorCI: Require 100% test coverage
+AppveyorCI: Install clang
+Use codecov-python
+logging.py: Remove unnecessary import
+circle.yml: Add csslint
+Bear.py: Modify get_non_optional_settings
+ConfParser: Deprecate settings without section
+ConfigurationGathering: Disable default section
+Setting: Raise exception if `to_append` is true
+requirements: Update coala-utils version
+CliParsing: Move to 'cli' section
+ConfWriter: Add setting appending
+LineParser: Add append(+=) parsing
+Section: Add settings appending
+ConfigurationGathering: Add section inheritance
+Redundancy.py: Change 'deco' to 'docs'
+.travis.yml: Add csslint
+Collectors.icollect: Cache results of globs
+TestUtilities.py: Return stdout, stderr separately
+Revert "LICENSE: Remove boilerplate stuff"
+PULL_REQUEST_TEMPLATE: Name newcomer friendlier
+Writing_Good_Commits.rst: Rephrase Line
+RELEASE_NOTES: Fix malformed heading
+RELEASE_NOTES: Add release ascii
+RELEASE_NOTES: Add 0.10 release notes
+Add ShellCommandResult for accessing return codes
+ConsoleInteraction: Remove format_str (deprecated)
+Shell: Use shutil.which on Windows
+coafile: Blacklist urllib
+coafile: Remove trailing whitespace
+.travis.yml: Replace 3.6-dev with 3.6
+.travis.yml: De-activate OSX job
+Fix bare exception clauses
+Revert "coafile: Fix shortlog regex"
+DefaultArgParser.py: Add Default log level to help
+Shell.py: Remove call_without_output
+requirements.txt: Pin requests properly
+Pull template: Add branch button warning
+ConsoleInteraction.py: Show file path of bear
+docs: Add link to Code of Conduct
+Useful_Links.rst: Add Useful_Links.rst
+Bear.py: Rewrite download_cached_file in requests
+.travis.yml: Move unsupported Python versions
+requirements.txt: Upgrade dependency_management
+Result.py: Format message if arguments provided
+Writing_Linter_Bears.rst: Grammatical Error fix
+Result.py: Add message arguments
+Diff.py: Deprecate change_line method
+Language.py: Ignore whitespace in lookup
+JSON Mode: Add ``--log-json`` flag
+coala_main: Remove Redundant Logging
+Logging: Move logging to own module
+Newcomers_Guide.rst: Add link for review
+Newcomers_Guide.rst: Improve in note
+coafile: Fix shortlog regex
+coafile: Change python3 to python 3
+coafile: Fix regex in commit section
+Globbing.py: Add test for the sequence ([xy])
+Bear.py: Correct the maintainers_emails method
+AppveyorCI: Remove unneeded cache directories
+Newcomers_Guide.rst: Emphasize on low issues
+CSharp: Add fields for CSharp language support
+AppveyorCI: Remove astroid as not used
+ConsoleInteraction.py: Use dot in place of bullet
+README.rst: Resize asciinema demo image
+.travis.yml: Add exclude branch 'sils/*'
+ConfigurationGathering: Rename default_section
+Writing_Native_Bears.rst: Change in filename
+OpenEditorAction: Add emacs
+ConsoleInteraction.py: Move join_names upstream
+Requirements.txt dependency_management 0.3.0
+Writing_Documentation.rst: Add rst tutorial
+Linter: Keep .__module__ name of decorated classes
+docs: Update review documentation
+README: Add mention about coala-quickstart
+Writing_Good_Commits.rst: Update Issue reference
+Newcomers_Guide.rst: Specify the assign command
+External_Bears.rst: Change to "coala --json"
+coala_settings: Remove "Documentation" from title
+CliParsingTest.py: Use unittest.asserRaisesRegex
+Bear.py: Remove a paragraph from docstring
+.coafile: Add css section
+custom.css: Fix linting errors
+ConfWriter.py: Check if param for -c is dir
+Newcomers_Guide.rst: Fix a typo
+IgnoreResultAction: Add Multiline comment support
+DefaultArgParser: Modify help string
+requirements.txt: Update dependecy_management
+OpenEditorAction: Emit warning for unknown editors
+Development_Setup.rst: Update commands
+Newcomers_Guide: Add command for test-requirements
+Newcomers_Guide: Add references to Dev Setup Notes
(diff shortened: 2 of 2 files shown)


pratyushprakash:
You have PEP8 issues. You can fix them by running `coala` before submitting a PR.

tests/coafileTest.py
```diff
@@ -0,0 +1,27 @@
...
+from coalib.output.printers.LogPrinter import LogPrinter
+from pyprint.ConsolePrinter import ConsolePrinter
+from coalib.settings.ConfigurationGathering import load_config_file
+from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
+
+
+class coafileTest(unittest.TestCase):
+
+    def setUp(self):
+        self.logprinter = LogPrinter(ConsolePrinter(), LOG_LEVEL.ERROR)
```
Mixih: I believe this function has been deprecated and you should use `logging` instead

affanv14: The function load_config_file requires logprinter.

Mixih: damn, the deprecation wasn't finished yet... sigh. ok this is fine until we finally deprecate it

affanv14: @Mixih So it does not work. logging.getLogger does not have warn and err methods,it has warning and error methods instead.
I think a better solution is to check the file exists beforehand,making the need for a logger unnecessary in load_config_file.

Mixih: Ok, this function still needs to be fully removed. I think this is ok until we get that

Udayan12167:
So the test file for a coafile change is not needed.

Udayan12167:
unack 2ca8506

tests/coafileTest.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+from coalib.misc.Shell import run_shell_command
+from coalib.settings.ConfigurationGathering import load_config_file
+
+
+class coafileTest(unittest.TestCase):
+
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.assertTrue(os.path.isfile(self.filepath),
```
Mixih: I don't see the point of this, plus you could have just used os.exists. Passing a logPrinter is fine for now until we get it fully removed.

Udayan12167:
ack 3fe8824

Mixih:
Also in this case, I think you should squash because the test file change was needed by the commit regex change so it is one change unit.

tests/coafileTest.py
affanv14: @Mixih the logprinter in load_config_file was only used for filenotfound exception.So i dont think it is necessary to pass a logprinter at all.It futureproofs it too.

yeah.. I think i overdid it in checking if the file exist though :)

Udayan12167:
unack 8b5b8e1

tests/coafileTest.py
```diff
@@ -0,0 +1,23 @@
...
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
+        git_command = 'git log -100 --pretty=%s'
+        git_log, stderr = run_shell_command(git_command)
+        self.assertEqual(stderr, '', msg=stderr)
+        commit_messages = git_log.split('\n')
```
Mixih: splitlines?

tests/coafileTest.py
```diff
@@ -0,0 +1,23 @@
...
+        commit_messages = commit_messages[:100]
```
Mixih: Why this slice?

affanv14: damn.i had some assumptions which i just realised were wrong.okay

tests/coafileTest.py
```diff
@@ -0,0 +1,22 @@
...
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
+        git_command = 'git log -100 --pretty=%s'
+        git_log, stderr = run_shell_command(git_command)
+        self.assertEqual(stderr, '', msg=stderr)
+        commit_messages = git_log.splitlines()
+        for message in commit_messages:
+            self.assertIsNotNone(re.fullmatch(shortlog_regex, message),
```
Mixih: why assertIsNotNone instead of assertNotEqual?

affanv14: re.fullmatch returns None if there is no match.else it returns a match object.

tests/coafileTest.py
```diff
@@ -0,0 +1,22 @@
...
+        self.filepath = '.coafile'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
+        git_command = 'git log -100 --pretty=%s'
+        git_log, stderr = run_shell_command(git_command)
+        self.assertEqual(stderr, '', msg=stderr)
+        commit_messages = git_log.splitlines()
+        for message in commit_messages:
```
Mixih: Why don't you use git_log.splitlines() directly here? Imo, it's shorter and still clear enough

jayvdb:
(sorry I forgot to publish the review before you uploaded a new version.)

affanv14:
@jayvdb Do I need to change something?

jayvdb:
I think my comments on https://github.com/coala/coala/issues/3587 and at https://github.com/coala/coala/pull/3599#discussion_r98600646 are still applicable.
The test is excellent, but I think we need more thought put into the changes to the regex.

affanv14:
@jayvdb @Mixih how about this regex?

.coafile
```diff
@@ -62,7 +62,7 @@ languagetool_disable_rules =
 [commit]
 bears = GitCommitBear
 shortlog_trailing_period = False
-shortlog_regex = ([^:]*|[^:]+[^ ]: [A-Z0-9*].*)
+shortlog_regex = [^:]{3,}|[^:]+[^: ]: [A-Z0-9*].{2,}
```
jayvdb: I suggest removing this part of your PR.
The regex is still not very good.  Background reading:
https://www.google.com.au/search?q=good+commit+messages

We can merge your test now.  It is good.

affanv14: Okay.give me a minute

affanv14:
@jayvdb done

sils:
this only adds the test right? The commit message tells me it'd fix the issue

affanv14:
okay done

tests/coafileTest.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
+        git_command = 'git log -100 --pretty=%s'
+        git_log, stderr = run_shell_command(git_command)
+        self.assertEqual(stderr, '', msg=stderr)
```
jayvdb: assertFalse.
`msg=stderr` is unhelpful. the value of stderr will be shown anyway.

affanv14: Okay.on it.

tests/coafileTest.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
+        git_command = 'git log -100 --pretty=%s'
+        git_log, stderr = run_shell_command(git_command)
+        self.assertEqual(stderr, '')
```
jayvdb: why didnt you use self.assertFalse as requested at https://github.com/coala/coala/pull/3599#pullrequestreview-20423278

affanv14:
wait just a minute

Makman2:
Actually I dislike that the test is flaky because the last 100 commits can change. With bad luck the last 100 messages are very simple, and a change in the regex gets not properly tested. Either check all commits or extract a fixed amount of them into a file/similar imo so we have always the same test conditions^^

Makman2:
Also we have to test against bad commit messages as well^^

affanv14:
@Makman2 How about taking the current last 350 messages and storing them in a file which is checked?

Makman2:
@affanv14 sounds good :)

jayvdb:
@Makman2, this isnt unit testing.  This is integration testing.
We still should have specific well designed tests for all corner cases we can think about, *especially* any bad commit messages.
This test is only to catch problems with the regex that we haven't thought about, using the existing dataset as a sanity check, because we've twice had a regex change which prevented sane commit messages, which held up PRs from being merged.

The commit messages regex should improve over time, and reject old commit messages which were not up to the new standard.  I don't mind if we store a static list of messages, but we will end up having to remove rows  occasionally.

Makman2:
but that's better and more deterministic than not doing so. If the regex changes, our test set has to change too ;) That would also apply when we get the most recent 100 commits, if that would fail we would have to start to fetch 1 commit and increase the number when we have enough commits :)

jayvdb:
Sure.  are you ok with "test against bad commit messages" being out of scope for this change?
If so, a stored list of 100 previous messages is the way forward (and will be quicker than fetching the last 100 out of git anyway).

affanv14:
So take the current last 100 commits only and store them in a file?

jayvdb:
Yup.

tests/coafileTest.py
```diff
@@ -0,0 +1,19 @@
...
+
+class coafileTest(unittest.TestCase):
+
+    def setUp(self):
+        self.filepath = '.coafile'
+        self.commits_filepath = 'tests/commits.txt'
+        self.sections = load_config_file(self.filepath, None, True)
+
+    def test_commit_regex(self):
+        shortlog_regex = self.sections['commit']['shortlog_regex'].value
```
Makman2: you could speed up the test by compiling the regex:
```python
shortlog_regex = re.compile(....value)
...
    self.assertIsNotNone(shortlog_regex.fullmatch(message))
```

tests/coafileTest.py
```diff
@@ -0,0 +1,19 @@
...
+        git_log = open(self.commits_filepath, 'r').read()
+        for message in git_log.splitlines():
+            self.assertIsNotNone(re.fullmatch(shortlog_regex, message),
+                                 msg='\"{0}\" does not match the '
```
Makman2: you don't need to escape `"` as you use single quotes.
\+
you don't need the `0` inside `{0}` :)

tests/coafileTest.py
```diff
@@ -0,0 +1,19 @@
...
+                                 'given regex'.format(message))
```
Makman2: maybe also print the regex?

tests/coafileTest.py
Makman2: you could maybe print the `repr` version of the string, then you don't need the double quotes `"` at all and python escapes stuff probably if necessary.
You can enable this behaviour by using `{!r}`.
Up to you :)

nemaniarjun:
@affanv14 The tests looks great! :)
But shouldn't there also be a commit which changes the shortlog?  Or is this PR just for the test?

Makman2:
> But shouldn't there also be a commit which changes the shortlog?

What do you mean @nemaniarjun ?

Makman2:
ah we actually need a `bad_commits.txt` file with some hand-written wrong shortlog messages. @affanv14 could you quickly craft one? :D
(btw: rename the current `commits.txt` to `good_commits.txt` :+1:)

nemaniarjun:
@Makman2 My bad. I meant shortlog **regex**.

Makman2:
@nemaniarjun we test whether our coafile regex works properly on shortlogs :)

nemaniarjun:
Then we should rename the original issue. 
It misled me :stuck_out_tongue_winking_eye:  :laughing:

Makman2:
It's just related to that issue, it does not fix it ;)

nemaniarjun:
Aaahhh. (Makes mental to read the commit message and not the PR text)

affanv14:
@Makman2 For bad commits,just the most common types of bad commits messages in the commit history are enough right?.Or should i add in random ones too?

Makman2:
@affanv14 what you think is enough :)

Issue #3558, opened 2017-01-14
provide a default value when a settings-question is left empty


Makman2:
From @abit2:
>http://pastebin.com/jkPW92mA
>
>it should be something like this:
>Please provide a value for `use_spaces` (True or False), when you don't enter something, the default is >`True`:
>
>To reproduce the problem just skip the use_space without giving any boolean value.

Makman2:
From @jayvdb:
> I suspect that this will need to be solved in [`coala`](https://github.com/coala/coala), rather than in the bear.

Makman2:
From @CosminIcatoiu:
> I would like to try this. From what i checked the changes have to be in ConsoleInteraction.py where some conditions have to be added. The only problem is that the message seems to be rather general so another one should be added if we want it displayed as @abit2 said. Should it be modified somewhere else, or changed in another way?

CosminIcatoiu:
is this the same issue? or what should i exactly do ?

Makman2:
yes it's the same issue :+1:

CosminIcatoiu:
did you check the solution i proposed? does it seem alright? how should i do with the string message. define another one?

Makman2:
don't understand it quite:

> The only problem is that the message seems to be rather general so another one should be added if we want it displayed as abit2 said.

CosminIcatoiu:
There is a variable which has that answer pattern Please enter a value for the setting {} etc. so, if we want to change that for this specific example i think another variable is needed with that answer pattern.

Makman2:
Still not getting it completely, but submit a PR and we will see how it looks :)

CosminIcatoiu:
Yep. I will do that :+1:

CosminIcatoiu:
Sry for not submitting a pull request yet. but i didn`t have network last 2 days and didn`t really have time either. Complicated period at university. In maximum 2 days i will submit the pull if it`s ok :-D

Makman2:
no hurry :)

Makman2:
@CosminIcatoiu still on it? :)

CosminIcatoiu:
@Makman2 yep. Sry, really had a busy period. Try to solve it now :)

CosminIcatoiu:
I tried to change something . Have a look when you have time and tell me what you believe :)

PR #3616, opened 2017-01-22
ConsoleInteraction.py: Add default value for use_spaces

CosminIcatoiu:
There was a problem when an user used SpaceConsistencyBear
because if he didn`t enter anything as a value for
use_spaces no default value would`ve been assigned. Adds
a default value and updates the message.

Fixes https://github.com/coala/coala/issues/3558

### Checklist

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -65,7 +65,7 @@ def highlight_text(no_color, text, lexer=TextLexer(), style=None):
 
 
 STR_GET_VAL_FOR_SETTING = ('Please enter a value for the setting \"{}\" ({}) '
-                           'needed by {} for section \"{}\": ')
+                           'needed by {} for section \"{}\":')
 STR_LINE_DOESNT_EXIST = ('The line belonging to the following result '
                          'cannot be printed because it refers to a line '
                          "that doesn't seem to exist in the given file.")
@@ -529,7 +529,10 @@ def acquire_settings(log_printer, settings_names_dict, section):
     for setting_name, arr in sorted(settings_names_dict.items(),
                                     key=lambda x: (join_names(x[1][1:]), x[0])):
         value = require_setting(setting_name, arr, section)
-        result.update({setting_name: value} if value is not None else {})
+        while value is None:
+            print('You have not entered any value.')
+            value = require_setting(setting_name, arr, section)
+        result.update({setting_name: value})
 
     return result
 


coalib/output/ConsoleInteraction.py
```diff
@@ -484,7 +484,7 @@ def require_setting(setting_name, arr, section):
                          following.
     :param section:      The section the action corresponds to.
     """
-    needed = join_names(arr[1:])
+    needed = join_names(arr[1:]) + 'True is the default value.'
```
pratyushprakash: `needed` is a list of all the bears which need said setting. I don't think the default message should go here. You probable want to add it at the end of the question. `Please enter a value for the setting "use_spaces" (True if spaces are to be used instead of tabs.) needed by SpaceConsistencyBear for section "Default" (True is the default):`

coalib/output/ConsoleInteraction.py
```diff
@@ -529,7 +529,8 @@ def acquire_settings(log_printer, settings_names_dict, section):
     for setting_name, arr in sorted(settings_names_dict.items(),
                                     key=lambda x: (join_names(x[1][1:]), x[0])):
         value = require_setting(setting_name, arr, section)
-        result.update({setting_name: value} if value is not None else {})
+        result.update({setting_name: value} if value is not None else
+                      True if setting_name == 'use_spaces' else {})
```
pratyushprakash: I think the issue required the default for all settings not only `use_spaces`

coalib/output/ConsoleInteraction.py
CosminIcatoiu: Ok. Will change. I think it looks better this way :)

coalib/output/ConsoleInteraction.py
CosminIcatoiu: From what i understood from @Makman2 in the end this is the issue that was first opened and it referred only the SpaceConsistencyBear. From what i understood from the previous chat it was not a real need for the other bears . Am i wrong ?

pratyushprakash: The same problem occurs when another setting is left blank. The issue is applicable to any bear.

CosminIcatoiu: OK. I guess that not all the settings are true/false settings right ? If they are not what default setting should they be assigned. If they are true/false it makes everything even easier.

CosminIcatoiu:
I have updated it. Although, as i said i am not really sure if all the settings are True/False. If they are i think this solves it. If they are not all i need to do is to add a method like the join_names which for a specific setting returns the default value, but i don`t know exactly where to add it. So if it is not ok this way can anybody tell me where to add that? in the file it says it is from coala_utils.string_processing.Core

CosminIcatoiu:
I also think that the error from cicircle is not from the program.

coalib/output/ConsoleInteraction.py
```diff
@@ -65,7 +65,8 @@ def highlight_text(no_color, text, lexer=TextLexer(), style=None):
 
 
 STR_GET_VAL_FOR_SETTING = ('Please enter a value for the setting \"{}\" ({}) '
-                           'needed by {} for section \"{}\": ')
+                           'needed by {} for section \"{}\" (True is the '
+                           'default value): ')
```
Makman2: This is bad, this happens now for every setting, not only boolean ones.

Makman2:
unack a9d4b10

Makman2:
maybe we should just reask the question. What we try to do is exactly the purpose of a normal default value. If there's no default value, the user has to enter something, as it's his preference.

@CosminIcatoiu what do you think?

CosminIcatoiu:
As i said previously, i guessed that not all the settings are boolean. @Makman2 i think that there should be something( like a map) that for every setting returns the default value of that value as a string. This would solve both problems( in the message sent and in the value set as default). But i don`t know where to write this and also i don`t know all the values for all the settings asked for all the bears. Can you give me an idea? Or would there be another solution?

Makman2:
The problem is that we don't know at that point which type a setting has (settings don't have actually a type associated with them inside a `Section`, the bear is responsible for interpreting them into a type). And it would get even more problematic as we allow arbitrary types (or more precise, we allow custom "conversion functions" that convert a setting into a type) inside bears.

> Or would there be another solution?
>> maybe we should just reask the question.

What do you think of that solution? Then the user at least gets no empty setting, like
```
use_spaces =
```

CosminIcatoiu:
Yes . I think something like: You have not entered any value for use_spaces(Select true if you want to use spaces instead of tabs). I think this is the easiest and most feasible solution. Will do it today or tommorrow. I am in the middle of my exam period this is why i am not so prompt . Sry @Makman2

CosminIcatoiu:
Obviously addapted for every bear :)

Makman2:
> You have not entered any value for use_spaces(Select true if you want to use spaces instead of tabs)

I would rather modify the default description, instead of prompting a "you have typed something invalid"-message, as this involves introducing new fields for the bear and makes bear-writing more complex.

CosminIcatoiu:
@Makman2 can you please explain me the problem of coverage travis and circleci are displaying? Don't really see what i did wrong. Also, is this what you wanted me to do? :)

coalib/output/ConsoleInteraction.py
```diff
@@ -65,7 +65,7 @@ def highlight_text(no_color, text, lexer=TextLexer(), style=None):
 
 
 STR_GET_VAL_FOR_SETTING = ('Please enter a value for the setting \"{}\" ({}) '
-                           'needed by {} for section \"{}\": ')
+                           'needed by {} for section \"{}\":')
```
Makman2: unrelated + unwanted change

coalib/output/ConsoleInteraction.py
```diff
@@ -529,7 +529,10 @@ def acquire_settings(log_printer, settings_names_dict, section):
     for setting_name, arr in sorted(settings_names_dict.items(),
                                     key=lambda x: (join_names(x[1][1:]), x[0])):
         value = require_setting(setting_name, arr, section)
-        result.update({setting_name: value} if value is not None else {})
+        while value is None:
```
Makman2: can `value` be actually `None`? `input()` always returns some string, imo you have to check for empty string `''`.

Makman2: and better check for `value.strip()`, entering spaces and \n's only should also be treated as an empty string

coalib/output/ConsoleInteraction.py
```diff
@@ -529,7 +529,10 @@ def acquire_settings(log_printer, settings_names_dict, section):
...
+            print('You have not entered any value.')
```
Makman2: just reask, I wouldn't print this^^ Most linux tools don't print such a message either^^

coalib/output/ConsoleInteraction.py
```diff
@@ -529,7 +529,10 @@ def acquire_settings(log_printer, settings_names_dict, section):
...
+            value = require_setting(setting_name, arr, section)
+        result.update({setting_name: value})
```
Makman2: though the code before did the same, it's faster + easier to do `result[setting_name] = value`

Makman2:
@CosminIcatoiu as you can see your code you introduced is not covered. You need to write a test for your stuff^^

Makman2:
Closing due to inactivity.

PR #3641, opened 2017-01-27
Collectors.icollect: Cache results of globs

underyx:
By adding a very simple caching mechanism to stop `icollect` from doing the same exact work for multiple sections, coala's runtime on a real world project went from 55.73s to 14.06s.

### Checklist

- [ ] I have included the issue URL with the 'Fixes' keyword if the commit fixes a *real bug* and the 'Closes' keyword if it adds a feature or enhancement. For details, check the [issue reference section in our commit guidelines](http://coala.io/commit#issue-reference).

### Reviewers

- @sils 
- @fneu

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -37,12 +37,13 @@ def _import_bears(file_path, kinds):
 
 
 @yield_once
-def icollect(file_paths, ignored_globs=None):
+def icollect(file_paths, ignored_globs=None, match_cache={}):
     """
     Evaluate globs in file paths and return all matching files.
 
     :param file_paths:    File path or list of such that can include globs
     :param ignored_globs: List of globs to ignore when matching files
+    :param match_cache:   Dictionary to use for caching results
     :return:              Iterator that yields tuple of path of a matching
                           file, the glob where it was found
     """
@@ -50,7 +51,10 @@ def icollect(file_paths, ignored_globs=None):
         file_paths = [file_paths]
 
     for file_path in file_paths:
-        for match in iglob(file_path):
+        if file_path not in match_cache:
+            match_cache[file_path] = list(iglob(file_path))
+
+        for match in match_cache[file_path]:
             if not ignored_globs or not fnmatch(match, ignored_globs):
                 yield match, file_path
 


underyx:
Marked WIP as there are no tests yet.

fneu:
sils and I consider the design ok, for the reasons you gave on Gitter. Also it seems like everything is already covered by our existing tests, I think this can go through as-is.

fneu:
ack https://github.com/coala/coala/pull/3641/commits/33393c50c3d827f4daeae5ff60b7faadcdc25f0c

fneu:
unack commit message

fneu:
@underyx if you remove the WIP in the commit message and maybe mention the issue (if applicable) we are good to go.

sils:
unack 33393c5

underyx:
Wow, I really didn't expect it to be this easy :D

Alright, ready for merging then I guess!

fneu:
almost.

> sets of iles.

you get to put an extra f into the commit message :)

underyx:
As always, the typos :D

@fneu done

fneu:
ack https://github.com/coala/coala/pull/3641/commits/b7a5c9e0eeddc83e1bb6bae8ab442e037dd1e8cf

jayvdb:
@rultor merge

sils:
reack https://github.com/coala/coala/pull/3641/commits/9c1a7c847df15920508f993b8bd102ffdcc7f182

sils:
just discussing with @userzimmermann would be way simpler to just use `lru_cache` on `glob` and use that

sils:
and deprecate iglob, logging the eprecation and defining iglob to just return `iter(glob(...)` so we don't have unneeded code and make use of the cache always

fneu:
maybe stupid questions but why wouldn't `lru_cache` work on `iglob` and return the respective iterator?
I'm not sure getting rid of iglob is a good idea, at least not without checking that it won't bite us.

sils:
@fneu you know how iterators are mutable objects? If you cache the iterator you're not going to be happy with the result the second time.

fneu:
yeah that's exactly why I'm asking, I know you cannot iter them twice. I had hoped `lru_cache` would save copies of mutables.

in any way we can't really let `iglob` return `iter(glob)`. It's literally implemented the other way around right now (`return list(iglob(pattern))`). We cannot easily, and shouldn't, change how glob works internally, and moving everything into `glob`, making a list there, and letting `iglob` iter that is insane.

So I'd guess we leave `iglob` be and add caching to `glob` only then? I would kinda make sense to have both options available, iterator vs caching. Probably add to the docstrings of both when you should use the other one.

sils:
@fneu point here is: iglob only exists for backwards compatability

fneu:
@sils have you looked at the code? Or is there some major work on it going on that I missed? The whole thing is built on top of iterators yielding from each other. `iglob` iters `_iter_alternatives` and yields from '_iglob', which in turn yields from several relative glob functions.

I see zero benefit in removing the ability to use results as they are coming in, especially since there is a noticable performance hit especially on recursive globs. Even if we run every glob multiple times because of sections and need a cached list and therefore `glob` in all places and never `iglob`: what do we gain?

sils:
I cherry picked this to https://github.com/coala/coala/pull/3706 so we can merge it in one go without so much rebase. You don't need to do anything and it's taken care of that this gets merged, thanks for your contribution!

Issue #2050, opened 2016-04-25
Linter: Support global bears


sils:

Makman2:
I'm planning a new parameter (maybe `kind` or `bear_base` or so), that you can set to `GlobalBear` (or explicitly to `LocalBear`, but this would be the default), and `@linter` rearranges the tool chain so that it receives the right params inside run and passes them to a slightly different version of `create_arguments` and `generate_config` (and maybe also `process_output`).

NiklasMM:
@Makman2 are you still working on this? Otherwise I'd like to give this a go over the weekend.

Makman2:
@NiklasMM go ahead :) If you have questions regarding the Linter, let me know, it's quite some code^^ :+1:

Makman2:
@NiklasMM how is it going? :)

NiklasMM:
I started looking at the code and it's a little more complex than I expected. I would still like to work on it, but have to find some time first ;)

NiklasMM:
Got it! :+1: Will move it up in my imaginary prioritized todo list ;)

NiklasMM:
@Makman2 sorry for the delay, but I've opened the PR now ;)

Issue #2067, opened 2016-04-25
create_arguments: Pass required parameters only.


tushar-rishav:
`config_file` isn't being used in `create_arguments`. We should pass only required arguments.
See [354](https://github.com/coala-analyzer/coala-bears/pull/354/files#diff-b120d691fa8cf7e31b53e225b6200cd6R26)

Makman2:
What makes this a bit harder is that we can't use `get_metadata` to filter the parameters like it's done when a user adds additional parameters to the linter signatures. We need another "inofficial" filtering for the mandatory arguments.

AbdealiJK:
I don't see this as a big deal actually. So what if it's not being used ?

AbdealiJK:
Sorry. I seem to have not been thinking properly when I said that. It'd be great if this is done :) (although if it's tough, it's not that important)

NiklasMM:
But how can we get rid of it, if all the bears have implemented it with that interface?

Makman2:
via deprecation :+1:

PR #3652, opened 2017-01-29
Linter: Add support for global Linters

NiklasMM:
The `linter` decorator now supports a new
parameter `global_bear`, which enables it
to wrap tools which operate on repository
level instead of file level.

Closes: https://github.com/coala/coala/issues/2050

Very sorry this took so long, but sometimes life gets in the way.
Anyway, this is my first contribution to such a core part of coala, so please be thorough when reviewing. I'm obviously very open to suggestions.

If you need something to try this out with, heres a little Bear I wrote while working on this. It wraps the [gitlint](https://pypi.python.org/pypi/gitlint/0.8.0) tool. (Once this PR is merged, I will make it into a proper Bear and add it to coala-bears).
```Python
from coalib.bearlib.abstractions.Linter import linter


@linter(executable='gitlint',
        global_bear=True,
        output_format='regex',
        use_stderr=True,
        output_regex=r'\d+: \S+ (?P<message>.*)')
class TestBear:
    LANGUAGES = {'Git'}
    AUTHORS = {'The coala developers'}
    AUTHORS_EMAILS = {'coala-devel@googlegroups.com'}
    LICENSE = 'AGPL-3.0'
    CAN_DETECT = {'Formatting'}

    def create_arguments(self, config_file):
        return []

```

### Checklist

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -9,11 +9,13 @@ from subprocess import check_call, CalledProcessError, DEVNULL
 from types import MappingProxyType
 
 from coalib.bears.LocalBear import LocalBear
+from coalib.bears.GlobalBear import GlobalBear
 from coala_utils.ContextManagers import make_temp
 from coala_utils.decorators import assert_right_type, enforce_signature
 from coalib.misc.Shell import run_shell_command
 from coalib.results.Diff import Diff
 from coalib.results.Result import Result
+from coalib.results.SourceRange import SourceRange
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.FunctionMetadata import FunctionMetadata
 
@@ -34,7 +36,8 @@ def _prepare_options(options, bear_class):
                        'use_stderr',
                        'config_suffix',
                        'executable_check_fail_info',
-                       'prerequisite_check_command'}
+                       'prerequisite_check_command',
+                       'global_bear'}
 
     if not options['use_stdout'] and not options['use_stderr']:
         raise ValueError('No output streams provided at all.')
@@ -64,6 +67,7 @@ def _prepare_options(options, bear_class):
             'origin',
             'message',
             'severity',
+            'filename',
             'line',
             'column',
             'end_line',
@@ -137,6 +141,10 @@ def _prepare_options(options, bear_class):
 
         allowed_options.add('prerequisite_check_fail_message')
 
+    if options['global_bear'] and options['use_stdin']:
+        raise ValueError('Incompatible arguments provided:'
+                         "'use_stdin' and 'global_bear' can't both be True.")
+
     # Check for illegal superfluous options.
     superfluous_options = options.keys() - allowed_options
     if superfluous_options:
@@ -155,7 +163,7 @@ def _create_linter(klass, options):
             return '<{} linter class (wrapping {!r})>'.format(
                 cls.__name__, options['executable'])
 
-    class LinterBase(LocalBear, metaclass=LinterMeta):
+    class LinterBase(metaclass=LinterMeta):
 
         @staticmethod
         def generate_config(filename, file):
@@ -180,26 +188,6 @@ def _create_linter(klass, options):
                 The config-file-contents as a string or ``None``.
             """
             return None
-
-        @staticmethod
-        def create_arguments(filename, file, config_file):
-            """
-            Creates the arguments for the linter.
-
-            You can provide additional keyword arguments and defaults. These
-            will be interpreted as required settings that need to be provided
-            through a coafile-section.
-
-            :param filename:
-                The name of the file the linter-tool shall process.
-            :param file:
-                The contents of the file.
-            :param config_file:
-                The path of the config-file if used. ``None`` if unused.
-            :return:
-                A sequence of arguments to feed the linter-tool with.
-            """
-            raise NotImplementedError
 
         @staticmethod
         def get_executable():
@@ -283,7 +271,8 @@ def _create_linter(klass, options):
             :param match:
                 The regex match object.
             :param filename:
-                The name of the file this match belongs to.
+                The name of the file this match belongs to or ``None`` for
+                project scope.
             :param severity_map:
                 The dict to use to map the severity-match to an actual
                 ``RESULT_SEVERITY``.
@@ -316,18 +305,30 @@ def _create_linter(klass, options):
                 groups['origin'] = '{} ({})'.format(klass.__name__,
                                                     groups['origin'].strip())
 
-            # Construct the result.
-            return Result.from_values(
-                origin=groups.get('origin', self),
-                message=(groups.get('message', '').strip()
-                         if result_message is None else result_message),
-                file=filename,
-                severity=groups['severity'],
-                line=groups['line'],
-                column=groups['column'],
-                end_line=groups['end_line'],
-                end_column=groups['end_column'],
-                additional_info=groups.get('additional_info', '').strip())
+            # GlobalBears do not pass a filename to the function. But they can
+            # still give one through the regex
+            if filename is None:
+                filename = groups.get('filename', None)
+
+            # Construct the result. If we have a filename, we
+            # use Result.from_values otherwise generate a project
+            # scope result.
+            result_params = {
+                'origin': groups.get('origin', self),
+                'message': (groups.get('message', '').strip()
+                            if result_message is None else result_message),
+                'severity': groups['severity'],
+                'additional_info': groups.get('additional_info', '').strip()
+            }
+
+            if filename:
+                range = SourceRange.from_values(filename,
+                                                groups['line'],
+                                                groups['column'],
+                                                groups['end_line'],
+                                                groups['end_column'])
+                result_params['affected_code'] = (range,)
+            return Result(**result_params)
 
         def process_output_corrected(self,
                                      output,
@@ -402,6 +403,8 @@ def _create_linter(klass, options):
                 of the following named groups (via ``(?P<name>...)``) to
                 provide a good result:
 
+                - filename - The name of the linted file. This is relevant for
+                    global bears only.
                 - line - The line where the issue starts.
                 - column - The column where the issue starts.
                 - end_line - The line where the issue ends.
@@ -471,7 +474,7 @@ def _create_linter(klass, options):
                 _processing_function = partialmethod(
                     process_output_regex, **_process_output_args)
 
-            def process_output(self, output, filename, file):
+            def process_output(self, output, filename=None, file=None):
                 """
                 Processes the output of the executable and yields results
                 accordingly.
@@ -484,9 +487,11 @@ def _create_linter(klass, options):
                     the selected output stream). If both are ``True``, a tuple
                     is placed with ``(stdout, stderr)``.
                 :param filename:
-                    The name of the file currently processed.
+                    The name of the file currently processed or ``None`` for
+                    project scope.
                 :param file:
-                    The contents of the file (line-splitted).
+                    The contents of the file (line-splitted) or ``None`` for
+                    project scope.
                 """
                 if isinstance(output, str):
                     output = (output,)
@@ -497,15 +502,17 @@ def _create_linter(klass, options):
 
         @classmethod
         @contextmanager
-        def _create_config(cls, filename, file, **kwargs):
+        def _create_config(cls, filename=None, file=None, **kwargs):
             """
             Provides a context-manager that creates the config file if the
             user provides one and cleans it up when done with linting.
 
             :param filename:
-                The filename of the file.
+                The filename of the file being linted. ``None`` for project
+                scope.
             :param file:
-                The file contents.
+                The content of the file being linted. ``None`` for project
+                scope.
             :param kwargs:
                 Section settings passed from ``run()``.
             :return:
@@ -521,7 +528,17 @@ def _create_linter(klass, options):
                         fl.write(content)
                     yield config_file
 
-        def run(self, filename, file, **kwargs):
+        def run(self, filename=None, file=None, **kwargs):
+            """
+            Runs the wrapped tool.
+
+            :param filename:
+                The filename of the file being linted. ``None`` for project
+                scope.
+            :param file:
+                The content of the file being linted. ``None`` for project
+                scope.
+            """
             # Get the **kwargs params to forward to `generate_config()`
             # (from `_create_config()`).
             generate_config_kwargs = FunctionMetadata.filter_parameters(
@@ -536,8 +553,16 @@ def _create_linter(klass, options):
                     FunctionMetadata.filter_parameters(
                         self._get_create_arguments_metadata(), kwargs))
 
-                args = self.create_arguments(filename, file, config_file,
-                                             **create_arguments_kwargs)
+                # The interface of create_arguments is different for local
+                # and global bears, therefore we must check here, what kind
+                # of bear we have.
+                if isinstance(self, LocalBear):
+                    args = self.create_arguments(filename,
+                                                 file, config_file,
+                                                 **create_arguments_kwargs)
+                else:
+                    args = self.create_arguments(config_file,
+                                                 **create_arguments_kwargs)
 
                 try:
                     args = tuple(args)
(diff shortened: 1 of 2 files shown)


coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +319,28 @@ def _convert_output_regex_match_to_result(self,
...
-                file=filename,
-                severity=groups['severity'],
-                line=groups['line'],
-                column=groups['column'],
-                end_line=groups['end_line'],
-                end_column=groups['end_column'],
-                additional_info=groups.get('additional_info', '').strip())
+            # Construct the result. For LocalBears we use Result.from_values
+            # and give all the file related info.
+            # For GlobalBears we consruct a project scope result.
```
sils: that's not such a good idea. For global bears we should rather expect the filename to be in the regex groups.

sils:
@NiklasMM holy awesome! Thanks a ton for helping us with this, this has been pending for so long.

sils:
remove colon after closes

coalib/bearlib/abstractions/Linter.py
Makman2: yes for global bears enable the `filename` group for the regex :+1:

NiklasMM: uhm, sorry guys, I don't get it =( which filename should be there? There might not even be a file involved.

Makman2: But we should support that, as often linters do emit messages like
```
my-file.py: L3C9: Error detected
another-file.txt: L10C9: Total bad
```
That's why we should support for global bears a `(?P<filename>.*): L\d+C\d+...`

coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +319,28 @@ def _convert_output_regex_match_to_result(self,
...
+            if filename is not None:
```
impmihai: This could be simplified in `if filename:`

NiklasMM: I like to check for None-ness explicitly like this. This way the reader understands quickly that I'm only checking for None-ness at this point and don't handle empty strings or `0`

coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +319,33 @@ def _convert_output_regex_match_to_result(self,
...
+                    end_line=groups['end_line'],
+                    end_column=groups['end_column'],
+                    additional_info=groups.get('additional_info', '').strip())
+            else:
+                return Result(
+                    origin=groups.get('origin', self),
+                    message=(groups.get('message', '').strip()
+                             if result_message is None else result_message),
+                    severity=groups['severity'],
+                    additional_info=groups.get('additional_info', '').strip())
```
sils: so this is kinda duplication, we could reduce that by making a dict and conditionally adding the file stuff, creating the result by just unpacking it into the args.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -497,15 +517,15 @@ def process_output(self, output, filename, file):
...
         @contextmanager
-        def _create_config(cls, filename, file, **kwargs):
+        def _create_config(cls, filename=None, file=None, **kwargs):
             """
             Provides a context-manager that creates the config file if the
             user provides one and cleans it up when done with linting.
 
             :param filename:
-                The filename of the file.
+                The filename of the file. None for project scope.
```
sils: maybe clarify that this file is not the config but the linted file

coalib/bearlib/abstractions/Linter.py
```diff
@@ -565,21 +591,82 @@ def run(self, filename, file, **kwargs):
...
+                return self.process_output(output, **process_output_kwargs)
+
+    if options['global_bear']:
+        LinterBaseClass = GlobalLinterBase
+    else:
+        LinterBaseClass = LocalLinterBase
     # Mixin the linter into the user-defined interface, otherwise
     # `create_arguments` and other methods would be overridden by the
     # default version.
     result_klass = type(klass.__name__, (klass, LinterBase), {
```
sils: the LinterBaseClass will inherit from LinterBase anyway so why double multi inheritance confuse people here? :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -775,5 +862,6 @@ def linter(executable: str,
     options['config_suffix'] = config_suffix
     options['executable_check_fail_info'] = executable_check_fail_info
     options['prerequisite_check_command'] = prerequisite_check_command
+    options['global_bear'] = global_bear
```
sils: please add documentation for this in the docstring above

coalib/bearlib/abstractions/Linter.py
```diff
@@ -565,21 +591,82 @@ def run(self, filename, file, **kwargs):
...
+
+                output = tuple(compress(
+                    output,
+                    (options['use_stdout'], options['use_stderr'])))
+                if len(output) == 1:
+                    output = output[0]
+
+                process_output_kwargs = FunctionMetadata.filter_parameters(
+                    self._get_process_output_metadata(), kwargs)
+                return self.process_output(output, **process_output_kwargs)
```
sils: a big portion of this routine is copied, I'm not fully sure to what extend it was modified but I *am* sure that there's a better way of doing this. I'm almost certain that within a year we will have a bugfix in one run method but forgetting to port it to the other with the importance of this class.

coalib/bearlib/abstractions/Linter.py
sils: ok just looked at this side by side it's only a few file args here and there passed through, that can probably be hacked into the kwargs or so so we don't even need so many branches. Most certainly they can be joined.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -565,21 +591,82 @@ def run(self, filename, file, **kwargs):
...
+    result_klass = type(klass.__name__, (klass, LinterBaseClass), {
+        '__module__': klass.__module__})
     result_klass.__doc__ = klass.__doc__ or ''
     return result_klass
 
 
 @enforce_signature
 def linter(executable: str,
+           global_bear: bool=False,
            use_stdin: bool=False,
```
sils: you'll probably want to document and catch the case global_bear=True and use_stdin=True which is incompatible.

coalib/bearlib/abstractions/Linter.py
NiklasMM: Can you elaborate on that? `from_values` can't be used without a file, so either we must do it like this or not use `from_values` at all, which would lead to duplication of code from `Result.py`

sils: yes, you can just use the normal result constructor and create the sourceranges if the filename is given, otherwise have an empty list; i.e. not use from_values at all. I don't see how that leads to duplication

NiklasMM: Well I'd have to duplicate the creation of the SourceRange, which is now done in `from_values`. In this case the duplication even ranges across files.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +319,33 @@ def _convert_output_regex_match_to_result(self,
...
-                additional_info=groups.get('additional_info', '').strip())
+            # GlobalBears do not pass a filename to the function. But they can
+            # still give one through the regex
+            if filename is None:
+                filename = groups.get('filename', None)
+
+            # Construct the result. If we have a filename, we
+            # use Result.from_values otherwise generate a project
+            # scope result.
+            if filename is not None:
```
Makman2: rather inverse the condition for more simplicity:
```
if filename is None:
    return Result(...)
else:
    return Result.from_values(...)
```

coalib/bearlib/abstractions/Linter.py
Makman2: I don't understand why you make a new "kind" of `Result`. `line`, `column` etc. should be still available, shouldn't it? :O

coalib/bearlib/abstractions/Linter.py
```diff
@@ -521,6 +541,12 @@ def _create_config(cls, filename, file, **kwargs):
                         fl.write(content)
                     yield config_file
 
+        def __repr__(self):
+            return '<{} linter object (wrapping {!r}) at {}>'.format(
+                type(self).__name__, self.get_executable(), hex(id(self)))
```
Makman2: separate commit :+1:

coalib/bearlib/abstractions/Linter.py
```diff
@@ -497,15 +517,15 @@ def process_output(self, output, filename, file):
...
             """
             Provides a context-manager that creates the config file if the
             user provides one and cleans it up when done with linting.
 
             :param filename:
-                The filename of the file.
+                The filename of the file. None for project scope.
             :param file:
-                The file contents.
+                The file contents. None for project scope.
```
Makman2: could you enclose all `None`s in double backticks so they get rendered as code? (`None` is a code keyword, so I think we should render it appropriately 😄  )

coalib/bearlib/abstractions/Linter.py
Makman2: yes we have to avoid this

coalib/bearlib/abstractions/Linter.py
Makman2: indeed :+1: good catch :)

NiklasMM:
@sils @Makman2 Very sorry this took so long, but thanks to your suggestions I was able to simplify this significantly. The only thing I don't like is that the `create_arguments` methods of all global linter bears will have to accept a `file` and `filename` argument, which will always be `None`. If you see any way around this let me know.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -137,6 +140,10 @@ def _prepare_options(options, bear_class):
 
         allowed_options.add('prerequisite_check_fail_message')
 
+    if options['global_bear'] and options['use_stdin']:
+        raise ValueError('Incompatible arguments provided:'
+                         "'use_stdin' and 'global_bear' can't both be True")
```
Makman2: missing period at end :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -569,17 +602,28 @@ def __repr__(self):
...
+    class LocalLinterBase(LinterBase, LocalBear):
+        pass
+
+    class GlobalLinterBase(LinterBase, GlobalBear):
+        pass
+
+    if options['global_bear']:
+        LinterBaseClass = GlobalLinterBase
+    else:
+        LinterBaseClass = LocalLinterBase
```
Makman2: I would rather do this:
```
bear_base_class = GlobalLinterBase if options['global_bear'] else LocalLinterBase
result_klass = type(klass.__name__, (klass, LinterBase, bear_base_class), ...
```

Makman2: 8 LOC -> 1 LOC :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -678,6 +722,7 @@ def linter(executable: str,
         command-line-interface.
     :param use_stdout:
         Whether to use the stdout output stream.
+        Incompatible with global_bear=True.
```
Makman2: enclose `global_bear=True` in double backticks so it gets rendered as code.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -695,6 +740,11 @@ def linter(executable: str,
         A custom message that gets displayed when ``check_prerequisites``
         fails while invoking ``prerequisite_check_command``. Can only be
         provided together with ``prerequisite_check_command``.
+    :param global_bear:
+        Wether the created bear should be a global bear or not. Global bears
```
Makman2: Wether -> Whether

and maybe I would also do
global bear -> GlobalBear (enclosed in double backticks)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -695,6 +740,11 @@ def linter(executable: str,
...
+        will be run only once instead of once per file. The file and filename
+        args of all their methods will default to ``None``.
+        Incompatible with use_stdin=True.
```
Makman2: same here: `use_stdin=True` in double backticks :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -695,6 +740,11 @@ def linter(executable: str,
         A custom message that gets displayed when ``check_prerequisites``
         fails while invoking ``prerequisite_check_command``. Can only be
         provided together with ``prerequisite_check_command``.
+    :param global_bear:
+        Wether the created bear should be a global bear or not. Global bears
+        will be run only once instead of once per file. The file and filename
```
Makman2: maybe enclose `file` and `filename` also in double backticks, as you refer to them as arguments^^

Makman2:
about your problem with `create_arguments`: you could try to use an indirection layer:
```
args = self.create_arguments(filename, file, config_file,
    ...
```
instead you invoke
```
args = self._create_arguments(filename, file, config_file...)
```
and `self._create_arguments` is for local_bears ` = self.create_arguments`, and for global bears it's
```
def _create_arguments(self, filename, file, config_file):
    return self.create_arguments(config_file)

# or better: use a partial!
```

`@linter` really is a beast...

PS:
Actually with this we could also omit `config_file` without much overhead :+1: (there's an issue for that lying around somewhere...)

coalib/bearlib/abstractions/Linter.py
NiklasMM: Well, it's more like 4 -> 1, but still you have a point there :)

NiklasMM:
Good idea with the partial, but since `file` and `filename` are positional args, I can't fix them with a partial. Or am I misunderstanding you there.

About getting rid of `config_file`: Are you referring to #2067? I'll add some comments there, as I'm not sure how to do this. In the meantime, I found another solution which isn't as elegant, but works :)

Makman2:
> Good idea with the partial, but since file and filename are positional args, I can't fix them with a partial. Or am I misunderstanding you there.

`partial` supports positional arguments too :+1:

> Are you referring to #2067?

yep exactly :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +324,29 @@ def _convert_output_regex_match_to_result(self,
...
-                severity=groups['severity'],
-                line=groups['line'],
-                column=groups['column'],
-                end_line=groups['end_line'],
-                end_column=groups['end_column'],
-                additional_info=groups.get('additional_info', '').strip())
+            # GlobalBears do not pass a filename to the function. But they can
+            # still give one through the regex
+            if filename is None:
+                filename = groups.get('filename', None)
```
Makman2: to make this working properly you have to add `filename` to the list of supported names if global-bear shall be used: https://github.com/coala/coala/pull/3652/commits/495db77764bb0bd0344599e867bd44957f5b511e#diff-6dee3b7ded9b28fd0db83b2e7d039b86R66

coalib/bearlib/abstractions/Linter.py
```diff
@@ -316,18 +324,29 @@ def _convert_output_regex_match_to_result(self,
...
+            # Construct the result. If we have a filename, we
+            # use Result.from_values otherwise generate a project
+            # scope result.
+            result_params = {
+                'origin': groups.get('origin', self),
+                'message': (groups.get('message', '').strip()
+                            if result_message is None else result_message),
+                'severity': groups['severity'],
+                'additional_info': groups.get('additional_info', '').strip()
+            }
```
Makman2: would add a newline here for sake of readability :)

coalib/bearlib/abstractions/Linter.py
Makman2: Actually `filename` is mandatory if you give `line`, `column` etc to the regex, right? Maybe you can add that logic too :+1: :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -569,17 +610,43 @@ def __repr__(self):
...
+            will be interpreted as required settings that need to be provided
+            through a coafile-section. This is the file agnostic version for
+            global bears.
+
+            :param config_file:
+                The path of the config-file if used. ``None`` if unused.
+            :return:
+                A sequence of arguments to feed the linter-tool with.
+            """
+            raise NotImplementedError
```
Makman2: maybe you should copy the default implementation for `LocalBear` also the `LocalLinterBase`?

coalib/bearlib/abstractions/Linter.py
NiklasMM: I'm not 100% sure that there can't be a case where `line` and `column` are given and `filename` isn't even for a global bear. E.g. a global bear which validates only a specific file in a repo (manifest, license, etc) might not give the name of the file in each part of the output.

sils:
ack https://github.com/coala/coala/pull/3652/commits/ed8c91d5a628e54e69588e248774d657227ab1f0

NiklasMM:
Anything left to do here for me?

sils:
I wanted to cherry pick it with the other coala PRS from coala.io/approved
of you want to speedt it up just put them all together in one rebased pr
and reack and merge it.

sils:

Issue #3655, opened 2017-01-30
Globbing: Don't even walk into directories that match Ignore globs


underyx:
A hacky proof of concept of this reduced runtime with no bears on a real-world project from 14 seconds to 2 seconds. (and ~50 seconds in 0.9, before #3641)

difficulty/medium

underyx:
I can do this on Friday, if someone else can do it earlier, go ahead and take it.

impmihai:
I have a possible solution here : https://github.com/coala/coala/pull/3674. I would like to know if it is an efficient solution, it is still WIP.

sils:
@impmihai tries to fix? Or fixes?

sils:
either it relates to it, i.e. goes into the direction, or it fixes it can you decide? ;)

impmihai:
@sils it is my first major contribution, i was a little scared to say it fixes. 😅

palash25:
@impmihai still working on this? If not then I would like to take this up as this might be related to a GSoC project that I'm preparing for. This will help me get a better understanding of Globbing.

Makman2:
Unassigning due to inactivity.

@palash25 I think you've enough issues in your queue, please process some before you take up new issues :+1: :smile:

PR #3674, opened 2017-02-01
Collectors.py: Pass ignored_globs to the globbing functions

impmihai:
The script originally collected all files it could find and
then check which of them should be ignored.
Now it sends the ignored globs to the iglob function so we
win some execution time.

Closes: #3655

Extreme speed test: http://imgur.com/a/7BUf3
*This is an extreme case scenario, but it is a good
example for big projects, with a lot of object files
that do not need checking. I just copied the coala
source over and over again in the same directory
and run with all files ignored. I really want a
better real world speed test, if anybody can help.

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -52,7 +52,7 @@ def icollect(file_paths, ignored_globs=None, match_cache={}):
 
     for file_path in file_paths:
         if file_path not in match_cache:
-            match_cache[file_path] = list(iglob(file_path))
+            match_cache[file_path] = list(iglob(file_path, ignored_globs))
 
         for match in match_cache[file_path]:
             if not ignored_globs or not fnmatch(match, ignored_globs):
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -252,16 +252,21 @@ def _absolute_flat_glob(pattern):
     return
 
 
-def _iter_relative_dirs(dirname):
+def _iter_relative_dirs(dirname, ignored_globs=None):
     """
     Recursively iterates subdirectories of all levels from dirname
 
-    :param dirname: Directory name
-    :return:        Iterator that yields files and directory from the given dir
-                    and all it's (recursive) subdirectories
+    :param dirname:       Directory name
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              Iterator that yields files and directory from the
+                          given dir all it's (recursive) subdirectories
     """
     if not dirname:
         dirname = os.curdir
+
+    if ignored_globs:
+        if fnmatch(dirname, ignored_globs):
+            return
     try:
         files_or_dirs = os.listdir(dirname)
     except os.error:
@@ -269,23 +274,25 @@ def _iter_relative_dirs(dirname):
     for file_or_dir in files_or_dirs:
         yield file_or_dir
         path = os.path.join(dirname, file_or_dir)
-        for sub_file_or_dir in _iter_relative_dirs(path):
+        for sub_file_or_dir in _iter_relative_dirs(path, ignored_globs):
             yield os.path.join(file_or_dir, sub_file_or_dir)
 
 
-def relative_wildcard_glob(dirname, pattern):
+def relative_wildcard_glob(dirname, pattern, ignored_globs=None):
     """
     Non-recursive glob for one directory. Accepts wildcards.
 
-    :param dirname: Directory name
-    :param pattern: Glob pattern with wildcards
-    :return:        List of files in the dir of dirname that match the pattern
+    :param dirname:       Directory name
+    :param pattern:       Glob pattern with wildcards
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              List of files in the dir of dirname that
+                          match the pattern
     """
     if not dirname:
         dirname = os.curdir
     try:
         if '**' in pattern:
-            names = list(_iter_relative_dirs(dirname))
+            names = list(_iter_relative_dirs(dirname, ignored_globs))
         else:
             names = os.listdir(dirname)
     except OSError:
@@ -299,33 +306,35 @@ def relative_wildcard_glob(dirname, pattern):
     return result
 
 
-def relative_flat_glob(dirname, basename):
+def relative_flat_glob(dirname, basename, ignored_globs=None):
     """
     Non-recursive glob for one directory. Does not accept wildcards.
 
-    :param dirname:  Directory name
-    :param basename: Basename of a file in dir of dirname
-    :return:         List containing Basename if the file exists
+    :param dirname:       Directory name
+    :param basename:      Basename of a file in dir of dirname
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              List containing Basename if the file exists
     """
     if os.path.exists(os.path.join(dirname, basename)):
         return [basename]
     return[]
 
 
-def relative_recursive_glob(dirname, pattern):
+def relative_recursive_glob(dirname, pattern, ignored_globs=None):
     """
     Recursive Glob for one directory and all its (nested) subdirectories.
     Accepts only '**' as pattern.
 
-    :param dirname: Directory name
-    :param pattern: The recursive wildcard '**'
-    :return:        Iterator that yields all the (nested) subdirectories of the
-                    given dir
+    :param dirname:       Directory name
+    :param pattern:       The recursive wildcard '**'
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              Iterator that yields all the (nested) subdirectories
+                          of the given dir
     """
     assert pattern == '**'
     if dirname:
         yield pattern[:0]
-    for relative_dir in _iter_relative_dirs(dirname):
+    for relative_dir in _iter_relative_dirs(dirname, ignored_globs):
         yield relative_dir
 
 
@@ -336,14 +345,14 @@ def has_wildcard(pattern):
     """
     Checks whether pattern has any wildcards.
 
-    :param pattern: Glob pattern that may contain wildcards
-    :return:        Boolean: Whether or not there are wildcards in pattern
+    :param pattern:       Glob pattern that may contain wildcards
+    :return:              Boolean: Whether or not there are wildcards in pattern
     """
     match = wildcard_check_pattern.search(pattern)
     return match is not None
 
 
-def _iglob(pattern):
+def _iglob(pattern, ignored_globs=None):
     dirname, basename = os.path.split(pattern)
     if not has_wildcard(pattern):
         for file in _absolute_flat_glob(pattern):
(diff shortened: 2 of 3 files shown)


coalib/parsing/Globbing.py
```diff
@@ -252,14 +252,20 @@ def _absolute_flat_glob(pattern):
...
     Recursively iterates subdirectories of all levels from dirname
 
+    :param ignored_globs: List of globs to ignore when matching files
     :param dirname: Directory name
     :return:        Iterator that yields files and directory from the given dir
                     and all it's (recursive) subdirectories
     """
+    if ignored_globs != None:
+        for ignored_dir in ignored_globs:
+            if dirname.endswith(ignored_dir):
```
underyx: I don't think this will work because

```python
ignored_dir == 'env/**'
```

or

```python
ignored_dir == 'vendor/*/node_modules/**'
```

etc.

impmihai: @underyx I was asking if the method i am parsing the ignored dirs is ok. I am working already on what you pointed.

underyx: Oh, right, sorry, I didn't realize.

I think this method is the best we can do indeed, though I'm not super familiar with this code. It might turn out to be a good idea to do the ignoring elsewhere. If I were implementing this I'd just try to handle the ignoring as soon as possible and run benchmarks to make sure I'm not missing some alternative code path that won't have the ignoring.

coalib/parsing/Globbing.py
```diff
@@ -252,14 +252,20 @@ def _absolute_flat_glob(pattern):
     return
```
Adrianzatreanu: commit message: globes-> globs ;p

coalib/parsing/Globbing.py
```diff
@@ -252,14 +252,20 @@ def _absolute_flat_glob(pattern):
...
+def _iter_relative_dirs(dirname, ignored_globs=None):
     """
     Recursively iterates subdirectories of all levels from dirname
 
+    :param ignored_globs: List of globs to ignore when matching files
     :param dirname: Directory name
     :return:        Iterator that yields files and directory from the given dir
                     and all it's (recursive) subdirectories
     """
+    if ignored_globs is not None:
```
Adrianzatreanu: wouldnt "if ignored_globs:" be enough?

impmihai:
@underyx Can you run a test with this change like you did in the issue? I don't have a large enough project at the moment to try a benchmark.

coalib/parsing/Globbing.py
```diff
@@ -252,14 +252,20 @@ def _absolute_flat_glob(pattern):
...
+        for ignored_dir in ignored_globs:
+            if fnmatch(dirname, ignored_globs):
```
userzimmermann: Should be `if fnmatch(dirname, ignored_dir):`

coalib/parsing/Globbing.py
userzimmermann: explicit `None` check is just more recommended

coalib/parsing/Globbing.py
```diff
@@ -252,14 +252,20 @@ def _absolute_flat_glob(pattern):
...
+                return
```
userzimmermann: Should be `if fnmatch(dirname, ignored_dir):`

coalib/parsing/Globbing.py
```diff
@@ -252,31 +252,37 @@ def _absolute_flat_glob(pattern):
     return
 
 
-def _iter_relative_dirs(dirname):
+def _iter_relative_dirs(dirname, ignored_globs=None):
     """
     Recursively iterates subdirectories of all levels from dirname
 
+    :param ignored_globs: List of globs to ignore when matching files
```
aptrishu: maintain the order of param definition according to the argument..
```Cython
:param dirname:          Directory name
:param ignored_globs: List of globs to ignore when matching files
:return:                         Iterator that yields files and directory from the given dir
                                    and all it's (recursive) subdirectories
```
Also maintain the alignment..

aptrishu: everywhere..^^

sils:
- you got conflicts
- https://github.com/coala/coala/issues/3655#issuecomment-277850830
- you seem to have a second shortlog right in the middle of your commit message, why? Maybe some squashing, forgot to remove it?

sils:
seems you lack some testing

coalib/parsing/Globbing.py
```diff
@@ -252,40 +252,47 @@ def _absolute_flat_glob(pattern):
...
+    :return:              Iterator that yields files and directory from the
+                          given dir all it's (recursive) subdirectories
     """
     if not dirname:
         dirname = os.curdir
+
+    if ignored_globs:
+        if fnmatch(dirname, ignored_globs):
+            return
```
RohanVB: Tests are failing due to coverage by this. 

> 268-269, 267->268

coalib/parsing/Globbing.py
```diff
@@ -252,40 +252,47 @@ def _absolute_flat_glob(pattern):
...
-                    and all it's (recursive) subdirectories
+    :param dirname:       Directory name
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              Iterator that yields files and directory from the
+                          given dir all it's (recursive) subdirectories
     """
     if not dirname:
         dirname = os.curdir
+
+    if ignored_globs:
```
Makman2: what about having an empty tuple as the default value for `ignored_globs`, then you don't need the if^^ `fnmatch` would return immediately because there aren't any globs^^

coalib/parsing/Globbing.py
```diff
@@ -358,49 +368,51 @@ def _iglob(pattern):
...
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              Iterator that yields all file names that match pattern
     """
     for pat in _iter_alternatives(pattern):
         pat = os.path.expanduser(pat)
         pat = os.path.normcase(pat)
 
         if pat.endswith(os.sep):
-            for name in _iglob(pat):
+            for name in _iglob(pat, ignored_globs):
```
Makman2: just for reference, we already supported ignoring globs directly, but we just did not make use of it? :O

impmihai:
Ignored globs were filtered after all files were collected. Initially we were looking for all files and directories in the project and then remove the files and directories that were matching ignore globs.

coalib/parsing/Globbing.py
impmihai: Ignored globs were filtered after all files were collected. Initially we were looking for all files and directories in the project and then remove the files and directories that were matching ignore globs.

coalib/parsing/Globbing.py
impmihai: I used it this way to make this parameter optional for functions. This approach means forcing all function calls to have an empty touple added.

coalib/parsing/Globbing.py
impmihai: Thank you for the help!

coalib/parsing/Globbing.py
Makman2: why that?
```
def _iter_relative_dirs(dirname, ignored_globs=tuple()):
    ...
```

coalib/parsing/Globbing.py
Makman2: ah never mind I understand it now :3 Underestimated the `fnmatch` :3

coalib/parsing/Globbing.py
impmihai: Oh, right, sorry. 
Should i change it like that? It is not that big of a difference.

Makman2: yeah please, even though it's just a single line, we avoid it :)

impmihai: The code that was already there had ignored_globs initializated with None. Should i change everywhere it's default value then? Because if i don't, it will end up as None sometimes in the _iter_relative_dirs function

Makman2: Where exactly?

impmihai: The first time it can be initializated with `None` is in `coalib/collecting/Collectors.py' on icollect function

Makman2: hm... okay then keep it :+1:

coalib/parsing/Globbing.py
Makman2: Where does the current file-filtering happen actually?

impmihai: Im collectors.py, right after my change in that file

Makman2: If possible, I think we should also filter out file names, to make the globbing functions work more consistently (as I respect from `glob` to filter out files too that don't match). Then you can remove the filtering from Collectors.py (which should only deal with collection, not with handling globbing).

impmihai: ok, I am working on it, but i think this will take a little time

impmihai: @Makman2 looks like it is pretty hard to do that because of cache, so i would have to change lots of tests to make this work the way you asked. That final filter would better stay there

Makman2: alright then let's keep it for now :) Be sure to file an issue after merge that we should also filter out files, like someone would expect from this function :+1:

coalib/parsing/Globbing.py
```diff
@@ -336,14 +345,15 @@ def has_wildcard(pattern):
     """
     Checks whether pattern has any wildcards.
 
-    :param pattern: Glob pattern that may contain wildcards
-    :return:        Boolean: Whether or not there are wildcards in pattern
+    :param pattern:       Glob pattern that may contain wildcards
+    :param ignored_globs: List of globs to ignore when matching files
```
Makman2: hah too enthusiastic :D
this function does not have an `ignored_globs` param

tests/parsing/GlobbingTest.py
```diff
@@ -222,8 +222,9 @@ class GlobTest(unittest.TestCase):
     def setUp(self):
         self.maxDiff = None
 
-    def _test_glob(self, pattern, file_list):
-        results = sorted([os.path.normcase(g) for g in glob(pattern)])
+    def _test_glob(self, pattern, file_list, ignored_globs=None):
```
Makman2: huh why do you add an `ignored_globs` param? :O

coalib/parsing/Globbing.py
impmihai: Whoops:D

tests/parsing/GlobbingTest.py
Makman2: aaaah nvm... why the heck is there just an underscore before the function name, I thought it was a test... not your fault :3

impmihai:
Anything else to add? 😄

coalib/collecting/Collectors.py
```diff
@@ -52,7 +52,7 @@ def icollect(file_paths, ignored_globs=None, match_cache={}):
 
     for file_path in file_paths:
         if file_path not in match_cache:
-            match_cache[file_path] = list(iglob(file_path))
+            match_cache[file_path] = list(iglob(file_path, ignored_globs))
 
         for match in match_cache[file_path]:
             if not ignored_globs or not fnmatch(match, ignored_globs):
```
sils: is this loop needed then?

coalib/parsing/Globbing.py
```diff
@@ -299,33 +306,35 @@ def relative_wildcard_glob(dirname, pattern):
     return result
 
 
-def relative_flat_glob(dirname, basename):
+def relative_flat_glob(dirname, basename, ignored_globs=None):
```
sils: this function doesn't need it!

coalib/parsing/Globbing.py
```diff
@@ -358,49 +367,51 @@ def _iglob(pattern):
...
+def iglob(pattern, ignored_globs=None):
     """
     Iterates all filesystem paths that get matched by the glob pattern.
     Syntax is equal to that of fnmatch.
 
-    :param pattern: Glob pattern with wildcards
-    :return:        Iterator that yields all file names that match pattern
+    :param pattern:       Glob pattern with wildcards
+    :param ignored_globs: List of globs to ignore when matching files
+    :return:              Iterator that yields all file names that match pattern
```
sils: you should elaborate on the ignore here - you still return some ignored files, right?You just don't traverse into ignored dirs.

Makman2: I already noticed this, and I think we can make a new issue for that. Let's improve speed for now, and increase consistency in a separate PR, so we get something merged.

coalib/parsing/Globbing.py
impmihai: yes, but in globbing.py in _iglob function, there is a variable names relative_glob_function which can get this function assigned. And that relative_glob_function is called with 2 parameters.

Makman2: don't understand that. If you don't need it, then you don't need, am I wrong? :D

impmihai: `if basename == '**':
    relative_glob_function = relative_recursive_glob
elif has_wildcard(basename):
    relative_glob_function = relative_wildcard_glob
else:
    relative_glob_function = relative_flat_glob
 
if not dirname:
    for file in relative_glob_function(dirname, basename, ignored_globs):
        yield fille`

This is how the function looks like. And in case the if exists on 3rd path, the `relative_glob_function` is equal to `relative_flat_glob`. After this, the `relative_glob_function` is called with 2 parameters. So `relative_flat_glob` gets called with 2 parameters. If it expects only one, then it starts complaining.

coalib/parsing/Globbing.py
```diff
@@ -336,14 +345,14 @@ def has_wildcard(pattern):
     """
     Checks whether pattern has any wildcards.
 
-    :param pattern: Glob pattern that may contain wildcards
-    :return:        Boolean: Whether or not there are wildcards in pattern
+    :param pattern:       Glob pattern that may contain wildcards
```
Asnelchristian: uhm... this should go in a different commit

coalib/parsing/Globbing.py
```diff
@@ -336,14 +345,14 @@ def has_wildcard(pattern):
...
+    :return:              Boolean: Whether or not there are wildcards in pattern
```
Asnelchristian: this change is not actually needed

Asnelchristian:
this is okay for me :+1:

Asnelchristian:
needs rebase

Makman2:
Closing due to inactivity.

PR #3680, opened 2017-02-02
Add ShellCommandResult for accessing return codes

userzimmermann:
`@linter` bears can produce wrong results when the external linting tools fail, which can't be checked with the current `coalib.misc.Shell` API. No chance to check the `returncode` of the process created by `run_shell_command`.

This is most probably also the reason for https://github.com/coala/coala-bears/issues/1235

Currently `run_shell_command` only returns the `(stdout, stderr)` tuple from `Popen.communicate`. I wrapped that in a backwards-compatible `tuple`-derived `ShellCommandResult` class which additionally stores the `Popen` instance and has a `.code` property to easily check the return code of the process.

### DIFF
--- a/coalib/misc/Shell.py
+++ b/coalib/misc/Shell.py
@@ -5,6 +5,43 @@ from subprocess import PIPE, Popen
 from shutil import which
 
 
+class ShellCommandResult(tuple):
+    """
+    The result of a :func:`coalib.misc.run_shell_command` call.
+
+    It is based on a ``(stdout, stderr)`` string tuple like it is returned
+    form ``subprocess.Popen.communicate`` and was originally returned from
+    :func:`coalib.misc.run_shell_command`. So it is backwards-compatible.
+
+    It additionally stores the return ``.code``:
+
+    >>> process = Popen(['python', '-c', 'print(input() + " processed")'],
+    ...                 stdin=PIPE, stdout=PIPE, stderr=PIPE,
+    ...                 universal_newlines=True)
+
+    >>> stdout, stderr = process.communicate(input='data')
+    >>> result = ShellCommandResult(process.returncode, stdout, stderr)
+    >>> result[0]
+    'data processed\\n'
+    >>> result[1]
+    ''
+    >>> result.code
+    0
+    """
+
+    def __new__(cls, code, stdout, stderr):
+        """
+        Creates the basic tuple from `stdout` and `stderr`.
+        """
+        return tuple.__new__(cls, (stdout, stderr))
+
+    def __init__(self, code, stdout, stderr):
+        """
+        Stores the return `code`.
+        """
+        self.code = code
+
+
 @contextmanager
 def run_interactive_shell_command(command, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


coalib/misc/Shell.py
```diff
@@ -4,6 +4,56 @@
...
+        Creates the basic tuple from `stdout` and `stderr`.
+        """
+        return tuple.__new__(cls, (stdout, stderr))
+
+    @enforce_signature
+    def __init__(self, process: Popen, stdout: str, stderr: str):
+        """
+        Stores the `process` instance.
+        """
+        self.process = process
```
Makman2: do we really need it?

Makman2:
it's good to maintain compatability. Though I'm thinking for sake of simplicity to just change the return value... what do you think?

userzimmermann:
@Makman2 `run_shell_command` is also directly used in certain bears :/

Makman2:
oh right... I'm opening an issue then to discuss a possible deprecation later...

coalib/misc/Shell.py
Makman2: I mean maybe we should just store the returncode directly^^

userzimmermann: You might want to get additional information from the process object later... like... hmm... well... do we really need it? :D

userzimmermann:
@Makman2 Now `ShellCommandResult` itself does the `.communicate()`, simplifying the whole thing a bit more.

coalib/misc/Shell.py
```diff
@@ -102,7 +140,7 @@ def run_shell_command(command, stdin=None, **kwargs):
     :return:        A tuple with ``(stdoutstring, stderrstring)``.
     """
     with run_interactive_shell_command(command, **kwargs) as p:
-        ret = p.communicate(stdin)
+        ret = ShellCommandResult(p, stdin)
```
Makman2: you think that makes sense to move the functionality into the result holding class? not sure...

userzimmermann: @Makman2 Not sure too... Now we have 2 choices :D

Makman2: I'm more for separating functionality, so let `ShellCommandResult` be a simple data-holder and do
```
for ...
    ret = p.communicate(stdin)
return ShellCommandResult(p.returncode, *ret)
```

Makman2: bump

userzimmermann:
@sils @Makman2 Added `pip install --upgrade pip setuptools` to `.travis.yml` for avoiding https://github.com/pypa/setuptools/issues/937 which (randomly) makes travis builds fail.

sils:
ack https://github.com/coala/coala/pull/3680/commits/dffca8c7e6c3d3a30731bfc76ee9bf766c819c4d

userzimmermann:
@sils @Makman2 Also added `pip install pip setuptools -U` to `circle.yml`

aptrishu:
ack 554036a6ef2cd

aptrishu:
@rultor merge

Issue #2693, opened 2016-08-31
"Warning: the bear ... failed" should include the filename on which it failed


sils:
later on the next gen bear design it should include the `str` conversion of the task object.

Otherwise it's like "hell coala failed on everything" when it only failed on a few files.

sils1297:
CC @Udayan12167 next gen related

Makman2:
the task object should appear in the debug logs imo, not by default^^

sils1297:
Its string representation should be short and it should be there by default
definitely because otherwise the warning is rather useless, like our
current exception warning which doesn't give any information and you always
have to get to debug level.

Nosferatul:
Can you guide me to the files that need to be changed?

Makman2:
@Nosferatul this is related to the next-gen-bears implementation, can't be done yet as we don't use "task" objects currently :)

Nosferatul:
Oh, my bad :)

Makman2:
oh wait I've misread a bit, you can already try to implement this (forget about the task object, but printing the filename should be possible) :D
@Nosferatul

Nosferatul:
Ok, from where should I start?

Makman2:
https://github.com/coala/coala/search?utf8=%E2%9C%93&q=failed+to+run

in specific
https://github.com/coala/coala/blob/4e94cdda8dd7793450c97f43c531173bff65bd24/coalib/processes/BearRunning.py
and
https://github.com/coala/coala/blob/7a1f395b162b64e6c4366298dedc17a8563e2dd2/coalib/bears/Bear.py
I think^^

Nosferatul:
Ok, I'll check it! Thank you!

prafful002:
@sils can I give this assigned to me if noone is working on it ?

sils:
@prafful002 there you go

nemaniarjun:
Can I work on this? It hasn't been updated in a while.

Nosferatul:
@prafful002 any progress?

yash-nisar:
Since a majority of the issues have been assigned, can I work on this ? @sils @Makman2

yash-nisar:
@sils Can you point me to the steps to reproduce the following warning ? 
`Warning: the bear ... failed`

sils:
from `git grep failed`: `coalib/bears/Bear.py:            self.warn('Bear {} failed to run. Take a look at debug messages'`

[...]

PR #3698, opened 2017-02-05
Bear.py: Include filename on which the bear failed

yash-nisar:
Mentions the absolute path of the file on which
 the local bear failed

Closes https://github.com/coala/coala/issues/2693

### Checklist

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -14,6 +14,7 @@ from coala_utils.decorators import (enforce_signature, classproperty,
 from dependency_management.requirements.PackageRequirement import (
     PackageRequirement)
 from dependency_management.requirements.PipRequirement import PipRequirement
+from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.output.printers.LogPrinter import LogPrinterMixin
 from coalib.results.Result import Result
 from coalib.settings.FunctionMetadata import FunctionMetadata
@@ -249,8 +250,14 @@ class Bear(Printer, LogPrinterMixin):
             result = self.run_bear_from_section(args, kwargs)
             return [] if result is None else list(result)
         except (Exception, SystemExit):
-            self.warn('Bear {} failed to run. Take a look at debug messages'
-                      ' (`-V`) for further information.'.format(name))
+            if self.kind() == BEAR_KIND.LOCAL:
+                self.warn('Bear {} failed to run on file {}. Take a look '
+                          'at debug messages (`-V`) for further '
+                          'information.'.format(name, args[0]))
+            else:
+                self.warn('Bear {} failed to run. Take a look '
+                          'at debug messages (`-V`) for further '
+                          'information.'.format(name))
             self.debug(
                 'The bear {bear} raised an exception. If you are the author '
                 'of this bear, please make sure to catch all exceptions. If '
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -6,6 +6,9 @@ import shutil
 from time import sleep
 
 from coalib.bears.Bear import Bear
+from coalib.bears.BEAR_KIND import BEAR_KIND
+from coalib.bears.GlobalBear import GlobalBear
+from coalib.bears.LocalBear import LocalBear
 from coalib.results.Result import Result
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.processes.communication.LogMessage import LogMessage
@@ -18,6 +21,10 @@ class BadTestBear(Bear):
     def __init__(self, section, queue):
         Bear.__init__(self, section, queue)
 
+    @staticmethod
+    def kind():
+        return BEAR_KIND.GLOBAL
+
     def run(self):
         raise NotImplementedError
 
@@ -28,6 +35,10 @@ class TestBear(Bear):
 
     def __init__(self, section, queue):
         Bear.__init__(self, section, queue)
+
+    @staticmethod
+    def kind():
+        return BEAR_KIND.GLOBAL
 
     def run(self):
         self.print('set', 'up', delimiter='=')
@@ -105,7 +116,7 @@ class BearTest(unittest.TestCase):
     def test_simple_api(self):
         self.assertRaises(TypeError, TestBear, self.settings, 2)
         self.assertRaises(TypeError, TestBear, None, self.queue)
-        self.assertRaises(NotImplementedError, self.uut.kind)
+        self.assertRaises(NotImplementedError, Bear.kind)
 
         base = Bear(self.settings, None)
         self.assertRaises(NotImplementedError, base.run)
@@ -128,6 +139,26 @@ class BearTest(unittest.TestCase):
                            'information.')
         # debug message contains custom content, dont test this here
         self.queue.get()
+
+    def test_print_filename_LocalBear(self):
+        self.uut = LocalBear(self.settings, self.queue)
+        self.uut.execute('filename.py', 'file\n')
+        self.check_message(LOG_LEVEL.DEBUG)
+        # Fails because of no run() implementation
+        self.check_message(LOG_LEVEL.WARNING,
+                           'Bear LocalBear failed to run on file filename.py. '
+                           'Take a look at debug messages (`-V`) for further '
+                           'information.')
+
+    def test_print_no_filename_GlobalBear(self):
+        self.uut = GlobalBear(None, self.settings, self.queue)
+        self.uut.execute()
+        self.check_message(LOG_LEVEL.DEBUG)
+        # Fails because of no run() implementation
+        self.check_message(LOG_LEVEL.WARNING,
+                           'Bear GlobalBear failed to run. Take a look at '
+                           'debug messages (`-V`) for further '
+                           'information.')
 
     def test_inconvertible(self):
         self.uut = TypedTestBear(self.settings, self.queue)


coalib/bears/Bear.py
```diff
@@ -249,8 +249,14 @@ def execute(self, *args, **kwargs):
             result = self.run_bear_from_section(args, kwargs)
             return [] if result is None else list(result)
         except (Exception, SystemExit):
-            self.warn('Bear {} failed to run. Take a look at debug messages'
-                      ' (`-V`) for further information.'.format(name))
+            if self.kind():
```
meetmangukiya: This is incorrect, it'll be true for both local and global bears as well as linter bears making the else unreachable. Instead check it against the constants in the `coalib/bears/BEAR_KIND`.

Also add tests

yash-nisar: Yeah, the tests are remaining.

yash-nisar: However, I have replaced it with the condition `bear.kind() == BEAR_KIND.GLOBAL`. Is that what you implied to say ?

yash-nisar:
Screwed up with the virualenv, idk how. :worried:

coalib/bears/Bear.py
meetmangukiya: Yeah, you are right.. I just assumed that BEAR_KIND.GLOBAL_BEAR would return something more informative, but it seems that we are using a it as a binary `0` is Local and `1` is global.

But I still suggest to use `self.kind() == BEAR_KIND.GLOBAL` as it is more readable and understandable then `if self.kind()`..

@Makman2 would you suggest the same ? ^^

Makman2: yes please definitely do `if self.kind() == BEAR_KIND.XYZ`, relying on those values is bad as we use an enum which auto-generates those values^^

Makman2: Btw `self.kind()` uses `NotImplemented` by default, maybe you should catch that^^

yash-nisar:
Does this look good ? Please have a look at it when you are free :smile:  @Makman2

coalib/bears/Bear.py
```diff
@@ -249,8 +250,19 @@ def execute(self, *args, **kwargs):
...
-                      ' (`-V`) for further information.'.format(name))
+            try:
+                if self.kind() == BEAR_KIND.GLOBAL:
+                    self.warn('Bear {} failed to run. Take a look '
+                              'at debug messages (`-V`) for further '
+                              'information.'.format(name))
+                else:
+                    self.warn('Bear {} failed to run on file {}. Take a look '
+                              'at debug messages (`-V`) for further '
+                              'information.'.format(name, args[0]))
```
Makman2: actually I would rephrase the condition a bit:
```
if self.kind() == BEAR_KIND.LOCAL:
    # print file path
else:
    # normal message
```
conceptually you want to print filenames only on local bears, regardless what other types exist^^

coalib/bears/Bear.py
```diff
@@ -249,8 +250,19 @@ def execute(self, *args, **kwargs):
...
+            except (Exception, SystemExit):
+                self.warn('Bear {} failed to run. Take a look '
+                          'at debug messages (`-V`) for further '
+                          'information.'.format(name))
```
Makman2: huh why do you have the same try-catch block again?

tests/bears/BearTest.py
```diff
@@ -129,6 +139,24 @@ def test_bad_bear(self):
         # debug message contains custom content, dont test this here
         self.queue.get()
 
+    def test_BadLocalBear(self):
+        self.uut = BadLocalBear(self.settings, self.queue)
```
Makman2: why not giving directly `LocalBear` with a comment that because of no `run()` implementation it fails?

tests/bears/BearTest.py
```diff
@@ -129,6 +139,24 @@ def test_bad_bear(self):
...
+        self.uut.execute('filename', 'file\n')
+        self.check_message(LOG_LEVEL.DEBUG)
+        self.check_message(LOG_LEVEL.WARNING,
+                           'Bear BadLocalBear failed to run on file filename. '
+                           'Take a look at debug messages (`-V`) for further '
+                           'information.')
+
+    def test_BadGlobalBear(self):
+        self.uut = BadGlobalBear(None, self.settings, self.queue)
```
Makman2: same here, `GlobalBear` vs `BadGlobalBear`?

coalib/bears/Bear.py
yash-nisar: To catch the `NotImplementedError` of the kind() method by default.

Makman2: but then catch for `NotImplementedError` + adapt the message :O

yash-nisar: What should the adapted message be from your point of view ?

Makman2: Actually I wouldn't catch it at all. If a bear without a kind makes it through this point, I don't know how to continue then. I mean without a kind coala doesn't know how to execute it (and me neither). A `NotImplementedError` with the stacktrace says everything relevant imo.

yash-nisar: Yes, what you're saying is absolutely right. So what should the custom stacktrace message be like ? `The bear hasn't been implemented` or something similar ?

Makman2: I would completely omit the try-catch^^

(Btw: Throwing an exception inside an except-block keeps the traceback alive, so don't `print`, but `raise` again :D)

yash-nisar: Completely omitting the second `try catch` inside the `except` block will result in failure of some tests since the bear [here](https://github.com/coala/coala/blob/master/tests/bears/BearTest.py#L16) doesn't have a kind. So while executing [this](https://github.com/coala/coala/blob/master/tests/bears/BearTest.py#L121) test, when it enters the if condition ie. `self.kind() == BEAR_KIND.LOCAL`, it will throw an exception IIUC.

yash-nisar: What do you suggest @Makman2 ?

tests/bears/BearTest.py
```diff
@@ -129,6 +131,24 @@ def test_bad_bear(self):
         # debug message contains custom content, dont test this here
         self.queue.get()
 
+    def test_LocalBear(self):  # Fails because of no run() implementation
```
meetmangukiya: Won't this failure be because of `NotImplementedError` ?

tests/bears/BearTest.py
```diff
@@ -129,6 +131,24 @@ def test_bad_bear(self):
...
+        self.uut = LocalBear(self.settings, self.queue)
+        self.uut.execute('filename', 'file\n')
+        self.check_message(LOG_LEVEL.DEBUG)
+        self.check_message(LOG_LEVEL.WARNING,
+                           'Bear LocalBear failed to run on file filename. '
+                           'Take a look at debug messages (`-V`) for further '
+                           'information.')
+
+    def test_GlobalBear(self):  # Fails because of no run() implementation
```
meetmangukiya: Even this one?

meetmangukiya: Because in that case it gets into except block, then again the except `NotImplementedError` block, and the if else will never be executed ?

meetmangukiya:
Rebase required

coalib/bears/Bear.py
Makman2: Try to give it a kind then :+1: :)

yash-nisar: So, basically, the default should be replaced by a kind instead of a NotImplementedError ?

coalib/bears/Bear.py
```diff
@@ -249,8 +250,19 @@ def execute(self, *args, **kwargs):
             result = self.run_bear_from_section(args, kwargs)
             return [] if result is None else list(result)
         except (Exception, SystemExit):
-            self.warn('Bear {} failed to run. Take a look at debug messages'
-                      ' (`-V`) for further information.'.format(name))
+            try:
+                if self.kind() == BEAR_KIND.LOCAL:
+                    self.warn('Bear {} failed to run on file {}. Take a look '
+                              'at debug messages (`-V`) for further '
+                              'information.'.format(name, args[0]))
```
Asnelchristian: you could maybe, use a variable to store  this message, so that you don't repeat yourself

coalib/bears/Bear.py
```diff
@@ -249,8 +250,19 @@ def execute(self, *args, **kwargs):
...
+                else:
+                    self.warn('Bear {} failed to run. Take a look '
+                              'at debug messages (`-V`) for further '
+                              'information.'.format(name))
+            except NotImplementedError:
+                self.warn('Bear {} failed to run. Take a look '
+                          'at debug messages (`-V`) for further '
+                          'information.'.format(name))
```
Asnelchristian: Maybe you should clearly state that, the bear doesn't exist or that it hasn't been implemented

coalib/bears/Bear.py
Asnelchristian: and have something like ``.format(name, 'on file' + args[0])``

Makman2: I would avoid assembling more "complex" grammatical structures using `format`. Though what makes sense is to append "Take a look at debug message ..." after the if I think^^

@Asnelchristian @yash-nisar opinions?

coalib/bears/Bear.py
yash-nisar: I removed the `try catch` since I gave a `DEFAULT` kind instead of `NotImplementedError`

coalib/bears/Bear.py
yash-nisar: I think we should keep 2 separate messages for better readability IMO @Makman2 @Asnelchristian

yash-nisar: I think this pretty much satisfies everyone's opinion ? Please let me know if it doesn't. :smile:

coalib/bears/BEAR_KIND.py
```diff
@@ -1,3 +1,3 @@
 from coalib.misc.Enum import enum
 
-BEAR_KIND = enum('LOCAL', 'GLOBAL')
+BEAR_KIND = enum('LOCAL', 'GLOBAL', 'DEFAULT')
```
Makman2: aaahh wait wait
what shall `DEFAULT`-kind mean?

I meant you shall add a kind to the bears defined in the tests :3

yash-nisar: I misinterpreted it. :(

yash-nisar: Is it okay now ? @Makman2

tests/bears/BearTest.py
```diff
@@ -105,7 +116,7 @@ def tearDown(self):
     def test_simple_api(self):
         self.assertRaises(TypeError, TestBear, self.settings, 2)
         self.assertRaises(TypeError, TestBear, None, self.queue)
-        self.assertRaises(NotImplementedError, self.uut.kind)
+        self.assertEqual(BEAR_KIND.GLOBAL, self.uut.kind())
```
Makman2: please create a new bear for it without an implementation for `kind`, this one explicitly tests the default behaviour^^

Makman2: or just do `self.assertRaises(NotImplementedError, Bear.kind)` :+1: :)

yash-nisar:
Looks good @Makman2 ?

tests/bears/BearTest.py
```diff
@@ -129,6 +140,24 @@ def test_bad_bear(self):
         # debug message contains custom content, dont test this here
         self.queue.get()
 
+    def test_LocalBear(self):  # Fails because of no run() implementation
```
Makman2: I would place the comment before `self.check_message(LOG_LEVEL.WARNING, ...)`, it somehow confuses me when I read it each time... sounds like the test case will always fail (which shouldn't happen, right :P)

Makman2: same for the comment below

tests/bears/BearTest.py
```diff
@@ -129,6 +140,24 @@ def test_bad_bear(self):
...
+        self.uut = LocalBear(self.settings, self.queue)
+        self.uut.execute('filename', 'file\n')
```
Makman2: add an extension, makes the warning message below better to read ;)

tests/bears/BearTest.py
Makman2: Also the name of the testcase: You don't test the whole `LocalBear`, you test the aspect of printing the filename. Rename it to sth that fits better :+1:

Makman2: same for testcase below

yash-nisar: I renamed them as `test_print_filename_LocalBear` and `test_print_filename_GlobalBear` but we don't print filenames for the GlobalBear ? so should I rename it to something else or are the changes good ? @Makman2

Makman2: then call it `test_print_no_filename_GlobalBear` :D

Makman2:
unack b306990

see https://github.com/coala/coala/pull/3698#discussion_r100858385 :)

tests/bears/BearTest.py
yash-nisar: Done and rebased as well ! :D

Makman2:
ack 1143282

Makman2:
@rultor merge

PR #3711, opened 2017-02-06
core: Add Bear v2

Makman2:
Part 1 - from https://github.com/coala/coala/pull/3523

### DIFF
--- a/coalib/core/Bear.py
+++ b/coalib/core/Bear.py
@@ -1,0 +1,460 @@
+from collections import defaultdict
+from functools import partial
+import inspect
+import logging
+from os import makedirs
+from os.path import join, abspath, exists
+
+from appdirs import user_data_dir
+
+from coala_utils.decorators import (enforce_signature, classproperty,
+                                    get_public_members)
+
+import requests
+
+from coalib.results.Result import Result
+from coalib.settings.ConfigurationGathering import get_config_directory
+from coalib.settings.FunctionMetadata import FunctionMetadata
+from coalib.settings.Section import Section
+
+
+class Bear:
+    """
+    A bear contains the actual subroutine that is responsible for checking
+    source code for certain specifications. However, it can actually do
+    whatever it wants with the files it gets.
+
+    This is the base class for every bear. If you want to write a bear, you
+    will probably want to look at the ``ProjectBear`` and ``FileBear`` classes
+    that inherit from this class.
+
+    To indicate which languages your bear supports, just give it the
+    ``LANGUAGES`` value which should be a set of string(s):
+
+    >>> class SomeBear(Bear):
+    ...     LANGUAGES = {'C', 'CPP', 'C#', 'D'}
+
+    To indicate the requirements of the bear, assign ``REQUIREMENTS`` a set
+    with instances of ``PackageRequirements``.
+
+    >>> from dependency_management.requirements.PackageRequirement import (
+    ...     PackageRequirement)
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = {
+    ...         PackageRequirement('pip', 'coala_decorators', '0.2.1')}
+
+    If your bear uses requirements from a manager we have a subclass from,
+    you can use the subclass, such as ``PipRequirement``, without specifying
+    manager:
+
+    >>> from dependency_management.requirements.PipRequirement import (
+    ...     PipRequirement)
+    >>> class SomeBear(Bear):
+    ...     REQUIREMENTS = {PipRequirement('coala_decorators', '0.2.1')}
+
+    To specify additional attributes to your bear, use the following:
+
+    >>> class SomeBear(Bear):
+    ...     AUTHORS = {'Jon Snow'}
+    ...     AUTHORS_EMAILS = {'jon_snow@gmail.com'}
+    ...     MAINTAINERS = {'Catelyn Stark'}
+    ...     MAINTAINERS_EMAILS = {'catelyn_stark@gmail.com'}
+    ...     LICENSE = 'AGPL-3.0'
+    ...     ASCIINEMA_URL = 'https://asciinema.org/a/80761'
+
+    If the maintainers are the same as the authors, they can be omitted:
+
+    >>> class SomeBear(Bear):
+    ...     AUTHORS = {'Jon Snow'}
+    ...     AUTHORS_EMAILS = {'jon_snow@gmail.com'}
+    >>> SomeBear.maintainers
+    {'Jon Snow'}
+    >>> SomeBear.maintainers_emails
+    {'jon_snow@gmail.com'}
+
+    If your bear needs to include local files, then specify it giving strings
+    containing relative file paths to the INCLUDE_LOCAL_FILES set:
+
+    >>> class SomeBear(Bear):
+    ...     INCLUDE_LOCAL_FILES = {'checkstyle.jar', 'google_checks.xml'}
+
+    To keep track easier of what a bear can do, simply tell it to the
+    ``CAN_FIX`` and the ``CAN_DETECT`` sets. Possible values are:
+
+    >>> CAN_DETECT = {'Syntax', 'Formatting', 'Security', 'Complexity',
+    ... 'Smell', 'Unused Code', 'Redundancy', 'Variable Misuse', 'Spelling',
+    ... 'Memory Leak', 'Documentation', 'Duplication', 'Commented Code',
+    ... 'Grammar', 'Missing Import', 'Unreachable Code', 'Undefined Element',
+    ... 'Code Simplification'}
+    >>> CAN_FIX = {'Syntax', ...}
+
+    Specifying something to ``CAN_FIX`` makes it obvious that it can be
+    detected too, so it may be omitted:
+
+    >>> class SomeBear(Bear):
+    ...     CAN_DETECT = {'Syntax', 'Security'}
+    ...     CAN_FIX = {'Redundancy'}
+    >>> sorted(SomeBear.can_detect)
+    ['Redundancy', 'Security', 'Syntax']
+
+    Every bear has a data directory which is unique to that particular bear:
+
+    >>> class SomeBear(Bear): pass
+    >>> class SomeOtherBear(Bear): pass
+    >>> SomeBear.data_dir == SomeOtherBear.data_dir
+    False
+
+    A bear can be dependent from other bears. ``BEAR_DEPS`` contains bear
+    classes that are to be executed before this bear gets executed. The results
+    of these bears will then be passed inside ``self.dependency_results`` as a
+    dict. The dict will have the name of the bear as key and a list of its
+    results as values:
+
+    >>> class SomeBear(Bear): pass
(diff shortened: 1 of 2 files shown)


Makman2:
Ahhhh damn I know why the docs build fail... it picks up the global coala installation (which doesn't seem to contain my new changes, looks like an older build). On my machine I get the same error, unless I install the local version via `sudo python3 setup.py install` (removing all other coala installs).

coalib/core/Bear.py
```diff
@@ -0,0 +1,501 @@
...
+    classes that are to be executed before this bear gets executed. The results
+    of these bears will then be passed inside ``self.dependency_results`` as a
+    dict. The dict will have the name of the bear as key and a list of its
+    results as values:
+
+    >>> class SomeBear(Bear): pass
+    >>> class SomeOtherBear(Bear):
+    ...     BEAR_DEPS = {SomeBear}
+    >>> SomeOtherBear.BEAR_DEPS
+    {<class 'coalib.core.Bear.SomeBear'>}
```
Asnelchristian:
```
Every bear resides in some directory which is specified by the
    source_location attribute:

    >>> class SomeBear(Bear): pass
    >>> SomeBear.source_location
    '...Bear.py'
```
this part of the docs had been removed, was it on purpose?

coalib/core/Bear.py
```diff
@@ -0,0 +1,501 @@
...
+    CAN_DETECT = set()
+    CAN_FIX = set()
+    ASCIINEMA_URL = ''
+    BEAR_DEPS = set()
+
+    @classproperty
+    def name(cls):
+        """
+        :return:
+            The name of the bear.
```
Asnelchristian: ``:return: The name of the bear`` as in the original file

coalib/core/Bear.py
```diff
@@ -0,0 +1,501 @@
...
+        """
+        return cls.__name__
+
+    @classproperty
+    def can_detect(cls):
+        """
+        :return:
+            A set that contains everything a bear can detect, including
+            information from what it can fix too.
+        """
```
Asnelchristian: same here

tests/core/BearTest.py
```diff
@@ -0,0 +1,262 @@
...
+        self.assertEqual(uut.dependency_results,
+                         {Bear1: [1, 2, 3]})
+
+        bear2 = Bear2(section, filedict)
+        uut.add_dependency_results(bear2, [4, 5, 6])
+
+        self.assertEqual(uut.dependency_results,
+                         {Bear1: [1, 2, 3], Bear2: [4, 5, 6]})
+
+        uut.add_dependency_results(bear1, [7, 8, 9])
```
Asnelchristian: do something like ``uut.add_dependency_results(bear1, [3, 8, 9])`` so that we see, that 3 is not duplicated at the same time

Asnelchristian:
ack https://github.com/coala/coala/pull/3711/commits/61d1e803a4f5a2f2201989775a026b346ca4b823

Asnelchristian:
ack https://github.com/coala/coala/pull/3711/commits/72228be2eadd09580f100ba880aa57050486f6b3

coalib/core/Bear.py
Makman2: Any reason for keeping it?
As soon as the docstyle bear is finished (I hope someday...), we switch to this format anyway^^

tests/core/BearTest.py
Makman2: it will be duplicated, results with the same content are not merged into a single one.

Makman2: shall I still add this case? :)

coalib/core/Bear.py
Asnelchristian: okay :sweat_smile:

tests/core/BearTest.py
Asnelchristian: do we want that duplication ?

Asnelchristian: i don't feel like we should allow it :sweat_smile:

Makman2: The bear shouldn't yield multiple results normally. And if he does, it shouldn't be a problem anyway, then it seems to have a reason.

Actually I wanted to use sets before, but: In my current design I allow on purpose arbitrary types as results (coala enforces to use normal `Result` objects to get something meaningful). This can be convenient for dependency results, I have the feeling `HiddenResult`s contain too much info the dependent bear does not need. So why not just giving a tuple or a different type with the exact info what's needed, not less not more :)
Arbitrary types are not always hashable (consider dicts), also the order of the results can be important in a few cases (rare, but we support it anyway).

Makman2:
reack 37beccd 9bd1d3f

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+import requests
+
+from coalib.results.Result import Result
+from coalib.settings.ConfigurationGathering import get_config_directory
+from coalib.settings.FunctionMetadata import FunctionMetadata
+from coalib.settings.Section import Section
+
+
+class Bear:
+    """
```
sils: so this documentation comment  is helpful to me. But not to somebody wanting to know what a bear is and what he has to look out for when implementing one - this should rather be a sort of tutorial for bears *or* at first link to one.

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+    ['Redundancy', 'Security', 'Syntax']
+
+    Every bear has a data directory which is unique to that particular bear:
+
+    >>> class SomeBear(Bear): pass
+    >>> class SomeOtherBear(Bear): pass
+    >>> SomeBear.data_dir == SomeOtherBear.data_dir
+    False
+
+    A bear can be dependant from other bears. ``BEAR_DEPS`` contains bear
```
sils: dependant -> dependent

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+
+        :param bear:
+            The bear instance to add results for.
+        :param dependency_results:
+            The results to add.
+        """
+        bear_type = type(bear)
+
+        if bear_type not in self._dependency_results:
+            self._dependency_results[bear_type] = list(dependency_results)
```
sils: why don't you just make it a defaultdict?

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+        else:
+            self._dependency_results[bear_type] += dependency_results
```
sils: I'm not sure I get the value provided by those two functions that just access a (default)dict

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+        {}
+        >>> dependency_bear = Bear(section, file_dict)
+        >>> bear.add_dependency_results(dependency_bear, [1, 2, 3])
+        >>> bear.dependency_results
+        {<class 'coalib.core.Bear.Bear'>: [1, 2, 3]}
+
+        :return:
+            A dictionary with bear-types as keys and their results received.
+        """
+        return self._dependency_results
```
sils: we're starting to use the attrs library a bit and I think this would be a good fit, remove those three functions which are just proxies for the dependency dict and make it an `_dependency_results = attr.ib(init=collections.defaultdict(list))`

sils: see also https://attrs.readthedocs.io/en/stable/examples.html

coalib/core/Bear.py
```diff
@@ -0,0 +1,502 @@
...
+        """
+        :return:
+            Metadata for the ``analyze`` function extracted from its signature.
+            Excludes parameter ``self``.
+        """
+        return FunctionMetadata.from_function(
+            cls.analyze,
+            omit={'self'})
+
+    # FIXME Make this a @classproperty.
```
sils: so... do it :)

coalib/core/Bear.py
Makman2: could we do this after merge? I mostly copied this from the old bear^^

coalib/core/Bear.py
Makman2: I dislike that it creates new items on access when you probably don't want them.
But I can use a defaultdict :+1:

coalib/core/Bear.py
Makman2: I just wanted to create a unified API usable from the bear. If you want I remove those functions again :+1:

coalib/core/Bear.py
Makman2: breaks behaviour with `ConsoleInteraction`. That's why I postponed it.

coalib/core/Bear.py
sils: IMO defaultdict is still way better

coalib/core/Bear.py
Makman2: doesn't this force the constructor to take in `dependency_results`?

Makman2: nvm

Makman2: `@attr.s` adds some magic to the class, like comparison operators. A bear shall not be comparable by dependency results. I think we should make a complete refactor with `attr.ib`s consistently. Let's make that after PR.

coalib/core/Bear.py
Udayan12167: We are anyway going to rewrite a lot of it I think

coalib/core/Bear.py
Udayan12167: You can choose to remove those operators.

Makman2: really let's do it after :) It's just a simple property :)

coalib/core/Bear.py
Makman2: but we have to integrate the core into the current system, and this one can be easily changed after integration -> smaller commits + less problems (imo)

Udayan12167:
looks good to me.

Udayan12167:
@Makman2 waiting for him to look at it too.

coalib/core/Bear.py
```diff
@@ -0,0 +1,458 @@
...
+        """
+        Gives the directory where the configuration file resides.
+
+        :return:
+            Directory of the config file.
+        """
+        return get_config_directory(self.section)
+
+    @classmethod
+    def download_cached_file(cls, url, filename):
```
Udayan12167: @jayvdb just added some changes to this method in another PR on the current Bear.py you might want to incorporate them. They do address some bugs.

Makman2:
Incorporated 422193a62fd70a21a94f4592b0ce61a16389c5c2

Udayan12167:
Lets get @sils approval and get this merged?

sils:
if you approve it just merge it, waiting for someone is usually just
holding up. If my comments were addressed I'm likely happy

Udayan12167:
ack 7cde8ca

Udayan12167:
@Makman2 I have approved the commit and made issues for the things required to be done post merge.

Udayan12167:
Rebase and we will merge this.

Makman2:
reack 65e45a8

Makman2:
@rultor merge

Issue #3020, opened 2016-11-18
FunctionMetadata.py: Do not hardcode the default value into a param's metadata


adtac:
Instead it'd be better have it as another field in the tuple to improve flexibility (for example, I'd like this feature because it'd help me solve https://github.com/coala/bear-docs/issues/20 cleanly rather than doing a hacky solution).

adtac:
We already have the default value in the last column, so basically remove that string concatenation (but maybe add a field to say that this is optional).

Makman2:
What do you mean exactly "hardcoded default value"?

adtac:
We add `(Optional, default value is 'True')` (or something similar) to each param's description. Instead we could have them as separate fields. See line 187 in FunctionMetadata.py

Yash17Agrawal:
@adtac can you explain little more about what do you mean on line 187 exactly? and these default value you want to add are only for line 187 or there are many such occurrences?

adtac:
I *think* that's the only place; anyway that's for you to figure out ;) it's a part of learning your way around a codebase

ap916:
As the default value is already there in the last column, would changing the `str_optional = "Optional, defaults to '{}'."` to just `"Optional"` do the trick. After removing the string concatenation, of course.  
The tests would need to be edited too.

BrateS:
Can you help me understand the code ? I want to help with this.

adtac:
@BrateS sure! I've assigned you to this issue.

* First take a look at coala.io/newcomer if you haven't already - that should help you get started a little.

* So this issue has something to do with the setting of each bear. Run `coala --show-bears --show-details`. This should give a list of all bears installed and the list of settings under each bear. Take a look at `YapfBear`, for example. The description for each setting ends with `(Optional, ...`. This is text is basically appended to the help string. What this issue asks for is to separate this into two different fields. Take a look at FunctionMetadata.py (L187) for more details.

BrateS:
@adtac Thanks a lot ! I look into it.

dracarys09:
@BrateS still working on this ?

BrateS:
Trying ,but you can unassign me if someone wants to fix it.

dracarys09:
@adtac so what this requires is to basically remove this concatenation ```doc_comment.param_dict.get(arg, cls.str_nodesc) + ' (' +
                    cls.str_optional.format(
                        defaults[i-num_non_defaults]) + ')'```

[...]

PR #3721, opened 2017-02-08
FunctionMetadata.py: Modify param's metadata

dracarys09:
Split "optional_params" description into two parts to ensure that hard coded
default values are not present. The first part corresponds to the description
of a parameter and the second part corresponds to the string "Optional".

Fixes https://github.com/coala/coala/issues/3020

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/settings/FunctionMetadata.py
+++ b/coalib/settings/FunctionMetadata.py
@@ -8,7 +8,6 @@ from coalib.settings.DocstringMetadata import DocstringMetadata
 
 class FunctionMetadata:
     str_nodesc = 'No description given.'
-    str_optional = "Optional, defaults to '{}'."
 
     @enforce_signature
     def __init__(self,
@@ -126,7 +125,7 @@ class FunctionMetadata:
 
         for param in self.optional_params:
             if param in section:
-                _, annotation, _ = self.optional_params[param]
+                _, annotation, _, _ = self.optional_params[param]
                 params[param] = self._get_param(param, section, annotation)
 
         return params
@@ -186,11 +185,10 @@ class FunctionMetadata:
                     argspec.annotations.get(arg, None))
             else:
                 optional_params[arg] = (
-                    doc_comment.param_dict.get(arg, cls.str_nodesc) + ' (' +
-                    cls.str_optional.format(
-                        defaults[i-num_non_defaults]) + ')',
+                    doc_comment.param_dict.get(arg, cls.str_nodesc),
                     argspec.annotations.get(arg, None),
-                    defaults[i-num_non_defaults])
+                    defaults[i-num_non_defaults],
+                    'Optional')
 
         return cls(name=func.__name__,
                    desc=doc_comment.desc,
--- a/tests/output/ConsoleInteractionTest.py
+++ b/tests/output/ConsoleInteractionTest.py
@@ -712,9 +712,7 @@ class ShowBearsTest(unittest.TestCase):
                              '  Needed Settings:\n'
                              '   * setting1: Required Setting.\n\n'
                              '  Optional Settings:\n'
-                             '   * setting2: Optional Setting. ('
-                             "Optional, defaults to 'None'."
-                             ')\n\n'
+                             '   * setting2: Optional Setting.\n\n'
                              '  Can detect:\n   * Formatting\n\n'
                              '  Can fix:\n   * Formatting\n\n  Path:\n   ' +
                              repr(TestBear.source_location) + '\n\n')
--- a/coalib/bearlib/abstractions/ExternalBearWrap.py
+++ b/coalib/bearlib/abstractions/ExternalBearWrap.py
@@ -83,8 +83,7 @@ def _create_wrapper(klass, options):
             if default_value is NoDefaultValue:
                 return (description, setting_type)
             else:
-                return (description + ' ' +
-                        FunctionMetadata.str_optional.format(default_value),
+                return (description,
                         setting_type, default_value)
 
         @classmethod
--- a/tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
+++ b/tests/bearlib/abstractions/external_bear_wrap_testfiles/ExternalBearWrapTest.py
@@ -99,15 +99,12 @@ class ExternalBearWrapComponentTest(unittest.TestCase):
         self.assertEqual(metadata.non_optional_params['asetting'][0],
                          FunctionMetadata.str_nodesc)
         self.assertEqual(metadata.optional_params['bsetting'][0],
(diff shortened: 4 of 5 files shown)


Asnelchristian:
this looks good to me

Asnelchristian:
ack https://github.com/coala/coala/pull/3721/commits/2bf34f8efee8a4ada9f3a3198c60ce1da526ce4d

Asnelchristian:
needs rebase

coalib/settings/FunctionMetadata.py
```diff
@@ -186,11 +185,10 @@ def from_function(cls, func, omit=frozenset()):
...
             else:
                 optional_params[arg] = (
-                    doc_comment.param_dict.get(arg, cls.str_nodesc) + ' (' +
-                    cls.str_optional.format(
-                        defaults[i-num_non_defaults]) + ')',
+                    doc_comment.param_dict.get(arg, cls.str_nodesc),
                     argspec.annotations.get(arg, None),
-                    defaults[i-num_non_defaults])
+                    defaults[i-num_non_defaults],
+                    'Optional')
```
adtac: Instead of just `Optional`, we can use the original string within brackets here?

tests/output/ConsoleInteractionTest.py
```diff
@@ -712,9 +712,7 @@ def test_show_bear_with_content(self):
                              '  Needed Settings:\n'
                              '   * setting1: Required Setting.\n\n'
                              '  Optional Settings:\n'
-                             '   * setting2: Optional Setting. ('
-                             "Optional, defaults to 'None'."
-                             ')\n\n'
+                             '   * setting2: Optional Setting.\n\n'
```
adtac: Hmm, I think we want this to be the same - the UI should have no real changes - the output for `--show-bears --show-description` should still have the `(Optional, defaults to 'whatever'.)` part. So store that string instead of just `Optional` (see the comment above), and append it just before printing. That way, we have access to both.

coalib/settings/FunctionMetadata.py
dracarys09: didn't you say that we want ```(Optional, defaults to 'whatever')``` to be split into two (1) Optional and (2) default value... and since already have default in the second last column, keeping only 'Optional' makes sense right ?

tests/output/ConsoleInteractionTest.py
dracarys09: ```(Optional, defaults to 'whatever'.)``` was coming because we were doing like 
```doc_comment.param_dict.get(arg, cls.str_nodesc) + ' (' +
                     cls.str_optional.format(
                         defaults[i-num_non_defaults]) + ')'```

Since the issue asks for this concatenation to be removed... that ```(Optional, defaults to 'whatever')``` part will not be displayed... If UI should remain same, then we have to have this concatenation. This part in the UI is coming because of that concatenation.

dracarys09: actually I was just thinking, why do we need that extra tuple with 'Optional'... since we are working with ```optional_params``` it is implicit that they are "Optional"

Or if you want to keep this 'Optional' then maybe we can just display ```<param_name> (defaults to 'something')``` instead of ```<param_name> (Optional, defaults to 'something')``` because we are displaying this info about params under the heading "Optional Parameters"... so it doesn't make sense to display "Optional" every time...

coalib/settings/FunctionMetadata.py
adtac: Or we can store just the default value alone - what I meant was, we could construct the "Optional, defaults to..." string whenever we want.

dracarys09: so that means no need for this additional 'Optional' tuple right ?

tests/output/ConsoleInteractionTest.py
adtac: The issue is more about separating the two. Changing the UI/output is a different thing. Feel free to add it as a separate commit :) (FWIW, I agree, we don't need the "Optional" qualifier for every setting.)

The reason I filed the issue is, I wanted to get the default value for each setting to display as a column in bear-docs.

Anyway, this is how I'd like to see it:

`self.optional_params["use_spaces"] = ("setting description", <class 'str'>, True)`

So we can just remove the fourth "Optional" string. But remember to reconstruct the original help text like this:

`   * tab_width: Number of spaces per indentation level. (Optional, defaults to '4'.)"`

dracarys09: @adtac What you want was already there... :worried: 
optional_params[arg] = (
                    doc_comment.param_dict.get(arg, cls.str_nodesc) + ' (' +
                    cls.str_optional.format(
                        defaults[i-num_non_defaults]) + ')',
                    argspec.annotations.get(arg, None),
                    defaults[i-num_non_defaults])

The value in tuple is ```setting description``` , second is like ```<class 'str'>``` and the third value is ```default value```

If you want that UI should remain same then the reason that we can-not remove that concatenation where we are saying ```"Optional, defaults to something``` is that we are printing this exact string using ```show_enumeration``` ```in ConsoleInteraction.py``` where we are just printing this by using ```value[0]``` for each setting and since here we concatenated this in the first value of tuple, that is why that ```Optional, defaults to something``` is coming. If we remove that concatenation then UI will change... But what you want was already there before I made changes...

sils:

sils:
This makes the tests fail and seems to need some work.

tests/output/ConsoleInteractionTest.py
zhongLxyz: @adtac Hi, I am the new assignee of the issue that this pull request is trying to solve. I read the conversation here and I am a bit confused about what the status of the issue is now. 

I understand that the goal is to keep the UI of the optional parameters output the same, as in a format like ```* tab_width: Number of spaces per indentation level. (Optional, defaults to '4'.)"```, and what we should do is somehow try to modify the ```((Optional, defaults to '4'.))``` which is concatenated to the string of the description of the parameter in FunctionMetadata.py. 

However, the part I don't understand is what is required for this modification. You mentioned that the reason you created the issue is that you wanted to get the default value for each setting to display as a column in bear-docs. But I thought the bear-docs is the cmd UI itself, and if we are displaying them in a column then the UI won't stay the same. Correct me if I am wrong for this part, but I am still trying to figure out what the difference that should be created is.

Also, if we discard the ```((Optional, defaults to '4'.))``` which is concatenated to the description of the parameter, then we need to modify the ```show_enumeration``` function in ```ConsoleInteraction.py``` just like @dracarys09 said. I think this is doable but we need to create a special case just to deal with ```optional parameters``` in ```show_enumeration```. I just want to make sure I understand the requirement before I make any attempt at it.

PR #3725, opened 2017-02-09
naming_conventions: Support kebab-case

gs0510:
Adds function to_kebabcase() to convert string to kebab-case.

Relates to https://github.com/coala/coala-bears/issues/1013

### DIFF
--- a/coalib/bearlib/naming_conventions/__init__.py
+++ b/coalib/bearlib/naming_conventions/__init__.py
@@ -130,3 +130,49 @@ def to_spacecase(string):
     return re.sub('(?<=[^\s])([A-Z])',
                   lambda match: ' ' + match.group(1),
                   string)
+
+
+def to_kebabcase(string):
+    """
+    Converts the given string to kebab-case.
+
+    >>> to_kebabcase('HelloWorld')
+    'hello-world'
+    >>> to_kebabcase('__Init__File__')
+    'init-file'
+    >>> to_kebabcase('')
+    ''
+    >>> to_kebabcase('already-kebab-case')
+    'already-kebab-case'
+    >>> to_kebabcase('   string  ')
+    'string'
+    >>> to_kebabcase('ABCde.F.G..H..IH')
+    'a-b-cde.f.g..h..i-h'
+
+    :param string: The string to convert.
+    :return:       The kebab-cased string.
+    """
+
+    string = re.sub('(\s)',
+                    lambda match: '-',
+                    string)
+    string = re.sub('(\_)',
+                    lambda match: '-',
+                    string)
+    string = re.sub('^(-*)([^-])',
+                    lambda match: match.group(1) + match.group(2).lower(),
+                    string)
+    string = re.sub('(\w*)([.]+)([A-Z])',
+                    lambda match: (match.group(1) + match.group(2) +
+                                   match.group(3).lower()),
+                    string)
+    string = re.sub('(?<=[^-])-+([^-])',
+                    lambda match: '-' + match.group(1).lower(),
+                    string)
+    string = re.sub('[A-Z]',
+                    lambda match: '-' + match.group(0).lower(),
+                    string)
+    string = re.sub(r'^-+|-+$',
+                    lambda match: ''*len(match.group()),
+                    string)
+    return re.sub('\-+', '-', string)


coalib/bearlib/naming_conventions/__init__.py
```diff
@@ -130,3 +130,45 @@ def to_spacecase(string):
...
                   lambda match: ' ' + match.group(1),
                   string)
+
+
+def to_kebabcase(string):
+    """
+    Converts the given string to kebab-case.
+
+    >>> to_kebabcase('HelloWorld')
+    'hello-world'
```
RohanVB: Wouldn't this be `-hello-world` ?
See: https://www.npmjs.com/package/kebab-case

coalib/bearlib/naming_conventions/__init__.py
```diff
@@ -130,3 +130,45 @@ def to_spacecase(string):
...
+    >>> to_kebabcase('__Init__File__')
+    '--init-file--'
+    >>> to_kebabcase('')
+    ''
+    >>> to_kebabcase('already-kebab-case')
+    'already-kebab-case'
+    >>> to_kebabcase('   string  ')
+    '---string--'
+    >>> to_kebabcase('ABCde.F.G..H..IH')
+    'a-b-cde.f.g..h..i-h'
```
RohanVB: `-` after caps, inclusive of the beginning of the string?
See: https://www.npmjs.com/package/kebab-case

coalib/bearlib/naming_conventions/__init__.py
gs0510: There's no uniform standard I could find for kebab-case. Different github repos that convert text to kebab-case do things differently. Can I assume the npm kebab-case converter as the standard going forward?

coalib/bearlib/naming_conventions/__init__.py
gs0510: IMO a - (hyphen) at the start of the string should not really be there. And since there is no standard for kebab case conversion, we can take this liberty I guess. If we were to parallel kebab case to snake case, In snake case also we have all lower case characters, it's just that the hyphen is replaced with the an underscore.

coalib/bearlib/naming_conventions/__init__.py
RohanVB: In that case I think there should be a standard that we use for coala in particular, and since npm does look to have them all down, maybe we should use it for the other naming conventions aswell?
Open for discussion :)

coalib/bearlib/naming_conventions/__init__.py
RohanVB: same as above.

coalib/bearlib/naming_conventions/__init__.py
gs0510: Um, I think it does not make sense to have a - preceding the leading Uppercase character if we were to think about this from a point of view of snake case. Let's say if the example was '   String  ', if we were to follow npm convention, it'll be ----string-- rather than ---string-- . What I feel is it's just really unnecessary, and not really serving any purpose.

RohanVB: ` String ` would be `--string-` rather than `-string-` there which looks alright. I still feel like we should follow a certain uniform standard which npm would do well for across all the naming conventions, it'd make it easier for conversions between them two conventions as well if ever required.
This may as well be a separate issue if you feel it's unnecessary.

gs0510: Okay, to finally answer my initial question, I can assume what npm module does as the standard? :sweat_smile:

RohanVB: That is my opinion, yeah. It'd require the other conventions that don't follow it to do so aswell for the sake of uniformity. With separate PRs.

coalib/bearlib/naming_conventions/__init__.py
Siddhartha1234: I agree with @gs0510 on the above point . I've used it in lisp and lisp's kebab case does not begin with a hyphen. But instead of multiple hyphens , its preferred to strip white-lines from both ends and it contains blocks of lower case words separated by single hyphen. It's better if you implement that.

coalib/bearlib/naming_conventions/__init__.py
gs0510: As @Siddhartha1234 I followed the haskell conventions. To follow the npm conventions for kebab-case, I think we can file another issue. Let me know how the changes look. @Siddhartha1234 @RohanVB

gs0510: As @Siddhartha1234 I followed the haskell conventions. To follow the npm conventions for kebab-case, I think we can file another issue. Let me know how the changes look. @Siddhartha1234 @RohanVB

Siddhartha1234: @gs0510 It looks fine to me  now :) .

RohanVB:
ack 39e7dc209210fc634b0513a4dcaeef6ab718f877

sils:
instead of `__init__.py` rather do `naming_conventions` which is way more helpful! `naming_conventions: Add kebab-case` would be short and gives more info IMO

sils:
why is this only related to and not fixing it?

sils:
code looks good, let's get the commit message a bit brushed and off it goes?

gs0510:
@sils I'll modify the commit message :sweat_smile: It is only relating to the issue because after adding this function here, I'll have to make changes to the FileNameBear to support kebab-case naming.

sils:
ack https://github.com/coala/coala/pull/3725/commits/9b552d8578f795d69b0f68ab9259bfdb91920c48 but needs a rebase

sils:
reack https://github.com/coala/coala/pull/3725/commits/88d558b6a84f4e4ec4477fd9e809932c2b21e5fe

sils:
@rultor merge

sils:
@rultor stop

sils:
what's going on here?

gs0510:
I had rebased against the upstream instead of origin and in between these two rebases the rultor tried to merge my branch. @sils

Makman2:
@gs0510 but why do you close this PR? rebase it to the correct master and repush :)

PR #3739, opened 2017-02-14
Add output format `unidiff` to Linter

satwikkansal:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -42,7 +42,8 @@ def _prepare_options(options, bear_class):
     if not options['use_stdout'] and not options['use_stderr']:
         raise ValueError('No output streams provided at all.')
 
-    if options['output_format'] == 'corrected':
+    if (options['output_format'] == 'corrected' or
+            options['output_format'] == 'unified-diff'):
         if (
                 'diff_severity' in options and
                 options['diff_severity'] not in RESULT_SEVERITY.reverse):
@@ -404,6 +405,42 @@ def _create_linter(klass, options):
                 result_message,
                 diff_distance)
 
+        def process_output_unified_diff(self,
+                                        output,
+                                        filename,
+                                        file,
+                                        diff_severity=RESULT_SEVERITY.NORMAL,
+                                        result_message='Inconsistency found.',
+                                        diff_distance=1):
+            """
+            Processes the executable's output as a unified diff.
+
+            :param output:
+                The output of the program as a string containing the
+                unified diff for correction.
+            :param filename:
+                The filename of the file currently being corrected.
+            :param file:
+                The contents of the file currently being corrected.
+            :param diff_severity:
+                The severity to use for generating results.
+            :param result_message:
+                The message-string to use for generating results.
+            :param diff_distance:
+                Number of unchanged lines that are allowed in between two
+                changed lines so they get yielded as one diff. If a negative
+                distance is given, every change will be yielded as an own diff,
+                even if they are right beneath each other.
+            :return:
+                An iterator returning results containing patches for the
+                file to correct.
+            """
+            return self.process_diff(Diff.from_unified_diff(output, file),
+                                     filename,
+                                     diff_severity,
+                                     result_message,
+                                     diff_distance)
+
         def process_output_regex(
                 self, output, filename, file, output_regex,
                 severity_map=MappingProxyType({
@@ -496,6 +533,16 @@ def _create_linter(klass, options):
 
                 _processing_function = partialmethod(
                     process_output_corrected, **_process_output_args)
+
+            elif options['output_format'] == 'unified-diff':
+                _process_output_args = {
+                    key: options[key]
+                    for key in ('result_message', 'diff_severity',
+                                'diff_distance')
+                    if key in options}
+
+                _processing_function = partialmethod(
+                    process_output_unified_diff, **_process_output_args)
 
             else:
                 assert options['output_format'] == 'regex'
@@ -837,6 +884,8 @@ def linter(executable: str,
           ``output_regex``.
         - ``'corrected'``: The output is the corrected of the given file. Diffs
           are then generated to supply patches for results.
+        - ``'unified_diff'``: The output is the unified diff of the corrections.
+          Patches are then supplied for results using this output.
 
         Passing something else raises a ``ValueError``.
     :param output_regex:
@@ -878,21 +927,21 @@ def linter(executable: str,
         used inside ``output_regex`` and this parameter is given.
     :param diff_severity:
         The severity to use for all results if ``output_format`` is
-        ``'corrected'``. By default this value is
+        ``'corrected'`` or ``'unified_diff'``. By default this value is
         ``coalib.results.RESULT_SEVERITY.NORMAL``. The given value needs to be
         defined inside ``coalib.results.RESULT_SEVERITY``.
     :param result_message:
         The message-string to use for all results. Can be used only together
-        with ``corrected`` or ``regex`` output format. When using
-        ``corrected``, the default value is ``"Inconsistency found."``, while
-        for ``regex`` this static message is disabled and the message matched
-        by ``output_regex`` is used instead.
+        with ``corrected`` or ``unified_diff`` or ``regex`` output format.
+        When using ``corrected`` or ``unified_diff``, the default value is
+        ``"Inconsistency found."``, while for ``regex`` this static message is
+        disabled and the message matched by ``output_regex`` is used instead.
     :param diff_distance:
         Number of unchanged lines that are allowed in between two changed lines
-        so they get yielded as one diff if ``corrected`` output-format is
-        given. If a negative distance is given, every change will be yielded as
-        an own diff, even if they are right beneath each other. By default this
-        value is ``1``.
+        so they get yielded as one diff if ``corrected`` or ``unified_diff``
+        output-format is given. If a negative distance is given, every change
+        will be yielded as an own diff, even if they are right beneath each
+        other. By default this value is ``1``.
     :raises ValueError:
         Raised when invalid options are supplied.
     :raises TypeError:
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -330,6 +330,42 @@ def _create_linter(klass, options):
                 result_params['affected_code'] = (range,)
             return Result(**result_params)
 
+        def process_diff(self,
+                         diff,
+                         filename,
+                         diff_severity,
+                         result_message,
+                         diff_distance):
+            """
+            Processes the given ``coalib.results.Diff`` object and yields
+            correction results.
+
+            :param diff:
+                An instance of ``coalib.results.Diff`` object containing
+                differences of the file named ``filename``.
+            :param filename:
+                The name of the file currently being corrected.
+            :param diff_severity:
+                The severity to use for generating results.
+            :param result_message:
+                The message to use for generating results.
+            :param diff_distance:
+                Number of unchanged lines that are allowed in between two
+                changed lines so they get yielded as one diff. If a negative
+                distance is given, every change will be yielded as an own diff,
+                even if they are right beneath each other.
+            :return:
+                An iterator returning results containing patches for the
+                file to correct.
+            """
+            for splitted_diff in diff.split_diff(distance=diff_distance):
+                yield Result(self,
+                             result_message,
+                             affected_code=splitted_diff.affected_code(
+                                 filename),
+                             diffs={filename: splitted_diff},
+                             severity=diff_severity)
+
         def process_output_corrected(self,
                                      output,
                                      filename,
@@ -359,15 +395,14 @@ def _create_linter(klass, options):
                 An iterator returning results containing patches for the
                 file to correct.
             """
-            for diff in Diff.from_string_arrays(
-                file,
-                output.splitlines(keepends=True)).split_diff(
-                    distance=diff_distance):
-                yield Result(self,
-                             result_message,
-                             affected_code=diff.affected_code(filename),
-                             diffs={filename: diff},
-                             severity=diff_severity)
+            return self.process_diff(
+                Diff.from_string_arrays(
+                    file,
+                    output.splitlines(keepends=True)),
+                filename,
+                diff_severity,
+                result_message,
+                diff_distance)
 
         def process_output_regex(
                 self, output, filename, file, output_regex,
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -2,6 +2,8 @@ import copy
 import difflib
 import logging
 
+from unidiff import PatchSet
+
 from coalib.results.LineDiff import LineDiff, ConflictError
 from coalib.results.SourceRange import SourceRange
 from coalib.results.TextRange import TextRange
@@ -68,6 +70,90 @@ class Diff:
                         result.delete_line(index)
 
         return result
+
+    @classmethod
+    def from_unified_diff(cls, unified_diff, original_file):
+        """
+        Creates a ``Diff`` object from given unified diff.
+
+        If the provided unified diff does not contain any patch,
+        the ``Diff`` object initialized from the original file is
+        returned.
+
+        :param unified_diff:  Unified diff string.
+        :param original_file: The contents of the original file
+                              (line-splitted).
+        :raises RuntimeError: Raised when the context lines or the
+                              lines to be removed do not match in
+                              the original file and the unified diff.
+        """
+        patch_set = PatchSet(unified_diff.splitlines())
+
+        diff = Diff(original_file)
+
+        if not patch_set:
+            return diff
+
+        # FIXME Handle patches consisting of changes in more than one file
+        file = patch_set[0]
+
+        for hunk in file:
+            file_line = hunk.source_start
+            hunk_iterator = iter(hunk)
+
+            try:
+                while True:
+                    line = next(hunk_iterator)
+                    source_code = str(line)[1:]
+                    if line.is_added:
+                        add_set = []
+                        # As ``Diff`` does not allow line additions to a
+                        # position more than one time, add all the
+                        # consecutive '+' lines at once.
+                        try:
+                            while line.is_added:
+                                add_set.append(source_code)
+                                line = next(hunk_iterator)
+                                source_code = str(line)[1:]
+
+                            diff.add_lines(file_line-1, add_set)
+
+                        except StopIteration:
+                            diff.add_lines(file_line-1, add_set)
+                            break
+
+                    original_line = original_file[file_line-1].rstrip('\n')
+
+                    if line.is_removed:
+                        if source_code != original_line:
+                            raise RuntimeError(
+                                'The line to delete does not match with '
+                                'the line in the original file. '
+                                'Line to delete: {!r}, '
+                                'Original line #{!r}: {!r}'.format(
+                                    source_code,
+                                    file_line,
+                                    original_line)
+                                )
+                        diff.delete_line(file_line)
+
+                    else:
+                        if source_code != original_line:
+                            raise RuntimeError(
+                                'Context lines do not match. '
+                                'Line from unified diff: {!r}, '
+                                'Original line #{!r}: {!r}'.format(
+                                    source_code,
+                                    file_line,
+                                    original_line)
+                                )
+
+                    file_line += 1
+
+            except StopIteration:
+                pass
+
+        return diff
 
     @classmethod
     def from_clang_fixit(cls, fixit, file):
--- a/docs/Developers/Writing_Linter_Bears.rst
+++ b/docs/Developers/Writing_Linter_Bears.rst
@@ -97,6 +97,8 @@ handling different output formats:
 - ``regex``: This parses issue messages yielded by the underlying executable.
 - ``corrected``: Auto-generates results from a fixed/corrected file provided by
   the tool.
+- ``unified-diff``: This auto-generates results from a unified-diff output
+  provided by the executable.
 
 In this tutorial we are going to use the ``regex`` output format. But before we
 continue with modifying our bear, we need to figure out how exactly output from
@@ -228,10 +230,10 @@ Normally, providing a severity-map is not needed, as coala has a default
 severity-map which recognizes many common words used for severities. Check out
 the API documentation for keywords supported!
 
-Suggest Corrections Using the ``corrected`` Output Format
----------------------------------------------------------
+Suggest Corrections Using the ``corrected`` and ``unified-diff`` Output Formats
+-------------------------------------------------------------------------------
 
-This output format is very simple to use and doesn't require further setup from
+These output formats are very simple to use and don't require further setup from
 your side inside the bear:
 
 ::
@@ -239,8 +241,16 @@ your side inside the bear:
     @linter(...
             output_format='corrected')
 
-If your underlying tool generates a corrected file, the class automatically
-generates patches for the changes made and yields results accordingly.
+or
+
+::
+
+    @linter(...
+            output_format='unified-diff')
+
+If your underlying tool generates a corrected file or a unified-diff of the
+corrections, the class automatically generates patches for the changes made and
+yields results accordingly.
 
 Adding Settings to our Bear
 ---------------------------
--- a/tests/bearlib/abstractions/LinterTest.py
+++ b/tests/bearlib/abstractions/LinterTest.py
@@ -91,6 +91,20 @@ class LinterComponentTest(unittest.TestCase):
 
         with self.assertRaisesRegex(ValueError,
                                     'Invalid keyword arguments provided: '
+                                    "'output_regex'"):
+            linter('some-executable',
+                   output_format='unified-diff',
+                   output_regex='.*')(self.EmptyTestLinter)
+
+        with self.assertRaisesRegex(ValueError,
+                                    'Invalid keyword arguments provided: '
+                                    "'severity_map'"):
+            linter('some-executable',
+                   output_format='unified-diff',
+                   severity_map={})(self.EmptyTestLinter)
+
+        with self.assertRaisesRegex(ValueError,
+                                    'Invalid keyword arguments provided: '
                                     "'prerequisite_check_fail_message'"):
             linter('some-executable',
                    prerequisite_check_fail_message='some_message'
@@ -200,6 +214,18 @@ class LinterComponentTest(unittest.TestCase):
             linter('some-executable',
                    output_format='corrected',
                    diff_severity=999888777)(self.EmptyTestLinter)
+
+        with self.assertRaisesRegex(TypeError, self.PARAM_TYPE_ERROR_RE):
+            linter('some-executable',
+                   output_format='unified-diff',
+                   result_message=list())(self.EmptyTestLinter)
+
+        with self.assertRaises(TypeError) as cm:
+            linter('some-executable',
+                   output_format='unified-diff',
+                   diff_severity=123456789)(self.EmptyTestLinter)
+        self.assertEqual(str(cm.exception),
+                         'Invalid value for `diff_severity`: 123456789')
 
         with self.assertRaisesRegex(TypeError, self.PARAM_TYPE_ERROR_RE):
             linter('some-executable',
@@ -323,6 +349,107 @@ class LinterComponentTest(unittest.TestCase):
                (self.section, None))
 
         results = list(uut.process_output(fixed_string,
+                                          'some-file.c',
+                                          original))
+        self.assertEqual(len(results), 2)
+
+    def test_process_output_unified_diff_simple_modifications(self):
+        uut = (linter(sys.executable, output_format='unified-diff')
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        original = ['void main()  {', 'return 09;', '}']
+
+        diff = ['--- a/some-file.c',
+                '+++ b/some-file.c',
+                '@@ -1,3 +1,4 @@',
+                '-void main()  {',
+                '-return 09;',
+                '+void main()'
+                '+{',
+                '+       return 9;',
+                ' }']
+
+        diff_string = '\n'.join(diff)
+
+        results = list(uut.process_output(diff_string,
+                                          'some-file.c',
+                                          original))
+
+        diffs = list(Diff.from_unified_diff(diff_string, original).split_diff())
+
+        expected = [Result.from_values(uut,
+                                       'Inconsistency found.',
+                                       'some-file.c',
+                                       1, None, 2, None,
+                                       RESULT_SEVERITY.NORMAL,
+                                       diffs={'some-file.c': diffs[0]})]
+
+        self.assertEqual(results, expected)
+
+        uut = (linter(sys.executable,
+                      output_format='unified-diff',
+                      diff_distance=-1)
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        results = list(uut.process_output(diff_string,
+                                          'some-file.c',
+                                          original))
+        self.assertEqual(len(results), 2)
+
+    def test_process_output_unified_diff_incomplete_hunk(self):
+        uut = (linter(sys.executable, output_format='unified-diff')
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        original = ['void main()  {',
+                    '// This comment is missing',
+                    '// in the unified diff',
+                    'return 09;',
+                    '}']
+
+        diff = ['--- a/some-file.c',
+                '+++ b/some-file.c',
+                '@@ -1,1 +1,2 @@',
+                '-void main()  {',
+                '+void main()',
+                '+{',
+                '@@ -4,2 +5,2 @@',
+                '-return 09;',
+                '+       return 9;',
+                ' }']
+
+        diff_string = '\n'.join(diff)
+
+        results = list(uut.process_output(diff_string,
+                                          'some-file.c',
+                                          original))
+
+        diffs = list(Diff.from_unified_diff(diff_string, original).split_diff())
+
+        expected = [Result.from_values(uut,
+                                       'Inconsistency found.',
+                                       'some-file.c',
+                                       1, None, 1, None,
+                                       RESULT_SEVERITY.NORMAL,
+                                       diffs={'some-file.c': diffs[0]}),
+                    Result.from_values(uut,
+                                       'Inconsistency found.',
+                                       'some-file.c',
+                                       4, None, 4, None,
+                                       RESULT_SEVERITY.NORMAL,
+                                       diffs={'some-file.c': diffs[1]})]
+
+        self.assertEqual(results, expected)
+
+        uut = (linter(sys.executable,
+                      output_format='unified-diff',
+                      diff_distance=-1)
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        results = list(uut.process_output(diff_string,
                                           'some-file.c',
                                           original))
         self.assertEqual(len(results), 2)
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -3,6 +3,8 @@ import logging
 import unittest
 import unittest.case
 
+from unidiff.errors import UnidiffParseError
+
 from coalib.output.JSONEncoder import create_json_encoder
 from coalib.results.Diff import ConflictError, Diff, SourceRange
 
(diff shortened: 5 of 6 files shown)


coalib/bearlib/abstractions/Linter.py
```diff
@@ -53,6 +53,17 @@ def _prepare_options(options, bear_class):
             assert_right_type(options['diff_distance'], int, 'diff_distance')
 
         allowed_options |= {'diff_severity', 'result_message', 'diff_distance'}
+    elif options['output_format'] == 'unidiff':
+        if (
+                'diff_severity' in options and
+                options['diff_severity'] not in RESULT_SEVERITY.reverse):
```
impmihai: Move the first condition on the same line with if

coalib/results/Diff.py
```diff
@@ -70,6 +71,23 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unidiff(cls, diff_array):
+        """
+        Creates a Diff object from given unified diff.
+        :param diff_array: list of lines of unified diff.
```
impmihai: You should leave a new line after the function description

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -93,6 +93,13 @@ def test_decorator_invalid_parameters(self):
 
         with self.assertRaises(ValueError) as cm:
             linter('some-executable',
+                   output_format='unidiff',
+                   diff_distance=1)(self.EmptyTestLinter)
```
impmihai: `)(self.EmptyTestLinter)` This would look better on a new line.

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -204,6 +211,18 @@ def test_decorator_invalid_parameter_types(self):
 
         with self.assertRaisesRegex(TypeError, self.PARAM_TYPE_ERROR_RE):
             linter('some-executable',
+                   output_format='unidiff',
+                   result_message=list())(self.EmptyTestLinter)
```
impmihai: `)(self.EmptyTestLinter)` on a new line

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -204,6 +211,18 @@ def test_decorator_invalid_parameter_types(self):
...
+
+        with self.assertRaises(TypeError) as cm:
+            linter('some-executable',
+                   output_format='unidiff',
+                   diff_severity=123456789)(self.EmptyTestLinter)
```
impmihai: `)(self.EmptyTestLinter)` on a new line

tests/bearlib/abstractions/LinterTest.py
impmihai: just my opinion, i see now that they are put in both ways in the code, sorry.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -368,6 +379,37 @@ def process_output_corrected(self,
                              diffs={filename: diff},
                              severity=diff_severity)
 
+        def process_output_unidiff(self,
+                                   output,
+                                   filename,
+                                   file,
+                                   diff_severity=RESULT_SEVERITY.NORMAL,
+                                   result_message='Inconsistency found.'):
```
aptrishu: Don't we need diff-distance ?
or showing whole in a single patch would be better ?

coalib/results/Diff.py
```diff
@@ -70,6 +71,23 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+        """
+        source_array = []
+        target_array = []
+        patch_set = unidiff.PatchSet(diff_array)
+        for patched_file in patch_set:
+            for hunk in patched_file:
+                for line in hunk.source_lines():
+                    source_array.append(str(line)[1:])
```
aptrishu: line would already be in string. no?

satwikkansal: No, line is an object of `Line` class that `Hunk` class uses in `unidiff` module

coalib/bearlib/abstractions/Linter.py
satwikkansal: I tried to match it with the style in the file

coalib/bearlib/abstractions/Linter.py
```diff
@@ -53,6 +53,17 @@ def _prepare_options(options, bear_class):
             assert_right_type(options['diff_distance'], int, 'diff_distance')
 
         allowed_options |= {'diff_severity', 'result_message', 'diff_distance'}
+    elif options['output_format'] == 'unidiff':
```
Makman2: either call it `diff` or `unified-diff` (I think 2nd is better)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -53,6 +53,17 @@ def _prepare_options(options, bear_class):
...
+        if (
+                'diff_severity' in options and
+                options['diff_severity'] not in RESULT_SEVERITY.reverse):
+            raise TypeError('Invalid value for `diff_severity`: ' +
+                            repr(options['diff_severity']))
+
+        if 'result_message' in options:
+            assert_right_type(options['result_message'], str, 'result_message')
+
+        allowed_options |= {'diff_severity', 'result_message'}
```
Makman2: the code is mostly duplicated with the if-stmt before, could you try to merge those?

coalib/bearlib/abstractions/Linter.py
Makman2: good question actually, we could still break the received diff into chunks...

We already had the discussion about meaningful diff-distances for the `corrected` output format, and this is highly non-trivial. Best would be if the program itself emits several patches addressing separate issues, though most tools don't do so (actually I haven't seen a single one).

For simplicity I can accept no diff-distance at all for this PR and emitting the whole patch file as a result. But we need an issue to discuss the introduction of `diff-distance`.

Makman2: Actually it's possible to disable diff-splitting by setting `diff-distance` to `0` or `-1`. So yeah let's introduce `diff-distance` :+1:

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -328,6 +347,30 @@ def test_process_output_corrected(self):
...
         self.assertEqual(len(results), 2)
 
+    def test_process_output_unidiff(self):
+        uut = (linter(sys.executable, output_format='unidiff')
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        diff = ['--- a/some-file.c', '+++ b/some-file.c', '@@ -1,3 +1,4 @@',
+                '-void main() {', '-       return 09;', '+void main()',
+                '+{', '+       return 9;', ' }']
```
Makman2: could you write it like this? :D
```
diff = ['line 1',
        'line2',
        'line3',
        ...]
```
then it's easier for a human to read the patch :)

satwikkansal: Sure :)

Makman2:
Ah just remembering: Could you maybe add another commit updating the documentation about this new format? :)

Makman2:
If you don't want to, it's okay, I'll file an issue after merge for that :+1:

satwikkansal:
I will update the documentation as well :smile:

coalib/bearlib/abstractions/Linter.py
```diff
@@ -39,7 +39,8 @@ def _prepare_options(options, bear_class):
     if not options['use_stdout'] and not options['use_stderr']:
         raise ValueError('No output streams provided at all.')
 
-    if options['output_format'] == 'corrected':
+    if options['output_format'] == 'corrected' or \
```
Udayan12167: Put it in a bracket and avoid using `\`.

satwikkansal: okay :)

coalib/results/Diff.py
```diff
@@ -70,6 +71,24 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unidiff(cls, diff_array):
+        """
+        Creates a Diff object from given unified diff.
```
Makman2: Enclose `Diff` in double backticks :)

coalib/results/Diff.py
```diff
@@ -70,6 +71,24 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+        :param diff_array: list of lines of unified diff.
```
Makman2: Maybe we should support just a string with the complete unified diff?

tests/results/DiffTest.py
```diff
@@ -206,6 +206,52 @@ def test_from_string_arrays(self):
...
+        self.uut = Diff.from_unidiff(diff)
+        self.assertEqual(self.uut.original, source)
+        self.assertEqual(self.uut.modified, target)
+
+        source = ['first', 'second']
+        target = ['first_changed', 'second_changed']
+        diff = ['--- a/testfile',
+                '+++ b/testfile',
+                '@@ -1,2 +1,2 @@',
+                '-first', '-second', '+first_changed',
```
Makman2: would be nice if you wrap those too ;)

tests/results/DiffTest.py
```diff
@@ -206,6 +206,52 @@ def test_from_string_arrays(self):
...
+        target = ['two lines']
+        diff = ['--- a/testfile',
+                '+++ b/testfile',
+                '@@ -1,2 +1 @@',
+                ' two lines',
+                '-to be removed']
+        self.uut = Diff.from_unidiff(diff)
+        self.assertEqual(self.uut.original, source)
+        self.assertEqual(self.uut.modified, target)
+
```
Makman2: could you add a testcase with two different position indicators?
```
@@ -1,2 +1 @@
 two lines
-to be removed
@@ -1,10 +10 @@
...
```

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -328,6 +347,47 @@ def test_process_output_corrected(self):
...
+               (self.section, None))
+
+        diff = ['--- a/some-file.c',
+                '+++ b/some-file.c',
+                '@@ -1,3 +1,4 @@',
+                '-void main() {',
+                '-       return 09;',
+                '+void main()',
+                '+{',
+                '+       return 9;', ' }']
```
Makman2: please wrap ` }` onto next line :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -39,7 +39,8 @@ def _prepare_options(options, bear_class):
     if not options['use_stdout'] and not options['use_stderr']:
```
Makman2: commit message:
maybe:
```
Linter.py: Add unified-diff ouput format
```

Makman2:
ack 8cbf4ac

coalib/results/Diff.py
satwikkansal: So for test cases, shall I leave them as arrays (and do join) or make them strings as well.

tests/results/DiffTest.py
satwikkansal: Sorry I forgot to commit few changes before pushing :( will fix them

coalib/results/Diff.py
Makman2: make them strings as well, breaking manually after `\n` :)

satwikkansal: Maybe we should keep them as arrays just for the sake of consistency and readability because rest of the tests define arrays (even in `LinterTest.py`), just a suggestion though :)

Makman2: hm yeah we often do that. alright 👍

satwikkansal:
@Makman2 I hope it's ready to be merged

coalib/bearlib/abstractions/Linter.py
```diff
@@ -368,6 +369,44 @@ def process_output_corrected(self,
...
+                Number of unchanged lines that are allowed in between two
+                changed lines so they get yielded as one diff. If a negative
+                distance is given, every change will be yielded as an own diff,
+                even if they are right beneath each other.
+            :return:
+                An iterator returning results containing patches for the
+                file to correct.
+            """
+            for diff in Diff.from_unidiff(
+                    output).split_diff(
```
Makman2: yes please :3

coalib/results/Diff.py
```diff
@@ -70,6 +71,25 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unidiff(cls, unified_diff):
+        """
+        Creates a `Diff` object from given unified diff.
+
+        :param unified_diff: unified diff string.
```
Makman2: Upper case letter after `:param xyz:`

coalib/results/Diff.py
```diff
@@ -1,6 +1,7 @@
 import copy
 import difflib
 import logging
+import unidiff
```
meetmangukiya: Leave a line in between, third-party package?

coalib/results/Diff.py
meetmangukiya: I don't understand how this works.. 
`Diff.from_string_arrays` expect two complete files(original and modified), one never gets a complete file in a unified diff, it is just the part that is changed. :confused:  ?

meetmangukiya:
unack 4ea5771

coalib/results/Diff.py
satwikkansal: > one never gets a complete file in a unified diff

Not necessarily, I guess it depends on the unified diff itself. But why do we need complete files to generate the `Diff` object?

satwikkansal: > it is just the part that is changed?

Yes,  `unidiff` generates patch files out of the unified diff we pass into it.

satwikkansal:
@Makman2 Please review :)

coalib/results/Diff.py
Makman2: hm but @meetmangukiya is right, I don't think the coala core is able to handle that / does not propose valid patches then for a file when we use that for linter (coala uses `Diff.modified`)^^ Maybe you should take in the original file as an argument?

coalib/bearlib/abstractions/Linter.py
```diff
@@ -368,6 +369,43 @@ def process_output_corrected(self,
...
+                An iterator returning results containing patches for the
+                file to correct.
+            """
+            for diff in Diff.from_unidiff(
+                    output).split_diff(distance=diff_distance):
+                yield Result(self,
+                             result_message,
+                             affected_code=diff.affected_code(filename),
+                             diffs={filename: diff},
+                             severity=diff_severity)
```
Makman2: I got an awesome idea to avoid code duplication:

Create a `process_diff` function, which takes in: `filename`, `diff_severity`, `result_message` and `diff_distance`. It will contain nearly the same code, but:
```
for splitted_diff in diff.split_diff(distance=diff_distance):
    yield Result(...)
```
and `diff` is another parameter.

Then you can refactor the current `process_output_corrected` and `process_output_unidiff`, to just call `process_diff`:
```python
return process_diff(Diff.from_string_arrays(...))
# and
return process_diff(Diff.from_unidiff(...))
```

coalib/results/Diff.py
meetmangukiya: > Not necessarily, I guess it depends on the unified diff itself. But why do we need complete files to generate the Diff object?

Because thats how it  is implemented.. It takes in the whole file in `self._file` and stores the modified one in `self.modified`, and the file is modified with calls to `delete_lines`, `modify_lines`, `change_line` mutators AFAIU. You have to atleast generate the modified file completely from the diff.. or come up with better implementation ideas IMO

coalib/bearlib/abstractions/Linter.py
```diff
@@ -368,6 +369,43 @@ def process_output_corrected(self,
                              diffs={filename: diff},
                              severity=diff_severity)
 
+        def process_output_unidiff(self,
```
Makman2: I would call the function `process_output_unified_diff`, I think that makes it easier for the user to find that function when he writes a custom `process_output` function^^

coalib/results/Diff.py
```diff
@@ -70,6 +72,25 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unidiff(cls, unified_diff):
```
Makman2: though the module name is `unidiff`, I would call it `from_unified_diff`. But up to you^^

docs/Developers/Writing_Linter_Bears.rst
```diff
@@ -90,6 +90,8 @@ handling different output formats:
 - ``regex``: This parses issue messages yielded by the underlying executable.
```
Makman2: commit message: `Writing_Linter_Bears: Add unified-diff` or `Writing_Linter_Bears: Add unified-diff format`:)

coalib/bearlib/abstractions/Linter.py
Makman2: I think the best is you prepend a commit to this one which does the refactor to `process_diff`, and this commit then introduces `unified-diff` format based on it :+1:

Makman2: Then it's easier to review what is refactor and what is relevant to `unified-diff` :)

Makman2:
Also: You have to update inline-documentation properly, like:
```text
 :param result_message: The message-string to use for all results. Can be used only together
        with ``corrected`` or ``regex`` output format. When using
        ``corrected``, the default value is ``"Inconsistency found."``, while
        for ``regex`` this static message is disabled and the message matched
        by ``output_regex`` is used instead.
```

coalib/bearlib/abstractions/Linter.py
satwikkansal: @Makman2 Do I need to give inline-documentation for the `filename`, `diff_severity`, `result_message` and `diff_distance` params in all three functions (`process_output_corrected`, `process_output_unidiff` and `process_diff`)

satwikkansal: I don't know if documentation duplication is considered or not :P

Makman2: in fact duplication is bad, but no documentation at all is worse :/

so yes please document them everywhere (as all of them are potential API), you can freely copy docs like needed 👍

satwikkansal:
@Makman2 review please :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -329,6 +329,44 @@ def _convert_output_regex_match_to_result(self,
...
+                         diff_severity,
+                         result_message,
+                         diff_distance):
+            """
+            Processes the given ``coalib.results.Diff`` object and yields
+            correction results.
+
+            :param diff:
+                An instance of ``coalib.results.Diff`` object representing
+                difference of the file named ``filename``.
```
Makman2: representing difference -> containing differences

coalib/bearlib/abstractions/Linter.py
```diff
@@ -329,6 +329,44 @@ def _convert_output_regex_match_to_result(self,
...
+            :param filename:
+                The name of the file currently being corrected.
+            :param diff_severity:
+                The severity to use for generating results, needs to be
+                defined inside ``coalib.results.RESULT_SEVERITY``.
+            :param result_message:
+                    The message-string to be used for all the generated
+                    results.
```
Makman2: indentation off.

coalib/bearlib/abstractions/Linter.py
```diff
@@ -340,15 +378,17 @@ def process_output_corrected(self,
...
                 The filename of the file currently being corrected.
             :param file:
                 The contents of the file currently being corrected.
             :param diff_severity:
-                The severity to use for generating results.
+                The severity to use for generating results, needs to be
+                defined inside ``coalib.results.RESULT_SEVERITY``.
             :param result_message:
-                The message to use for generating results.
+                The message-string to use for generating results.
```
Makman2: 1. different commit as unrelated change
2. actually I'm not sure this change is improving docs :/

coalib/bearlib/abstractions/Linter.py
```diff
@@ -358,15 +398,13 @@ def process_output_corrected(self,
...
-                file,
-                output.splitlines(keepends=True)).split_diff(
-                    distance=diff_distance):
-                yield Result(self,
-                             result_message,
-                             affected_code=diff.affected_code(filename),
-                             diffs={filename: diff},
-                             severity=diff_severity)
+            diff = Diff.from_string_arrays(file,
+                                           output.splitlines(keepends=True))
```
Makman2: up to you: I would inline this value into the call below^^

coalib/results/Diff.py
```diff
@@ -70,6 +72,65 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unified_diff(cls, unified_diff, original_file):
+        """
+        Creates a `Diff` object from given unified diff.
```
Makman2: use double backticks. Not sure whether RST does support that, but with double backticks we are on the safe side :) (and consistent too)

coalib/results/Diff.py
```diff
@@ -70,6 +72,65 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+        If this Diff is applied to the original file, the corrected
```
Makman2: `Diff` into double backticks.

coalib/results/Diff.py
```diff
@@ -70,6 +72,65 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+        file will the created.
```
Makman2: `the corrected file will the created.` is no sentence :O

coalib/results/Diff.py
```diff
@@ -70,6 +72,65 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+        :param unified_diff:  Unified diff string.
+        :param original_file: The contents of the original file
+                              (line-splitted).This is required to
+                              construct the corrected file.
```
Makman2: 1. Missing space after period
2. Omit "This is required to construct...", it's redundant. If it wouldn't have been required, you wouldn't put a parameter ;)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -406,6 +407,45 @@ def process_output_corrected(self,
...
+            :param diff_distance:
+                Number of unchanged lines that are allowed in between two
+                changed lines so they get yielded as one diff. If a negative
+                distance is given, every change will be yielded as an own diff,
+                even if they are right beneath each other.
+            :return:
+                An iterator returning results containing patches for the
+                file to correct.
+            """
+            diff = Diff.from_unified_diff(output, file)
```
Makman2: I would inline this value :+1:

Makman2:
You're still not updating the documentation inside `Linter.py` accordingly^^

coalib/bearlib/abstractions/Linter.py
satwikkansal: umm, the `output` one is useful I think :)

coalib/results/Diff.py
satwikkansal: Ah, sorry :(

satwikkansal:
@Makman2 Addressed all your suggestions :)

coalib/bearlib/abstractions/Linter.py
```diff
@@ -401,6 +402,42 @@ def process_output_corrected(self,
                                      result_message,
                                      diff_distance)
 
+        def process_output_unidiff(self,
```
Makman2: didn't we agree on calling it `process_output_unified_diff`?

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -328,6 +347,50 @@ def test_process_output_corrected(self):
                                           original))
         self.assertEqual(len(results), 2)
 
+    def test_process_output_unidiff(self):
```
Makman2: would call it `test_process_output_unified_diff` too :)

tests/bearlib/abstractions/LinterTest.py
```diff
@@ -328,6 +347,50 @@ def test_process_output_corrected(self):
...
+        uut = (linter(sys.executable,
+                      output_format='unified-diff',
+                      diff_distance=-1)
+               (self.EmptyTestLinter)
+               (self.section, None))
+
+        results = list(uut.process_output(diff_string,
+                                          'some-file.c',
+                                          original))
+        self.assertEqual(len(results), 2)
```
Makman2: could you add a test for a patch with a larger file that doesn't contain some parts of the origin file? As your `from_unified_diff` did not work before (correctly), we should cover that.

coalib/bearlib/abstractions/Linter.py
satwikkansal: Yeah, my bad

coalib/bearlib/abstractions/Linter.py
```diff
@@ -358,15 +394,12 @@ def process_output_corrected(self,
...
-                             result_message,
-                             affected_code=diff.affected_code(filename),
-                             diffs={filename: diff},
-                             severity=diff_severity)
+            return self.process_diff(Diff.from_string_arrays(
+                file, output.splitlines(keepends=True)),
+                                     filename,
+                                     diff_severity,
+                                     result_message,
+                                     diff_distance)
```
Makman2: would format it like this:
```
return self.process_diff(
    Diff.from_string_arrays(file, output.splitlines(True)),
    filename,
    diff_severity,
    result_message,
    diff_distance)
```

Makman2: if `Diff.from_string_array(...)` does not fit into the line:
```
return self.process_diff(
    Diff.from_string_arrays(
        file,
        output.splitlines(True)),
    filename,
    diff_severity,
    result_message,
    diff_distance)
```

coalib/results/Diff.py
```diff
@@ -70,6 +72,64 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                        # find occurrence of this line in original file
+                        new_pos = original_file.index(str(line)[1:], pos)
+                        # between pos and new_pos everything is unchanged
+                        for n in range(pos, new_pos+1):
+                            corrected_file.append(original_file[n])
+                        pos = new_pos+1
+                    elif line.is_added:
+                        corrected_file.append(str(line)[1:])
+                    else:
+                        pos += 1
```
Makman2: damn that looks complicated, is that really necessary? I thought we use a library for that :O

satwikkansal: Actually that's the only way I could think of for generating complete "modified file" from partial unified-diff, and as per previous disussion, it is necessary to generate complete files to create `Diff` object correctly.

tests/bearlib/abstractions/LinterTest.py
satwikkansal: Yeah, there should be one. Also, I've added tests for unified diffs with some unchanged part of files missing in `DiffTest.py` as well.

coalib/results/Diff.py
Makman2: phew... I have the feeling this should be easier, will try something soon^^

tests/bearlib/abstractions/LinterTest.py
Makman2: you can copy it 1:1 and adapt for `@linter` usage :+1:

coalib/results/Diff.py
satwikkansal: @Makman2 Any updates? I tried different approaches but couldn't come up with better one, neither does `unidiff` supports something like this. I guess the code is good time complexity wise, may be simplified a bit by combining the `for` and the `while` loop somehow.

Makman2: Have you tried yet to use the basic functionality of the snippet I sent you?

satwikkansal: Sorry, wasn't able to manage enough time in past 2 weeks or so. Will definitely try it and update the PR within next few days.

satwikkansal: Updated the code, please have a look

tests/results/DiffTest.py
```diff
@@ -207,6 +209,164 @@ def test_from_string_arrays(self):
...
+        source = ['first', 'invalid starting character']
+        target = ['first', 'invalid starting character']
+        diff = ['--- a/testfile',
+                '+++ b/testfile',
+                '@@ -1,2 +1,2 @@',
+                ' first',
+                '*invalid_starting_character']
+        diff_string = '\n'.join(diff)
+        with self.assertRaises(UnidiffParseError):
```
satwikkansal: Is it okay to test Invalid unified diffs explicitly or should we expect `unidff` to throw an error

Makman2: definitely okay :+1: it asserts correct behaviour of your class failing properly too :+1:

satwikkansal: Okay :) Then should I also add test Invalid hunks, and other such cases?

Makman2: if you want to support proper errors, sure :+1:
Though you can request from the users to supply valid diffs, otherwise undefined behaviour. If you want :)

coalib/results/Diff.py
```diff
@@ -70,6 +72,58 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unified_diff(cls, unified_diff, original_file):
```
meetmangukiya: This is awesome :tada:

coalib/results/Diff.py
```diff
@@ -70,6 +72,58 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+        patch_set = PatchSet(unified_diff.splitlines())
+
+        diff = Diff(original_file)
+
+        for file in patch_set:
+            for hunk in file:
+                file_line = hunk.source_start
+                i = 0
+                while i < len(hunk):
+                    # FIXME: The stuff may be faster when using single or
```
satwikkansal: To be removed in next iteration as it doesn't apply here.

coalib/results/Diff.py
satwikkansal: That's @Makman2's magic, he came up with this simple approach :P

satwikkansal:
@Makman2 review please :)

coalib/results/Diff.py
```diff
@@ -70,6 +72,55 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                        add_set = []
+                        # Add all the consecutive '+' lines at once
+                        while(i < len(hunk) and hunk[i].is_added):
+                            source_code = str(hunk[i])[1:]
+                            add_set.append(source_code)
+                            i += 1
+                        diff.add_lines(file_line-1, add_set)
+
+                    elif line.is_removed:
+                        if source_code != original_file[file_line-1].strip("\n"):
```
satwikkansal: Had to add `strip("\n")` because newlines are stripped of by `unidiff` but the `original_file` contains newlines.
(even 
```py
 PatchSet(unified_diff.splitlines(keepends=True))
``` 
doesn't help)

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
         return result
 
     @classmethod
+    def from_unified_diff(cls, unified_diff, original_file):
+        """
+        Creates a ``Diff`` object from given unified diff.
+
+        If this ``Diff`` is applied to the original file, the corrected
+        file will be created.
```
Makman2: What do you mean with this? Imo it's normal that you can apply a diff and that it changes file :)

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+        :param unified_diff:  Unified diff string.
+        :param original_file: The contents of the original file
+                              (line-splitted).
+        """
+        patch_set = PatchSet(unified_diff.splitlines())
+
+        diff = Diff(original_file)
+
+        for file in patch_set:
```
Makman2: so currently you fuse all files together into one diff, though `Diff` can only handle a single file. I think there are two possibilities:
1. Extract the first diff for a single file
2. Return a list/set of Diffs

coalib/bearlib/abstractions/Linter.py
```diff
@@ -827,6 +874,9 @@ def linter(executable: str,
           ``output_regex``.
         - ``'corrected'``: The output is the corrected of the given file. Diffs
           are then generated to supply patches for results.
+        - ``'unified_diff'``: The ouptut is the unified diff of the original
```
Makman2: typo ;)
ouptut -> output

coalib/bearlib/abstractions/Linter.py
```diff
@@ -827,6 +874,9 @@ def linter(executable: str,
...
+          file and the corrected file. Patches are then supplied for results
```
Makman2: "and the corrected file"?

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+    def from_unified_diff(cls, unified_diff, original_file):
+        """
+        Creates a ``Diff`` object from given unified diff.
+
+        If this ``Diff`` is applied to the original file, the corrected
+        file will be created.
+
+        :param unified_diff:  Unified diff string.
+        :param original_file: The contents of the original file
+                              (line-splitted).
```
Makman2: you should document `RuntimeError` with `:raises RuntimeError:`

tests/results/DiffTest.py
```diff
@@ -207,6 +209,183 @@ def test_from_string_arrays(self):
...
+                ' B',
+                '-C',
+                '+Z',
+                '@@ -6,2 +5,2 @@',
+                ' F',
+                '-G',
+                '+K']
+        diff_string = '\n'.join(diff)
+        with self.assertRaises(UnidiffParseError):
+            self.uut = Diff.from_unified_diff(diff_string, source)
```
Makman2: could you split up each test setup into different test functions? I know finding names for them is hard, though try to name the aspect you try to test inside the test-function, like
`test_from_unified_diff_invalid_hunk`, `test_from_unified_diff_simple`, `test_from_unified_diff_2_changes_beneath` etc etc

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+            for hunk in file:
+                file_line = hunk.source_start
+                i = 0
+                while i < len(hunk):
+                    line = hunk[i]
+                    source_code = str(line)[1:]
+
+                    if line.is_added:
+                        add_set = []
+                        # Add all the consecutive '+' lines at once
```
Makman2: you should describe why you do this shortly: `Diff` doesn't allow to make changes to a place more than once :+1:

coalib/bearlib/abstractions/Linter.py
satwikkansal: My bad, "The ouptut is the unified diff of the corrections" is better imo.

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                              (line-splitted).
+        """
+        patch_set = PatchSet(unified_diff.splitlines())
+
+        diff = Diff(original_file)
+
+        for file in patch_set:
+            for hunk in file:
+                file_line = hunk.source_start
+                i = 0
```
Makman2: as you only use `i` to control the loop and only increment by a fixed amount, you could try to use iterators so you don't have to do `i += 1`, like
```python
hunk_iterator = iter(hunk)
line = next(hunk_iterator)
```

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                            add_set.append(source_code)
+                            i += 1
+                        diff.add_lines(file_line-1, add_set)
+
+                    elif line.is_removed:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
+                            raise RuntimeError(
+                                'Line to delete does not match in '
+                                'the original file.')
```
Makman2: you should display where exactly, and maybe the contents of the lines that do not fit together :+1:

coalib/results/Diff.py
```diff
@@ -70,6 +72,57 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                        diff.delete_line(file_line)
+                        i += 1
+                        file_line += 1
+
+                    else:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
+                            raise RuntimeError('Context lines do not match.')
```
Makman2: Also here, you should display where and the lines :)

coalib/results/Diff.py
Makman2: And to check when the iterator is finished, just wrap the whole thing into a try-except:
```
try:
    # process hunk...
except StopIteration:
    pass
```

coalib/results/Diff.py
satwikkansal: Copied it from `from_string_arrays` function, will remove it :P

coalib/results/Diff.py
satwikkansal: As of now, added the behavior assuming the `patch_set` contains single patched file. Let's keep the multiple file patch thing for a separate issue.

coalib/bearlib/abstractions/Linter.py
satwikkansal: Yep, I always do that :(

satwikkansal:
@Makman2 Please review whenever you get some free time :)

coalib/results/Diff.py
```diff
@@ -70,6 +72,89 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                            while line.is_added:
+                                source_code = str(line)[1:]
+                                add_set.append(source_code)
+                                line = next(hunk_iterator)
+                            diff.add_lines(file_line-1, add_set)
+
+                        except StopIteration:
+                            # The hunk is empty but the lines are
+                            # still to be added to diff.
+                            diff.add_lines(file_line-1, add_set)
```
Makman2: -->
```python
try:
    while line.is_added:
        ...
except StopIteration:
    pass

diff.add_lines(file_line-1, add_set)
```
As the `add_lines` function is executed regardless whether you enter the exception or not

coalib/results/Diff.py
```diff
@@ -70,6 +72,89 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+                    elif line.is_removed:
```
Makman2: I think you have to make this a plain `if`. When a couple of lines get added, you execute in the end `next(hunk_iterator)`. If you now step out of the `if`, this line doesn't get handled as the overall loop is continued and in the beginning again `next(hunk_iterator)` is called. If you make this a plain `if`, the last retrieved line is available for further processing.

Makman2: Also add tests for this case, like
```
 Context line
+ A
+ B
 This context line isn't checked as it's skipped.
```
and
```
 Context line
+A
-B
```

coalib/results/Diff.py
satwikkansal: Just realized it should be "break" instead or "pass", so now the code looks like 
```py
                      try:
                            while line.is_added:
                                source_code = str(line)[1:]
                                add_set.append(source_code)
                                line = next(hunk_iterator)

                            diff.add_lines(file_line-1, add_set)
                            source_code = str(line)[1:]

                        except StopIteration:
                            diff.add_lines(file_line-1, add_set)
                            break
```

coalib/results/Diff.py
satwikkansal: Yes, you are right. Fixed this and added `test_from_unified_diff_multiple_additions_different_orderings` and `test_from_unified_diffrent_beginning_line_types` tests to DiffTest

coalib/results/Diff.py
Makman2: You need tests for this, as your old tests did not cover this case :+1:

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                    source_code = str(line)[1:]
+                    if line.is_added:
+                        add_set = []
+                        # As ``Diff`` does not allow line additions to a
+                        # position more than one time, add all the
+                        # consecutive '+' lines at once.
+
+                        try:
+                            while line.is_added:
+                                source_code = str(line)[1:]
```
Makman2: I think you can move this directly below `line = next(hunk_iterator)`, because at this stage `source_code` does already point to the correct line (see L106).
-->
```python
while line.is_added:
    add_set.append(source_code)
    line = next(hunk_iterator)
    source_code = str(line)[1:]

diff.add_lines(file_line-1, add_set)
#source_code = str(line[1:]) <-- can be removed now
```

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+                        except StopIteration:
+                            diff.add_lines(file_line-1, add_set)
+                            break
+
+                    if line.is_removed:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
+                            raise RuntimeError(
+                                'The line to delete does not match with'
```
Makman2: Missing space after "with"
Message would look like
`... does not match withthe line in the original file.`

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                                'the line in the original file. \n'
```
Makman2: unnecessary space after period

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                                'Line to delete: {} \n'
```
Makman2: unnecessary space before `\n`

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                                'Original line #{}: {}'.format(
```
Makman2: Up to you, but for exception messages I usually try to avoid `\n`, as this applies some kind of formatting to the messages and imo exception messages should be only information. I try to separate sentences with periods and commas instead of `\n`. But as we don't have some general guideline for this, up to you ;)

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                        diff.delete_line(file_line)
+                        file_line += 1
+
+                    else:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
+                            raise RuntimeError(
+                                'Context lines do not match. \n'
+                                'Line from unified diff: {} \n'
+                                'Original line #{}: {}'.format(
```
Makman2: same here, please look out for unnecessary spaces before `\n` ;)

coalib/results/Diff.py
satwikkansal: This ^^ test case is covered in `test_from_unified_diff_multiple_additions_different_orderings` (last test case)

coalib/results/Diff.py
satwikkansal: Ah okay, actually I didn't find any usage of `\n` for exceptions in our codebase, so I guess this should be our preferred style then  :)

satwikkansal:
@Makman2 Thanks for your extensive reviews, I'll try my best to make fewer mistakes from next PR onwards. I think this PR should be ready to merge in this iteration or next iteration.

coalib/results/Diff.py
```diff
@@ -70,6 +72,90 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                            break
+
+                    if line.is_removed:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
+                            raise RuntimeError(
+                                'The line to delete does not match with '
+                                'the line in the original file. '
+                                'Line to delete: {}, '
+                                'Original line #{}: {}'.format(
```
Makman2: Ah maybe you want to use `repr` on the lines when you print them?
```
'Line to delete: {!r}, '
'Original line #{}: {!r}'
```
Up to you again ;)

tests/results/DiffTest.py
```diff
@@ -207,6 +209,305 @@ def test_from_string_arrays(self):
...
+    def test_from_unified_diff_unmatched_context_line(self):
+        source = ['first', 'second']
+        diff = ['--- a/testfile',
+                '+++ b/testfile',
+                '@@ -1,2 +1,2 @@',
+                ' context_line_is_not_same',
+                ' second']
+        diff_string = '\n'.join(diff)
+        with self.assertRaises(RuntimeError):
+            self.uut = Diff.from_unified_diff(diff_string, source)
```
Makman2: this is something worth to test I think: Could you assert for the exception messages? As you print the lines, line-numbers etc it should be tested imo :+1:

satwikkansal: Done

coalib/results/Diff.py
```diff
@@ -70,6 +72,90 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+
+                            diff.add_lines(file_line-1, add_set)
+
+                        except StopIteration:
+                            diff.add_lines(file_line-1, add_set)
+                            break
+
+                    if line.is_removed:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
```
Makman2: you use this value quite often. You should move that into a variable before `if line.is_removed:`. + You can just do `rstrip('\n')`, as from the left side there won't be any newline characters^^

coalib/results/Diff.py
```diff
@@ -70,6 +72,90 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                            raise RuntimeError(
+                                'Context lines do not match. '
+                                'Line from unified diff: {!r}, '
+                                'Original line #{!r}: {!r}'.format(
+                                    source_code,
+                                    file_line,
+                                    original_file[file_line-1].strip('\n'))
+                                )
+                        file_line += 1
```
Makman2: we should move this out of the if, as for both branches you execute `file_line += 1`

satwikkansal:
@Makman2 ping :P

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                              the original file and the unified diff.
+        """
+        patch_set = PatchSet(unified_diff.splitlines())
+
+        diff = Diff(original_file)
+
+        if not patch_set:
+            return diff
+
+        # FIXME Handle patches consisting of change in more that one files
```
Makman2: I know it's a fixme, but let's do it right :P
```
Handle patches consisting of changes in more than one file
```
:3

coalib/results/Diff.py
```diff
@@ -70,6 +72,91 @@ def from_string_arrays(cls, file_array_1, file_array_2, rename=False):
...
+                                'Original line #{!r}: {!r}'.format(
+                                    source_code,
+                                    file_line,
+                                    original_line)
+                                )
+                        diff.delete_line(file_line)
+
+                    else:
+                        if (source_code !=
+                                original_file[file_line-1].strip('\n')):
```
Makman2: use `original_line` :)

coalib/results/Diff.py
satwikkansal: Oops, sorry :)

coalib/results/Diff.py
satwikkansal: Done

Makman2:
ack 81b327e d6788a7 7659c56 07a6547

Makman2:
@rultor merge

PR #3742, opened 2017-02-14
LineParser.py: Print warning for comments

Nosferatul:
Adds warning if comment does not have
whitespace before or after #.

Closes https://github.com/coala/coala/issues/2926

### DIFF
--- a/coalib/parsing/LineParser.py
+++ b/coalib/parsing/LineParser.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from coala_utils.string_processing.StringConverter import StringConverter
 from coala_utils.string_processing import (unescape, convert_to_raw,
@@ -78,6 +79,15 @@ class LineParser:
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        for separator in self.comment_separators:
+            if (re.match('[^ ]' + separator, line)
+                    or re.match(separator + '[^ ]', line)):
+                logging.warning('This comment does not have whitespace' +
+                                ' before or after ' + separator + ' in: ' +
+                                repr(line.replace('\n', '')) + '. If you ' +
+                                'didn\'t mean to make a comment, use a ' +
+                                'backslash for escaping.')
+
         line, comment = self.__separate_by_first_occurrence(
             line,
             self.comment_separators)
--- a/tests/parsing/LineParserTest.py
+++ b/tests/parsing/LineParserTest.py
@@ -14,6 +14,7 @@ class LineParserTest(unittest.TestCase):
         self.check_data_set('\n \n \n')
 
     def test_comment_parsing(self):
+        logger = logging.getLogger()
         self.check_data_set('# comment only$§\n',
                             output_comment='# comment only$§')
         self.check_data_set('   ; comment only  \n',
@@ -21,6 +22,23 @@ class LineParserTest(unittest.TestCase):
         self.check_data_set('   ; \\comment only  \n',
                             output_comment='; comment only')
         self.check_data_set('#', output_comment='#')
+        with self.assertLogs(logger, 'WARNING') as warn:
+            self.check_data_set('##\n', output_comment='##')
+        self.assertEqual(len(warn.output), 1)
+        self.assertEqual(warn.output[0], 'WARNING:root:This comment does ' +
+                                         'not have whitespace before or ' +
+                                         'after # in: ' + repr('##') +
+                                         '. If you didn\'t mean to make ' +
+                                         'a comment, use a backslash for ' +
+                                         'escaping.')
+        with self.assertLogs(logger, 'WARNING') as warn:
+            self.check_data_set('#A\n', output_comment='#A')
+        self.assertEqual(warn.output[0], 'WARNING:root:This comment does ' +
+                                         'not have whitespace before or ' +
+                                         'after # in: ' + repr('#A') +
+                                         '. If you didn\'t mean to make ' +
+                                         'a comment, use a backslash for ' +
+                                         'escaping.')
 
     def test_section_override(self):
         self.check_data_set(r'a.b, \a\.\b\ c=',


impmihai:
Unless you have `'` in your text, always use the double quotes `"` :)

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
+            comment_index = comment.index('#')
+            if comment_index + 1 < len(comment) or comment_index - 1 > 0:
+                if comment[comment.index('#') + 1] != ' ' \
```
impmihai: the trailing `\` is not needed

Techievena: If your if statement is spanning across multiple lines you may use braces with if `if( )`

Techievena:
Running coala on your system before pushing will automatically fix the errors gitmate is showing right now. 😄

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
+            comment_index = comment.index('#')
```
adtac: Wouldn't this always be 0? Every comment starts with a `#`.

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
...
+            if comment_index + 1 < len(comment) or comment_index - 1 > 0:
+                if comment[comment.index('#') + 1] != ' '\
+                   or comment[comment.index('#') - 1] != ' ':
+                    logging.warning('The comment does not have whitespace' +
```
adtac: Imagine getting a random "The comment does not have whitespace" out of nowhere!

Since we have no access to the line number, we should at least print the line so that the user knows where to look.

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
+            comment_index = comment.index('#')
+            if comment_index + 1 < len(comment) or comment_index - 1 > 0:
+                if comment[comment.index('#') + 1] != ' '\
+                   or comment[comment.index('#') - 1] != ' ':
```
adtac: Why calculate this so often? You already have this result ;)

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
+            comment_index = comment.index('#')
+            if comment_index + 1 < len(comment) or comment_index - 1 > 0:
+                if comment[comment.index('#') + 1] != ' '\
```
adtac: Same here, don't recalculate ;)

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
+            comment_index = comment.index('#')
+            if comment_index + 1 < len(comment) or comment_index - 1 > 0:
```
adtac: You check if `comment_index + 1` is within bounds OR `comment_index - 1` is within bounds. So even if one of the two passes, we'll enter the conditional branch. But the next line uses both `comment[comment.index('#') + 1]` and `comment[comment.index('#') - 1]`. This will definitely raise an exception at the extremes. Let's just have two separate conditions.

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
+        if '#' in comment:
```
adtac: Are you checking if the line has a comment? Just a `if comment:` would do.

coalib/parsing/LineParser.py
```diff
@@ -82,6 +82,15 @@ def _parse(self, line):
             line,
             self.comment_separators)
         comment = unescape(comment)
+
```
adtac: Anyway, I don't really think this is the best way to solve this. How about a simple regex on the line before we split it with `__separate_by_first_occurrence` in line 81? `[^ ]#[^ ]` will catch all possible issue :)

adtac:
Sigh, Github messed up the order of my review comments.

@Nosferatul If it's confusing, ping me :)

coalib/parsing/LineParser.py
Nosferatul: You are right :+1:

coalib/parsing/LineParser.py
Techievena: Suppose the statement is "C# is cool" or something like "#This is a comment", it won't match this regex. Basically you need an or instead of and. Using two separate regex [^ ]# and #[^ ] in if elif statements would do but if there is a better regex use it preferably.

coalib/parsing/LineParser.py
```diff
@@ -1,5 +1,5 @@
 import logging
-
+import re
```
Techievena: Give a blank space after this import

coalib/parsing/LineParser.py
```diff
@@ -78,10 +78,14 @@ def _parse(self, line):
...
                      append delimiter is found else False), comment
         """
+        if re.match('[^ ]#[^ ]', line) is not None:
+            logging.warning('The comment does not have whitespace' +
+                            ' before or after # in: ' + line.replace('\n', ''))
         line, comment = self.__separate_by_first_occurrence(
             line,
             self.comment_separators)
         comment = unescape(comment)
+
```
Techievena: Remove this extra blank space

tests/parsing/LineParserTest.py
```diff
@@ -14,13 +14,19 @@ def test_empty_line(self):
...
         self.check_data_set('   ; comment only  \n',
                             output_comment='; comment only')
         self.check_data_set('   ; \\comment only  \n',
                             output_comment='; comment only')
         self.check_data_set('#', output_comment='#')
+        with self.assertLogs(logger, 'WARNING') as warn:
+            self.check_data_set('##\n', output_comment='##')
+        self.assertEqual(warn.output[0][13:], 'The comment does not have ' +
+                                              'whitespace before or after # ' +
+                                              'in: ##')
```
Techievena: Tests are also not sufficient to detect the cases I mentioned and preferably use  this
```
 self.assertEqual(len(warn.output), 1)
self.assertIn(warn.output[0], 'WARNING:root:The comment does not have whitespace before or after # in: ##')
```

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,10 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        if re.match('[^ ]#', line) is not None or\
```
adtac: `is not None` isn't required :) Just a `if re.match(...) or re.match(...):` will do :)

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,10 @@ def _parse(self, line):
...
+           re.match('#[^ ]', line) is not None:
+            logging.warning('The comment does not have whitespace' +
+                            ' before or after # in: ' + line.replace('\n', ''))
```
adtac: "This comment" instead of "The comment".

Also maybe quotes around the comment? use `repr(line.replace('\n', ''))` so that you don't have to worry about internal quoting.

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,10 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        if re.match('[^ ]#', line) is not None or\
+           re.match('#[^ ]', line) is not None:
```
adtac: I think we should allow for lines with only a `#` in them: for example, if you're writing a long comment and want a paragraph break, this will be useful. Example:

```
# Line 1
#
# Line 2
```

But currently, this will be flagged.

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,12 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        if (re.match('[^ ]#', line) or re.match('#[^ ]', line))\
```
Udayan12167: Avoid breaking the line with a `\`. Use brackets over the whole expression.

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,12 @@ def _parse(self, line):
...
+           and len(line) > 2:
+            logging.warning('This comment does not have whitespace' +
```
Udayan12167: Warning should contain a way to silence it. So tell the user to put it or give a fixed representation of the line in the warning. I think a correct representation should be easy in this case...

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,13 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        if ((re.match('[^ ]#', line) or re.match('#[^ ]', line))
```
Udayan12167: Are you saying with this that a comment like `code #comment` is acceptable?

sils: the LineParser has a parameter comment_seperator so it's not necessarily #, we should obey that setting (or remove it)

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,13 @@ def _parse(self, line):
...
+                and len(line) > 2):
+            logging.warning('This comment does not have whitespace' +
+                            ' before or after # in: ' +
+                            repr(line.replace('\n', '')) + '. The line ' +
+                            'should be: ' + repr('# <comment>'))
```
sils: maybe add something like "If you didn't mean to make a comment, use a backslash for escaping" as this is the case of `C#` we're talking about right?

coalib/parsing/LineParser.py
Nosferatul: #[^ ] finds `code #comment`

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,16 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        for separator in self.comment_separators:
+            if ((re.match('[^ ]' + separator, line)
+                    or re.match(separator + '[^ ]', line))
+                    and len(line) > 2):
```
adtac: Why do we need to test the length of the line? If it matches either of the two regexes, then it's automatically at least 2 characters long.

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,16 @@ def _parse(self, line):
...
+                logging.warning('This comment does not have whitespace' +
+                                ' before or after # in: ' +
```
adtac: before or after `separator` instead of hardcoding the `#`

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,16 @@ def _parse(self, line):
...
+                                repr(line.replace('\n', '')) + '. If you ' +
+                                'didn\'t mean to make a comment, use a '
```
adtac: Is `"... didn't ..."` disallowed by the quotes bear? Just curious.

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,16 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        for separator in self.comment_separators:
+            if ((re.match('[^ ]' + separator, line)
+                    or re.match(separator + '[^ ]', line))
+                    and len(line) > 2):
+                logging.warning('This comment does not have whitespace' +
+                                ' before or after # in: ' +
+                                repr(line.replace('\n', '')) + '. If you ' +
```
adtac: and `"... {} ...".format(...)` is a better way to do stuff generally

coalib/parsing/LineParser.py
Nosferatul: i had errors with that

coalib/parsing/LineParser.py
```diff
@@ -78,6 +79,15 @@ def _parse(self, line):
                      [(section_override, key), ...], value, to_append (True if
                      append delimiter is found else False), comment
         """
+        for separator in self.comment_separators:
+            if (re.match('[^ ]' + separator, line)
+                    or re.match(separator + '[^ ]', line)):
+                logging.warning('This comment does not have whitespace' +
+                                ' before or after # in: ' +
```
adtac: Don't hard-code the `#` - use separator (did I mention this before?)

Nosferatul: Yes you did,  I believed that I did the change

adtac:
ack https://github.com/coala/coala/pull/3742/commits/c14fab1661335532bd8214e01a79e0540a3ecc9d

adtac:
@rultor merge

Issue #3154, opened 2016-12-05
--format does not work in non-interative mode


jayvdb:
Using `--ci` and `--format` together discards the --format argument.

namanyadav12:
Hi. Can i be assigned be this issue ?

jayvdb:
I've assigned you to https://github.com/coala/docker-coala-base/issues/34 first.  Once that is done, this would be a good second bug to complete.  Happy hacking.

meetmangukiya:
I digged a bit, and it seems to be a problem in https://github.com/coala/coala/blob/b02883a7a4abb9e98d85d28fa21617def1e88268/coalib/coala.py#L77
The `if` block makes it either `--ci` or `--format`, since they are considered two different modes..

shrutishrm512:
Hello, please assign this bug to me !

shrutishrm512:
Changing line 77 as
`if args.format or args.ci:`
then this statement can be correct?

jayvdb:
line 77 of which file?  Please give a link so I can quickly answer.

jayvdb:
AH; ignore me; I see your PR now.  sorry.

shrutishrm512:
Line 77 in coala/coalib/coala.py
https://github.com/coala/coala/blob/b02883a7a4abb9e98d85d28fa21617def1e88268/coalib/coala.py#L77

jayvdb:
ok, that wont work.  Above it is `if args.non_interactive: return ..` so line 77 is never executed.

shrutishrm512:
`if args.format:` line never executes due to `if args.non_interactive:` so i made changes in if block such that both will execute as per my understanding, but i also want to ask about that
> coala --format=anyfilename(say:README.rst) can be executable one? since i got error in that!

jayvdb:
Anyone picking up this issue, please first do a little analysis before coding.

Nosferatul:
@andreiraduta1101 are you still working on this?

Nosferatul:
unassigned due to inactivity

meetmangukiya:
The PR raised for this issue doesn't allow the user to use both the arguments together, and lets the user know that one of them will be ignored.

Nosferatul:
Closing because 3999 was merged

PR #3750, opened 2017-02-16
coala.py: use --format and --ci

Nosferatul:
Adds case where --ci and --format are used together.

Fixes https://github.com/coala/coala/issues/3154


tests/coalaTest.py
```diff
@@ -65,6 +65,16 @@ def test_did_nothing(self):
...
 
+    def test_format(self):
+        retval, stdout, stderr = execute_coala(coala.main, 'coala', '--ci',
+                                               '--format', '-b',
+                                               'SpaceConsistencyTestBear',
+                                               '-f', '*.py'
+                                               )
+        self.assertEqual(retval[0], 0)
+        self.assertEqual(retval[1], 0)
+        self.assertIn('Executing section cli...', stdout)
```
aptrishu: Please assert stderr also as empty. we separated stdout and stderr return for the reason that both can be used individually.

tests/coalaTest.py
```diff
@@ -65,6 +65,17 @@ def test_did_nothing(self):
...
+        self.assertTrue(stderr)
```
aptrishu: If there is a content - 
validate it , either full validation or partially using `assertIn` or regex.
If its empty - 
simply assert it empty or false.

So, we need to validate contents of stderr.

coalib/coala.py
```diff
@@ -71,10 +71,11 @@ def main():
     except BaseException as exception:  # pylint: disable=broad-except
         return get_exitcode(exception, log_printer)
 
-    if args.non_interactive:
+    if args.non_interactive and args.format:
+        return mode_non_interactive(console_printer, args), mode_format()
```
sils: shouldn't this rather throw an exception? We handle conflicts somewhere and this is a conflict, you can't just run it twice.

Nosferatul:
done @sils

coalib/coala.py
```diff
@@ -71,10 +71,11 @@ def main():
     except BaseException as exception:  # pylint: disable=broad-except
         return get_exitcode(exception, log_printer)
 
-    if args.non_interactive:
+    if args.non_interactive and args.format:
+        return Exception('Error: do not use --format and --ci in the same time')
```
sils: his should happen in CliParsing:check_conflicts

coalib/parsing/CliParsing.py
```diff
@@ -119,4 +119,10 @@ def check_conflicts(sections):
             ArgumentParser().error(
                 "'output' or 'relpath' cannot be used without `--json`.")
 
+        if (
+                section.get('non_interactive', False) and
+                (str(section.get('format', ''))) != ''):
+            ArgumentParser().error('do not use --format and --ci in ' +
```
aptrishu: in -> at

coalib/parsing/CliParsing.py
```diff
@@ -119,4 +119,10 @@ def check_conflicts(sections):
             ArgumentParser().error(
                 "'output' or 'relpath' cannot be used without `--json`.")
 
+        if (
```
aptrishu: Add a test in cliParsingTest to test this conflict.

Nosferatul:
Done :+1:

coalib/parsing/CliParsing.py
```diff
@@ -119,4 +119,10 @@ def check_conflicts(sections):
...
+                section.get('non_interactive', False) and
```
Mixih: This empty line seems a bit strange.... Would it be possible to remove the empty newline?

Nosferatul: i just followed the pattern

coalib/parsing/CliParsing.py
```diff
@@ -119,4 +119,9 @@ def check_conflicts(sections):
             ArgumentParser().error(
                 "'output' or 'relpath' cannot be used without `--json`.")
 
+        if (section.get('non_interactive', False) and
```
Mixih: Crap, I see that was the general formatting. Please revert, I'm sorry.

Nosferatul:
Done @Mixih

meetmangukiya:
@Nosferatul I don't think this really solves the issue :/ This just notifies user that either of them will be discarded. I think we want to be able to format the output of CI mode.

Though, we can still merge it, so at least user knows it will not work as user intended. Just create a new issue and change the issue reference?

/cc @jayvdb (since you are the author of the issue)

Nosferatul:
Yea, that's right. I'll create a new issue @meetmangukiya

userzimmermann:
Hi @Nosferatul :) Thanks for the effort you already put into this topic! Please try to rebase and solve conflicts with merged https://github.com/coala/coala/pull/4036 . And please try to align this with #4055 , whereby I think that `--ci --format` and `--ci --json` combinations don't necessarily need to raise errors. But let's discuss this further...

userzimmermann:
This now fixes https://github.com/coala/coala/issues/4055

Nosferatul:
Hey @mixih, could you review it?

tests/parsing/CliParsingTest.py
```diff
@@ -71,3 +71,8 @@ def test_check_conflicts(self):
         with self.assertRaisesRegex(SystemExit, '2') as cm:
             check_conflicts(sections)
             self.assertEqual(cm.exception.code, 2)
+
+        sections = parse_cli(arg_list=['--format', '--json'])
+        with self.assertRaisesRegex(SystemExit, '2') as cm:
+            check_conflicts(sections)
+            self.assertEqual(cm.exception.code, 2)
```
userzimmermann: This `self.assertEqual` won't be reached when inside the `with` block. Whereby that's strangely done in all tests above o.O This is confusing. Something must raise a `SystemExit` inside the `with` block. Otherwise the `self.assertRaisesRegex` would fail. And the `self.assertEqual` doesn't raise `SystemExit`. So `check_conflicts` surely raises that, and all `cm.exception.code` checks must be moved out of the `with` blocks.

Correct me if I'm wrong :D 

And please also check the error message :)

**CC** @aptrishu

coalib/coala.py
```diff
@@ -37,7 +37,8 @@ def main():
 
         console_printer = ConsolePrinter(print_colored=not args.no_color)
 
-        if args.json:  # needs to be checked in order to display bears in json
+        # needs to be checked in order to display bears in json
+        if args.json and not args.format:
```
userzimmermann: It shouldn't be possible that both are `True` here. Should already have been handled by `check_conflicts` before. So better `assert` here if you want to get sure...

Nosferatul: I can't agree with you on this:) I checked it, when I use `coala --format -- json` both of them are true

tests/parsing/CliParsingTest.py
Nosferatul: Made the changes :+1:

meetmangukiya:
lowercase c for coala, please in the commit message

Adrianzatreanu:
ack b4599561d46b2da8679a840b0b7354e111f5750b

Adrianzatreanu:
@userzimmermann can you take a look and approve/request changes? as you requested some

tests/parsing/CliParsingTest.py
userzimmermann: @Nosferatul Except the error message check ;)

coalib/coala.py
userzimmermann: @Nosferatul We have a fundamental problem here! After...

```python
ArgumentParser().error('do not use --format and --json at ' +
                       'the same time')
```

... `coala.main()` should not continue!

@sils @Adrianzatreanu @Mixih @aptrishu @meetmangukiya

meetmangukiya: I agree with terminating coala execution.

coalib/parsing/CliParsing.py
```diff
@@ -119,4 +119,10 @@ def check_conflicts(sections):
             ArgumentParser().error(
                 "'output' or 'relpath' cannot be used without `--json`.")
 
+        if (
+                section.get('json', False) and
```
meetmangukiya: move this in previous line and remove one level of indent

userzimmermann:
unack https://github.com/coala/coala/commit/b4599561d46b2da8679a840b0b7354e111f5750b

@Adrianzatreanu @Nosferatul We need to solve https://github.com/coala/coala/pull/3750#discussion_r112845326 first

userzimmermann:
unack https://github.com/coala/coala/commit/b4599561d46b2da8679a840b0b7354e111f5750b

sils:
ack b459956 this is a test for @userzimmermann

sils:
unack b459956 this

also works

with more lines

userzimmermann:
@Adrianzatreanu @Nosferatul We need to solve https://github.com/coala/coala/pull/3750#discussion_r112845326 first

jayvdb:
I force rebased this to check https://github.com/coala/coala/pull/3750#pullrequestreview-32716725 , and even after a little fiddling the problem remains.  argparser errors are not automatically being rippled out.
That needs to be fixed first.

coalib/parsing/CliParsing.py
```diff
@@ -127,4 +127,10 @@ def check_conflicts(sections):
...
+                (str(section.get('format', ''))) != ''):
+            ArgumentParser().error('do not use --format and --json at ' +
+                                   'the same time')
```
Makman2: "do not use" --> "can't use". This is not a recommendation but an impossibility ;)

tests/coalaTest.py
```diff
@@ -323,6 +323,15 @@ def test_coala_without_color(self):
             self.assertEqual(
                 retval, 0, 'coala must return zero when there are no errors')
 
+    def test_format_and_json(self):
+        retval, stdout, stderr = execute_coala(coala.main, 'coala', '--json',
+                                               '--format', '-b',
+                                               'SpaceConsistencyTestBear',
+                                               '-f', '*.py'
+                                               )
```
bekicot: Move the closng brackets to the previous line

jayvdb: No, do not do that.  Instead, add a trailing comma on the previous line

jayvdb:
@Nosferatul needs rebase and a small manual conflict resolved.

PR #3751, opened 2017-02-16
AppveyorCI: Use pip caching

jayvdb:
Replaces five pip commands performing custom caching
with one command which uses pip's inbuilt caching.

### DIFF
--- a/.misc/appveyor.yml
+++ b/.misc/appveyor.yml
@@ -17,7 +17,7 @@ environment:
       LLVM_VERSION: 3.4.1
 
 cache:
-  - "C:\\wheels"
+  - "C:\\pip_cache"
 
 branches:  # Only build official branches, PRs are built anyway.
   only:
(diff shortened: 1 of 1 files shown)


jayvdb:
Same as https://github.com/coala/coala-bears/pull/1424

jayvdb:
Note that this has the small side-effect of not having cached wheels for some packages which dont provide wheels. See https://github.com/coala/coala/issues/3753

The elapsed times for the builds using pip caching is approx the same as using the custom caching: 1 m 30 s.

The benefit is that the Appveyor build uses the same system as the Unix builds , and it is less complex.
IMO it is even a benefit that the packages without wheels are shown differently, as we should be working with upstream to resolve that, and ideally then we can enable some flag to turn of non-wheel packages, forcing us to not use packages that do not provide wheels.

sils:
ack https://github.com/coala/coala/pull/3751/commits/75b255dc79b55905893569d9574daad253603919

sils:
(needs a rebase though)

sils:
reack https://github.com/coala/coala/pull/3751/commits/eb5e70d7bae2d94a78db846aa7916f863c9407a6

sils:
@rultor merge

Issue #2377, opened 2016-06-21
generate-ordering: Not working


abhsag24:
```
>>> file = ["1"]
>>> a = AbsolutePosition(file, 0)
>>> a.line
1
>>> a.column
1
>>> b = SourceRange.from_values("A", 1,1,1,1)
>>> a == b.start.line
False
>>> a<=b.start
False
```
but suprisingly 
```
>>> a>=b.start
True
```

fneu:
I can... more than just confirm:
```python
>>> a > b.start
True
>>> b.start < a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/coala_decorators/decorators.py", line 226, in lt
    raise TypeError("Comparison with unrelated classes is "
TypeError: Comparison with unrelated classes is unsupported.
```

This most likely has to do with `a` being an `AbsolutePosition` and `b.start` being a `SourcePosition`.
While both inherit from `TestPosition`, `SourcePosition` is the only one to have a `file` param and is defined with:

```python
@generate_ordering("file", "line", "column")
class SourcePosition(TextPosition):
```
compared to 
```
@generate_ordering("line", "column")
class TextPosition:
```

If this is not the source of this problem this bug has to be filed against coala_decorators.

fneu:
```python
>>> b.start
<SourcePosition object(file='/home/fabian/devel/coala/A', line=1, column=1) at 0x7f7de98ccac8>
>>> a
<AbsolutePosition object(line=1, column=1) at 0x7f7de98ccb00>
```

abhsag24:
what kind of ordering is required for "file" ? i do think that is actually the source of the problem @fneu

fneu:
well, we have this in decorators:

```python
def generate_ordering(*members):
    """
    Decorator that generates ordering operators for the decorated class based
    on the given member names. All ordering except equality functions will
    raise a TypeError when a comparison with an unrelated class is attempted.
    (Comparisons with child classes will thus work fine with the capabilities
    of the base class as python will choose the base classes comparison
    operator in that case.)
    Note that this decorator modifies the given class in place!
    :param members: A list of members to compare, ordered from high priority to
                    low. I.e. if the first member is equal the second will be
                    taken for comparison and so on. If a member is None it is
                    considered smaller than any other value except None.
    """
    def decorator(cls):
        def lt(self, other):
            if not isinstance(other, cls):
[...]

PR #3773, opened 2017-02-20
AbsolutePosition.py: Use generate_ordering

aptrishu:
This ensures that comparison of AbsolutePosition with any
incompatible type like SourcePosition yields an error.

Fixes https://github.com/coala/coala/issues/2377

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


tests/results/AbsolutePositionTest.py
```diff
@@ -73,3 +73,18 @@ def test_instantiation(self):
...
+        uut2 = AbsolutePosition(text1, 8)
+        uut3 = AbsolutePosition(text2, 4)
+        uut4 = AbsolutePosition(text2, 4)
+
+        self.assertTrue(uut1 <= uut2)
+        self.assertTrue(uut1 < uut2)
+        self.assertTrue(uut1 >= uut3)
+        self.assertTrue(uut2 > uut1)
+        self.assertFalse(uut1 == uut2)
+        self.assertTrue(uut1 == uut4)
```
RohanVB: Are multiple asserts like this a good idea? Like if the first assertTrue fails, the others won't be executed and you won't get anything from them. 
See : http://softwareengineering.stackexchange.com/questions/7823/is-it-ok-to-have-multiple-asserts-in-a-single-unit-test maybe.

aptrishu: Having multiple asserts here is completely valid. Because we already ensure that a test covers only a single functionality like say this test checks comparison of `AbsolutePosition` so if the first assert fails, why would I care about the other asserts? The ordering is not working that's it then and I'd modify the code to make the state valid.

aptrishu:
Yeah I had thought to do that, ofcourse the issue is about this thing but that'd not fit anywhere in tests. I mean putting a crossfunctionality checking would not be good. It is not supposed to be compared with SourcePosition. Though it solves that.(tested locally).
@Makman2 views ?

Makman2:
maybe we should still compare using `row` and `column`?

Makman2:
I mean comparing by absolute position just gives a performance boost, doesn't it? But it depends on which classes are compared^^

aptrishu:
@Makman2 
It's ok to compare them with position and yes it gives the performance boost. Also It prohibits its comparison with classes which are not compatible with it.

aptrishu:
Is it good to merge ? @Makman2

Makman2:
actually I'm not sure we need this, the ordering from the base class should work here too.
\+ as `AbsolutePosition` uses a text variable, we need to be careful whether to compared instances use the same text.

aptrishu:
> the ordering from the base class should work here too.

Yes, ofc it works. The only motive of this PR is to raise an error when compared with classes which are not compatible. Also, performance improvement.

Makman2:
Hm for such scenarios I would override `__eq__` and others. Imo you should include a check for the `text`. If they don't match, comparison is impossible^^

aptrishu:
Yes, I'll do that.

tests/results/AbsolutePositionTest.py
```diff
@@ -73,3 +73,18 @@ def test_instantiation(self):
...
+
+    def test_ordering(self):
+        text1 = ('A Lannister always pays his debts',)
+        text2 = ('Hello from the other side',)
+        uut1 = AbsolutePosition(text1, 4)
+        uut2 = AbsolutePosition(text1, 8)
+        uut3 = AbsolutePosition(text2, 4)
+        uut4 = AbsolutePosition(text2, 4)
+
+        self.assertTrue(uut1 <= uut2)
```
rhemon: is this needed: `uut1 <= uut2` ? when you have `uut1 < uut2` and  `uut1 == uut2` (which is assertFalse)
also if you have `uut1 < uut2` its the same thing for `uut2 > uut1`

Makman2: don't understand what you mean :D

rhemon: `self.assertTrue(uut1 < uut2)`  and `self.assertFalse(uut1 == uut2)` and `self.assertTrue(uut1 <= uut2)` if `uut1 == uut2` is supposed to be false, then why does it need the `uut1 <= uut2` ? 
and then `self.assertTrue(uut1 < uut2)` and `self.assertTrue(uut2 > uut1)` are testing the same thing right?

Makman2: It's too see whether `uut1 <= uut2` is also working as expected. You can override behaviour in Python for operators like `<` or `>` or `<=` (and actually we are doing that with this commit by using `@generate_ordering`).

That's the whole purpose of the test, to see whether really everything works as expected (although it seems clear). "Should work" and "does work" are two different things ;D

rhemon: ow. sorry i didnt know that. thanks for explaining :)

coalib/results/AbsolutePosition.py
```diff
@@ -1,7 +1,8 @@
 from coalib.results.TextPosition import TextPosition
-from coala_utils.decorators import enforce_signature
+from coala_utils.decorators import enforce_signature, generate_ordering
 
 
+@generate_ordering('position')
```
sils: that means there's equality between the same positions in different texts? That doesn't make too much sense to me :/

PR #3784, opened 2017-02-24
docs/conf.py: Generate from local source

Makman2:


Makman2:
reack 3a734aa

jayvdb:
unack 37156f5
unack 3a734aa

jayvdb:
You "reack 3a734aa" but I dont see someone else ack'ing that commit here.  Was it done somewhere else?

Is there an issue raised about this CI problem?

Makman2:
Yeah the reacked commit comes from here: https://github.com/coala/coala/pull/3711

And no issue yet, shall I create one and link to it?

Makman2:
removing the doctest-fix commit 37156f5, definitely obsoleted by the latest master.

jayvdb:
Yes, please create an issue.  The underlying problem is that a second copy of coala is installed, which is a build time perf hit, and that should be fixed, not worked around.

Makman2:
We do install two versions in CI? :O

Still this is no workaround, this allows to install the docs from the repo, not from the installed packages. Creating an issue for both^^

jayvdb:
The commit message should at least start with ` docs/conf.py: ...` as it affects any use of sphinx, not just invocation via `setup.py docs`

jayvdb:
ack 9daff0e

jayvdb:
@rultor merge

PR #3789, opened 2017-02-25
ConfParser.py: Add override settings warning

yash-nisar:
Add warning if the user defines a setting in the
coafile that is already defined.

Closes https://github.com/coala/coala/issues/3316

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/parsing/ConfParser.py
+++ b/coalib/parsing/ConfParser.py
@@ -123,6 +123,12 @@ class ConfParser:
                 if key == '':
                     continue
 
+                if key in current_section.contents and keys != []:
+                    logging.warning('{} setting has already been defined in '
+                                    'section {}. The previous setting will be '
+                                    'overridden.'.format(key,
+                                                         current_section.name))
+
                 if section_override == '':
                     current_section.add_or_create_setting(
                         Setting(key,
--- a/tests/parsing/ConfParserTest.py
+++ b/tests/parsing/ConfParserTest.py
@@ -34,6 +34,11 @@ class ConfParserTest(unittest.TestCase):
     A = a, b, c
     B = a, ,, d
     C = ,,,
+
+    [name]
+    key1 = value1
+    key2 = value1
+    key1 = value2
     """
 
     def setUp(self):
@@ -176,3 +181,13 @@ class ConfParserTest(unittest.TestCase):
 
     def test_config_directory(self):
         self.uut.parse(self.tempdir)
+
+    def test_settings_override_warning(self):
+        self.assertEqual(self.cm.output[1], 'WARNING:root:test setting has '
+                                            'already been defined in section '
+                                            'foo. The previous setting will '
+                                            'be overridden.')
+        self.assertEqual(self.cm.output[2], 'WARNING:root:key1 setting has '
+                                            'already been defined in section '
+                                            'name. The previous setting will '
+                                            'be overridden.')


satwikkansal:
Commit message may be
```
Warn user when a setting is redefined in the same
section.
```

coalib/settings/Section.py
```diff
@@ -1,6 +1,7 @@
 import copy
 import os
 import sys
+import logging
 from collections import OrderedDict
```
nemaniarjun: We should make this alphabetically sorted. :)

tests/settings/SectionTest.py
```diff
@@ -1,5 +1,6 @@
 import unittest
 import os
+import logging
```
nemaniarjun: Here too.

tests/settings/SectionTest.py
```diff
@@ -22,6 +23,16 @@ def test_append(self):
         self.assertEqual(str(uut.get('5 ')), '5')
         self.assertEqual(int(uut.get('nonexistent', 5)), 5)
 
+    def test_settings_override_warning(self):
+        logger = logging.getLogger()
+        with self.assertLogs(logger, 'WARNING') as cm:
+            uut = Section('name', None)
+            uut.add_or_create_setting(Setting('key1', 'value1'), None)
+            uut.add_or_create_setting(Setting('key1', 'value2'), None, False)
+        self.assertRegex(cm.output[0], 'WARNING:root:key1 setting has already '
```
nemaniarjun: Why `assertRegex`? Why not `assertEqual`? or even `assertIn`

coalib/settings/Section.py
yash-nisar: My bad :(

tests/settings/SectionTest.py
yash-nisar: assertEqual will be much better to read. :+1:

coalib/settings/Section.py
```diff
@@ -159,6 +160,10 @@ def add_or_create_setting(self,
             val.value = str(val._value) + '\n' + setting._value
             self.append(val, custom_key=key)
         else:
+            if key in list(self.contents.keys()):
```
yash-nisar: `allow_appending` will not always be `True`. It depends on whether `keys` is an empty list or not. See https://github.com/coala/coala/blob/master/coalib/parsing/ConfParser.py#L137. It will be an empty list if the setting in a section spans multiple lines. The `.coafile` is parsed line by line, `allow_appending` will only be `True` if the setting spans multiple lines and all the lines need to be mapped to a single key.

tests/settings/SectionTest.py
```diff
@@ -22,6 +23,16 @@ def test_append(self):
         self.assertEqual(str(uut.get('5 ')), '5')
         self.assertEqual(int(uut.get('nonexistent', 5)), 5)
 
+    def test_settings_override_warning(self):
+        logger = logging.getLogger()
+        with self.assertLogs(logger, 'WARNING') as cm:
+            uut = Section('name', None)
+            uut.add_or_create_setting(Setting('key1', 'value1'), None)
+            uut.add_or_create_setting(Setting('key1', 'value2'), None, False)
```
yash-nisar: > Maybe we should have tests for the situation when allow_appending is True and the user defines the setting multiple times

This has been done. Thanks ! However adding a test for custom keys would be irrelevant IMO.

tests/settings/SectionTest.py
```diff
@@ -22,6 +23,24 @@ def test_append(self):
...
+            uut = Section('name', None)
+            uut.add_or_create_setting(Setting('key1', 'value1'), None, False)
+            uut.add_or_create_setting(Setting('key1', 'value2'), None, False)
+        self.assertEqual(cm.output[0], 'WARNING:root:key1 setting has already '
+                                       'been defined in section name. The '
+                                       'previous setting will be overridden.')
+        with self.assertLogs(logger, 'WARNING') as cm:
+            uut = Section('name', None)
+            uut.add_or_create_setting(Setting('key2', 'value1'), None, False)
+            uut.add_or_create_setting(Setting('key2', 'value2'), None, True)
```
aptrishu: swap the order.
No warning if the `allow_appending` is True, assert empty.
assert warning if False.

yash-nisar: Since we parse each line for the settings, the first time we parse a setting, `allow_appending` will be False. Now suppose our setting spans multiple lines, and when we parse the second line, `allow_appending` will be set to `True` and no warning will be issued. If the setting is encountered again, `allow_appending` will be `False` and a warning will be issued if the setting is already defined(checking `self.contents.keys()` for already defined settings). Is that okay @aptrishu ?

yash-nisar:
@Makman2 Can you have a look at this when you are free ?

tests/settings/SectionTest.py
aptrishu: If there will be no warning after this line, then verify `cm.output` as empty.

yash-nisar: Sure. :)

yash-nisar: Addressed your comment @aptrishu :)

coalib/settings/Section.py
Makman2: Can't you do `if key in self.contents`?

tests/settings/SectionTest.py
```diff
@@ -1,5 +1,6 @@
-import unittest
+import logging
 import os
+import unittest
```
Makman2: would be nice if you put the relocation of `unittest` into a different commit :)

yash-nisar: Sure. :)

yash-nisar: @Makman2 doneeeee. :D

yash-nisar: Is this what we want ?

tests/settings/SectionTest.py
```diff
@@ -1,4 +1,4 @@
-import unittest
```
Makman2: this has also to go to the relocation commit :+1:

tests/settings/SectionTest.py
```diff
@@ -1,5 +1,6 @@
 import logging
```
Makman2: commit message: The `Closes XXXX` reference is wrong :+1:

yash-nisar: I thought gitmate would report errors without it. :P

yash-nisar:
@Makman2 Is this ready for merge ?

coalib/settings/Section.py
```diff
@@ -159,6 +160,10 @@ def add_or_create_setting(self,
             val.value = str(val._value) + '\n' + setting._value
             self.append(val, custom_key=key)
         else:
+            if key in self.contents:
+                logging.warning('{} setting has already been defined in '
+                                'section {}. The previous setting will be '
+                                'overridden.'.format(key, self.name))
```
Makman2: shouldn't this be inside `ConfParser.py`? The function docstring itself says

> Adds the value of the setting to an existing setting if there is
> already a setting  with the key. Otherwise creates a new setting.

So this is intended behaviour and we shall not raise a warning here. Instead we should do so during parsing of the file.

yash-nisar: Can we modify the docstring instead ? @Makman2

Makman2: Actually the function does what it says, I'm not sure we should alter that behaviour. Imo it belongs to `ConfParser`. Any reason not to do it inside `ConfParser`?

yash-nisar:
@Makman2 Incorporated the changes suggested by you, ie.moved everything to `ConfParser.py`.

coalib/parsing/ConfParser.py
```diff
@@ -123,6 +123,12 @@ def __parse_lines(self, lines, origin):
                 if key == '':
                     continue
 
+                if key in current_section.contents and keys != []:
```
Makman2: is it possible to omit `keys != []`? either it's empty (so loop won't entered anyway) or it's `keys` which is then not empty.
As far as I understand^^

yash-nisar: It won't be possible to omit `keys != []` as that is the condition for `allow_appending` to be `False`. The loop will be entered even when `keys` is empty since we don't have any explicit condition that handles `keys` to be empty. CC @Makman2

Asnelchristian: this for loop is entered if and only if, current_keys is not empty ( hence keys ) right, ?

Asnelchristian: in that case you should probably remove ``keys != []``

yash-nisar: Nope, `current_keys` is not equal to `keys` all the time. This is the output that I receive when I remove `keys != []`
http://paste.ubuntu.com/24234922/

yash-nisar: CC @Asnelchristian @Makman2

yash-nisar: This is the output that I get when I put 
```
logging.warning(current_keys)
logging.warning(keys)
```
http://paste.ubuntu.com/24234945/

Asnelchristian:
ack https://github.com/coala/coala/pull/3789/commits/ab28131f08a230130b17ef0bc71343e6b8a0fb5b

Asnelchristian:
@rultor merge

Issue #3804, opened 2017-02-28
download_cached_file should retry if there is a http error code


jayvdb:
Follows on from https://github.com/coala/coala/issues/3803

The interface of `Bear.download_cached_file` isnt suited to failure.  It needs to retry if possible.

jayvdb:
Note unlike #3803 , this is not a regression as the pre-requests code didn't do this.

jayvdb:
Note that `requests` has native `max_retries` logic which we can use.
It requires working at the requests session & adapter level.

I've also looked for a sane requests wrapper which provides simplified retry, and oddly not found one.

jayvdb:
Note, if there is a retry, then surely there needs to be a new status code check somewhere, but this doesnt appear to be documented anywhere in requests; see https://github.com/coala/coala/pull/3805#discussion_r103427077.

jayvdb:
It looks like requests now supports `max_retries` everywhere, so this should be fairly simple.

jayvdb:
Optional: if a *transient error* occurred, such as 500+, the bear should be disabled with a warning.

However this could complexity can be deferred to be added later in https://github.com/coala/coala/issues/3332

xurror:
I would like to be assigned this issue

gr455:
What value of max_retries is expected? Also what should be an optimal backoff_factor value for the retries?

gr455:
If max_retries is hit and the error code persists, a RetryError is raised which is causing the pytest to fail. How do I handle the pytest failing?

gr455:
How about handling the RetryError by attempting to download once more(in the except statement) so that HTTPError can be raised with the latest status code

gr455:
Please review the PR and suggest changes

PR #3805, opened 2017-02-28
Bear: Raise exception on problematic HTTP status

jayvdb:
requests.Response.raise_for_status raises an exception
when the HTTP status code is not 200.

Mocked test only verifies exceptions exception is raised.
Testing of mid-stream failures not included.

Fixes https://github.com/coala/coala/issues/3803
Related to https://github.com/coala/coala-bears/issues/1461

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -423,6 +423,8 @@ class Bear(Printer, LogPrinterMixin):
                   .format(filename=filename, bearname=self.name, url=url))
 
         response = requests.get(url, stream=True, timeout=20)
+        response.raise_for_status()
+
         with open(filename, 'wb') as file:
             for chunk in response.iter_content(125):
                 file.write(chunk)
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,9 +1,11 @@
 import multiprocessing
 import unittest
 from os.path import abspath, exists, isfile, join, getmtime
-import requests_mock
 import shutil
 from time import sleep
+
+import requests
+import requests_mock
 
 from coalib.bears.Bear import Bear
 from coalib.bears.BEAR_KIND import BEAR_KIND
@@ -102,7 +104,7 @@ class DependentBear(Bear):
         yield w
 
 
-class BearTest(unittest.TestCase):
+class BearTestBase(unittest.TestCase):
 
     def setUp(self):
         self.queue = multiprocessing.Queue()
@@ -112,6 +114,9 @@ class BearTest(unittest.TestCase):
     def tearDown(self):
         if exists(self.uut.data_dir):
             shutil.rmtree(self.uut.data_dir)
+
+
+class BearTest(BearTestBase):
 
     def test_simple_api(self):
         self.assertRaises(TypeError, TestBear, self.settings, 2)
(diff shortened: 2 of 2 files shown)


jayvdb:
For mid-stream tests, we should use httpbin.

coalib/bears/Bear.py
```diff
@@ -423,9 +423,14 @@ def download_cached_file(self, url, filename):
                   .format(filename=filename, bearname=self.name, url=url))
 
         response = requests.get(url, stream=True, timeout=20)
+        response.raise_for_status()
+
         with open(filename, 'wb') as file:
             for chunk in response.iter_content(125):
                 file.write(chunk)
+
+        response.raise_for_status()
```
satwikkansal: Just wondering why to use `response.raise_for_status()` twice :sweat_smile:

satwikkansal:
Regarding commit message,
>requests.Response.raise_for_status raises an exception
when the HTTP status code is not 200.

I think it raises an exception only when the HTTP status code is above 400.

tests/bears/BearTest.py
```diff
@@ -262,6 +270,35 @@ def test_new_result(self):
...
+            self.uut.download_cached_file(
+                'http://httpbin.org/status/418', self.filename)
+
+    def test_read_failure_httpbin(self):
+        exc = requests.exceptions.HTTPError
+        with pytest.raises(exc):
+            self.uut.download_cached_file(
+                'http://httpbin.org/drip?duration=5&numbytes=5&code=500',
+                self.filename)
+
```
satwikkansal: Should we check `3XX` URL's as well?

jayvdb: This doesnt need to check all status.  While it is using a real server, it is still a very fake test covering only a basic scenario.  The same basic scenario with different codes is not worth testing.

Perhaps a useful addition is verifying that the status code `500` does appear in the exception message.  I'll add that.

coalib/bears/Bear.py
jayvdb: The first one uses the initial status code, raising an exception before opening the file.  Obviously not a problem.

I added the second one as super defensive programming (almost as a TODO item for me and the reviewer, so I am glad you questioned it).  If there is any chance of a new response during the `iter_content`, a new response code will occur, and maybe the response code is not the same as the original one.

One aspect that may trigger a new request is retries  -- surely a new request is needed to retry the fetch.  I havent thought of any other scenario where the response status code might change during the content fetch phase.

And I can leave this second (and probably useles?) `raise_for_status` out of this PR, and tackle the retries problem in the separate bug https://github.com/coala/coala/issues/3804.

tests/bears/BearTest.py
```diff
@@ -262,13 +269,44 @@ def test_new_result(self):
...
+            self.uut.download_cached_file(
+                'http://httpbin.org/status/418', self.filename)
+
+    def test_read_timeout(self):
+        exc = requests.exceptions.ReadTimeout
+        with self.assertRaisesRegexp(exc, 'Read timed out.'):
+            self.uut.download_cached_file(
+                'http://httpbin.org/delay/10',
+                self.filename,
+                timeout=5)
```
Mixih: Isn't this going to delay the test by 5 seconds?

jayvdb: The test will consume 5 seconds.  I can shorten it a bit, and it should still work ok?

fwiw, it appears that mocking is not working ; see https://github.com/coala/coala/issues/3811#issuecomment-283214574

jayvdb: OK, I've changed it to a connection timeout of 5, and a read timeout of 2, so the test should finish within 2 seconds if the connection was ok.

tests/bears/BearTest.py
```diff
@@ -1,10 +1,14 @@
...
 import unittest
+
 from os.path import abspath, exists, isfile, join, getmtime
-import requests_mock
-import shutil
 from time import sleep
 
+import requests
+
+import requests_mock
```
satwikkansal: Should there be a space b/w the imports (not aware of the import styles of coala though :sweat_smile: , I guess they're not documented anywhere) But according to PEP8
> Imports should be grouped in the following order:
    standard library imports
    related third party imports
    local application/library specific imports

jayvdb: For test modules, I like to keep functionality (i.e. requests) vs test components (requests-mock) in separate groups.

tests/bears/BearTest.py
```diff
@@ -262,13 +269,44 @@ def test_new_result(self):
         expected = Result.from_values(bear, 'test message', '/tmp/testy')
         self.assertEqual(result, expected)
 
+
+class BearDownloadTest(BearTestBase):
+
+    def setUp(self):
+        super().setUp()
+        self.mock_url = 'https://test.com'
+        self.filename = 'test.html'
```
satwikkansal: Wondering why to use `'https://test.com'` that actually resolved to some certification website in my browser :P

satwikkansal: Okay, just realized that `requests_mock.Mocker()` is a loading mechanism to ensure the adapter is correctly in place to intercept calls from requests. Ignore ^^ above comment.

satwikkansal:
LGTM :+1:

jayvdb:
> I would introduce the timeout param in a separate commit

Do we actually want a `timeout` feature?
I only added it so we can have tests that dont consume 20 seconds.
I'd rather not have a multi-commit fix for a regression that must be backported.
I'll try to think of another way to do the tests without adding the `timeout` parameter.

jayvdb:
The fix for the timeout feature is simple : remove it from this patch and come back to it later : https://github.com/coala/coala/issues/3820

jayvdb:
Note there were review comments placed on https://github.com/jayvdb/coala/commit/016125c90 by @Makman2 regarding the `import` ordering, and requesting the `timeout` feature be moved to a separate commit.

Mixih:
ack 422193a

Mixih:
ack 422193a

jayvdb:
@rutlor merge

jayvdb:
@rultor merge

Issue #3820, opened 2017-03-01
Add download read timeout failure tests


jayvdb:
An extra test has been pulled out of https://github.com/coala/coala/pull/3805 , as it was not part of the fix.

```
    def test_read_timeout(self):
        exc = requests.exceptions.ReadTimeout
        with self.assertRaisesRegexp(exc, 'Read timed out.'):
            self.uut.download_cached_file(
                'http://httpbin.org/delay/10',
                self.filename,
                timeout=(5, 2))
```

There should be tests like this, but it is best to consider them separately, as http://httpbin.org/delay/10 has a maximum 10 second delay, and the coala function has a 20 second timeout, so that httpbin feature can never be used without improving coala.

Issue #3831, opened 2017-03-02
Delete file when exception in read occurs


jayvdb:
In the same vein as https://github.com/coala/coala/issues/3803 was for status codes, if the read fails for some reason, a corrupt file will still exist. (identified by @Mixih )

This isnt a regression; the same occurred for before the rewrite https://github.com/coala/coala/pull/3612 .

jayvdb:
https://github.com/coala/coala/pull/3833 provides a test for this scenario

Mixih:
Update: still working out the tests. Something is not working with the message queue check. I should be done in a day or so.

jayvdb:
Eh?  Why is the test that I have created not good enough?  Why is it not merged yet?

Mixih:
Not your test, I was working on the test for the log output.

jayvdb:
https://github.com/coala/coala/issues/3332 should be done first.

jayvdb:
unassigning due to inactivity

PR #3833, opened 2017-03-02
BearTest: Test exception during HTTP fetch

jayvdb:
This test shows that a file is created and
still exists after a corrupted download.

Related to https://github.com/coala/coala-bears/issues/1461
Closes https://github.com/coala/coala/issues/3820

### DIFF
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,3 +1,4 @@
+from io import BytesIO
 import multiprocessing
 import unittest
 from os.path import abspath, exists, isfile, join, getmtime
@@ -268,6 +269,25 @@ class BearTest(BearTestBase):
         self.assertEqual(result, expected)
 
 
+class BrokenReadHTTPResponse(BytesIO):
+
+    def __init__(self, chunks, *args, **kwargs):
+        self.read_count = 0
+        self.chunks = chunks
+
+    def read(self, *args, **kwargs):
+        # A HTTPResponse will return an empty string when you read from it
+        # after the socket has been closed.
+        if self.closed:
+            return b''
+
+        if self.read_count == len(self.chunks):
+            raise requests.exceptions.ReadTimeout('Fake read timeout')
+
+        self.read_count += 1
+        return self.chunks[self.read_count - 1]
+
+
 class BearDownloadTest(BearTestBase):
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


tests/bears/BearTest.py
```diff
@@ -268,6 +269,26 @@ def test_new_result(self):
         self.assertEqual(result, expected)
 
 
+class BrokenReadHTTPResponse(BytesIO):
+    """A HTTPResponse-like reader that emits chunks and then fails."""
```
jayvdb: Another coala bears bug!

jayvdb:
fwiw, I've have raised a related enhancement request at https://bugs.launchpad.net/requests-mock/+bug/1669458 , for making it easier to mock a mid-stream failure.

Makman2:
I thought the idea is to remove the file again on bad download, or do you just want to add a test so we have defined behaviour?

jayvdb:
@Makman2, yes, this is just the test, preparing the way for https://github.com/coala/coala/issues/3831 .
I built it separately as it is quite difficult to generate a read failure without creating a parameter `timeout` which I believe is an anti-pattern: Bears shouldn't be doing network problem negotiation/fallback like timeouts. The download function should just work (TM), or the bear be disabled.

tests/bears/BearTest.py
```diff
@@ -268,6 +269,25 @@ def test_new_result(self):
...
+        # A HTTPResponse will return an empty string when you read from it
+        # after the socket has been closed.
+        if self.closed:
+            return b''
+
+        if self.read_count == len(self.chunks):
+            raise requests.exceptions.ReadTimeout('Fake read timeout')
+
+        self.read_count += 1
+        return self.chunks[self.read_count - 1]
```
Makman2: why reading from the opposite direction?

jayvdb: It doesnt.  `read_count` starts at 0, but is incremented on the line above here, so we need a `-1` here to index from 0.
Anyway, I've added a second chunk so that is more clear that the writer and reader are seeing the same content in the same order, until the exception happens.

Makman2: oh yeah sorry totally confused myself...

Makman2:
ack b99d2c3

Makman2:
@rultor merge

Issue #3854, opened 2017-03-06
Use clang 3.8


jayvdb:
coala should support the most recent llvm/clang, at least 3.8

jayvdb:
The CI for coala and coala-bears repos need to use a later version

jayvdb:
We also need a coala-bears PR using the same approach.

PR #3857, opened 2017-03-06
Add support for clang-3.8

gaurav1911:
Related to https://github.com/coala/coala/issues/3854

### DIFF
--- a/.misc/deps.sh
+++ b/.misc/deps.sh
@@ -14,8 +14,19 @@ case $CIRCLE_NODE_INDEX in
 esac
 
 # apt-get commands
-deps="libclang1-3.4"
+deps="libclang1-3.8"
 sudo apt-get install $deps
+
+LIBCLANG_PATH="$(find /usr/lib/ -name "libclang.so")"
+echo "$LIBCLANG_PATH"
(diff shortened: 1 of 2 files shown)


gaurav1911:
@jayvdb  should i also check clang-3.8 support in windows??

jayvdb:
I'd like to see it also work on the bears repo. Then we are in business.

gaurav1911:
on it

Mixih:
@gaurav1911 ping! can you please work on this soonish so the other PR can get unblocked? Thanks!

Mixih:
@jayvdb @gaurav1911 could you please add why this is blocked? it's currently blocking another pending pr.

jayvdb:
@Mixih , it is blocked per the comment added when blocked : https://github.com/coala/coala/pull/3857#issuecomment-284428340
i.e. we need a PR to the coala-bears repo also, demonstrating that the same voodoo works with the Clang bears, which use more of clang.

IMO we also need to now check that 3.8 works on Appveyor and Travis-macOS, otherwise we'll be having different versions of clang on each platform, which will be a different nightmare.

But first step is confirming that 3.8 works on Linux.  Then we can start moving forward on this.

Is there anything crucial depending on this?  If so, I can help push it forward.

PR #3870, opened 2017-03-07
Issue3809

aptrishu:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


aptrishu:
@Makman2 It's ready.

coalib/results/TextPosition.py
```diff
@@ -3,7 +3,6 @@
 
 
 @generate_repr('line', 'column')
-@generate_ordering('line', 'column')
```
Makman2: what about `__eq__`?

tests/results/TextPositionTest.py
```diff
@@ -56,3 +56,33 @@ def test_comparison(self):
         uut2 = TextPosition(4, 8)
         self.assertTrue(uut1 >= uut2)
         self.assertTrue(uut2 <= uut1)
+
+        uut1 = TextPosition(1, 2)
+        uut2 = TextPosition(1, 3)
+        self.assertTrue(uut1 < uut2)
+        self.assertFalse(uut1 > uut2)
```
Makman2: Maybe add those assertions for the previous cases with `<=` and `>=` too?

coalib/results/TextPosition.py
Makman2: ah you implement it. Though the `__eq__` overload has to come before this commit :+1:

aptrishu:
@Makman2 updated.

tests/results/TextPositionTest.py
```diff
@@ -36,23 +36,48 @@ def test_comparison(self):
...
         self.assertTrue(uut1 <= uut2)
+        self.assertFalse(uut1 < uut2)
+        self.assertFalse(uut1 > uut2)
 
         uut1 = TextPosition(4, 8)
         uut2 = TextPosition(4, 8)
         self.assertTrue(uut1 >= uut2)
         self.assertTrue(uut2 <= uut1)
+        self.assertFalse(uut1 > uut2)
+        self.assertFalse(uut2 < uut1)
```
Makman2: also try to assert for `==` :)

adtac:
ack https://github.com/coala/coala/pull/3870/commits/6053bc33299a4c806a4cab16dda0c7c71177ef07

coalib/results/TextPosition.py
```diff
@@ -32,6 +32,54 @@ def line(self):
...
+        consider whole file.
+
+        :param other: ``TextPosition`` to compare with.
+        :return:      Whether this ``TextPosition`` is behind the other
+                      one.
+        """
+        if self.line is None or other.line is None:
+            return False
+        if self.line == other.line:
+            return (False
```
adtac: Can you please make these regular if statements? Really hurts readability.

tests/results/TextPositionTest.py
```diff
@@ -36,23 +36,48 @@ def test_comparison(self):
...
+        uut2 = TextPosition(4, 3)
+        self.assertTrue(uut1 < uut2)
+        self.assertFalse(uut1 > uut2)
+
+        uut1 = TextPosition(1, 4)
+        uut2 = TextPosition(1, 4)
+        uut3 = TextPosition(2, 8)
+        uut4 = TextPosition(3, None)
+        uut5 = TextPosition(3, None)
+        self.assertTrue(uut1 == uut2)
```
adtac: Use `assertEqual` (and `assertNotEqual`) rather - it's generally more helpful when tests fail

adtac:
https://github.com/coala/coala/pull/3870/commits/2545c0f7a2198c3c53465385ecd8569f18aff4f8 needs work

coalib/results/TextPosition.py
aptrishu: by regular if statements you mean `if else` ?

tests/results/TextPositionTest.py
aptrishu: I thought it's easy to show operators are working properly this way.

coalib/results/TextPosition.py
adtac: yeah

tests/results/TextPositionTest.py
adtac: I dunno about that, if `a` and `b` are different, `assertTrue(a == b)` will complain that `a == b` is simply not true. But `assertEqual` will actually print the two objects.

adtac: ... which is easier to debug IMO.

Makman2: `assertEqual` uses `==` internally and the behaviour is documented officially, thus it's okay to rely on that.

coalib/results/TextPosition.py
aptrishu: @adtac 
Using regular if else may seem dummy here, as we'll be using something like
```Cython
if a:
  return True
else:
  if..
```
here if `a` is true, it'd return anyway. no need to put that in if else block.
Also, one more reason why I used separate if's is to clarify things when should it return true and when to return false according to None values, as discussed with makman2 and sils.

aptrishu:
Is this PR ok to merge?
cc @adtac @Makman2

coalib/results/TextPosition.py
adtac: Sorry for the late reply.

How about:

```python
if a:
    return True

if next_condition:
    return something_else

# and as many conditions as you want - the first one that's true will return control to the calling function

return final
```

But if this is the only thing holding up this PR, please go ahead and merge it.

aptrishu: That's what I'm doing. right? or am I missing something!!

adtac: noo, I meant the huge return statement - that logic requires more time to understand

adtac: Instead just use regular if else statements

aptrishu:
@Makman2 @adtac I have updated the return statements to make it simple.
(sorry about gitmate comments)

tests/results/TextPositionTest.py
userzimmermann: @adtac @Makman2 @aptrishu Or just use `assert uut1 == uut2` and let **pytest** do the fancy error output :)

tests/results/TextPositionTest.py
userzimmermann: @aptrishu @Makman2 And don't forget `!=`. It's always good to have this test too...

aptrishu:
cc @Makman2

aptrishu:
reack ff2e4a6af0c1b02f646ce1e83d60961ea496ed3e

aptrishu:
ack ff2e4a6af0c1b02f646ce1e83d60961ea496ed3e

aptrishu:
ah commit hashes have changed.

aptrishu:
not sure why my ack is not working.. @adtac

aptrishu:
reack ff2e4a6af0c1b02f646ce1e83d60961ea496ed3e

aptrishu:
actually, other commits need review..
and they need to be in separate commits. :)

coalib/results/TextPosition.py
```diff
@@ -32,6 +32,60 @@ def line(self):
...
+            return False
+
+        if self.line == other.line:
+            if self.column is None or other.column is None:
+                return False
+            else:
+                if self.column < other.column:
+                    return True
+                else:
+                    return False
```
Makman2: --> `return self.column < other.column`

coalib/results/TextPosition.py
```diff
@@ -32,6 +32,60 @@ def line(self):
...
+            return False
+
+        if self.line == other.line:
+            if self.column is None or other.column is None:
+                return False
+            else:
+                if self.column > other.column:
+                    return True
+                else:
+                    return False
```
Makman2: --> `return self.column > other.column`

coalib/results/TextPosition.py
```diff
@@ -47,12 +101,15 @@ def __le__(self, other):
...
-                    other.column is None
-                    else
-                    self.column <= other.column)
+            if self.column is None or other.column is None:
+                return True
+            else:
+                if self.column <= other.column:
+                    return True
+                else:
+                    return False
```
Makman2: different commit

coalib/results/TextPosition.py
```diff
@@ -71,11 +128,25 @@ def __ge__(self, other):
...
-                    other.column is None
-                    else
-                    self.column >= other.column)
+            if self.column is None or other.column is None:
+                return True
+            else:
+                if self.column >= other.column:
+                    return True
+                else:
+                    return False
```
Makman2: different commit

Makman2: --> `return self.column >= other.column`

coalib/results/TextPosition.py
Makman2: --> `return self.column <= other.column`

Makman2:
ack d77509e

Makman2:
unack f746fc4

tests/results/TextPositionTest.py
```diff
@@ -36,23 +36,54 @@ def test_comparison(self):
...
         uut1 = TextPosition(4, 8)
         uut2 = TextPosition(4, 8)
         self.assertTrue(uut1 >= uut2)
         self.assertTrue(uut2 <= uut1)
+        self.assertFalse(uut1 > uut2)
+        self.assertFalse(uut2 < uut1)
+
+        uut1 = TextPosition(1, 2)
+        uut2 = TextPosition(4, 3)
+        self.assertTrue(uut1 < uut2)
```
userzimmermann: If we wanna stick to `unittest` style then better use `self.assertLess` here and in all other similar expressions according to the other suggestions from @adtac and @Makman2

tests/results/TextPositionTest.py
```diff
@@ -36,23 +36,54 @@ def test_comparison(self):
...
+        self.assertFalse(uut1 > uut2)
```
userzimmermann: @adtac @Makman2 There is no `self.assertNotGreater` :/ Is there something else for such specific cases in `unittest`?

Makman2: hm doesn't look like. Normally you would use `assertLess` in that case, but that's not the goal here ;)

tests/results/TextPositionTest.py
aptrishu: we had this discussion that we should use `<` as assertLess also uses `<` . @Makman2

coalib/results/TextPosition.py
aptrishu: This is what I had before, then I changed it as adtac asked to use simple if else. (as it was confusing)

tests/results/TextPositionTest.py
adtac: I don't think so - although I can think of a contrived hack: wrap a `assertGreater` around an `assertRaises` :D :D (a joke - do NOT do this!)

NiklasMM: why not just use `assert uut1 > uut2` ? We use pytest anyway, why use the clunky unittest methods?

userzimmermann: @NiklasMM I love you ;P for defending `pytest` just like me :) But it seems we have no chance here! Also because people are interested in using **green** as alternative test runner: https://github.com/coala/coala/issues/4060

So join my aspects team and support the separate **coala-aspects** repo via https://github.com/coala/coala/issues/4038 . Then we can do everything as pytestastic as possible there :)

NiklasMM: Remember, how you once told me that I should be a politician? I can now safely say: You too. :grin:

userzimmermann: @NiklasMM So let's unite our political ambitions :P

tests/results/TextPositionTest.py
Makman2: actually I don't care, both are fine to me ;)

tests/results/TextPositionTest.py
Makman2: guys lgtm, I have the feeling this pytest-vs-unittest discussion goes out of control, this is a separate issue. As long as tests do what they should do, it's fine imo.

coalib/results/TextPosition.py
Makman2: this is way better to read imo + a bit faster

aptrishu:
Updated.. @Makman2 @userzimmermann @NiklasMM

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+            if self.column is None or other.column is None:
+                return False
+            else:
+                return self.column < other.column
+        else:
+            return self.line < other.line
+
+    def __gt__(self, other):
+        """
+        Test whether ``self`` is ahead of the other
```
userzimmermann: `__gt__` actually checks if `self` is *behind* `other`. Same applies to `__ge__`

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+            if self.column is None or other.column is None:
+                return True
+            else:
+                return self.column >= other.column
         else:
             return self.line > other.line
+
+    def __lt__(self, other):
+        """
+        Test whether ``self`` is behind the other
```
userzimmermann: `__lt__` actually checks if `self` is *ahead* (or *before*) `other`. Same applies to `__le__`

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+
+        If the column in a ``TextPosition`` is ``None``, consider
+        whole line. If the line in a ``TextPosition`` is ``None``,
+        consider whole file.
+
+        :param other: ``TextPosition`` to compare with.
+        :return:      Whether this ``TextPosition`` is behind the other
+                      one.
+        """
+        if self.line is None or other.line is None:
```
userzimmermann: You should really decide here what is *"smaller"*. IMO `None` should be smaller than not `None`. Same applies to `__le__`

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+
+        If the column in a ``TextPosition`` is ``None``, consider
+        whole line. If the line in a ``TextPosition`` is ``None``,
+        consider whole file.
+
+        :param other: ``TextPosition`` to compare with.
+        :return:      Whether this ``TextPosition`` is ahead of the other
+                      one.
+        """
+        if self.line is None or other.line is None:
```
userzimmermann: You should really decide here what is *"greater"*. IMO not `None` should be greater than `None`. Same applies to `__ge__`

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+            return False
+
+        if self.line == other.line:
+            if self.column is None or other.column is None:
```
userzimmermann: Same as comment on `if self.line is None ...`

coalib/results/TextPosition.py
```diff
@@ -71,11 +70,70 @@ def __ge__(self, other):
...
+
+        :param other: ``TextPosition`` to compare with.
+        :return:      Whether this ``TextPosition`` is behind the other
+                      one.
+        """
+        if self.line is None or other.line is None:
+            return False
+
+        if self.line == other.line:
+            if self.column is None or other.column is None:
```
userzimmermann: Same as comment on `if self.line is None ...`

userzimmermann:
ack https://github.com/coala/coala/pull/3870/commits/c62459b56679150a0402a06a2f95d31200eec8fc

coalib/results/TextPosition.py
aptrishu: if line is None, we consider whole file. so we really cant make any decisions as some part of `None` is ahead of `Not none` and some part before.. It has been discussed alot :D

userzimmermann: @aptrishu We have a logically inconsistent situation here:

If any `.line is None`, then `__lt__` returns `False` and `__le__` returns `True`. So the logical consequence is now that `__eq__` must always return `True` in all of those cases. But in fact, `__eq__` only returns `True` in exactly one of those cases: if both `.line` and both `.column` attributes are `None`

@sils @Makman2 @adtac @NiklasMM

aptrishu: for `__eq__` I return true only when they are exactly equal.
for `__le__` if line is none then basically one of the textposition kind of lies under the other position's range so it returns true.

Makman2: because of the possibilities of `None` values, we discussed the meaning of `None` fields:
- `column = None` -> Means any position in line
- `line = None` -> Means any position in file

So a `TextPosition` is not precise any more, it can mark many possible positions, it nearly indicates a range. Like @aptrishu said, because those `None`-ranges might span over the `TextPosition` to compare, we agreed on the current behaviour. This means in this case that for example `A < B` is not the same as `not A >= B`. The same applies for similar logical constructs.

Let's consider two `TextPositions` A and B:
- We consider A to be lower than B, if all marked positions by A are ahead of B and not touching each other. The same principle applies if A is greater than B.
- We consider A to be lower or equal than B, if nearly all positions of A are ahead of B. The remaining part is only allowed to be past B, if it was caused by a `None` value.
- We consider A to be equal to B, if they exactly match.

I think the crucial part is 2., and I can agree on another reasonable behaviour^^

userzimmermann: @aptrishu And that's the problem... Because `__lt__` returns `False`, and...

(`__lt__(other) is False` && `__le__(other) is True`) ==> `__eq__(other) is True`

... is a logical rule that is broken when only some `.line` or `.column` is `None` but not all of them

Makman2: do you have a suggestion?

userzimmermann: @Makman2 I will have soon ;)

sils:
unack https://github.com/coala/coala/pull/3870/commits/1c4cd8447c068d9c5fc84fd6063c078b7d178111 I'd change the commit message to `Simplify` instead of `Modify` because it's way easier to understand, code of that one is good though

tests/results/TextPositionTest.py
sils: I'm for using assertEqual

sils: ah

> guys lgtm, I have the feeling this pytest-vs-unittest discussion goes out of control, this is a separate issue. As long as tests do what they should do, it's fine imo.

ok let's keep this

coalib/results/TextPosition.py
sils: None "consider the whole line" so None can't be smaller than not None

coalib/results/TextPosition.py
sils: @userzimmermann so what is your suggestion?

I think to get the correct behaviour here we have to think about when this function will be used and what for. That's where this bug came from in the first place right?

sils:
unack https://github.com/coala/coala/pull/3870/commits/57fae1de0887c969eded12f78323a70668cf7557

PR #3871, opened 2017-03-07
core: Add nextgen core routines

Makman2:
The nextgen-core is a scheduler-based system. Bears can offload tasks
via `generate_tasks()` which get executed by a Python pool (in this
case a multiprocess pool).

A dependency mechanism is not implemented yet.

See https://github.com/coala/cEPs/blob/master/cEP-0002.md


coalib/core/Core.py
```diff
@@ -0,0 +1,167 @@
...
+    # FIXME   (for example Mesos).
+
+    # Set up event loop and executor.
+    event_loop = asyncio.SelectorEventLoop()
+    executor = concurrent.futures.ProcessPoolExecutor(
+        max_workers=get_cpu_count())
+
+    # Let's go.
+    schedule_bears(bears, result_callback, event_loop, {}, executor)
+    try:
```
Nosferatul: Maybe we should have an `except` here too?

Makman2: No if something crashes here it should bubble up. Exceptions here are likely core related and not bear related and shall not be swallowed

coalib/core/Core.py
```diff
@@ -0,0 +1,167 @@
+import asyncio
+import concurrent.futures
+import functools
+import logging
+import multiprocessing
+
+
+def get_cpu_count():
```
sils: if this method is only used here, can you mark it with an underscore?

Makman2: hm it's still functionality someone can use, imo it doesn't have to be "private"

Makman2: nvm will do so :+1:

coalib/core/Core.py
```diff
@@ -0,0 +1,167 @@
...
+    event_loop = asyncio.SelectorEventLoop()
+    executor = concurrent.futures.ProcessPoolExecutor(
+        max_workers=_get_cpu_count())
+
+    # Let's go.
+    schedule_bears(bears, result_callback, event_loop, {}, executor)
+    try:
+        event_loop.run_forever()
+    finally:
+        event_loop.close()
```
Udayan12167: If we catch all the exceptions properly we can show results for other bears even if one exits out. Is this needed in that case?

Makman2: If a bear exits or raises an error, other bears still continue to run. Happens here: https://github.com/coala/coala/pull/3871/commits/bb19717550d61898d2c6703354c5ca32c10fe5ba#diff-94f6470cc4dcfa1e33ff5eb050fd5191R112

tests/core/CoreTest.py
```diff
@@ -0,0 +1,184 @@
...
+
+        run({bear}, on_result)
+
+        on_result.assert_has_calls([unittest.mock.call(i) for i in range(10)],
+                                   any_order=True)
+
+    def test_run_bear_exception(self):
+        # Test exception in bear. Core needs to shutdown directly and not wait
+        # forever.
+        self.execute_run({FailingBear(self.section1, self.filedict1)})
```
Udayan12167: Check exception handling as well?

coalib/core/Core.py
Udayan12167: 👍

tests/core/CoreTest.py
Makman2: you mean I shall assert for the log message the core emits?

Makman2: maybe I should also run another bear in another test case and see whether it gets executed properly although the other one crashed^^

Udayan12167: 👍

Udayan12167:
ack https://github.com/coala/coala/pull/3871/commits/ada14dcc9cfd482ac48efe6cc61b9abcb42ed5fb

Udayan12167:
@rultor merge

Issue #3808, opened 2017-02-28
pytest.raises or self.assertRaises


jayvdb:
There are currently six(6) cases of `pytest.raises` in the coala repo, and none in the coala-bears repo, but there are many, many more uses of the `unittest` `self.assertRaises` (and friends).  These six are odd, and we should either replace them with assertRaises , or use the pytest functionality.

`pytest.raises` has one use that `unittest`'s `self.assertRaises` cant do, which is that they work in `pytest` tests functions that are not in a class, which have no `self`.
All six uses of `pytest.raises` are in class methods.

`unittest` has `self.assertRaisesRegexp` which is better than `assertRaises` (see https://github.com/coala/coala/issues/3374) .  There is https://github.com/kissgyorgy/pytest-raisesregexp to fill that void originally , but then pytest added a different pattern of doing the re match on the exception. (see https://github.com/pytest-dev/pytest/pull/1502)

If we are looking broadly, there is also a pytest decorator https://github.com/Authentise/pytest-raises , and a similar one could be built for raises-with-regexp-support.

jayvdb:
IMO, we should replace the pytest specific functionality with unittest.

NiklasMM:
What's the general consensus on the test framework issue in coala? Since pytest is used as testrunner, but unittest-style tests are generally written, I'm assuming it started out with unittest and was moved to pytest.

Personally, I prefer pytest and would encourage people to write pytest style tests. There are some cool things like test parametrization, that don't work with unittest-style tests, even when running with pytest.

Makman2:
As we don't use pytest features currently, I would use `unittest` functions to keep compatability. We should rather file issues for pytest features on demand and introduce pytest functions then if appropriate.

RJ722:
Hi!
I was trying to make the required changes, but when changing to `assertRaisesRegex`, what should be the value for `expected_regex`?

Should it be left as `.*` for now

Thanks in advance! :smile:

jayvdb:
Hi @RJ722 , `.*` would be bad.  See #3374.

RJ722:
Hi @jayvdb ! 
When changing [InstanceTest.py](https://github.com/coala/coala/blob/master/tests/bearlib/aspects/InstanceTest.py#L17-L25) ([changes](http://pastebin.com/yd5kkJED)) in, but an [error] shows up! (https://gist.github.com/RJ722/f3e852563d34c8ab1204e22d2359f0f3).

RJ722:
[...]

PR #3894, opened 2017-03-11
change pytest.raise to assertRaisesRegex

RJ722:
In tests/bearlib/aspects/InstanceTest.py change pytest.raises to
unittest.TestCase.assertRaisesRegex to check for the correct error
message. It also makes the testing consistent across the project.

up your stuff for merge!

### Checklist

- [X ] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [X ] I ran coala over my code locally. (*All commits have to pass
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


RJ722:
Partially fixes: #3808

Siddhartha1234:
Follow commit guidelines ( http://coala.io/commit ) properly.

jayvdb:
Very interesting approach.
Before 'accepting' this approach, I would like to see if the other causes of pytest.raises are as problematic.

RJ722:
>Before 'accepting' this approach, I would like to see if the other causes of pytest.raises are as problematic.

So, should we leave this issue open for a while??

jayvdb:
Yea, or you could try to fix the other ones yourself. I've upgraded the issue from newcomer to low, but you're welcome to keep working on it, as you appear to know what you are doing here.

RJ722:
There is also this factory method:
```python
testcase = unittest.TestCase.__new__(unittest.TestCase)
with testcase.assertRaisesRegex(AtributeError, any_error_str):
    # test code
    pass
```
IMO: this way it would be more intuitive as to why the ```testcase``` was generated!
Should we use this @jayvdb @userzimmermann?

userzimmermann:
@RJ722 @jayvdb I would still prefer `pytest` style like already stated in https://github.com/coala/coala/issues/3808#issuecomment-286836123

```python
with pytest.raises(AttributeError) as exc:
    # test code
assert re.search(any_error_str, str(exc.value))
```

RJ722:
Any opinion? @jayvdb @sils

madhukar01:
@userzimmermann It's just that we can validate the asserted string in the same command `assertRaisesRegex` where as in `pytest.raises` we are creating an object and then testing them by `re.search` which is searching the value inside the string ;)
I feel `assertRaisesRegex` saves more of space and time here, What do you say !

userzimmermann:
@madhukar01 Thanks for your statement! Yes... might be a bit more code... But! Look at what I just found out via https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions : 

```python
with pytest.raises(AttributeError) as exc:
    # test code
    pass
exc.match(r'...')
```

Looks structurally clearer IMO :)

@RJ722 @jayvdb

RJ722:
Yes, I believe that we can make this trade-off in performance for the sake of clarity .
So, should we start changing the tests, as proposed by @userzimmermann or does this needs more discussion?

tests/bearlib/aspects/BaseTest.py
```diff
@@ -1,7 +1,7 @@
 from coalib.bearlib.aspects import aspectclass
 from coalib.bearlib.aspects.base import aspectbase
 
-import pytest
+import unittest
 
 
 class AspectBaseTest:
```
CleanCut: `class AspectBaseTest(unittest.TestCase):`

The class containing the test methods needs to subclass `unittest.TestCase` in order to be able to use the `TestCase` methods such as `assertRaises`.  Documentation and examples are here: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises

tests/bearlib/aspects/BaseTest.py
```diff
@@ -15,5 +15,7 @@ def test_class(self):
         assert not isinstance(aspectbase, aspectclass)
 
     def test_init_needs_aspectclass(self):
-        with pytest.raises(AttributeError):
+        error_str = 'object has no attribute'
```
CleanCut: Just `with self.assertRaises(AttributeError):`

The other two lines are unnecessary, and you don't have a complicated set of Exceptions whose names you want to match via a regular expression, you are just checking for a specific exception raised (so use `assertRaises` instead).

tests/bearlib/aspects/ClassTest.py
```diff
@@ -1,12 +1,14 @@
 from coalib.bearlib.aspects import aspectclass
 
-import pytest
+import unittest
 
 
 class AspectClassTest:
```
CleanCut: `class AspectClassTest(unittest.TestCase):`

tests/bearlib/aspects/ClassTest.py
```diff
@@ -1,12 +1,14 @@
...
 
     def test_subaspect_without_definition(self, RootAspect):
-        with pytest.raises(TypeError):
+        testcase = unittest.TestCase()
+        error_str = 'must be an instance of one of'
+        with testcase.assertRaisesRegex(TypeError, error_str):
```
CleanCut: Just `with self.assertRaises(TypeError):` -- the other two lines are unnecessary.

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -1,6 +1,6 @@
 from itertools import chain
 
-import pytest
+import unittest
 
 
 class AspectInstanceTest:
```
CleanCut: `class AspectInstanceTest(unittest.TestCase):`

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -16,13 +16,12 @@ def test_tastes(
 
     def test__setattr__(self, SubAspect, SubAspect_tastes):
         aspect = SubAspect('py')
+        testcase = unittest.TestCase()
```
CleanCut: These three lines are unnecessary.

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -16,13 +16,12 @@ def test_tastes(
...
+        error_taste = "can't set taste values of aspectclass instances"
+        error_aspectless = "can't set attributes of aspectclass instances"
         for name in SubAspect_tastes:
-            with pytest.raises(AttributeError) as exc:
+            with testcase.assertRaisesRegex(AttributeError, error_taste):
```
CleanCut: `with self.assertRaises(AttributeError) as exc:`

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -16,13 +16,12 @@ def test_tastes(
...
                 setattr(aspect, name, 'value')
-            assert str(exc.value) \
```
CleanCut: You can leave this check in now.

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -16,13 +16,12 @@ def test_tastes(
...
-                == "can't set taste values of aspectclass instances"
         for name in ['docs', 'subaspects', 'tastes', '_tastes']:
-            with pytest.raises(AttributeError) as exc:
```
CleanCut: `with self.assertRaises(AttributeError) as exc:`

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -16,13 +16,12 @@ def test_tastes(
...
+            with testcase.assertRaisesRegex(AttributeError, error_aspectless):
                 setattr(aspect, name, 'value')
-            assert str(exc.value) \
```
CleanCut: You can leave this in now as well.

tests/bearlib/aspects/SubTest.py
```diff
@@ -1,7 +1,7 @@
 from coalib.bearlib.aspects import Root, aspectclass, TasteError
 from coalib.bearlib.aspects.base import aspectbase
 
-import pytest
+import unittest
 
 
 class SubAspectTest:
```
CleanCut: `class SubAspectTest(unittest.TestCase):`

tests/bearlib/aspects/SubTest.py
```diff
@@ -22,7 +22,8 @@ def test_class_tastes_items(self, SubAspect, SubAspect_tastes):
     def test__init__unavailable_taste(
             self, SubAspect, SubAspect_taste_values
     ):
-        with pytest.raises(TasteError):
+        testcase = unittest.TestCase()
```
CleanCut: This line can be deleted.

tests/bearlib/aspects/SubTest.py
```diff
@@ -22,7 +22,8 @@ def test_class_tastes_items(self, SubAspect, SubAspect_tastes):
...
+        with testcase.assertRaisesRegex(TasteError, 'is not available'):
```
CleanCut: `with self.assertRaises(TasteError):`

tests/bearlib/aspects/SubTest.py
```diff
@@ -34,7 +35,10 @@ def test_tastes(self, SubAspect, SubAspect_tastes):
                     assert getattr(aspect, name) == taste_values[name] \
                         == taste.default
                 else:
-                    with pytest.raises(TasteError):
+                    testcase = unittest.TestCase()
```
CleanCut: This line can be removed.

tests/bearlib/aspects/SubTest.py
```diff
@@ -34,7 +35,10 @@ def test_tastes(self, SubAspect, SubAspect_tastes):
...
+                    with testcase.assertRaisesRegex(
```
CleanCut: These three lines can just be `with self.assertRaises(TasteError):`

CleanCut:
I hope you all don't mind that I jumped in and reviewed this PR.  Just thought I could be helpful.  I'm quite...well-versed...in unittest internals.

sils:
@CleanCut we don't mind, we're very open and we allow newcomers to review right when they start contributing because we believe that sitting on both sides of the table is good for learning - and that "the old grumpy maintainers" can very much learn something fresh from our excited newcomers however experienced they are! In that sense: thanks a lot for your support!

CleanCut:
Heh, I *am* an old grumpy maintainer.

Okay, well, I guess  I'm not *always* grumpy... ;-)

tests/bearlib/aspects/BaseTest.py
jayvdb: I believe @RJ722 tried that in a previous iteration of this patch, and it caused strange errors.
This area of the code has an extreme level of meta-programming, which may be part of the problem.

tests/bearlib/aspects/ClassTest.py
jayvdb: Actually, we want to lock down the exception message as well, so we have a more precise knowledge of which exception is being triggered.
Due to lots of meta-programming, there is always the chance that a different TypeError is being hit, and test failures are a great way to help experienced users quickly guess what has happened when a newcomer has inadvertently broken something beyond their understanding.

tests/bearlib/aspects/BaseTest.py
CleanCut: Ah, well meta-programming can certainly complicate things.

RJ722:
@gitmate-bot  is showing up errors because it is running current tests on old commits also.

tests/bearlib/aspects/BaseTest.py
```diff
@@ -15,5 +15,7 @@ def test_class(self):
...
+        with pytest.raises(AttributeError) as exc:
```
jayvdb: The issue is about removing dependency on pytest-specific functionality in the tests.

RJ722: @userzimmermann deprecated the use of unittest and since pytest is the test runner and is also backwards compatible. So, IMHO, it would be good to reconsider as there won't be any visible performance improvements, neither would it make code any more readable.
I'ld highly respect your decision @jayvdb

thisch:
Note that with pytest 3.1 the usage of assertRaisesRegex generates a deprecation warning. Therefore, I've created #4276, which replaces assertRaisesRegex by pytest.raises.

RJ722:
Should I close the PR?

tests/bearlib/aspects/BaseTest.py
jayvdb: There was a period where lots of pytest-only tests where created in aspects.
We had already decided to use unittest compatible tests, but were still trying to understand and fix the pytest-only test problems.
This was a case of left-hand not talking with right hand.
The result is we are still approving pytest-only tests in aspects code, but trying to prevent the problem growing elsewhere.

jayvdb:
I thought this PR had a working solution to removing some of the `pytest.raises`, which is necessary, but the current code changes are not particularly useful.

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -13,14 +13,14 @@ def test_tastes(
...
+        error_taste = "A 'taste' value for this aspectclass instance "\
+            'exists already.'
+        error_aspectless = "can't set attributes of aspectclass instances"
         for name in SubAspect_tastes:
             with pytest.raises(AttributeError) as exc:
                 setattr(aspect, name, 'value')
-            assert str(exc.value) \
-                == "A 'taste' value for this aspectclass instance "\
-                   'exists already.'
+            assert exc.match(error_taste)
```
jayvdb: This change doesnt change anything. Actually it is less precise that the previous test code.

.coafile
```diff
@@ -28,6 +28,16 @@ bears = PyUnusedCodeBear
 language = Python
 remove_all_unused_imports = true
 
+[flakes]
```
jayvdb: this looks unrelated to your change.

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -31,12 +31,13 @@ def test_tastes(
 
     def test__setattr__(self, SubAspect, SubAspect_tastes):
         aspect = SubAspect('py')
+        error_taste = "A 'taste' value for this aspectclass instance "\
+            'exists already.'
+        error_aspectless = "can't set attributes of aspectclass instances"
```
jayvdb: unused.

tests/bearlib/aspects/InstanceTest.py
jayvdb: (It does at least get rid of the trailing `\`, so I'll accept it ;-) )

tests/bearlib/aspects/InstanceTest.py
```diff
@@ -31,12 +31,13 @@ def test_tastes(
 
     def test__setattr__(self, SubAspect, SubAspect_tastes):
         aspect = SubAspect('py')
+        error_taste = "A 'taste' value for this aspectclass instance "\
```
jayvdb: use round brackets, not `\`

tests/bearlib/aspects/InstanceTest.py
RJ722: Actually, I squashed an additional commit by mistake which did the overwriting on this commit. I've rectified it.

Thanks for the quick review @jayvdb! :-)

jayvdb:
ack https://github.com/coala/coala/pull/3894/commits/91b0ea27a32a32a3122f9cd4d011b36473729152

jayvdb:
@gitmate-bot rebase

Issue #3898, opened 2017-03-12
misc/DictUtilities.py: Switch into defaultdict


corona10:
`inverse` is now `dict`.
If we switch into defaultdict, we don't have to check whether it is None or not.

PR #3899, opened 2017-03-12
misc/DictUtilities.py: Use defaultdict

corona10:
defaultdict is more useful in this use-case.
we don't have to check whether it is None or not.
Also now, we can remove add_pair_to_dict method.

Closes https://github.com/coala/coala/issues/3898

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/misc/DictUtilities.py
+++ b/coalib/misc/DictUtilities.py
@@ -1,4 +1,4 @@
-from collections import Iterable, OrderedDict
+from collections import defaultdict, Iterable, OrderedDict
 
 
 def inverse_dicts(*dicts):
@@ -11,31 +11,21 @@ def inverse_dicts(*dicts):
     No order is preserved.
 
     :param dicts: The dictionaries to invert.
+    :type dicts:  dict
     :return:      The inversed dictionary which merges all dictionaries into
                   one.
+    :rtype: defaultdict
     """
-    inverse = {}
+    inverse = defaultdict(list)
 
     for dictionary in dicts:
         for key, value in dictionary.items():
             if isinstance(value, Iterable):
                 for item in value:
-                    add_pair_to_dict(item, key, inverse)
+                    inverse[item].append(key)
             else:
-                add_pair_to_dict(value, key, inverse)
-
+                inverse[value].append(key)
     return inverse
-
-
-def add_pair_to_dict(key, value, dictionary):
-    """
-    Add (key, value) pair to the dictionary. The value is added to a list of
-    values for the key.
-    """
-    if key in dictionary:
-        dictionary[key].append(value)
-    else:
-        dictionary[key] = [value]
 
 
 def update_ordered_dict_key(dictionary, old_key, new_key):


coalib/misc/DictUtilities.py
```diff
@@ -32,10 +32,7 @@ def add_pair_to_dict(key, value, dictionary):
     Add (key, value) pair to the dictionary. The value is added to a list of
     values for the key.
     """
-    if key in dictionary:
-        dictionary[key].append(value)
-    else:
-        dictionary[key] = [value]
+    dictionary[key].append(value)
```
jayvdb: This is a public function, which means it can be used by other code which you cant see.
Your change requires that the parameter `dictionary` is a `defaultdict(list)` , which is not a guarantee.

You should use `dictionary[key].append(value)` in `inverse_dicts` , and deprecate `add_pair_to_dict`.

coalib/misc/DictUtilities.py
```diff
@@ -11,26 +11,31 @@ def inverse_dicts(*dicts):
...
-                    add_pair_to_dict(item, key, inverse)
+                    inverse[item].append(key)
             else:
-                add_pair_to_dict(value, key, inverse)
-
+                inverse[value].append(key)
     return inverse
 
 
 def add_pair_to_dict(key, value, dictionary):
```
sils: actually I think you should just remove this function, it's not used by anything!

sils:
otherwise good, commit message I'd go for "Use defaultdict" instead of S"Switch into" if I was nitpicky :) Great stuff, let's remove that function though

coalib/misc/DictUtilities.py
```diff
@@ -11,26 +11,31 @@ def inverse_dicts(*dicts):
     No order is preserved.
 
     :param dicts: The dictionaries to invert.
+    :type: dict
     :return:      The inversed dictionary which merges all dictionaries into
```
NiklasMM: the `return` string should state that a defaultdict is returned.

coalib/misc/DictUtilities.py
```diff
@@ -11,26 +11,31 @@ def inverse_dicts(*dicts):
...
+                inverse[value].append(key)
     return inverse
 
 
 def add_pair_to_dict(key, value, dictionary):
     """
     Add (key, value) pair to the dictionary. The value is added to a list of
     values for the key.
+    :param key: A key for value.
+    :param value: A value for append.
```
NiklasMM: These doc strings could be clearer: 
```
:param key: The key to append the value to.
:param value: The value to be appended.
```

coalib/misc/DictUtilities.py
jayvdb: In the previous review, I requested that it is deprecated.

corona10: @jayvdb @sils 
Which is better? Removing this function? or notice to a user that It is deprecated and will be removed?

corona10:
@sils @NiklasMM @jayvdb Updated!! PTAL.

NiklasMM:
[this](https://github.com/coala/coala/pull/3899#discussion_r105552597) remains unadressed ;)

coalib/misc/DictUtilities.py
sils: this function is not marked as private but I don't see anyone externally seriously importing stuff from coalib.misc.DictUtilities, and it's not in bears. I'm voting for removal definitely although idealisticly you would have to deprecate but realistically I'm super sure nobody is and would not want to maintain this fuction further.

If we deprecate it it should be decorated or so by a deprecation decorator.

adtac: FWIW I think we should remove it - it just adds to code baggage. It's specific to this file and it's not used anywhere.

corona10: @sils @adtac I removed it and updated it :)

coalib/misc/DictUtilities.py
```diff
@@ -11,33 +11,23 @@ def inverse_dicts(*dicts):
     No order is preserved.
 
     :param dicts: The dictionaries to invert.
+    :type: defaultdict
```
jayvdb: This syntax is not correct

corona10:
Updated PTAL.

corona10:
@sils @jayvdb 
I updated code. please take a look. 
Circle ci needs to retry the build. because it was failed with timeout.

jayvdb:
ack 32e2b5b

jayvdb:
@rultor merge

PR #3905, opened 2017-03-12
coala_main: Non-zero exit code if errors logged

NiklasMM:
Adds a new logging handler `CounterHandler`, which
counts the number of logged errors.
Adds logic in coala_main to exit with a non-zero
code if any errors were logged during the run.

Closes https://github.com/coala/coala/issues/3087

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/output/Logging.py
+++ b/coalib/output/Logging.py
@@ -1,8 +1,35 @@
 from datetime import datetime
+from collections import Counter
 import json
 import io
 import logging
 import logging.config
+
+
+class CounterHandler(logging.Handler):
+    """
+    A logging handler which counts the number of calls
+    for each logging level.
+    """
+    _call_counter = Counter()
+
+    @classmethod
+    def reset(cls):
+        """
+        Reset the counter to 0 for all levels
+        """
+        cls._call_counter.clear()
+
+    @classmethod
+    def emit(cls, record):
+        cls._call_counter[record.levelname] += 1
+
+    @classmethod
+    def get_num_calls_for_level(cls, level):
+        """
+        Returns the number of calls registered for a given log level.
+        """
+        return cls._call_counter[level]
 
 
 def configure_logging():
@@ -11,6 +38,9 @@ def configure_logging():
     """
     import sys
 
+    # reset counter handler
+    CounterHandler.reset()
+
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
@@ -18,11 +48,14 @@ def configure_logging():
                 'class': 'logging.StreamHandler',
                 'formatter': 'colored',
                 'stream': sys.stderr
+            },
+            'counter': {
+                'class': 'coalib.output.Logging.CounterHandler'
             }
         },
         'root': {
             'level': 'DEBUG',
-            'handlers': ['colored']
+            'handlers': ['colored', 'counter']
         },
         'formatters': {
             'colored': {
@@ -48,6 +81,9 @@ def configure_json_logging():
     """
     stream = io.StringIO()
 
+    # reset counter handler
+    CounterHandler.reset()
+
     logging.config.dictConfig({
         'version': 1,
         'handlers': {
@@ -55,11 +91,14 @@ def configure_json_logging():
                 'class': 'logging.StreamHandler',
                 'formatter': 'json',
                 'stream': stream
+            },
+            'counter': {
+                'class': 'coalib.output.Logging.CounterHandler'
             }
         },
         'root': {
             'level': 'DEBUG',
-            'handlers': ['json']
+            'handlers': ['json', 'counter']
         },
         'formatters': {
             'json': {
--- a/tests/coalaCITest.py
+++ b/tests/coalaCITest.py
@@ -23,6 +23,8 @@ class coalaCITest(unittest.TestCase):
             coala_ci.main, 'coala-ci', '--help')
         self.assertIn('usage: coala', stdout)
         self.assertIn('Use of `coala-ci` binary is deprecated', stderr)
+        self.assertEqual(retval, 0,
+                         'coala must return zero when successful')
 
     def test_nonexistent(self):
         retval, stdout, stderr = execute_coala(
@@ -31,6 +33,8 @@ class coalaCITest(unittest.TestCase):
         self.assertRegex(
             stderr,
             ".*\\[ERROR\\].*The requested coafile '.*' does not exist. .+\n")
+        self.assertNotEqual(retval, 0,
+                            'coala must return nonzero when errors occured')
 
     def test_find_no_issues(self):
         with bear_test_module(), \
@@ -46,7 +50,7 @@ class coalaCITest(unittest.TestCase):
             self.assertIn('Executing section cli', stdout)
             self.assertFalse(stderr)
             self.assertEqual(retval, 0,
-                             'coala-ci must return zero when successful')
+                             'coala must return zero when successful')
 
     def test_find_issues(self):
         with bear_test_module(), \
@@ -61,7 +65,7 @@ class coalaCITest(unittest.TestCase):
                           'The output should report count as 1 lines')
             self.assertIn('This result has no patch attached.', stderr)
             self.assertNotEqual(retval, 0,
-                                'coala-ci was expected to return non-zero')
+                                'coala must return nonzero when errors occured')
 
     def test_show_patch(self):
         with bear_test_module(), \
@@ -75,7 +79,7 @@ class coalaCITest(unittest.TestCase):
             self.assertIn('Line contains ', stdout)  # Result message is shown
             self.assertIn("Applied 'ShowPatchAction'", stderr)
             self.assertEqual(retval, 5,
-                             'coala-ci must return exitcode 5 when it '
+                             'coala must return exitcode 5 when it '
                              'autofixes the code.')
 
     def test_fail_acquire_settings(self):
@@ -86,3 +90,5 @@ class coalaCITest(unittest.TestCase):
                                                    '-c', os.devnull)
             self.assertFalse(stdout)
             self.assertIn('During execution, we found that some', stderr)
+            self.assertNotEqual(retval, 0,
+                                'coala was expected to return non-zero')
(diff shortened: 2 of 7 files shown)


NiklasMM:
Great catch @jayvdb ! I added a bunch of asserts.

tests/coalaCITest.py
```diff
@@ -86,3 +86,5 @@ def test_fail_acquire_settings(self):
                                                    '-c', os.devnull)
             self.assertFalse(stdout)
             self.assertIn('During execution, we found that some', stderr)
+            self.assertNotEqual(retval, 0,
+                                'coala-ci was expected to return non-zero')
```
jayvdb: coala-ci -> coala

NiklasMM:
@jayvdb I found most of the successful runs already had assertions for exit code 0.

coalib/output/Logging.py
```diff
@@ -1,10 +1,26 @@
...
+class CounterHandler(logging.Handler):
+    """
+    A logging handler which counts the number of calls
+    for each logging level.
+    """
+    level_called = None
+
+    def __init__(self, *args, **kwargs):
+        super(CounterHandler, self).__init__(*args, **kwargs)
+        CounterHandler.level_called = defaultdict(lambda: 0)
```
userzimmermann: @NiklasMM Hi :)

**[0]** Since we only support Python 3, `super().__init__(...)` is enough here

**[1]** Why initializing `level_called` in `__init__` when it is always used on class level? Just initialize it directly on defining it above. Then `__init__` also becomes completely obsolete. And **[0]** becomes so too :) And **most important**, then `level_called` is not re-initialized when `CounterHandler` is intantiated more than once for whatever reason...

**CC** @jayvdb

coalib/output/Logging.py
```diff
@@ -1,10 +1,26 @@
...
+
+    def emit(self, record):
+        CounterHandler.level_called[record.levelname] += 1
```
userzimmermann: Better use `type(self)` instead of `CounterHandler`. So you get independent from the class name.

Or even more better ;) Turn `emit` into:

```python
@classmethod
def emit(cls, record):
    cls.level_called[record.levelname] += 1
```

coalib/output/Logging.py
NiklasMM: 0) Whoops. Old habits ;)

1) You're right. I initially didn't have it as a class member, but a regular instance member. But it turns out there is no convenient way to access the handlers of a logger, so I redesigned it. I'll implement your suggestions.

coalib/output/Logging.py
NiklasMM: Good catch! Thanks!

coalib/output/Logging.py
NiklasMM: Hm, so I had to leave the `__init__` in, so the counter would get reset for each run. It would probably work "in the wild" because coala will only run once per import, but for the tests, I see no other way.

userzimmermann: @NiklasMM U could also define a `pytest.fixture` in a **tests/conftest.py** that resets `level_called` for every test where it's needed:

```python
@pytest.fixture(scope='function')
def reset_logging_counter(request):
    CounterHandler.level_called = defaultdict(lambda: 0)
    return CounterHandler
```

Then you just add a `reset_logging_counter` argument to every `test_...` method where the dict shall be reset:

```python
def test_...(self, reset_logging_counter):
    ...
```

And you can then also perform additional asserts on the `reset_logging_counter` class itself, like checking if the counting really works correctly for all levels...

It would also be generally good to define an explicit `reset` method:

```python
class CounterHandler(...):
    ...
    @classmethod
    def reset(cls):
        cls.level_called = defaultdict(lambda: 0)
    ...

CounterHandler.reset()
```

Then you can also use `reset` in the fixture and not reimplement logic there which has actually to be tested :)

```python
@pytest.fixture(scope='function')
def reset_logging_counter(request):
    CounterHandler.reset()
    return CounterHandler
```

You can read more about fixtures here: http://doc.pytest.org/en/latest/fixture.html

I still think it's not good when every instantiation of `CounterHandler` resets the counters.

**CC** @jayvdb @sils

userzimmermann: Or you write a `test_...` method decorator that resets the counters...

So many possibilities ;)

NiklasMM: Yes, that would work, but I don't understand your issue with the resetting of the counter when an instance of the counter is created. The instance is create once for a run, when the logging is set up. Isn't resetting the counter then exactly what we want?

userzimmermann: @NiklasMM Well... I was just thinking about that `CounterHandler` could maybe be derived and used for additional logging to other interfaces like web or whatever, with explicit extra messages for those interfaces, what would cause multiple instantiation...

**CC** @sils

NiklasMM: The cleanest thing to do would be to have it as a regular instance member. But as I pointed it's a bit hard to access the handlers via the logger. I'm gonna go into a bit of detail, maybe you have a better idea :smile: 

So in `coala_main` I could get the logger and then access the list of handlers in `logger.handlers`. I could loop through it and select the one that is an instance of `CounterHandler`. But that's kind of awkward to do and it creates a problem with branch coverage, as I would need to add a test for the case where the CounterHandler isn't there, which should never be the case. 

Hope that's not too confusing. Would appreciate any ideas :)

NiklasMM:
This PR is somewhat stuck and I'd like to get it going again, so I'm gonna try to summarize the issues I see:

* @userzimmermann and I were debating over the design of the `CounterHandler` [here](https://github.com/coala/coala/pull/3905#discussion_r106227061). I outlined [here](https://github.com/coala/coala/pull/3905#discussion_r106514646) why it's awkward to have the `level_called` member on instance level instead of class level. But taking the `__init__` away destroys the tests, because the counter is only reset on import.

* @jayvdb has pointed out a number of tests which mere missing assertions and I have added a bunch. There are still some occurrences of `execute_coala` in `CachingTest.py` without them, but I'd argue that checking the exit code is not in the scope of those tests.

So yeah, if you two (and maybe others) could chime in again, that would help me get this going again :)

userzimmermann:
@NiklasMM @jayvdb I would still prefer the explicit `.reset()` classmethod as proposed in https://github.com/coala/coala/pull/3905#discussion_r106275802 , together with the simplest way to get it called on every test that needs a reset. I makes also the testing procedure clearer instead of explicitly having to know that the class-based counter gets implicitly reset by creating a new instance.

NiklasMM:
I'd still argue that as a user I would expect the behavior that the counter is reset for each run of coala. This is now only a problem for tests and your suggestion would of course fix that, but I could see this greatly confuse people if we ever have more than one run for each import.

I'd suggest this as a comprimise: I will do a `reset()` method, but I will call it in `configure_logging` which will fix the problem for both tests and production. What do you say, @userzimmermann ?

userzimmermann:
@NiklasMM I like your compromise! And can't wait to see it implemented :)

NiklasMM:
Wait no more! It's here!

NiklasMM:
I milestoned this, since I really would like this to be in the next release :)

userzimmermann:
> Wait no more! It's here!

This I adore! I owe you a beer!

But WTF'ing coverage is going on there:

```
coalib/output/Logging.py      25      0      4      1    97%   14->exit
```

Starting from https://github.com/coala/coala/pull/3905/files#diff-4a49662a37e70525cb34f6b722cff846R14

How is that possible? :D

coalib/output/Logging.py
```diff
@@ -1,28 +1,52 @@
...
+class CounterHandler(logging.Handler):
+    """
+    A logging handler which counts the number of calls
+    for each logging level.
+    """
+    level_called = defaultdict(lambda: 0)
+
+    @classmethod
+    def reset(cls):
+        '"" Reset the counter to 0 for all levels ""'
```
userzimmermann: And one last thing... quite strange doc string delimiters ;)

NiklasMM: lol? QuotesBear must have been drunk

coalib/output/Logging.py
```diff
@@ -1,28 +1,54 @@
...
 import logging
 import logging.config
 
 
+class CounterHandler(logging.Handler):
+    """
+    A logging handler which counts the number of calls
+    for each logging level.
+    """
+    level_called = None
```
jayvdb: Have you tried using a `collections.Counter` here?

And add a custom 'get' and 'set' method so that `level_called` can be private?

NiklasMM:
@jayvdb Thanks for the suggestion! I changed it to use a `Counter` object and a getter method.

userzimmermann:
ack https://github.com/coala/coala/pull/3905/commits/df0804cb82008670beec7748e27a58307131130d

userzimmermann:
@NiklasMM From my POV it's done :)

@jayvdb What do you think?

userzimmermann:
@NiklasMM As long as @jayvdb doesn't approve... What do you think about the following idea? Not that you should implement it! ;) I'm just interested in your opinion :)

Instead of `get_num_calls_for_level`:

```python
class Meta(type(logging.Handler)):
    def __getitem__(cls, level):
        return cls._call_counter[level]

class CounterHandler(logging.Handler, metaclass=Meta):
    ...
```

```python
>>> CounterHandler['ERROR']
42
```

jayvdb:
I would like `__getitem__` instead of `get_num_calls_for_level`, but we can play with other ideas in followup patches.

No need for a metaclasses to achieve that.

It also could use instance variables instead of class variables, and then one instance is stored somewhere.  You never know when another instance of the class is needed, to count something else, or a subset of messages, etc.

jayvdb:
ack df0804c

jayvdb:
reack b56590c

jayvdb:
reack b56590c

jayvdb:
@rultor merge

userzimmermann:
@jayvdb AFAIK you would need to fetch the handler instances out of the `logging` framework. And then all this *"storing somewhere"*... Instead of creating other instances of the class to count something else you could also work with subclasses. But those are of course just ideas... and all that is up to future discussions... when it becomes really useful :) Good that we have this merged for now!

Issue #3906, opened 2017-03-12
Fail to install and py.test on docker environment.


corona10:
When I try to install by `python setup.py install`, it is failed with this message.
`UnicodeEncodeError: 'ascii' codec can't encode character '\xfc' in position 15224: ordinal not in range(128)`
Also, the same happening when I try to run unit test on local.
It needs to be fixed.

jayvdb:
Which docker image are you using?

corona10:
@jayvdb I am using ubuntu by `docker pull ubuntu:latest`.

Makman2:
Confirmed :+1:

Launch docker: `docker run -it ubuntu:latest bash`
Inside docker:
```
root@8fcd31b218fc:/# apt-get update
root@8fcd31b218fc:/# apt-get install -y git python3 python3-pip
root@8fcd31b218fc:/# git clone https://github.com/coala/coala
root@8fcd31b218fc:/# cd coala/
root@8fcd31b218fc:/coala# pip3 install -r requirements -r test-requirements.txt
root@8fcd31b218fc:/coala# python3 setup.py install
```

Makman2:
That even happens on latest `fedora` image...

corona10:
@Makman2 
I wrote little description why this happening on https://github.com/coala/coala/pull/3908.
It looks like that coala expected utf-8 setting locale on [this side](https://github.com/coala/coala/blob/master/setup.py#L18).
But unfortunately, They do not raise (ValueError, UnicodeError) but just return
`(None, None)`. 
This is why this issue is happening. Also, for the unit tests, there are some codes which are reading `utf-8` files. so it should be set locale before running unit tests.

PR #3908, opened 2017-03-12
setup.py, tests/__init__.py: Catch (None, None) locale

corona10:
Sometimes docker images have locale problem.
I found that `locale.getlocale()` can
return (None, None).
So I fixed that when it return (None, None) then setlocale for it.
* setup.py: Catch (None, None) locale
* tests/__init__.py: Catch (None, None) locale

Fixes https://github.com/coala/coala/issues/3906

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,10 @@ from coalib import VERSION, assert_supported_version, get_version
 from coalib.misc.BuildManPage import BuildManPage
 
 try:
-    locale.getlocale()
+    lc = locale.getlocale()
+    pf = platform.system()
+    if pf != 'Windows' and lc == (None, None):
+        locale.setlocale(locale.LC_ALL, 'C.UTF-8')
 except (ValueError, UnicodeError):
     locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
 
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,11 @@
+import locale
+import platform
+
+
+try:
+    lc = locale.getlocale()
+    ps = platform.system()
+    if ps != 'Windows' and lc == (None, None):
+        locale.setlocale(locale.LC_ALL, 'C.UTF-8')
+except (ValueError, UnicodeError):
+    locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')


setup.py
```diff
@@ -15,7 +15,10 @@
 from coalib.misc.BuildManPage import BuildManPage
 
 try:
-    locale.getlocale()
+    lc = locale.getlocale()
+    pf = platform.system()
+    if pf != 'Windows' and lc == (None, None):
```
Makman2: why checking explicitly for windows?

tests/__init__.py
```diff
@@ -0,0 +1,10 @@
+import locale
+import platform
+
```
Makman2: add a newline ;)

Makman2:
I would maybe split up your commit into two parts, one changing that for installation, one for the tests

\+

Commit message suggestion (when split up):
- `setup.py: Catch (None, None) locale`
- `tests/__init__.py: Catch (None, None) locale`

setup.py
corona10: @Markman2 
I tested on both Linux platform and osx platform.
They worked fine.
but on windows, it looks like has different locale settings.
But I don't have windows machine.

supergr35:
Have you tested if this issue occurs on windows?

corona10:
@supergr35 
I do not test on Windows.
First of all, I don't have Windows machine. 
Second, But it looks like they also return `(None, None)` on windows.
Because when I removed checking platform codes, it was failed on `appveyor/pr` with this message.
```
tests\__init__.py:6: in <module>
    locale.setlocale(locale.LC_ALL, 'C.UTF-8')
C:\Python34\lib\locale.py:592: in setlocale
    return _setlocale(category, locale)
E   locale.Error: unsupported locale setting
```

So I guessed locale problem is not affected on Windows but affected on Linux/Unix OS.

setup.py
Makman2: yeah but you have no reason to exclude it. Windows might too produce `(None, None)` and then the error slips through again :+1:

Makman2: Nvm just read your post in the PR...

Makman2:
ugh that's weird. As we don't have tests for setup.py I believe, I would test that myself too (or someone else).

--> **Please don't merge until we have tested this properly! Either manually or even better by unit tests!**

corona10:
@Makman2 That's not setup.py. It is tests/__init__.py. Python will execute __init__.py before running test script.

corona10:
@markman2 
Also, this issue is relevant with when we execute `py.test`. 
Because unit test files read `*.rst` file by utf-8 decode.
I also wrote at https://github.com/coala/coala/issues/3906.
as like this. `Also, the same happening when I try to run unit test on local.`

corona10:
@Makman2 And I tested this codes on docker environment and it was succeeded. 
Is there any way to add unit test codes for this environment-sensitive issue?

Makman2:
Yep, we have to add a test file inside `tests/` and reconstruct the conditions for the environment to return `(None, None)` as locale. Then running `setup.py` while installing into a temporary folder. Also we would need to test test-execution itself somehow (without running all tests again)... not that easy, but possible

supergr35:
ping @corona10 , I realize that this is going to take a while, but just checking if you're still working on this.

supergr35:
@Makman2 wouldn't testing test-execution be more of a different issue?

corona10:
@supergr35 Yeah, I looking into this issue. I definitely agree with @Makman2 's opinion. I currently pause for working on this issue for mid-exam. If you can, cherry-pick and proceed it.

Makman2:
@supergr35 thing is we modify the test execution code with something that might break. Although it really is a bit hard, let's try. If it's too hard, I would ack stuff without this "test-for-tests", as the same code for `setup.py` needs definitely tests.

supergr35:
I'm not sure if I've got enough knowledge to accomplish something like that, so I probably shouldn't be the one to cherry pick.

jayvdb:
ack 302bd96 f7da89b

jayvdb:
I think tests-for-tests is not worth the effort here.
Once it is merged, it gets a lot of testing.
If it breaks, we learn something and maybe revert.

jayvdb:
reack a0c9666 d8de173

jayvdb:
@rultor merge

Issue #3929, opened 2017-03-15
TasteTest: Add missing field


Techievena:
```
>>> taste.__dict__
{'suggested_values': (), 'languages': (), 'default': None, 'description': ''}
```
`taste.languages` field missing in `test__init__defaults` function

PR #3936, opened 2017-03-16
TasteTest.py: Add missing field

madhukar01:
taste.languages field was added to test_init_defaults function

Fixes https://github.com/coala/coala/issues/3929

### DIFF
--- a/tests/bearlib/aspects/TasteTest.py
+++ b/tests/bearlib/aspects/TasteTest.py
@@ -1,5 +1,6 @@
 from coalib.bearlib.aspects import Taste
 from coalib.bearlib.aspects.taste import TasteMeta
+from coalib.bearlib.languages import Languages
 
 
 class TasteTest:
@@ -17,6 +18,10 @@ class TasteTest:
         assert taste.description is ''
         assert taste.suggested_values is ()
         assert taste.default is None
+        assert type(taste.languages) is Languages
+        assert not taste.languages
+        assert not len(taste.languages)
+        assert taste.languages == ()
 
     def test__get__(
             self, SubAspect, SubAspect_tastes, SubAspect_taste_values):


Techievena:
Fix the tests as suggested in the test logs.

corona10:
`tests/bearlib/aspects/TasteTest.py:19: AssertionError`
Please run `py.test` on your local environment. 😄

corona10:
@Madhukar-SH I recommend you to run `$ coala` on your local environment before pushing. :smile:

Madhukar-SH:
@corona10 I did run py.test and was success, forgot to run coala :p 
Will run both from next time :)

tests/bearlib/aspects/TasteTest.py
```diff
@@ -1,5 +1,6 @@
 from coalib.bearlib.aspects import Taste
 from coalib.bearlib.aspects.taste import TasteMeta
+from coalib.bearlib.languages import Languages
```
pratyushprakash: Why is this being imported here?

arjunsinghy96:
Do not merge from `coala/master`. Instead, do a `rebase`. You need to `reset` last merge commit and `rebase` your branch properly.

Madhukar-SH:
@arjunsinghy96 Done, Ty ;)

tests/bearlib/aspects/TasteTest.py
Madhukar-SH: Cleared ;)

tests/bearlib/aspects/TasteTest.py
```diff
@@ -17,6 +17,7 @@ def test__init__defaults(self):
         assert taste.description is ''
         assert taste.suggested_values is ()
         assert taste.default is None
+        assert taste.languages in [(), ('C#',)]
```
pratyushprakash: Why is `C#` in the default languages?

Madhukar-SH: https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/taste.py#L38

adhikasp: @Madhukar-SH the line you referencing is just docstring. The Taste code itself define only empty tuple `()` as default languages.

Madhukar-SH: @adhikasp Look at this https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/taste.py#L75

Madhukar-SH: I suggest you to run the code and check the output before preferring changes :)

supergr35:
Quick thing in the commit body. Make sure you're in the imperative mood. For example: `x was added to y` should become `Add x to y`. I'm pretty sure you can figure out the change you need to make 😉 but let me know if you need help.

supergr35:
3ced873

supergr35:
ack 3ced873

userzimmermann:
unack https://github.com/coala/coala/pull/3936/commits/3ced873cbd94fe6afe71c2cedccc4cb08c4214ba

Madhukar-SH:
@userzimmermann Did you run the code and check the outputs as I suggested?

userzimmermann:
@Madhukar-SH The expected situation for default `languages` on instantiating `Taste()` without arguments is:

```python
assert type(taste.languages) is Languages
assert not taste.languages
assert not len(taste.languages)
assert taste.languages == ()
```

Whereby the latter 2 checks should be moved to a `LanguagesTest`.

If there would be any other state possible, then it wouldn't be correct and couldn't be allowed in the tests. It would be a bug. But even that is not the case. The code is working correctly.

Can you anyway please provide an example what you mean with *"run the code and check the outputs"*. Please show that run command and output you are referring to.

userzimmermann:
@Madhukar-SH BTW: Now that I realized that tests for the `Languages` class are missing you could go on with https://github.com/coala/coala/issues/4037 after this PR :)

madhukar01:
@userzimmermann Yes I agree with you. But for the tests being run. what you told isnt appropriate !

First I checked what would be the value by adding `print(self.laguages)` and creating a `Taste()` object.
Result was `()` empty tuple. See here -> https://snag.gy/RHI4sq.jpg

And then again I ran the `py.test` with `print(self.languages)` present, the output is here -> https://snag.gy/KBMV6E.jpg

Making it `assert taste.languages is ()` would just return an assertion error and wouldn't help us.

userzimmermann:
@madhukar01 `TasteTest.test__init__defaults` is an explicit test for instantiating `Taste()` without any arguments. As @adhikasp already mentioned in https://github.com/coala/coala/pull/3936#discussion_r108041178 , the second case from your `print(...)`s comes from the doc string example which you correctly identified in https://github.com/coala/coala/pull/3936#discussion_r107747924 . But `TasteTest.test__init__defaults` is not related in any way to that example. That example is implicitly verified via `doctest`. `.languages` will always depend on what you put in via `Taste(..., languages=[...])`. That can be any combination of any supported language, not just **C#**. But that's not the business of `TasteTest.test__init__defaults`. Sorry, but you are mixing up unrelated things...

madhukar01:
@userzimmermann Can you please explain me why did the output come out to be like that as mentioned [here](https://github.com/coala/coala/pull/3936#issuecomment-291056376) ?

madhukar01:
See this https://snag.gy/KBMV6E.jpg

userzimmermann:
@madhukar01 Because the **C#** comes from the `Taste` instantiation in the doc string:

```python
>>> Taste[bool](
    ...     'Ignore ``using`` directives in C#.',
    ...     (True, False), default=False,
    ...     languages=('CSharp', )
    ... ).languages
    (C#,)
```

But `TasteTest.test__init__defaults` doesn't operate on that `Taste` instance. The test method is only run once and only operates on the local `taste` object created in that test method at https://github.com/coala/coala/blob/master/tests/bearlib/aspects/TasteTest.py#L16 . That's the only purpose of that test method.

madhukar01:
@userzimmermann If we change it to `assert taste.languages is ()` it's simply returning assertion error while running py.test

userzimmermann:
@madhukar01 Because `taste.languages is not ()` (the empty tuple object). `is` checks if you have exactly the same object (at the same memory address). `taste.languages` is an empty `Languages` object. `Languages` is derived from `tuple`, but `Languages([]) is not ()`. That's why appropriate checks are (as mentioned before):

```python
assert type(taste.languages) is Languages
assert not taste.languages
assert not len(taste.languages)
assert taste.languages == ()
```

madhukar01:
Well @userzimmermann Thanks for making things clear :) 
Now shall I just add `assert type(taste.languages) is Languages` or do I need to add all of them?

userzimmermann:
@madhukar01 You are welcome! I'm glad that we found that common understanding :) Both of us learned something... I learned how to better explain ;)

As I already told in https://github.com/coala/coala/pull/3936#issuecomment-290832814 , not all those `asserts`s are really necessary at this point, since the integrity of a `Languages` object should be checked in a `LanguagesTest`, which is already getting implemented via https://github.com/coala/coala/pull/4047

But let's keep the current state. Better testing a bit too much than too less :)

I hope this PR wasn't too exhaustive ;) and you will pick the next coala issue ASAP

userzimmermann:
ack https://github.com/coala/coala/pull/3936/commits/e8df8eb90c3988edfcfe5e072d909cb9a0183079

userzimmermann:
@rultor merge

Issue #3940, opened 2017-03-17
Missing function in coalib/misc/Shell.py


madhukar01:
Fucntion call_without_output missing in shell.py 
Which is needed fot installation of cib
` File "/home/helios/coala_env/lib/python3.5/site-packages/coala_package_manager/installation_tool.py", line 29, in <module>
    from coalib.misc.Shell import call_without_output
ImportError: cannot import name 'call_without_output'
`

jayvdb:
The comment later added here is correct

https://github.com/coala/coala/pull/3942#issuecomment-288910998

> Since cib is broken and fixing this issue will not be solving the problem, I'm closing this PR

This issue should also be closed.

PR #3942, opened 2017-03-17
Shell.py: Add missing function

Madhukar-SH:
Function call_without_output was added to Shell.py
which was found missing during cib installation

Fixes https://github.com/coala/coala/issues/3940


Madhukar-SH:
Since cib is broken and fixing this issue will not be solving the problem, I'm closing this PR

Issue #3864, opened 2017-03-07
Attrs in Bear v2.


Udayan12167:
> we're starting to use the attrs library a bit and I think this would be a good fit, remove those three functions which are just proxies for the dependency dict and make it an _dependency_results = attr.ib(init=collections.defaultdict(list))

#3711

bhavishyagopesh:
@Udayan12167 Could you please elaborate the issue here.

Makman2:
Have you looked at the `attrs` library? It allows to define conveniently fields and behaviour of a class. The idea is to use that for `coalib.core.Bear`.

RaiVaibhav:
@Siddhartha1234 working on issue ,as i would like to work on it ??

RaiVaibhav:
@Siddhartha1234 still working?

Makman2:
Reassigning due to inactivity.

RaiVaibhav:
Kindly unassign me I am not working on this anymore.

palash25:
I can give it a crack. Maybe after GSoC.

Makman2:
Since it's quite some time till end of GSoC, leaving it open for now :+1:

PR #3959, opened 2017-03-20
Bear.py: Add attrs to Bears v2

Siddhartha1234:
Add attrs for section, file_dict and _dependency_results

Closes https://github.com/coala/coala/issues/3864

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/core/Bear.py
+++ b/coalib/core/Bear.py
@@ -7,10 +7,10 @@ from os.path import join, abspath, exists
 
 from appdirs import user_data_dir
 
-from coala_utils.decorators import (enforce_signature, classproperty,
-                                    get_public_members)
+from coala_utils.decorators import classproperty, get_public_members
 
 import requests
+import attr
 
 from coalib.results.Result import Result
 from coalib.settings.ConfigurationGathering import get_config_directory
@@ -18,6 +18,7 @@ from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
 
 
+@attr.s
 class Bear:
     """
     A bear contains the actual subroutine that is responsible for checking
@@ -27,6 +28,16 @@ class Bear:
     This is the base class for every bear. If you want to write a bear, you
     will probably want to look at the ``ProjectBear`` and ``FileBear`` classes
     that inherit from this class.
+
+    To instantiate a bear, pass a section which contains the bear setting and
+    a file-dictionary (that maps filenames to the according file contents):
+
+    >>> section = Section('my-section')
+    >>> file_dict = {'file.py': ['This is a one-liner\\n']}
+    >>> bear = Bear(section, file_dict)
+
+    On instantiation, a ``RuntimeError`` is raised when the bear requirements
+    are not fulfilled.
 
     To indicate which languages your bear supports, just give it the
     ``LANGUAGES`` value which should be a set of string(s):
@@ -131,6 +142,13 @@ class Bear:
     ASCIINEMA_URL = ''
     BEAR_DEPS = set()
 
+    section = attr.ib(hash=False,
+                      validator=attr.validators.instance_of(Section))
+    file_dict = attr.ib(hash=False,
+                        validator=attr.validators.instance_of(dict))
+    _dependency_results = attr.ib(init=False, default=defaultdict(list),
+                                  hash=False)
+
     @classproperty
     def name(cls):
         """
@@ -179,24 +197,11 @@ class Bear:
                 if cls.MAINTAINERS_EMAILS == set() else
                 cls.MAINTAINERS_EMAILS)
 
-    @enforce_signature
-    def __init__(self, section: Section, file_dict: dict):
+    def __attrs_post_init__(self):
         """
-        Constructs a new bear.
-
-        :param section:
-            The section object where bear settings are contained.
-        :param file_dict:
-            The file-dictionary containing a mapping of filenames to the
-            according file contents.
-        :raises RuntimeError:
-            Raised when bear requirements are not fulfilled.
+        This function is called after all attributes are initialized.
+        It checks for prerequisites.
         """
-        self.section = section
-        self.file_dict = file_dict
-
-        self._dependency_results = defaultdict(list)
-
         self.setup_dependencies()
         cp = type(self).check_prerequisites()
         if cp is not True:
@@ -277,7 +282,8 @@ class Bear:
         """
         # Those members get duplicated if they aren't excluded because they
         # exist also as fields.
-        excluded_members = {'can_detect', 'maintainers', 'maintainers_emails'}
+        excluded_members = {'can_detect', 'maintainers', 'maintainers_emails',
+                            'file_dict', 'section'}
 
         # json cannot serialize properties, so drop them
         data = {
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 appdirs~=1.4
+attrs~=16.3.0
 coala_utils~=0.6.2
 colorlog~=2.7
 dependency_management~=0.3.1


ashwini0529:
@Siddhartha1234 try always to run `coala` on your code before pushing it. 😄

coalib/core/Bear.py
```diff
@@ -179,25 +185,11 @@ def maintainers_emails(cls):
                 if cls.MAINTAINERS_EMAILS == set() else
                 cls.MAINTAINERS_EMAILS)
 
-    @enforce_signature
-    def __init__(self, section: Section, file_dict: dict):
+    def __attrs_post_init__(self):
```
adhikasp: Does this mean the rather helpful docs string on how to construct a bear and it's argument get lost? Is there anyway to keep the doc string?

Siddhartha1234: @adhikasp I can keep the Docstring If you wish . Any tip @Makman2 @Udayan12167 ?

Makman2: Definitely don't delete the docstring

Makman2: Or try to incorporate it into the auto-generated constructor.

coalib/core/Bear.py
```diff
@@ -179,23 +196,24 @@ def maintainers_emails(cls):
...
+        """
+        This function is called after all attributes are initialized.
+        It checks for pre-requisites.
         """
-        self.section = section
-        self.file_dict = file_dict
 
-        self._dependency_results = defaultdict(list)
+        if not isinstance(self.section, Section):
```
Siddhartha1234: added manual isinstance check as enforce_signature is not allowed for Attributes and tests were there to assert TypeErrors. This can be improved by including enforce_signature support for Attributes in coala-utils. @Makman2  is it okay for now?

requirements.txt
```diff
@@ -8,3 +8,5 @@ Pygments~=2.1
 PyPrint~=0.2.6
 requests~=2.12
 setuptools>=17.0
+attrs~=16.3.0
+
```
Makman2: redundant newline

coalib/core/Bear.py
```diff
@@ -130,6 +132,21 @@ class Bear:
...
+        Attributes for a new bear.
+
+        :param section:
+            The section object where bear settings are contained.
+        :param file_dict:
+            The file-dictionary containing a mapping of filenames to the
+            according file contents.
+        :raises RuntimeError:
+            Raised when bear requirements are not fulfilled.
+    """
```
Makman2: Hm I want to have this documentation in the constructor somehow... or somewhere available where everyone would look instantly. Is there some possibility? @Siddhartha1234

coalib/core/Bear.py
```diff
@@ -130,6 +132,21 @@ class Bear:
...
+    section = attr.ib(hash=False)
+    file_dict = attr.ib(hash=False)
+    _dependency_results = attr.ib(init=False, default=defaultdict(list),
+                                  hash=False)
```
Makman2: actually where's the difference over assigning those inside a custom `__init__`?

requirements.txt
Siddhartha1234: Travis was not recognizing attrs~=16.3.0 unless there was  a newline .It was saying module not found . Other CIs had no issues.

coalib/core/Bear.py
Siddhartha1234: @Makman2 attr is a decorator which generates an __init__ flexibly . The whole point is not only about not having to write an init. We can utilize the properties of attrs .Like , we can obtain all attributes as a dict or json .We can attach hooks to each attribute which will execute after that is initialized etc. That should be a separate PR IMO . I'll be glad to add that in a new PR. This PR just adds the attrs, not It's usecases :) .

requirements.txt
Makman2: that sounds really strange and this shouldn't happen. Please remove this line. If CI fails I'll look at it :+1:

coalib/core/Bear.py
Makman2: > we can obtain all attributes as a dict or json

this is in fact already possible by using `vars(instance)` or `dir(instance)` ;)

> We can attach hooks to each attribute which will execute after an attribute is initialized 

so that's what we would do inside the `__init__` anyway, right?

> This PR just adds the attrs, not It's usecases :)

Yeah I understood that intention already from @sils :) But the bear is in fact working, so I don't really see to change something in terms of "layout".

Don't want to sound too strict or so, I just want to settle things with `attrs`, as it came accross me from some people, telling me like "oh man super cool stuff", but I can't see that for `coalib.core.Bear` :/

all3fox: @Makman2 just two cents here: `attrs` generates comparisons/string representations "quickly" and only gives slight overhead at class definition but not at runtime. These might be the "selling points" you are trying to hear.

As for this particular use case, `attrs` probably does not differ much from just writing the `__init__` manually. You could argue, that you do not have to implement `__repr__` and `__str__` and can even compare bears with `__le__` and the like.

Makman2: for `repr` we have thankfully `@generate_repr` :D

> ... and can even compare bears with __le__ and the like.

The problem is: it makes no sense to compare them, except `==` (though we need to be also careful with that, the new core defines whether considering two bears equal, this plays a large role in the bear-dependency system). The default instance check is the safest currently imo^^

Many other classes would really benefit from `attr.s`, though it's a bit tricky for this one. It's used in multiple processes and contains partially unhashable content^^

Siddhartha1234: @Makman2 As Lasse said , even though the usecases might be limited or tricky as you mentioned.It's a good idea to store all data members which are holding important data as an Attribute. So that the `attrs.as_dict` of the entire core repo will contain the important data members.This can be a starting point to improve our Data Model. Distinguishing between data_members is the most important aspect of `attrs` , other features have been implemented in some way or the other in the repo already.

coalib/core/Bear.py
```diff
@@ -179,23 +196,24 @@ def maintainers_emails(cls):
...
+        """
+        This function is called after all attributes are initialized.
+        It checks for pre-requisites.
         """
-        self.section = section
-        self.file_dict = file_dict
 
-        self._dependency_results = defaultdict(list)
+        if not isinstance(self.section, Section):
+            raise TypeError('section must be a Section')
```
sils: you can add a validator in attrs

coalib/core/Bear.py
sils: FWIW I really like this declarative style vs. having a custom init

coalib/core/Bear.py
Siddhartha1234: @sils That way i have to define 3 different functions outside the class having the same content as these 3 lines separately as validators don't auto-raise exceptions and just accept a callable entity. Is that okay ?

coalib/core/Bear.py
```diff
@@ -179,24 +198,19 @@ def maintainers_emails(cls):
                 if cls.MAINTAINERS_EMAILS == set() else
                 cls.MAINTAINERS_EMAILS)
 
-    @enforce_signature
-    def __init__(self, section: Section, file_dict: dict):
+    @staticmethod
+    def setup_dependencies():
```
Makman2: hm where does this one come from?

coalib/core/Bear.py
Makman2: Still not sure, though let's do it and see how it turns out :+1:

coalib/core/Bear.py
```diff
@@ -130,6 +132,23 @@ class Bear:
...
+    """
+        Attributes for a new bear.
+
+        :param section:
+            The section object where bear settings are contained.
+        :param file_dict:
+            The file-dictionary containing a mapping of filenames to the
+            according file contents.
+        :raises RuntimeError:
+            Raised when bear requirements are not fulfilled.
```
Makman2: as we can't document `__init__` any more, we should move that to class level and embed it into the text somehow, like
```
On instantiation, a `RuntimeError` is raised when the bear requirements are not fulfilled.
```

coalib/core/Bear.py
```diff
@@ -179,24 +198,19 @@ def maintainers_emails(cls):
...
-        self.section = section
-        self.file_dict = file_dict
-
-        self._dependency_results = defaultdict(list)
 
+    def __attrs_post_init__(self):
+        """
+        This function is called after all attributes are initialized.
+        It checks for pre-requisites.
+        """
```
Makman2: As `__attrs_post_init__` is known how to work for `attrs`, you could omit the docstring. But up to you :)

requirements.txt
```diff
@@ -8,3 +8,4 @@ Pygments~=2.1
 PyPrint~=0.2.6
 requests~=2.12
 setuptools>=17.0
+attrs~=16.3.0
```
Makman2: could you place it in the alphabetically right position? makes it easier to find ;)

coalib/core/Bear.py
Siddhartha1234: I misplaced that function while deleting init, Will fix it in next update.

coalib/core/Bear.py
Siddhartha1234: @Makman2 You mean in the main docstring? Or in a class property ?

coalib/core/Bear.py
Siddhartha1234: Okay, I will . :)

coalib/core/Bear.py
Makman2: in the docstring of the class :+1:

coalib/core/Bear.py
```diff
@@ -28,6 +30,15 @@ class Bear:
     will probably want to look at the ``ProjectBear`` and ``FileBear`` classes
     that inherit from this class.
 
+    Attributes for a new bear:
+
+    section:   The section object where bear settings are contained.
+    file_dict: The file-dictionary containing a mapping of filenames to the
+    according file contents.
+
+    On instantiation, a `RuntimeError` is raised when the bear requirements
```
Makman2: `RuntimeError` in double backticks :+1:

coalib/core/Bear.py
```diff
@@ -28,6 +30,15 @@ class Bear:
     will probably want to look at the ``ProjectBear`` and ``FileBear`` classes
     that inherit from this class.
 
+    Attributes for a new bear:
```
Makman2: Maybe you should incorporate a doctest, something like
```
To instantiate a bear, pass a section which contains the bear setting and a file-dictionary (that maps filenames to the according file contents):

>>> section = Section('my-section')
>>> file_dict = {'file.py': ['This is a one-liner\n'']}
>>> Bear(section, file_dict)
```
As when I read "Attributes for a new bear", I wouldn't relate that I have to pass the following values into the constructor.

requirements.txt
```diff
@@ -7,4 +8,4 @@ libclang-py3~=3.4.0
 Pygments~=2.1
 PyPrint~=0.2.6
 requests~=2.12
-setuptools>=17.0
+setuptools>=17.0
```
Makman2: hm this change doesn't seem to change anything

coalib/core/Bear.py
```diff
@@ -179,24 +197,11 @@ def maintainers_emails(cls):
...
-
-        :param section:
-            The section object where bear settings are contained.
-        :param file_dict:
-            The file-dictionary containing a mapping of filenames to the
-            according file contents.
-        :raises RuntimeError:
-            Raised when bear requirements are not fulfilled.
+        This function is called after all attributes are initialized.
+        It checks for pre-requisites.
```
Makman2: "prerequisite" is a legal word, so I would remove the `-` if you want to keep the docstring ;)

requirements.txt
Makman2: Please put this one into a new commit with a nice commit message containing that it screws up your CI ;)

coalib/core/Bear.py
```diff
@@ -28,6 +29,16 @@ class Bear:
     will probably want to look at the ``ProjectBear`` and ``FileBear`` classes
     that inherit from this class.
 
+    To instantiate a bear, pass a section which contains the bear setting and
```
adtac: IMO this should be in its own commit. A commit should do exactly one thing.

adtac: I fully welcome the addition :)

coalib/core/Bear.py
adtac: Was the docstring somewhere else? If so, ignore my previous comment.

requirements.txt
adtac: what's the change here?

coalib/core/Bear.py
Makman2: Imo it's okay, as the constructor docstring is moved into the class docstring.

coalib/core/Bear.py
```diff
@@ -131,6 +142,13 @@ class Bear:
     ASCIINEMA_URL = ''
     BEAR_DEPS = set()
 
+    section = attr.ib(hash=False,
+                      validator=attr.validators.instance_of(Section))
+    file_dict = attr.ib(hash=False,
+                        validator=attr.validators.instance_of(dict))
+    _dependency_results = attr.ib(init=False, default=defaultdict(list),
```
sils: careful this is mutable

sils: use attr.Factory as per https://attrs.readthedocs.io/en/stable/examples.html#defaults

sils:
@Siddhartha1234 let us know if we should take this over, it's almost ready for merge and just needs small changes, a rebase and the splitting of the commits

Makman2:
Closing due to inactivity.

Issue #3952, opened 2017-03-19
One test fails locally: OpenEditorActionTest.test_environ_editor


all3fox:
Here is how to reproduce it on my machine:

```
virtualenv3 coala-venv-3.6.0
source ./coala-venv-3.6.0/bin/activate
pip3 install -r docs-requirements.txt
pytest
```

Then, it collects 620 items to test, runs the tests and informs me of a single failure:

```
======================================================= short test summary info ========================================================
FAIL tests/results/result_actions/OpenEditorActionTest.py::OpenEditorActionTest::test_environ_editor

=============================================================== FAILURES ===============================================================
_______________________________________________ OpenEditorActionTest.test_environ_editor _______________________________________________

self = <tests.results.result_actions.OpenEditorActionTest.OpenEditorActionTest testMethod=test_environ_editor>

    def test_environ_editor(self):
        old_environ = os.environ
    
        file_dict = {self.fb: ['1\n', '2\n', '3\n']}
        diff_dict = {}
        subprocess.call = self.fake_edit
        with open(self.fb, 'w') as handle:
            handle.writelines(file_dict[self.fb])
        result = Result.from_values('origin', 'msg', self.fb)
    
        # Currently an ``editor`` param is required, so this will raise
        # a ``TypeError``.
        from coalib.results.result_actions.OpenEditorAction import (
            OpenEditorAction)
        action = OpenEditorAction()
        with self.assertRaises(TypeError):
>           action.apply(result, file_dict, diff_dict)
E           AssertionError: TypeError not raised

tests/results/result_actions/OpenEditorActionTest.py:161: AssertionError
================================================ 1 failed, 619 passed in 32.91 seconds =================================================
```

Alright, this test seems to be doing something with a spawned editor. My `.zshrc` defines a function called `emacs` (named exactly like the editor):

```
function emacs() {/usr/bin/emacs $@ -mm &}
```

This is my convenience function that spawns `/usr/bin/emacs`, passes all of the parameters to it through `$@`, then forces maximized mode with `-mm` and finally detaches it from console with `&`.

So, I commented it out and restarted my shell. The changes took effect because `emacs` was:

```
alex@tardis ~/Code/coala (git)-[master] % type emacs
emacs is /usr/bin/emacs
```
However, the tests still failed on that single test case.

Finally, I did `echo $EDITOR` and for some reason it showed me `vi`.

[...]

PR #3960, opened 2017-03-21
OpenEditorActionTest: Unset EDITOR before test

adtac:
Some users have the `EDITOR` environment variable set before
running the tests.

Fixes https://github.com/coala/coala/issues/3952

### DIFF
--- a/tests/results/result_actions/OpenEditorActionTest.py
+++ b/tests/results/result_actions/OpenEditorActionTest.py
@@ -152,6 +152,12 @@ class OpenEditorActionTest(unittest.TestCase):
             handle.writelines(file_dict[self.fb])
         result = Result.from_values('origin', 'msg', self.fb)
 
+        # Some users have the ``EDITOR`` environment variable set, so
+        # we should remove that temporarily.
+        if 'EDITOR' in os.environ:
+            del os.environ['EDITOR']
+            reload(coalib.results.result_actions.OpenEditorAction)
+
         # Currently an ``editor`` param is required, so this will raise
         # a ``TypeError``.
         from coalib.results.result_actions.OpenEditorAction import (


all3fox:
That was me, this issue fixes my problem (but many ppl have `$EDITOR=vi` set by default from what I hear, not just me).

ack d68de3d71d9abc38f2c19bed989d5896431239c0

all3fox:
For the record, appveyor seems to be failing for unrelated reasons: absence of `curl` and absence of some `*.gcno` file.

adtac:
Yeah, all PRs are failing.

tests/results/result_actions/OpenEditorActionTest.py
```diff
@@ -152,8 +152,14 @@ def test_environ_editor(self):
...
         result = Result.from_values('origin', 'msg', self.fb)
 
+        # Some users have the ``EDITOR`` environment variable set, so
+        # we should remove that temporarily.
+        if 'EDITOR' in os.environ:
+            del os.environ['EDITOR']
+
         # Currently an ``editor`` param is required, so this will raise
         # a ``TypeError``.
+        reload(coalib.results.result_actions.OpenEditorAction)
```
NiklasMM: Could the `reload` not go into the if?

adtac: Right, it should.

tests/results/result_actions/OpenEditorActionTest.py
```diff
@@ -152,6 +152,12 @@ def test_environ_editor(self):
             handle.writelines(file_dict[self.fb])
         result = Result.from_values('origin', 'msg', self.fb)
 
+        # Some users have the ``EDITOR`` environment variable set, so
+        # we should remove that temporarily.
```
NiklasMM: The "temporarily" doesn't really apply here, we never set it back. I think just "we should remove it, to test the behavior if it is missing" or something should suffice

adtac: actually, we do: see `old_environ`

NiklasMM:
ack dc78d41

adtac:
@rultor merge

PR #3966, opened 2017-03-21
Define bear aspects the magic metaclass way

userzimmermann:
Implementation for https://github.com/coala/coala-bears/issues/1517

* Following my proposal in https://github.com/coala/coala-bears/pull/1391#discussion_r106233373 - slightly modified from list to dict ;)
* Allowing:

```python
from coalib.bearlib.aspects.Metadata import CommitMessage
```

```python
class SomeAspectsBear(Bear, aspects={
    'detect': [CommitMessage.Shortlog.ColonExistence],
    'fix': [CommitMessage.Shortlog.TrailingPeriod],
 }):
    ...
```

```python
>>> SomeAspectsBear.aspects
{'detect': [<aspectclass 'Root.Metadata.CommitMessage.Shortlog.ColonExistence'>],
 'fix': [<aspectclass 'Root.Metadata.CommitMessage.Shortlog.TrailingPeriod'>]}
```

@sils @Asnelchristian @Techievena @adhikasp @andreiraduta1101

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -21,8 +21,10 @@ from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
 from coalib.settings.ConfigurationGathering import get_config_directory
 
+from .meta import bearclass
 
-class Bear(Printer, LogPrinterMixin):
+
+class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
     """
     A bear contains the actual subroutine that is responsible for checking
     source code for certain specifications. However it can actually do
@@ -139,6 +141,24 @@ class Bear(Printer, LogPrinterMixin):
     ...     SEE_MORE = 'https://www.pylint.org/'
     >>> PyLintBear.SEE_MORE
     'https://www.pylint.org/'
+
+    In the future, bears will not survive without aspects. aspects are defined
+    as part of the ``class`` statement's parameter list. According to the
+    classic ``CAN_DETECT`` and ``CAN_FIX`` attributes, aspects can either be
+    only ``'detect'``-able or also ``'fix'``-able:
+
+    >>> from coalib.bearlib.aspects.Metadata import CommitMessage
+
+    >>> class aspectsCommitBear(Bear, aspects={
+    ...         'detect': [CommitMessage.Shortlog.ColonExistence],
+    ...         'fix': [CommitMessage.Shortlog.TrailingPeriod],
+    ... }):
+    ...     pass
+
+    >>> aspectsCommitBear.aspects['detect']
+    [<aspectclass 'Root.Metadata.CommitMessage.Shortlog.ColonExistence'>]
+    >>> aspectsCommitBear.aspects['fix']
+    [<aspectclass 'Root.Metadata.CommitMessage.Shortlog.TrailingPeriod'>]
     """
 
     LANGUAGES = set()
--- a/coalib/bears/meta.py
+++ b/coalib/bears/meta.py
@@ -1,0 +1,25 @@
+from collections import defaultdict
+
+
+class bearclass(type):
+    """
+    Metaclass for :class:`coalib.bears.Bear.Bear` and therefore all bear
+    classes.
+
+    Pushing bears into the future... ;)
+    """
+
+    # by default a bear class has no aspects
+    aspects = defaultdict(lambda: [])
+
+    def __new__(mcs, clsname, bases, clsattrs, *varargs, aspects=None):
+        return type.__new__(mcs, clsname, bases, clsattrs, *varargs)
+
+    def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None):
+        """
+        Initializes the ``.aspects`` dict on new bear classes from the mapping
+        given to the keyword-only `aspects` argument.
+        """
+        type.__init__(cls, clsname, bases, clsattrs, *varargs)
+        if aspects is not None:
+            cls.aspects = defaultdict(lambda: [], aspects)
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from io import BytesIO
 import multiprocessing
 import unittest
@@ -8,6 +9,7 @@ from time import sleep
 import requests
 import requests_mock
 
+from coalib.bearlib.aspects.Metadata import CommitMessage
 from coalib.bears.Bear import Bear
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.bears.GlobalBear import GlobalBear
@@ -105,6 +107,25 @@ class DependentBear(Bear):
         yield w
 
 
+class aspectsTestBear(Bear, aspects={
+        'detect': [CommitMessage.Shortlog.ColonExistence],
+        'fix': [CommitMessage.Shortlog.TrailingPeriod],
+}):
+    pass
+
+
+class aspectsDetectOnlyTestBear(Bear, aspects={
+        'detect': [CommitMessage.Shortlog.ColonExistence],
+}):
+    pass
+
+
+class aspectsFixOnlyTestBear(Bear, aspects={
+        'fix': [CommitMessage.Shortlog.TrailingPeriod],
+}):
+    pass
+
+
 class BearTestBase(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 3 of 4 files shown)


userzimmermann:
Oops! Why is https://github.com/coala/coala/pull/3966/commits/b0db667c33c1adc00e38b2b9e040f4a845c844bf part of this PR? :D

userzimmermann:
Foreign commit removed! :D

userzimmermann:
@pratyushprakash You should also take a look at this PR and tell your opinion :)

coalib/bears/Bear.py
```diff
@@ -130,6 +132,24 @@ class Bear(Printer, LogPrinterMixin):
     >>> class SomeBear(Bear): pass
     >>> SomeBear.source_location
     '...Bear.py'
+
+    In the future, bears won't survive without aspects. Aspects are defined
+    as part of the ``class`` statement's parameter list. According to the
+    classic ``CAN_DETECT`` and ``CAN_FIX`` attributes, aspects can either be
+    only ``'detect'``able or also ``'fix'``able:
```
Asnelchristian: ``'detec'``

coalib/bears/meta.py
```diff
@@ -0,0 +1,25 @@
...
+
+class bearclass(type):
+    """
+    Metaclass for :class:`coalib.bears.Bear.Bear` and therefore all bear
+    classes.
+
+    Pushing bears into the future... ;)
+    """
+
+    # by default a bear class has no aspects
```
Techievena: why not use Root aspect as default aspect for bear class. So when user wants to run bears explicitly, all the aspects covered by bear will run as Root aspect is parent aspect of every aspect.

coalib/bears/meta.py
```diff
@@ -0,0 +1,25 @@
...
+    aspects = defaultdict(lambda: [])
```
pratyushprakash: Won't `Bear.aspects` be of different type (`defaultdict` or `dict`) depending on weather or not the bear was passed the `aspects` dict?

coalib/bears/meta.py
adhikasp: If a bear have "Root" aspect doesn't that will mean that "this bear can handle ALL possible aspect"?

coalib/bears/meta.py
adhikasp: [From what I understand](http://stackoverflow.com/questions/6589814/what-is-the-difference-between-dict-and-collections-defaultdict) they are same type of things but defaultdict have... default value. CMIIW, in this case if `aspects` is not defined in bear declaration then we can safely call SomeAspectsBear.aspects['detect'] without explicitly handling dict key not found error.

pratyushprakash: Yeah I was just thinking of explicitly defining `aspects['detect'] = []` to maintain consistency of `type(aspects)`.

userzimmermann: @pratyushprakash @adhikasp Thanks :) Good points! I chose the way of consistently using `defaultdict`: https://github.com/coala/coala/pull/3966/files#diff-7544b01c348d9f3cfa526b33644dc42aR25

userzimmermann:
@adhikasp IMO we should avoid any overlap and handle things according to your first option:

> Exclusively use metaclass to define aspect, metadata only for backward compatibility.

coalib/bears/meta.py
Techievena: I would like rephrase your last line "this bear can handle ALL possible aspects it can". With the help of aspects we can restrict the analysis of a bear. We can force it to analyse only a particular aspect, but we never can do an analysis the bear cannot do because bear is just wrapping the tool. So my point was if by default the aspect was root aspect, it can be beneficiary at the point when bears are called explicitly. We will divide bears into segments and a segment is run only when we pass that aspect and as Root aspect is parent aspect every aspect the segment will run even when no aspects are passed.

Techievena: It might be confusing because I am not able to express my thoughts properly. Please correct me. :smiley:

pratyushprakash: I don't think this is a good idea. Aspects are supposed to be things that the bear can analyze and fix. Root is just the parent of all aspects. We don't expect the user to ever mention `Root` as the aspect to be analyzed. It makes no sense to put `Root` in `aspects.detect` as it is not some physical aspect of the code which the bear analyzes and is hence counter-productive. 
When bears are called eexplcitly, we would have to have a different bear collection methodology as when called via aspects so adding `Root` won't help much anyway.

Techievena: I apologise for that. What I should be talking about is passing the aspects not initiating them. This is surely not the right place to toggle this discussion because this is the metaclass and its job is to instantiate the bear by defining its aspects. I realise it will be problematic in these cases.
```
>>> SomeAspectsBear.aspects['detect']
 +    [<aspectclass 'Root.Metadata.CommitMessage.Shortlog.ColonExistence'>]
```
 We do not expect user ever to mention `Root` as an aspect. I am asking to pass that as aspect of the bear __only__ when user doesn't passes any aspect. My point is we can use same bear collection methodology in both the cases if we pass `Root` aspect as the default aspect of any bear.

pratyushprakash: coala does not pass any aspects to the bears. The bear writer has to decide the aspects the bear can handle while writing the bear. The aspects a bear handles does not change dynamically. 
IMO the bear collie toon methodology will have to be different as in one case we are given only the aspects and have to find the appropriate bears while in the other we are given the names of the bears themselves.

Techievena: No I meant user passes aspects. `allow` or `disallow` tastes are passed to the Bear. :smiley: Yup its the bear writer who will decide the aspects I quite agree to that but we can restrict the bear to analyze one aspect out of two right (if it analyzes two aspects of code). So we can control that with these with allow or disallow tastes. So if no aspects are passed and bear is called explicitly, if we pass Root aspect, `allow` value will be true for all the aspects of the bear and it will do the aspect checks __it can__.

userzimmermann: The biggest problem with defining `Root` as default aspect is that a bear must be able to generate a result for every aspect is supports. And as @adhikasp correctly mentioned, `Root` would mean that the bear supports all aspects. Interpreting `Root` in any other way would break aspects definition consistency, since it is very beneficial to define a whole aspects sub-tree supported by a bear, like mentioned in #3977

userzimmermann: But @Techievena ... you are right that we should move this discussion to another place to get this PR merged. Because it's so fundamental and necessary for all further aspects PRs :) And existing ones like @Asnelchristian's https://github.com/coala/coala-bears/pull/1391

tests/bears/BearTest.py
```diff
@@ -119,6 +140,18 @@ def tearDown(self):
 
 class BearTest(BearTestBase):
 
+    def test_default_aspects(self):
```
adhikasp: Why we don't have test if Bear is assigned an aspect correctly?
`assert AspectsBear.aspects['detect'] == [CommitMessage.Shortlog.ColonExistence]`

userzimmermann: @adhikasp Done :)

tests/bears/BearTest.py
```diff
@@ -119,6 +140,29 @@ def tearDown(self):
...
+        assert type(Bear.aspects) is defaultdict
+        assert Bear.aspects['detect'] == Bear.aspects['fix'] == []
+
+    def test_no_fix_aspects(self):
+        assert type(DetectOnlyAspectsBear.aspects) is defaultdict
+        assert DetectOnlyAspectsBear.aspects['fix'] == []
+        assert DetectOnlyAspectsBear.aspects['detect'] \
```
Techievena: @userzimmermann Very trivial one but I'd like to mention that people at coala really don't like `\`

userzimmermann: @Techievena Strange coala people ;)

userzimmermann:
@coala/aspects-developers ACK! ACK! ACK! We need this stuff :D

userzimmermann:
@coala/aspects-developers OMG :D Found doc string issue that made CircleCI fail on `python setup.py docs`: Alphanumeric characters are not allowed after closing double backticks! But hyphens are:

https://github.com/coala/coala/pull/3966/commits/f93b9587373c86c73e5ed9b1e133e84edaa2599f#diff-7b888ece875aa379bf8cfcebcd7cdef1R148

So! This is done! I also rebased and made all occurrences of *"aspects"* lowercase according to https://github.com/coala/cEPs/issues/69

**ACK and MERGE to get forward** :)

@sils When could we do a new **coala** release to use this for migrating bears and writing new ones based on aspects?

adhikasp:
ack f93b9587373c86c73e5ed9b1e133e84edaa2599f

userzimmermann:
@rultor merge

PR #3991, opened 2017-03-25
TextRangeTest.py : Use assertRaisesRegex

mariaandreicosmin:
Addresses https://github.com/coala/coala-bears/issues/1194


andreidinu:
You have to run `coala` and then update your PR.

andreidinu:
Take care of line lenghts (they can not exceed 80 chars per line) and also take care of commit message title. I think the commit error is because you have a whitespace before `:`

tests/results/TextRangeTest.py
```diff
@@ -7,10 +7,10 @@
 class TextRangeTest(unittest.TestCase):
 
     def test_fail_instantation(self):
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError,"End position can't be less than start position."):
```
RJ722: Try packaging the error message inside a variable name.

tests/results/TextRangeTest.py
```diff
@@ -7,10 +7,12 @@
 class TextRangeTest(unittest.TestCase):
 
     def test_fail_instantation(self):
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, "End position can't be" +
+                                    ' less than start position.'):
```
satwikkansal: Better to write it this way
```py
error_message = "End position can't be less than start position"
with self.assertRaisesRegex(ValueError, error_message):
        # do stuff
```

tests/results/TextRangeTest.py
```diff
@@ -7,10 +7,12 @@
...
             TextRange(TextPosition(3, 4), TextPosition(2, 8))
 
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, "End position can't be" +
```
satwikkansal: Also, only use single quote, escape with "\" if required.

tests/results/TextRangeTest.py
```diff
@@ -7,10 +7,12 @@
 class TextRangeTest(unittest.TestCase):
 
     def test_fail_instantation(self):
-        with self.assertRaises(ValueError):
+        error_message = "End position can't be less than start position"
```
sils: can we give this a better name? Also maybe make use of the regex and only do part of the message like `".*less than start.*"` to keep the crucial part or so?

tests/results/TextRangeTest.py
```diff
@@ -7,10 +7,12 @@
...
+        with self.assertRaisesRegex(ValueError, error_message):
             TextRange(TextPosition(3, 4), TextPosition(2, 8))
 
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, "End position can't be" +
+                                    ' less than start position.'):
```
sils: duplication of previous string :/

tests/results/TextRangeTest.py
```diff
@@ -35,7 +37,8 @@ def test_properties(self):
 
     def test_from_values(self):
         # Check if invalid ranges still fail.
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, "End position can't be" +
+                                    ' less than start position.'):
```
sils: again

tests/results/TextRangeTest.py
```diff
@@ -150,14 +152,17 @@ def setUp(self):
 
     def test_fails(self):
         # need to pass ranges
-        with self.assertRaises(TypeError):
+        with self.assertRaisesRegex(TypeError, 'only instances of TextRange' +
```
jayvdb: store this string in a variable

tests/results/TextRangeTest.py
```diff
@@ -150,14 +152,17 @@ def setUp(self):
...
             TextRange.join(self.pos[0], self.pos[1])
 
-        with self.assertRaises(TypeError):
+        with self.assertRaisesRegex(TypeError, 'only instances of TextRange' +
+                                    ' can be joined'):
             TextRange.join(TextRange(self.pos[0], self.pos[1]), self.pos[1])
 
         # ranges must overlap
-        with self.assertRaises(ValueError):
+        with self.assertRaisesRegex(ValueError, 'TextRanges must overlap to' +
```
jayvdb: store this string in a variable

tests/results/TextRangeTest.py
jayvdb: Please put this string in a class variable called `END_LT_START_RE` , and then you can access it in this method, and in method `test_from_values`.

tests/results/TextRangeTest.py
```diff
@@ -150,14 +152,16 @@ def setUp(self):
 
     def test_fails(self):
         # need to pass ranges
-        with self.assertRaises(TypeError):
+        error_message1 = 'only instances of TextRange can be joined'
```
jayvdb: class level variable, called `JOIN_INSTANCEOF_RE`

tests/results/TextRangeTest.py
```diff
@@ -150,14 +152,16 @@ def setUp(self):
...
+        error_message2 = 'TextRanges must overlap to be joined'
```
jayvdb: again, class variable, with a short meaningful name.

Adrianzatreanu:
seems the changes requested here may be outdated. @sils @jayvdb can you check and dismiss/keep your review?

Adrianzatreanu:
also, @mariaandreicosmin please rebase.

tests/results/TextRangeTest.py
```diff
@@ -35,7 +36,8 @@ def test_properties(self):
 
     def test_from_values(self):
         # Check if invalid ranges still fail.
-        with self.assertRaises(ValueError):
+        END_LT_START_RE = "End position can't be less than start position"
```
shreyans800755: Add `END_LT_START_RE` and `JOIN_INSTANCEOF_RE` as class level attributes, as suggested by @jayvdb and not at function level to remove redundant declaration in multiple functions.

meetmangukiya:
@mariaandreicosmin please fix the commits. I suspect you didn't rebase correctly

Issue #3946, opened 2017-03-18
deprecate_settings doesn't return __doc__ of the wrapping function


meetmangukiya:
Try:
```py
from bears.python.PyImportSortBear import PyImportSortBear
from coalib.settings.Section import Section
from queue import Queue

bear = PyImportSortBear(Section(''), Queue())
bear.__doc__ # is '', doc string is lost

from bears.python.PEP8NotebookBear import PEP8NotebookBear
bear = PEP8NotebookBear(Section(''), Queue())
bear.__doc__ # this has doc string
```

So this mostly confirms it is the decorator

Makman2:
This is bad!

all3fox:
Hey @Flap-Py, what progress have you made so far? Are you stuck/do you need help?

Flap-Py:
@all3fox Sorry I got involved in my GSOC proposal. And thanks for your way to isolate the problem :smile: that was really helpful for me in getting a solution.
So I came up with a fix for the bug for the following [link](http://stackoverflow.com/questions/1782843/python-decorator-handling-docstrings#answer-1782888)
 @meetmangukiya check it out if its fine. I am creating a PR too. Let me know if something is wrong.

yash-nisar:
@jayvdb I'm thinking of getting this opened again and undoing https://github.com/coala/coala/commit/017bcfc2 so we can get rid of [this](https://travis-ci.org/coala/coala-bears/jobs/215711108) bug introduced in coala-bears. I need your approval.

jayvdb:
I would rather than this, and https://github.com/coala/coala/issues/3786, are fixed quickly, but if that doesnt happen, I agree that undoing 017bcfc is necessary

meetmangukiya:
That yapf build failure was because of this :/ ?

PR #3992, opened 2017-03-25
bearlib: Add @wraps to wrapping_function

Flap-Py:
@wraps(func) prevents the loss of doc string
of func, which occurs due to the use of
decorator on func.

Fixes https://github.com/coala/coala/issues/3946

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/__init__.py
+++ b/coalib/bearlib/__init__.py
@@ -5,6 +5,7 @@ while offering the best possible flexibility.
 """
 
 import logging
+from functools import wraps
 
 from coalib.settings.FunctionMetadata import FunctionMetadata
 
@@ -98,6 +99,7 @@ def deprecate_settings(**depr_args):
 
         logged_deprecated_args = set()
 
+        @wraps(func)
         def wrapping_function(*args, **kwargs):
             for arg, depr_value in wrapping_function.__metadata__.depr_values:
                 deprecated_arg = depr_value[0]
--- a/tests/bearlib/DeprecateSettingsTest.py
+++ b/tests/bearlib/DeprecateSettingsTest.py
@@ -1,0 +1,17 @@
+import unittest
+
+from coalib.bearlib import deprecate_settings
+
+
+@deprecate_settings(new='old')
+def func(new):
+    """
+    This docstring will not be lost.
+    """
+
+
+class DeprecateSettingsTest(unittest.TestCase):
+
+    def test_docstring(self):
+        self.assertEqual(func.__doc__.strip(),
+                         'This docstring will not be lost.')


all3fox:
How about writing a test-case for it? Something like `tests/bearlib/DeprecateSettingsTest.py` where you check that applying this decorator to a function/a class does not swallow the docstring?

meetmangukiya:
@Flap-Py you need a test case.

I really think that your patch doesn't fix the bug. Since it is calling the function right away, thought I not completely sure. So, write a test case and let's see if it fixes. 

Usually test cases are used to prove that my stuff *works* ;) giving reviewer little space for doubts

Flap-Py:
@meetmangukiya I had one doubt too.
Are your sure it should be 
"""bear = PyImportSortBear(Section(''), Queue())
bear.__doc__ # is '', doc string is lost"""

and not

"""bear = PyImportSortBear(Section(''), Queue())
bear.run.__doc__ # is '', doc string is lost"""

all3fox:
@Flap-Py you are correct, you are wrapping `.run` with the deprecation, so you should be preserving its docstring, not the bear's docstring. However, I would strongly suggest that you make an independent/simple test case:

```
@deprecate_settings()
def foo():
    """This docstring must survive the decorator."""

@deprecate_settings()
class Foo:
    """This cass docstring must survive it too."""
    pass
```

Keep it as simple as possible, no bears please.

all3fox:
@meetmangukiya no-no, wrapping it into `functools.wraps` actually does the trick and the bug is actually fixed.

But I am a simple person, I see a fixed bug --- I need to see a unit (regression) test for it too :)

Flap-Py:
@all3fox I am just working on simple testcase (without any bear). And yes the problem is fixed I have checked it :smile:

meetmangukiya:
https://github.com/coala/coala/pull/3992#issuecomment-289211621 You are correct :sweat_smile: that's what I meant

meetmangukiya:
Also, maybe use `bearlib` instead of `__init__.py` in the commit, since `__init__.py` is confusing.

coalib/bearlib/__init__.py
```diff
@@ -7,6 +7,7 @@
 import logging
 
 from coalib.settings.FunctionMetadata import FunctionMetadata
+from functools import wraps
```
SanketDG: put this import after `import logging` and a newline

tests/bearlib/DeprecateSettingsTest.py
```diff
@@ -0,0 +1,17 @@
...
+from coalib.bearlib import deprecate_settings
+
+
+@deprecate_settings(new='old')
+def func(new):
+    """
+    This docstring will not be lost.
+    """
+    print(new)
+    pass
```
SanketDG: any need for `pass`? 😄

tests/bearlib/DeprecateSettingsTest.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class DeprecateSettingsTest(unittest.TestCase):
+    def test(self):
```
SanketDG: provide a better test name

all3fox:
Nicely done. I would have also added a class together with a deprecation, but current state is fine for me too:

ack 5cdbfaf46f66563e6880bd929c332f57df982582

all3fox:
ack 5cdbfaf46f66563e6880bd929c332f57df982582

SanketDG:
@Flap-Py needs rebase

meetmangukiya:
ack 017bcfc

meetmangukiya:
@rultor merge

PR #3993, opened 2017-03-25
tests/bears/GlobalBearTest.py: Add more tests

all3fox:
In response to https://github.com/coala/coala/issues/3986 I have taken a look at the `GlobalBear` class. It is true that there were just a couple of unit-tests for this class. There is frankly not that much to test: it is an "abstract" class that other bears inherit from and reimpement its methods.

@jayvdb, here is a quick run down of tests that use `GlobalBear`:

1. `processes/BearRunningTest.py` that is all about dependencies between bears and the results of them running successfully/unsuccessfully.
2. `processes/section_executor_test_files/ProcessingGlobalTestBear.py` which is nearly empty and therefore hard for me to understand.
3. `settings/SectionFillingTest.py` which is all about appending/changing settings of bears (so, it is more about the `Settings` class, sections and so on)
4. `bears/BearTest.py` which contains both very broad tests (like [simple api][1]) and very specific tests (a bear [that downloads something with `requests_mock`][2])
5. `bears/GlobalBearTest.py` the file that I ended up modifying.
6. `collecting/collectors_test_dir/bears_local_global` that just has a couple of bears for, presumably, testing the bear-collecting logic.

So, this PR just adds a boring test to `GlobalBearTest.py`. Do you perhaps have some other/more specific test types in mind?

Weakly related question: many tests use the name `uut` for objects being tested. What does it stand for? "Unit Under Test"?

[1]: https://github.com/coala/coala/blob/b586c0a8e1d8417afc06621b20d0115f899e63ee/tests/bears/BearTest.py#L122

[2]: https://github.com/coala/coala/blob/b586c0a8e1d8417afc06621b20d0115f899e63ee/tests/bears/BearTest.py#L291

### DIFF
--- a/coalib/bears/GlobalBear.py
+++ b/coalib/bears/GlobalBear.py
@@ -4,22 +4,28 @@ from coalib.bears.BEAR_KIND import BEAR_KIND
 
 class GlobalBear(Bear):
     """
-    A GlobalBear is able to analyze semantic facts across several file.
+    A GlobalBear analyzes semantic facts across several files.
 
     The results of a GlobalBear will be presented grouped by the origin Bear.
-    Therefore Results spanning above multiple files are allowed and will be
-    handled right.
+    Therefore Results spanning across multiple files are allowed and will be
+    handled correctly.
 
-    If you only look at one file at once anyway a LocalBear is better for your
-    needs. (And better for performance and usability for both user and
-    developer.)
+    If you are inspecting a single file at a time, you should consider
+    using a LocalBear.
     """
 
     def __init__(self,
-                 file_dict,  # filename : file contents
+                 file_dict,
                  section,
                  message_queue,
                  timeout=0):
+        """
+        Constructs a new GlobalBear.
+
+        :param file_dict: The dictionary of {filename: file contents}.
+
+        See :class:`coalib.bears.Bear` for other parameters.
+        """
         Bear.__init__(self, section, message_queue, timeout)
         self.file_dict = file_dict
 
@@ -34,7 +40,11 @@ class GlobalBear(Bear):
         """
         Handles all files in file_dict.
 
+        :param dependency_results: The dictionary of {bear name:
+                                   result list}.
         :return: A list of Result type.
+
+        See :class:`coalib.bears.Bear` for `run` method description.
         """
         raise NotImplementedError(
             'This function has to be implemented for a runnable bear.')
--- a/tests/bears/GlobalBearTest.py
+++ b/tests/bears/GlobalBearTest.py
@@ -6,9 +6,25 @@ from coalib.settings.Section import Section
 
 class GlobalBearTest(unittest.TestCase):
 
-    def test_api(self):
-        test_object = GlobalBear(0, Section('name'), None)
-        self.assertRaises(NotImplementedError, test_object.run)
+    def test_file_dict(self):
+        file_dict_0 = {
+            'filename1': 'contents1', 'filename2': 'contents2'
+        }
 
-    def test_kind(self):
+        # check that bear does not modify original dictionary
+        file_dict_1 = file_dict_0.copy()
+
+        bear = GlobalBear(file_dict_0, Section(''), None)
+
+        self.assertEqual(bear.file_dict, file_dict_0)
+        self.assertEqual(bear.file_dict, file_dict_1)
+
+    def test_run_raises(self):
+        bear = GlobalBear(None, Section(''), None)
+        self.assertRaises(NotImplementedError, bear.run)
+
+    def test_kind_is_staticmethod(self):
         self.assertEqual(GlobalBear.kind(), BEAR_KIND.GLOBAL)
+
+        bear = GlobalBear(None, Section(''), None)
+        self.assertEqual(bear.kind(), BEAR_KIND.GLOBAL)


tests/bears/GlobalBearTest.py
```diff
@@ -6,9 +6,24 @@
...
+            'filename1': 'contents1', 'filename2': 'contents2'
+        }
 
-    def test_kind(self):
+        file_dict_1 = file_dict_0.copy()
+
+        bear = GlobalBear(file_dict_0, Section(''), None)
+
+        self.assertEqual(bear.file_dict, file_dict_0)
+        self.assertEqual(bear.file_dict, file_dict_1)
```
adhikasp: Why duplicate the file dict and check the bear against these 2 variables?

Udayan12167: I guess maybe that the bear doesn't modify the file_dict?

all3fox: @Udayan12167 exactly. @adhikasp I agree that is overkill here, but just to make sure there are no modifications.

Udayan12167: @all3fox can you confirm?

all3fox: @Udayan12167 yes, a copy is there just to make sure the bear does not actually modify/erase it.

Udayan12167:
Use just the filename in the commit shortlog. Complete path isn't required please :)

tests/bears/GlobalBearTest.py
adhikasp: Maybe add a comment so coala future generation is not confused about it 👍

all3fox: Ok, done

supergr35:
Is 45b52e3 part of the issue?

all3fox:
Most definitely yes.

Otherwise, splitting the small doc change and the small test change is just additional burden on the maintainer. Especially since we are talking about the same class.

coalib/bears/GlobalBear.py
```diff
@@ -34,7 +40,12 @@ def run(self,
         """
         Handles all files in file_dict.
 
+        :param dependency_results: The dictionary of {bear name:
+                                   result list}.
+
```
satwikkansal: No space required b/w `:param` and `:return`

coalib/bears/GlobalBear.py
```diff
@@ -4,22 +4,28 @@
...
-    Therefore Results spanning above multiple files are allowed and will be
-    handled right.
+    Therefore Results spanning across multiple files are allowed and will be
+    handled correctly.
 
-    If you only look at one file at once anyway a LocalBear is better for your
-    needs. (And better for performance and usability for both user and
-    developer.)
+    If you are inspecting a single file at a time, you should consider
+    using a LocalBear.
```
satwikkansal: I guess we should still mention the "performance and usability" benefits?

coalib/bears/GlobalBear.py
```diff
@@ -4,22 +4,28 @@
...
-                 file_dict,  # filename : file contents
+                 file_dict,
                  section,
                  message_queue,
                  timeout=0):
+        """
+        Constructs a new GlobalBear.
+
+        :param file_dict: The dictionary of {filename: file contents}.
+
```
satwikkansal: document the `section` and `message_queue` params as well :)

coalib/bears/GlobalBear.py
```diff
@@ -4,22 +4,28 @@
...
+        See :class:`coalib.bears.Bear` for other parameters.
```
satwikkansal: we use double ticks in the inline documentation, so ``coalib.bears.Bear``

satwikkansal: Ah.. can't get double ticks to show up around `coalib.bears.Bear` :P

coalib/bears/GlobalBear.py
all3fox: No, that would be duplication of documentation. See documentation for `Bear`

coalib/bears/GlobalBear.py
all3fox: No, that would be wrong. See `grep -r :class:`, that is syntax for `reSful` links which are done with single backticks

coalib/bears/GlobalBear.py
all3fox: No, we should not. Performance --- there are no measurements. Usability is too abstract. It is just architecture: `GlobalBear` is for several files, `LocalBear` is just for one. That point is clear.

coalib/bears/GlobalBear.py
all3fox: Agreed

all3fox:
Thanks @satwikkansal for review, I have addressed your comments, rebased of master and pushed.

coalib/bears/GlobalBear.py
satwikkansal: Duplication of documentation isn't bad at all. It has been intentionally done in some modules as well (see `Linter.py` for example).

all3fox: > Duplication of documentation isn't bad at all

Disagree. When changing documentation, you have to track down all the duplicated places. Duplication is always bad, be it either code or docs.

> It has been intentionally done in some modules as well (see Linter.py for example).

I have taken a look at that 880 lines file. Indeed, there is some duplicated documentation. However, there is some "metaclass conflict" which is probably the root cause of that (I have no idea though). So, please either be more specific and/or let's hear a third opinion about duplication.

satwikkansal: > When changing documentation, you have to track down all the duplicated places. Duplication is always bad, be it either code or docs.

imo, it's a usability v/s maintainability tradeoff, more documentation will help new developers and bears writers find and implement things easily.

> However, there is some "metaclass conflict" which is probably the root cause of that (I have no idea though). So, please either be more specific and/or let's hear a third opinion about duplication.

I've not heard of "metaclass conflict" before, so yeah, let's wait for third opinion :)

all3fox: > I've not heard of "metaclass conflict" before, so yeah, let's wait for third opinion :)

By that I mean the `LocalLinterMeta` and `GlobalLinterMeta` which respectively have `LocalLinterBase` and `GlobalLinterBase` classes as derivatives. These latter classes each have `create_arguments` staticmethods that have very similar documentation. The phrase "metaclass conflict" is in the docstring of both `*Meta` classes and that is what I was referring too.

So, bottomline, I understand the copy-pastedness of those docs as consequence of that "metaclass conflict". No idea if I am right/wrong though.

And the bottomline for current conflict: is it necessary to copy-paste docs from `Bear` into `GlobalBear`? Or does the doc reference from GlobalBear into Bear work?

all3fox: > more documentation will help new developers and bears writers find and implement things easily.

The problem with duplication is that overtime previously identical places begin to look differently and/or start contradicting each other. No matter how attentive/accurate the maintainer is, duplication will get them in the end and slip through.

jayvdb:
ack 21e0da7 cd6a8eb

Mixih:
This pr is getting batch merged at  https://github.com/coala/coala/pull/4142 as it is release critical. Thanks for your contribution!

Issue #3977, opened 2017-03-23
issubaspect()


userzimmermann:
We need some quick way of checking if an aspect is a subaspect of another aspect, to allow defining bear aspects by some parent, like...

```python
class AspectsBear(Bear, aspects={
    'detect': [CommitMessage],
}
```

... which means that our bear should support all subaspects of `CommitMessage`. Then we need to check if user-requested aspects, like `CommitMessage.Shortlog.ColonExistence`, are subaspects of the defined parent.

So I propose an `issubaspect` function to handle that task. It should basically work like python's builtin `issubclass` function, and also have those important two features:

* It should support aspectclasses as well as their instances as subaspect argument:

```python
>>> issubaspect(ColonExistence, CommitMessage)
True
>>> issubaspect(ColonExistence(shortlog_colon=False), CommitMessage)
True
```

* It should raise an exception if the arguments are no aspectclasses (or the first argument is neither class nor instance) and not just return `False` in that case:

```python
>>> issubaspect(ColonExistence, dict)
TypeError: <class 'dict'> is no aspectclass
```

And according to `issubclass`, every aspect should of course also be a subaspect of itself.

Makman2:
Maybe we implement this function also for each aspect itself, so you can do something like this?
```
CommitMessage.issubaspect(ColonExistence)
```

userzimmermann:
@Makman2 IMO something like `CommitMessage.issubaspect(ColonExistence)` would unnecessarily mess things up and cause confusion, since it is untypical in python and could also be read like check if `CommitMessage` is a subaspect of `ColonExistence`, whereby the argument order in the global function is always clear when you are used to builtin `issubclass`.

Makman2:
@userzimmermann good point. Maybe a different name would work, though you are right. Then we would use different names -> things get more complicated -> user confused :)

PR #3996, opened 2017-03-25
Add issubaspect function

Asnelchristian:
This function helps to determine whether or not
an aspect is a subaspect of another one.

Closes https://github.com/coala/coala/issues/3977

### DIFF
--- a/coalib/bearlib/aspects/meta.py
+++ b/coalib/bearlib/aspects/meta.py
@@ -1,3 +1,5 @@
+import re
+
 from inspect import getmembers, signature
 
 from coala_utils.decorators import generate_repr
@@ -72,3 +74,25 @@ class aspectclass(type):
 
     def __repr__(cls):
         return '<%s %s>' % (type(cls).__name__, repr(cls.__qualname__))
+
+
+def issubaspect(subaspect, aspect):
+    """
+    This function checks whether or not ``subaspect`` is a subaspect of
+    ``aspect``.
+    """
+    if (not isinstance(subaspect, aspectclass)
+            and not isinstance(subaspect, aspectbase)):
+        raise TypeError(
+            '{} is not an aspectclass or an instance of an '
+            'aspectclass'.format(repr(subaspect)))
+    if (not isinstance(aspect, aspectclass)
+            and not isinstance(aspect, aspectbase)):
+        raise TypeError(
+            '{} is not an aspectclass or an instance of an '
+            'aspectclass'.format(repr(aspect)))
+    aspect_qualname = (aspect.__qualname__ if isinstance(
+                    aspect, aspectclass) else type(aspect).__qualname__)
+    subaspect_qualname = (subaspect.__qualname__ if isinstance(
+                    subaspect, aspectclass) else type(subaspect).__qualname__)
+    return re.match(aspect_qualname+'(\.|$)', subaspect_qualname) is not None
--- a/tests/bearlib/aspects/ClassTest.py
+++ b/tests/bearlib/aspects/ClassTest.py
@@ -1,4 +1,6 @@
+from coalib.bearlib.aspects import Root
 from coalib.bearlib.aspects import aspectclass
+from coalib.bearlib.aspects.meta import issubaspect
 
 import pytest
 
@@ -31,3 +33,26 @@ class AspectClassTest:
                 example = 'Example'
 
         assert not SubAspect.docs.check_consistency()
+
+
+class IssubaspectFunctionTest:
+
+    def test_issubaspect(self, RootAspect):
+        @RootAspect.subaspect
+        class SubAspect:
+            """
+            Description
+            """
+
+        assert issubaspect(SubAspect, RootAspect)
+        assert not issubaspect(Root, RootAspect)
+        assert issubaspect(RootAspect, RootAspect)
+        with pytest.raises(TypeError) as exc:
+            issubaspect('String', SubAspect)
+        assert (str(exc.value) == "'String' is not an aspectclass or "
+                'an instance of an aspectclass')
+        with pytest.raises(TypeError) as exc:
+            issubaspect(RootAspect, str)
+        assert (str(exc.value) == "<class 'str'> is not an aspectclass or "
+                'an instance of an aspectclass')
+        assert issubaspect(SubAspect('Python'), RootAspect)


coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +75,21 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+    """
+    if not isinstance(aspect, aspectclass):
+        logging.warning(
+            '{} is not an aspectclass instance'.format(aspect.__name__))
+    if not isinstance(subaspect, aspectclass):
+        logging.warning(
+            '{} is not an aspectclass instance'.format(
+                                                       subaspect.__name__))
+    parent_aspect_qualname = aspect.__qualname__ + '.+'
+    if re.match(parent_aspect_qualname, subaspect.__qualname__) is not None:
```
pratyushprakash: Isn't `is not None` redundant here?

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +75,21 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+def issubaspect(aspect, subaspect):
+    """
+    This function checks whether or not ``subcls`` is a subaspect of cls
+    """
+    if not isinstance(aspect, aspectclass):
+        logging.warning(
+            '{} is not an aspectclass instance'.format(aspect.__name__))
+    if not isinstance(subaspect, aspectclass):
+        logging.warning(
+            '{} is not an aspectclass instance'.format(
```
pratyushprakash: There are no tests for this.

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +75,24 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
 
     def __repr__(cls):
         return '<%s %s>' % (type(cls).__name__, repr(cls.__qualname__))
+
+
+def issubaspect(aspect, subaspect):
+    """
+    This function checks whether or not ``subcls`` is a subaspect of cls
```
Techievena: If the arguments passed here are `aspect` and `subaspect`, shouldn't it be better if they are used in the docstring instead of `subcls` and `cls`

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +75,24 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+    """
+    if not isinstance(aspect, aspectclass):
+        logging.warning(
```
Techievena: Won't it be better if we raise `TypeError` instead of `logging.warning`?

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,8 @@ class docs:
                 example = 'Example'
 
         assert not SubAspect.docs.check_consistency()
+        assert issubaspect(RootAspect, SubAspect)
```
Techievena: IMO all the tests should be put together in a function

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,8 @@ class docs:
...
+        assert not issubaspect(RootAspect, Root)
+        assert not issubaspect(RootAspect, RootAspect)
+        assert issubaspect('String', SubAspect) is None
```
Techievena: No checks for the warning message displayed for invalid `aspectname`

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +75,24 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+            '{} is not an aspectclass instance'.format(
+                type(aspect).__name__))
+        return None
+    if not isinstance(subaspect, aspectclass):
+        logging.warning(
+            '{} is not an aspectclass instance'.format(
+                type(subaspect).__name__))
+        return None
+    parent_aspect_qualname = aspect.__qualname__ + '.+'
+    if re.match(parent_aspect_qualname, subaspect.__qualname__) is not None:
```
Techievena: Why `re.match` ? Wont this be OK. `if(aspect.__qualname__ in subaspect.__qualname__):`

coalib/bearlib/aspects/meta.py
Asnelchristian: oo yeah i forgot that

tests/bearlib/aspects/ClassTest.py
Techievena:
```
>>> issubaspect(ColonExistence(shortlog_colon=False), CommitMessage)
True
```
This case is missing

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+
+def issubaspect(aspect, subaspect):
+    """
+    This function checks whether or not ``subaspect`` is a subaspect of
+    ``aspect``.
+    """
+    if not isinstance(aspect, aspectclass):
+        raise TypeError(
+            "{} is not an aspectclass' instance".format(
+                type(aspect).__name__))
```
userzimmermann: Better use `repr(aspect)` to see what actual object we got here.

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+    if not isinstance(subaspect, aspectclass):
```
userzimmermann: Since `issubaspect` should also support instances of aspectclasses as `subaspect` you also have to check if otherwise `isinstance(subaspect, aspectbase)`

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+        raise TypeError(
+            "{} is not an aspectclass' instance".format(
```
userzimmermann: According to the comments above the message should be `"{} is not an aspectclass or instance of an aspectclass"`

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+                type(subaspect).__name__))
```
userzimmermann: Again, better use `repr(subaspect)` here.

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+    if (aspect.__qualname__ in subaspect.__qualname__
+            and aspect.__qualname__ != subaspect.__qualname__):
```
userzimmermann: `subaspect.__qualname__.startswith(aspect.__qualname__)` would be a bit clearer.

BTW: You could also recursively check the `.parent` attribute. But that's just a matter of taste ;)

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,23 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+        return True
+    else:
+        return False
```
userzimmermann: You could also just implicitly `return` the logical expression instead of explicit `True` or `False`.

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,9 @@ class docs:
                 example = 'Example'
 
         assert not SubAspect.docs.check_consistency()
+        assert issubaspect(RootAspect, SubAspect)
+        assert not issubaspect(RootAspect, Root)
+        assert not issubaspect(RootAspect, RootAspect)
+        with pytest.raises(TypeError):
+            issubaspect('String', SubAspect)
+            issubaspect(RootAspect, 'String')
```
userzimmermann: You should also check other class objects here and not only instance objects of other classes. So for example also `str`.

coalib/bearlib/aspects/meta.py
Asnelchristian: yeah i thought about that too, but this solution seemed to be alot simpler to me :)

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,10 @@ class docs:
                 example = 'Example'
 
         assert not SubAspect.docs.check_consistency()
+        assert issubaspect(SubAspect, RootAspect)
+        assert not issubaspect(Root, RootAspect)
+        assert not issubaspect(RootAspect, RootAspect)
+        with pytest.raises(TypeError):
+            issubaspect('String', SubAspect)
```
Techievena: I think would be better if : 
```
with pytest.raises(TypeError) as log:
    issubaspect('String', SubAspect)
    assertEqual(str(log.exception), 'String is not an aspectclass or an instance of an aspectclass')
```

userzimmermann: @Techievena Right! But... :)

**[0]** The assert statement must not be inside the `with` block. Otherwise it won't be reached after the exception occurred.

**[1]** We should not use `unittest`'s `assertEqual` here. Let's keep pythonic `pytest` style and just use `assert ... == ...` :)

**[2]** The `ExceptionInfo` object returned from `pytest.raises` holds the actual exception in its `.value` attribute. It has no `.exception`.

So:

```python
with pytest.raises(TypeError) as exc:
    issubaspect('String', SubAspect)
assert str(exc.value) \
    == "'String' is not an aspectclass or an instance of an aspectclass"
```

@Asnelchristian @shrutishrm512

coalib/bearlib/aspects/meta.py
```diff
@@ -72,3 +73,21 @@ class Sub(subcls, aspectbase, metaclass=aspectclass):
...
+            and not isinstance(subaspect, aspectbase)):
+        raise TypeError(
+            '{} is not an aspectclass or an instance of an '
+            'aspectclass'.format(repr(aspect)))
+    if (not isinstance(aspect, aspectclass)
+            and not isinstance(aspect, aspectbase)):
+        raise TypeError(
+            '{} is not an aspectclass or an instance of an '
+            'aspectclass'.format(repr(subaspect)))
+    return subaspect.__qualname__.startswith(aspect.__qualname__, 0, -1)
```
userzimmermann: There are 2 drawbacks here:

**[0]** One idea of `issubaspect` was that aspects are also considered as subaspects of themselves, just like `issubclass(str, str) is True`. The exclusion of the last character with stop index `-1` prevents this.

**[1]** Instances of aspectclasses don't have `__qualname__` anymore. You would need to use `type(subaspect).__qualname__` in that case.

**[2]** The `.startswith` approach creates false positives, just like the original `in` approach. Why didn't I realize that earlier? Shame on me! :D Just consider the following:

```python
@Root.subaspect
class SubAspect:
    """
    """

@Root.subaspect
class SubAspectYEAH:
    """
    """
```

```python
>>> issubaspect(SubAspectYEAH, SubAspect)
True
```

To solve **[2]**, it must be additionally checked if the `__qualname__` of the parent is followed by a dot in or the end of the subaspect's `__qualname__`. IMO the simplest way would be a regex here:

```python
    return bool(re.match(r'^%s(\.|$)' % aspect.__qualname__, subaspect.__qualname__))
```

Meanwhile I think the recursive `.parent` check approach would actually be simpler in the end ;) Since `.parent` is available on aspectclasses AND their instances.

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,10 @@ class docs:
                 example = 'Example'
 
         assert not SubAspect.docs.check_consistency()
+        assert issubaspect(SubAspect, RootAspect)
+        assert not issubaspect(Root, RootAspect)
+        assert not issubaspect(RootAspect, RootAspect)
```
userzimmermann: As mentioned in https://github.com/coala/coala/pull/3996#discussion_r108486591 this should be `True`.

tests/bearlib/aspects/ClassTest.py
```diff
@@ -31,3 +33,10 @@ class docs:
...
+        with pytest.raises(TypeError):
+            issubaspect('String', SubAspect)
+        with pytest.raises(TypeError):
+            issubaspect(RootAspect, str)
```
userzimmermann: As @Techievena mentioned in https://github.com/coala/coala/pull/3996#discussion_r108064910 all the `issubaspect` tests should really be put in a separate function outside of `AspectClassTest`, maybe even in a separate test module.

userzimmermann: And as @Techievena mentioned in https://github.com/coala/coala/pull/3996#discussion_r108152329 the tests for aspectclass instances are still missing...

coalib/bearlib/aspects/meta.py
Asnelchristian: oooh crap, that is exactly why i used regex, on the first trial,

tests/bearlib/aspects/ClassTest.py
```diff
@@ -30,4 +32,23 @@ class SubAspect:
             class docs:
                 example = 'Example'
 
+    def test_issubaspect(self, RootAspect):
```
userzimmermann: Please move it out of `AspectClassTest`, since `issubaspect` is not part of `aspectclass`. It is a global function. So `test_issubaspect` should also be global.

tests/bearlib/aspects/ClassTest.py
```diff
@@ -30,4 +32,23 @@ class SubAspect:
...
+        @RootAspect.subaspect
+        class SubAspect:
+            """
+            Description
+            """
+
         assert not SubAspect.docs.check_consistency()
```
userzimmermann: And please move this line back to `test_subaspect_without_enough_docs` :)

tests/bearlib/aspects/ClassTest.py
```diff
@@ -30,4 +32,23 @@ class SubAspect:
             class docs:
                 example = 'Example'
 
+    def test_issubaspect(self, RootAspect):
+        @RootAspect.subaspect
+        class SubAspect:
```
userzimmermann: It is also fine to use `SubAspect` fixture here. But again... just a matter of personal taste ;)

tests/bearlib/aspects/ClassTest.py
Asnelchristian: ooh my, how did this happened :sweat_smile:

userzimmermann:
@coala/aspects-developers So! Let's merge it! Things are going forward :)

Thanks @Asnelchristian and all reviewers!

**#aspectsYEAH**

userzimmermann:
ack https://github.com/coala/coala/pull/3996/commits/e78d2accfc5f122ec9bd244e5fd649bedc17de30

userzimmermann:
@rultor merge

Asnelchristian:
i guess i need to rebase

Asnelchristian:
reack https://github.com/coala/coala/pull/3996/commits/1a821def767651be533ef0b8fa75d517f73903f7

userzimmermann:
ack https://github.com/coala/coala/pull/3996/commits/1a821def767651be533ef0b8fa75d517f73903f7

userzimmermann:
@rultor merge

Asnelchristian:
Oops the tests have to finish first

userzimmermann:
@Asnelchristian Let's see what happens... :)

userzimmermann:
Oh! :D So... again :)

userzimmermann:
C'mon Circle!

userzimmermann:
@rultor merge

Issue #3999, opened 2017-03-26
Format the output of CI mode when is used both --format and --ci


Nosferatul:
https://github.com/coala/coala/pull/3750

jayvdb:
@Nosferatul , isnt this a dup of https://github.com/coala/coala/issues/3154

meetmangukiya:
This issue is about formatting the output of `--ci`

sils:
this is hard to understand. Wrote an explanation on gitter after managing to wrap my head around it, documenting here:

ssentially when you run coala --format it will give you a custom formatted output but it might ask you questions in the beginning; when you run coala --ci you don't get any questions asked. Now this issue requires that coala --ci --format doesn't ask questions and uses the format string output.

Nosferatul:
my bad @sils 😟

MariosPanag:
--format is for continuous integration and it doesnt asky you questions so i guess what you want to implement is that when user gives both arguments then the mode.format() should be executed. Am i right ?

PR #4036, opened 2017-03-31
coala.py: Format the output of CI mode when both --format and --ci are given

MariosPanag:
Give --format flag precedence over --ci when
both --format --ci args are given.

Closes https://github.com/coala/coala/issues/3999

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -71,11 +71,11 @@ def main():
     except BaseException as exception:  # pylint: disable=broad-except
         return get_exitcode(exception, log_printer)
 
-    if args.non_interactive:
-        return mode_non_interactive(console_printer, args)
-
     if args.format:
         return mode_format()
+
+    if args.non_interactive:
+        return mode_non_interactive(console_printer, args)
 
     return mode_normal(console_printer, log_printer)
 
--- a/tests/coalaFormatTest.py
+++ b/tests/coalaFormatTest.py
@@ -37,3 +37,19 @@ class coalaFormatTest(unittest.TestCase):
                              'coala-format must return exitcode 1 when it '
                              'yields results')
             self.assertFalse(stderr)
+
+    def test_format_ci_combination(self):
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(coala.main, 'coala',
+                                                   '--format', '--ci', '-c',
+                                                   os.devnull, '-f',
+                                                   re.escape(filename),
+                                                   '-b', 'LineCountTestBear')
+            self.assertRegex(stdout, r'message:This file has [0-9]+ lines.',
+                             'coala --format --ci output for line count should '
+                             'not be empty')
+            self.assertEqual(retval, 1,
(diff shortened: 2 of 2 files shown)


supergr35:
In the commit shortlog and message: `Combines` and `Checks` should be `Combine` and `Check`

coalib/coala.py
```diff
@@ -71,6 +71,9 @@ def main():
     except BaseException as exception:  # pylint: disable=broad-except
         return get_exitcode(exception, log_printer)
 
+    if args.non_interactive and args.format:
+        return mode_format()
+
     if args.non_interactive:
```
RohanVB: Shouldn't this be `elif args.non_interactive:` and the statement on L80 be `elif args.format:` ?
We're returning after checking for format/ci so if `args.non_interactive==true` we don't need to check if `args.format==true` since the case for both being true is handled by L74.

coalib/coala.py
```diff
@@ -71,10 +71,11 @@ def main():
...
 
-    if args.non_interactive:
+    if args.non_interactive and args.format:
+        return mode_format()
+    elif args.non_interactive:
         return mode_non_interactive(console_printer, args)
-
-    if args.format:
+    elif args.format:
         return mode_format()
```
userzimmermann: You would get exactly the same result by just switching the original check order to:

```python
if args.format:
    return mode_format()

if args.non_interactive:
    return mode_non_interactive(console_printer, args)
```

userzimmermann:
BTW: There is also no check if `--format` and `--json` are used together, and `--json` just has higher precedence. That must definitely raise an error! I will create an issue...

coalib/coala.py
MariosPanag: I didnt think of it wow :) i i will re-amend my commit

userzimmermann:
ack https://github.com/coala/coala/pull/4036/commits/64ea784ddb6590255c2ba360160b9cfbc5ccd010

userzimmermann:
@rultor merge

MariosPanag:
Rebased !

userzimmermann:
ack https://github.com/coala/coala/pull/4036/commits/c929a8b45bdb0e326014cb702ffcd54e5f8d1df6

userzimmermann:
@rultor merge

meetmangukiya:
> But better change commit message to something like giving --format flag precedence over --ci

Haven't followed this PR, but AFAIK the issue is not about precedence... IIUC the format option formats the messages, while the ci options applies patches automatically..

Setting up a precedance is just like ignoring another mode over another mode, isn't it?

Setting --format expects input of users, making it completely opposite of ci I.e. non-interactive, doesn't it?

userzimmermann:
@meetmangukiya This PR was just a first step to clean up the whole mess around this topic. I created #4055 for dealing further with it, mostly directing exactly that ignorance of other flags.

But regarding your other concern: `--format` and `--json` are always non-interactive. So there is no conflict with `--ci` regarding this.

PR #4083, opened 2017-04-10
Makman2/nextgen3-4

Makman2:
Adds the `DependencyTracker` and two utility functions. More to come.

### DIFF
--- a/coalib/core/DependencyTracker.py
+++ b/coalib/core/DependencyTracker.py
@@ -1,0 +1,122 @@
+class DependencyTracker:
+    """
+    A ``DependencyTracker`` allows to register and manage dependencies between
+    objects.
+
+    This class uses a directed graph to track relations.
+
+    Add a dependency relation between two objects:
+
+    >>> object1 = object()
+    >>> object2 = object()
+    >>> tracker = DependencyTracker()
+    >>> tracker.add(object2, object1)
+
+    This would define that ``object1`` is dependent on ``object2``.
+
+    If you define that ``object2`` has its dependency duty fulfilled, you can
+    resolve it:
+
+    >>> resolved = tracker.resolve(object2)
+    >>> resolved
+    {<object object at ...>}
+    >>> resolved_object = resolved.pop()
+    >>> resolved_object is object1
+    True
+
+    This returns all objects that are now freed, meaning they have no
+    dependencies any more.
+
+    >>> object3 = object()
+    >>> tracker.add(object2, object1)
+    >>> tracker.add(object3, object1)
+    >>> tracker.resolve(object2)
+    set()
+    >>> tracker.resolve(object3)
+    {<object object at ...>}
+
+    The ones who instantiate a ``DependencyTracker`` are responsible for
+    resolving dependencies in the right order. Dependencies which are itself
+    dependent will be forcefully resolved and removed from their according
+    dependencies too.
+    """
+
+    def __init__(self):
+        self._dependency_dict = {}
+
+    def add(self, dependency, dependant):
+        """
+        Add a dependency relation.
+
+        This function does not check for circular dependencies.
+
+        >>> tracker = DependencyTracker()
+        >>> tracker.add(0, 1)
+        >>> tracker.add(0, 2)
+        >>> tracker.resolve(0)
+        {1, 2}
+
+        :param dependency:
+            The object that is the dependency.
+        :param dependant:
+            The object that is the dependant.
+        """
+        if dependency not in self._dependency_dict:
+            self._dependency_dict[dependency] = set()
+
+        self._dependency_dict[dependency].add(dependant)
+
+    def resolve(self, dependency):
+        """
+        Resolves all dependency-relations from the given dependency, and frees
+        and returns dependants with no more dependencies. If the given
+        dependency is itself a dependant, all those relations are also removed.
+
+        >>> tracker = DependencyTracker()
+        >>> tracker.add(0, 1)
+        >>> tracker.add(0, 2)
+        >>> tracker.add(2, 3)
+        >>> tracker.resolve(0)
+        {1, 2}
+        >>> tracker.resolve(2)
+        {3}
+        >>> tracker.resolve(2)
+        set()
+
+        :param dependency:
+            The dependency.
+        :return:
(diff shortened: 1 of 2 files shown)


coalib/core/DependencyTracker.py
```diff
@@ -138,3 +138,25 @@ def check_circular_dependencies(self):
         traverse_graph(
             self._dependency_dict.keys(),
             lambda node: self._dependency_dict.get(node, frozenset()))
+
+    @property
+    def all_dependencies_resolved(self):
```
Makman2: From @SanketDG:

> I somehow feel that since this something that is mutated based on operation of other functions, it should be a function.
>
> properties are properties, features, attributes of an object.
>
> also, a name like are_dependencies_resolved() looks more suitable.

1. imo it describes a state, so properties should be suitable.
2. Hmm agree, `are_dependencies_resolved` sounds better :+1:

Makman2: (Sorry btw for missing your comment :/)

tests/core/DependencyTrackerTest.py
```diff
@@ -53,3 +53,16 @@ def test_resolve(self):
...
+
+        self.assertTrue(uut.all_dependencies_resolved)
+
+        uut.add(0, 1)
+
+        self.assertFalse(uut.all_dependencies_resolved)
+
+        uut.resolve(0)
+
+        self.assertTrue(uut.all_dependencies_resolved)
```
satwikkansal: imo, for completeness, we should also add a test case here in which some of the dependencies are forcefully resolved

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,162 @@
...
+    >>> tracker.add(object2, object1)
+
+    This would define that ``object1`` is dependent on ``object2``.
+
+    If you define that ``object2`` has its dependency duty fulfilled, you can
+    resolve it:
+
+    >>> resolved = tracker.resolve(object2)
+    >>> resolved
+    {<object object at ...>}
```
RohanVB: Is the `object object` intentional?

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,162 @@
...
+        >>> tracker.resolve(0)
+        {1, 2}
+
+        :param dependency:
+            The object that is the dependency.
+        :param dependant:
+            The object that is the dependant.
+        """
+        if dependency not in self._dependency_dict:
+            self._dependency_dict[dependency] = set()
```
Siddhartha1234: Why not use `` self._dependency_dict = defaultdict(set) `` ?

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,162 @@
...
+        >>> tracker.resolve(2)
+        set()
+
+        :param dependency:
+            The dependency.
+        :return:
+            Returns a set of dependants whose dependencies were all resolved.
+        """
+        # Check if dependency has itself dependencies which aren't resolved,
+        # these need to be removed too. The ones who instantiate a
```
Siddhartha1234: >  The ones who instantiate a DependencyTracker are responsible for resolving dependencies in the right order.
 
This should be part of the docstring as it explains the right usage of resolve function  IMO.

coalib/core/DependencyTracker.py
Makman2: The problem with defaultdicts that accesses do create new sets and those get placed into the dict. That makes it non-empty any more and `bool(dc)` would then yield `True` though there are effectively no dependency relations. I rely on that behaviour :+1:

coalib/core/DependencyTracker.py
Makman2: yes that's the `repr` of an object created by the built-in function `object()`. The class name of such an object is `object` ;)

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,166 @@
...
+        Resolves all dependency-relations from the given dependency, and frees
+        and returns dependants with no more dependencies. If the given
+        dependency is itself a dependant, all those relations are also removed.
+
+        >>> tracker = DependencyTracker()
+        >>> tracker.add(0, 1)
+        >>> tracker.add(0, 2)
+        >>> tracker.add(2, 3)
+        >>> tracker.resolve(0)
+        {1, 2}
```
Nosferatul: shouldn't also be `0` ?  because it isn't a dependent?

coalib/core/DependencyTracker.py
```diff
@@ -0,0 +1,166 @@
...
+        >>> tracker.add(1, 0)
+        >>> tracker.check_circular_dependencies()
+        Traceback (most recent call last):
+         ...
+        coalib.core.CircularDependencyError.CircularDependencyError: ...
+
+        :raises CircularDependencyError:
+            Raised on circular dependency conflicts.
+        """
+        traverse_graph(
```
Nosferatul: Why we just rise an Error if we have circular dependencies? Is it something that shouldn' ever happen? (Sorry if it's a dumb question :) )

Makman2: circular dependencies can't be resolved, so yes, it should never happen :) The core would deadlock^^

coalib/core/DependencyTracker.py
Makman2: `0` was never a dependant here, so it can't be "freed". It is in principle already freed. And `resolve` only returns newly freed dependent objects.

The idea is that we would make bear instances dependent on each other (depending what the user has specified inside his bear), and the core resolves a bear once it has finished. The returned "free" bears can now be scheduled for execution, as dependencies are resolved.

Nosferatul:
ack 819f30a 25fcd95 4173a1c

Nosferatul:
ack 25fcd95

Nosferatul:
ack 4173a1c

Nosferatul:
@Makman2 rebase and if you like it, merge. I read quite a few times your code and I didn't find something that should be changed :+1:

Makman2:
reack 305b5a5 954db8b 725b8db

Makman2:
@rultor merge

PR #4104, opened 2017-04-13
Add exclude-sections

Nosferatul:
This is an argument for cli to ignore section(s).

Closes https://github.com/coala/coala/issues/4019

### DIFF
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -105,6 +105,10 @@ To run coala without user interaction, run the `coala --non-interactive`,
     config_group = arg_parser.add_argument_group('Configuration')
 
     config_group.add_argument(
+        '-e', '--exclude-sections', nargs='*', metavar='STR',
+        help='exclude sections for the current run')
+
+    config_group.add_argument(
         '-c', '--config', nargs=1, metavar='FILE',
         help='configuration file to be used, defaults to {}'.format(
             Constants.default_coafile))
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -13,6 +13,7 @@ from coalib.parsing.ConfParser import ConfParser
 from coalib.settings.Section import Section
 from coalib.settings.SectionFilling import fill_settings
 from coalib.settings.Setting import Setting, path
+from coalib.parsing.DefaultArgParser import default_arg_parser
 
 
 def merge_section_dicts(lower, higher):
@@ -199,6 +200,18 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
                             'coafile to avoid any unexpected behavior.')
 
         sections = merge_section_dicts(sections, cli_sections)
+
+    arg_parser = default_arg_parser() if arg_parser is None else arg_parser
+
+    exclude_sections = vars(arg_parser.parse_args(arg_list))['exclude_sections']
+
+    if exclude_sections is not None:
+        for section_name in exclude_sections:
+            if section_name in sections:
+                del sections[section_name]
+            else:
+                logging.warning('The {} section doesn\'t exist'
+                                .format(section_name))
 
     for name, section in list(sections.items()):
         section.set_default_section(sections)
--- a/tests/settings/ConfigurationGatheringTest.py
+++ b/tests/settings/ConfigurationGatheringTest.py
@@ -20,6 +20,9 @@ from coalib.settings.ConfigurationGathering import (
 @pytest.mark.usefixtures('disable_bears')
 class ConfigurationGatheringTest(unittest.TestCase):
 
+    CURRENT_DIR = os.path.abspath(os.path.dirname(__file__))
+    TEST_DIR = os.path.join(CURRENT_DIR, 'section_manager_test_files')
+
     def setUp(self):
         self.log_printer = LogPrinter(NullPrinter())
 
@@ -31,6 +34,37 @@ class ConfigurationGatheringTest(unittest.TestCase):
 
     def test_gather_configuration(self):
         args = (lambda *args: True, self.log_printer)
+        logger = logging.getLogger()
+
+        with change_directory(self.TEST_DIR), \
+                self.assertLogs(logger, 'WARNING') as cm:
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile', '-e', 'java.test'])
+
+        self.assertRaises(KeyError, lambda: sections['java.test'])
+
+        with change_directory(self.TEST_DIR), \
+                self.assertLogs(logger, 'WARNING') as cm:
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile', '-e', 'java.test',
+                          'all.python'])
+
+        self.assertRaises(KeyError, lambda: sections['java.test'])
+        self.assertRaises(KeyError, lambda: sections['all.python'])
+
+        with change_directory(self.TEST_DIR), \
+                self.assertLogs(logger, 'WARNING') as cm:
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile', '-e', 'not_section', '-f',
+                          '*.java'])
+
+        self.assertRaises(KeyError, lambda: sections['not_section'])
 
         # Using incomplete settings (e.g. an invalid coafile) will error
         with self.assertRaises(SystemExit):
@@ -223,8 +257,7 @@ class ConfigurationGatheringTest(unittest.TestCase):
         self.assertEqual(targets, ['cli', 'test1', 'test2'])
 
     def test_find_user_config(self):
-        current_dir = os.path.abspath(os.path.dirname(__file__))
-        c_file = os.path.join(current_dir,
+        c_file = os.path.join(self.CURRENT_DIR,
                               'section_manager_test_files',
                               'project',
                               'test.c')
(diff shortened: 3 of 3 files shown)


coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +385,19 @@ def gather_configuration(acquire_settings,
...
+    exclude = False
+    for x in arg_list:
+        if x == '-ex' or x == '--exclude-sections':
+            exclude = True
+        elif exclude and '-' not in x:
+            if x in sections:
+                del sections[x]
+            else:
+                logging.warning('The {} section doesn\'t exists'.format(x))
+        elif '-' in x:
```
adhikasp: Why check for a dash?

Nosferatul: Because we can have multiple sections that a user can exclude at a run. So I need to know where the list is ending

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -31,6 +31,28 @@ def tearDown(self):
...
+
+        with change_directory(test_dir), \
+                self.assertLogs(logger, 'WARNING') as cm:
+            sections, _, _, _ = gather_configuration(
+                lambda *args: True,
+                self.log_printer,
+                arg_list=['-c', 'inherit_coafile', '-ex', 'java.test'])
+
+        self.assertRaises(KeyError, lambda: sections['java.test'])
+
```
adhikasp: Please add test for multiple section exclude too.

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +385,20 @@ def gather_configuration(acquire_settings,
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings,
                                               log_printer)
+
+    exclude = False
+    for x in arg_list:
```
adhikasp: I think it's better to use the `arg_parser` instance to do the parsing here instead of "manually" iterate through `arg_list` like this. There are existing example for how to do this https://github.com/coala/coala/blob/master/coalib/parsing/CliParsing.py#L52 or you can refer to their official docs.

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +385,20 @@ def gather_configuration(acquire_settings,
...
+        if x == '-ex' or x == '--exclude-sections':
+            exclude = True
+        elif exclude and '-' not in x:
```
adhikasp: What if the section name contain `-`? This could be avoided if we use `arg_parser`.

coalib/settings/ConfigurationGathering.py
Nosferatul: I saw that :) the problem is that we can have multiple sections, and arg_parser hold only the first section. with `coala -ex sec1, sec2` arg_parser will have only `sec1`

adhikasp: Hm the argparser docs https://docs.python.org/3/library/argparse.html#nargs seems able to correctly parse something like that as a list. Maybe you must use `nargs='*'` for it to works.

Nosferatul: ok, i'll check that

tests/settings/ConfigurationGatheringTest.py
Nosferatul: added

coalib/settings/ConfigurationGathering.py
Nosferatul: it worked, i just changed '?' with '*' in DefaultArgParser, thanks !

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,21 @@ def gather_configuration(acquire_settings,
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings,
                                               log_printer)
+
+    arg_parser = default_arg_parser() if arg_parser is None else arg_parser
+
+    for arg_key, arg_value in sorted(
+            vars(arg_parser.parse_args(arg_list)).items()):
+        if arg_key == 'exclude_sections':
+            if arg_value is not None:
```
adhikasp: This check is not needed because if `arg_value` is None, the `for` iteration below wouldn't run anyway.

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,21 @@ def gather_configuration(acquire_settings,
...
+                for x in arg_value:
+                    if x in sections:
+                        del sections[x]
+                    else:
+                        logging.warning(
+                            'The {} section doesn\'t exists'.format(x))
+                break
```
adhikasp: Is this break used?

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,21 @@ def gather_configuration(acquire_settings,
...
                                               acquire_settings,
                                               log_printer)
+
+    arg_parser = default_arg_parser() if arg_parser is None else arg_parser
+
+    for arg_key, arg_value in sorted(
+            vars(arg_parser.parse_args(arg_list)).items()):
+        if arg_key == 'exclude_sections':
+            if arg_value is not None:
+                for x in arg_value:
```
adhikasp: I would change `x` with `section_name` for better clarity.

coalib/settings/ConfigurationGathering.py
Nosferatul: it will run, i had some errors without this if :)

coalib/settings/ConfigurationGathering.py
Nosferatul: it will iterate through all content of the list without it, we don't need args after the excluded sections.

coalib/settings/ConfigurationGathering.py
adhikasp: Oh well okay then 👍 

How about combine it `if arg_key == 'exclude_sections' and not arg_value:`?

adhikasp:
I just thinking, is it possible to do `vars(arg_parser.parse_args(arg_list))['exclude_sections']`? We don't need to iterate the whole dicts object.

coalib/settings/ConfigurationGathering.py
adhikasp: I just realize that it's a dict....

coalib/parsing/DefaultArgParser.py
```diff
@@ -105,6 +105,10 @@ def default_arg_parser(formatter_class=None):
     config_group = arg_parser.add_argument_group('Configuration')
 
     config_group.add_argument(
+        '-ex', '--exclude-sections', nargs='*', metavar='STR',
```
meetmangukiya: Usually it is a one letter word with `-` because `-abc` is usually interpreted as `-a -b -c`. I suggest you use a one letter too...

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,20 @@ def gather_configuration(acquire_settings,
...
+
+    sections_name = vars(arg_parser.parse_args(arg_list))['exclude_sections']
+
+    if sections_name is not None:
+        for section_name in sections_name:
+            if section_name in sections:
+                del sections[section_name]
+            else:
+                logging.warning(
+                    'The {} section doesn\'t exists'
```
meetmangukiya: I think you can shift this line up without exceeding the limit. Check, please?

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -31,6 +31,39 @@ def tearDown(self):
 
     def test_gather_configuration(self):
         args = (lambda *args: True, self.log_printer)
+        current_dir = os.path.abspath(os.path.dirname(__file__))
+        test_dir = os.path.join(current_dir, 'section_manager_test_files')
+        logger = logging.getLogger()
+
+        with change_directory(test_dir), \
+                self.assertLogs(logger, 'WARNING') as cm:
```
meetmangukiya: Add a new line?

Nosferatul: I don't think is needed

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,19 @@ def gather_configuration(acquire_settings,
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings,
                                               log_printer)
+
+    arg_parser = default_arg_parser() if arg_parser is None else arg_parser
```
adhikasp: @Nosferatul What is your opinion about moving this whole deleting section into `load_configuration()` https://github.com/coala/coala/blob/master/coalib/settings/ConfigurationGathering.py#L139?

IMO it's better placed here because that function is intended to "parses the CLI args and loads the config file accordingly".

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +386,19 @@ def gather_configuration(acquire_settings,
...
+
+    sections_name = vars(arg_parser.parse_args(arg_list))['exclude_sections']
```
adhikasp: I would name this var `exclude_sections` too for better clarity.

coalib/settings/ConfigurationGathering.py
```diff
@@ -385,6 +398,7 @@ def gather_configuration(acquire_settings,
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings,
                                               log_printer)
+
```
adhikasp: Welp unneeded change

adhikasp:
ack 54aab8c566c6d7e39d238dddbb88d93f218db440

Nosferatul:
thx @adhikasp for the review

coalib/settings/ConfigurationGathering.py
```diff
@@ -200,6 +201,18 @@ def load_configuration(arg_list, log_printer, arg_parser=None):
...
+    arg_parser = default_arg_parser() if arg_parser is None else arg_parser
+
+    exclude_sections = vars(arg_parser.parse_args(arg_list))['exclude_sections']
+
+    if exclude_sections is not None:
+        for section_name in exclude_sections:
+            if section_name in sections:
+                del sections[section_name]
+            else:
+                logging.warning('The {} section doesn\'t exists'
```
Adrianzatreanu: exist*

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -31,6 +31,39 @@ def tearDown(self):
 
     def test_gather_configuration(self):
         args = (lambda *args: True, self.log_printer)
+        current_dir = os.path.abspath(os.path.dirname(__file__))
+        test_dir = os.path.join(current_dir, 'section_manager_test_files')
```
Adrianzatreanu: should make these constants of the test class i think

Adrianzatreanu:
minor tweaks tbh, but rest looks clean.

Adrianzatreanu:
@meetmangukiya can you approve/disapprove your request of changes? I dont think a newline is needed there either tbh.

jayvdb:
Re-assigned to someone else.

PR #4106, opened 2017-04-15
.travis.yml: Escape script after unsupported check

jayvdb:
Add an ``exit`` if UNSUPPORTED==true, and move the other
commands up to the top level of script block.

Follows style of https://github.com/coala/coala-bears/commit/a0f258e

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ addons:
 
 before_install:
   - if [[ $TRAVIS_OS_NAME == osx ]]; then TRAVIS_PYTHON_VERSION=3.5; fi
-  - export TRAVIS_PYTHON_VERSION_MAJOR=${TRAVIS_PYTHON_VERSION%.[0-9]}
-  - export TRAVIS_PYTHON_VERSION_MINOR=${TRAVIS_PYTHON_VERSION#[0-9].}
 
   - >
     if [[ $TRAVIS_OS_NAME == osx ]]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,24 +60,24 @@ before_script:
 
 script:
   - >
-    if [[ $TRAVIS_PYTHON_VERSION_MINOR < 4 || $TRAVIS_PYTHON_VERSION_MAJOR == 2 ]]; then
+    if [[ "$UNSUPPORTED" == true ]]; then
       coverage run setup.py install | grep -q 'coala supports only python 3.4 or later'
+    fi
+  - if [[ "$UNSUPPORTED" == true ]]; then exit; fi
```
RohanVB:
```
if [[ "$UNSUPPORTED" == true ]]; then 
exit; 
fi
```
or just the `fi` in next line?

.travis.yml
```diff
@@ -62,24 +60,24 @@ before_script:
...
+  - bash .misc/tests.sh
+  - python setup.py bdist_wheel
+  - pip install ./dist/coala-*.whl
+  - pip install coala-bears[alldeps] --pre -U
+  # https://github.com/coala/coala-bears/issues/1037
```
RohanVB: Why is this here?

.travis.yml
jayvdb: why?  Travis shows only the first line, so ideally the `exit` is there so the reader can see what is being done.
Why hide the `if`?

.travis.yml
jayvdb: This is an indentation only change, and should not be changed in this commit.
You can read more about it at https://github.com/coala/coala-bears/issues/1037

.travis.yml
RohanVB: IMO on the next line because it just made more sense in my mind as a block for readability.

.travis.yml
```diff
@@ -62,24 +60,24 @@ before_script:
...
+  - >
+    if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+      python3 -m nltk.downloader punkt maxent_treebank_pos_tagger averaged_perceptron_tagger
```
RohanVB: Can we load from cache here? We have `pip install coala-bears[alldeps] --pre -U` with the nltk merge

jayvdb: Waiting for https://github.com/coala/coala/pull/4128

.travis.yml
```diff
@@ -62,22 +60,22 @@ before_script:
...
-      pip install coala-bears[alldeps] --pre -U
-      # https://github.com/coala/coala-bears/issues/1037
-      if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
-        sed -i.bak '/bears = GitCommitBear/d' .coafile
-      fi
-      coala --non-interactive
-      make -C docs clean
-      python setup.py docs
     fi
+  - if [[ "$UNSUPPORTED" == true ]]; then codecov || true; exit; fi
```
Mixih: why don't we add this to the if block above?

jayvdb: because then we need `set -x -e` , and need to add trailing `;`, etc

.travis.yml
```diff
@@ -62,22 +60,22 @@ before_script:
...
+  - bash .misc/tests.sh
```
Mixih: intentional omission of the `set -e`?

jayvdb: Yes, travis does `set -e` by checking the error status of each entry in the build script.  It needs to be manually added only if there are multiple bash commands in a single entry of the build script, which travis cant 'see'.

Mixih:
ack 3b258e7

jayvdb:
@rultor merge

Issue #4094, opened 2017-04-12
Connect aspectlist with holder bear to match supported languages


userzimmermann:
To be able to also check in `coalib.bearlib.aspects.collections.aspectlist.__contains__` if languages of aspectclass instances match with languages supported by the bear holding the `aspectlist`

@coala/aspects-developers

jayvdb:
This is IMO very wrong.  An aspectlist could contain a list of languages, but it should not depend on bears.  Depending on bears creates a circular loop, and as far as I can see the only component of the bear that is needed is the bears languages.

adhikasp:
Yes the aspectlist only need to use bear's language and I don't see any downside of changing it to avoid circular loop. :+1:

userzimmermann:
@jayvdb @adhikasp IMO this is exactly what I proposed :) Having a back reference to the own holder is nothing unusual in OOP. But if you disagree, feel free! to only reference bear's languages accordingly...

PR #4109, opened 2017-04-15
Language.py: Change operators behavior to handle comparisons between Language instances as well

MariosPanag:
Add `compare` method to operators and apply normal
comparison in case we have 2 Language instances,
in order to avoid possible errors raised by calling
`limit_versions` method.

Fixes https://github.com/coala/coala/issues/4048

**For short term contributors:** we understand that getting your commits well
defined like we require is a hard task and takes some learning. If you
look to help without wanting to contribute long term there's no need
for you to learn this. Just drop us a message and we'll take care of brushing
up your stuff for merge!

### Checklist

- [ ] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [ ] I ran coala over my code locally. (*All commits have to pass
      individually.* It is not sufficient to have "fixup commits" on your PR,
      our bot will still report the issues for the previous commit.) You will
      likely receive a lot of bot comments and build failures if coala does not
      pass on every single commit!

After you submit your pull request, **DO NOT click the 'Update Branch' button.**
When asked for a rebase, consult [coala.io/rebase](https://coala.io/rebase)
instead.

Please consider helping us by reviewing other peoples pull requests as well:

- pick up any PR at <https://coala.io/review>
- review it (check <https://coala.io/reviewing> for more info)
- if you are sure that it needs work, use `cobot mark wip <URL>` to get it out
  of the review queue.

The more you review, the more your score will grow at coala.io and we will
review your PRs faster!

PR #4111, opened 2017-04-16
aspectlist: Connect aspectlist with holder bear

pratyushprakash:
Connect `asepectlist` with the holder bear to match aspects using
`__contains__` only if language of the aspect being checked matches
with a language that the bear handles.

Closes https://github.com/coala/coala/issues/4094


coalib/bearlib/aspects/collections.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, seq=()):
         list.__init__(self, seq)
 
     def __contains__(self, aspect):
+        if 'bear' in dir(self) and isinstance(aspect, aspectbase):
+            compatible = False
+            for language in self.bear.LANGUAGES:
+                if str(aspect.language).split()[0] == language:
```
pratyushprakash: Or instead of this we could maybe wait for https://github.com/coala/coala/pull/4109

coalib/bearlib/aspects/collections.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, seq=()):
         list.__init__(self, seq)
 
     def __contains__(self, aspect):
+        if 'bear' in dir(self) and isinstance(aspect, aspectbase):
```
adhikasp: How about using `hasattr()` instead for checking bear attribute?

tests/bears/BearTest.py
```diff
@@ -180,6 +180,10 @@ def test_detect_and_fix_aspects(self):
         assert (CommitMessage.Shortlog.TrailingPeriod in
                 aspectsTestBear.aspects['fix'])
 
+    def test_aspectlist_bear(self):
+        assert aspectsTestBear.aspects['fix'].bear == aspectsTestBear
```
adhikasp: Use `is` to assert the same object reference https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python

adhikasp:
I'll put this here so it's not hidden 👍 

@pratyushprakash  https://github.com/coala/coala/pull/4111#discussion-diff-111683153R22 regarding language comparison.
> Or instead of this we could maybe wait for #4109

tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -2,6 +2,14 @@
 
 from coalib.bearlib.aspects.collections import aspectlist
 from coalib.bearlib.aspects.Metadata import Metadata
+from coalib.bears.LocalBear import LocalBear
```
Techievena: IMO it is better to better to define this bear in `conftest.py` because we may further need this bear and its settings for future tests.

coalib/bearlib/aspects/collections.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, seq=()):
         list.__init__(self, seq)
 
     def __contains__(self, aspect):
+        if hasattr(self, 'bear') and isinstance(aspect, aspectbase):
+            compatible = False
+            for language in self.bear.LANGUAGES:
```
userzimmermann: We also have to extend `coalib.bears.meta.bearclass` for making this work properly. It might somehow work now this way, but `aspect.language` is a `coalib.bearlib.languages.Language.Language` instance, and we want to match it with `Language` instances the bear supports. Old-style `bear.LANGUAGES` will never hold `Language` instances. So I propose to add an additional `aspects=`-accompanying meta keyword-only `languages=` attribute to `bearclass` for adding `Language` instances to a bear and store them as lower-case `cls.languages` attribute

coalib/bearlib/aspects/collections.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, seq=()):
...
+                if str(aspect.language).split()[0] == language:
+                    compatible = True
+            if not compatible:
+                return False
```
userzimmermann: Even for now this can be simplified to:

```python
for ... in ...:
    if ... == ...:
        break
else:
   return False
```

coalib/bears/meta.py
```diff
@@ -27,3 +27,5 @@ def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None):
             cls.aspects = defaultdict(
                 lambda: aspectlist([]),
                 ((k, aspectlist(v)) for (k, v) in dict(aspects).items()))
+            for aspect_list in cls.aspects.values():
+                aspect_list.bear = cls
```
userzimmermann: Like mentioned in `aspectlist.__contains__`, better add `bear` to `aspectlist.__init__` and change `defaultdict` parameters to `lambda: aspectlist(cls, [])` and `((k, aspectlist(cls, v) for ...)`

tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -2,6 +2,14 @@
...
+
+
+class aspectsTestBear(LocalBear, aspects={
+        'detect': [Metadata.CommitMessage.Shortlog],
+        'fix': [Metadata.CommitMessage.Shortlog.TrailingPeriod],
+}):
+    LANGUAGES = {'Python'}
```
userzimmermann: Like mentioned in `aspectlist.__contains__`, better define a new meta keyword-only `languages=` attribute for `class` definition in `bearclass`

userzimmermann:
@pratyushprakash @adhikasp #4109 is not actually related to this, because there is already a properly working `Languages.__contains__`, and a `Languages` instance will be available as lower-case `bear.languages` after extending `bearclass` according to my proposal in https://github.com/coala/coala/pull/4111#discussion_r111831780 . Now is only the question if we want to implicitly realize that proposal as part of this PR or create a new issue/PR as prerequisite for this PR

@coala/aspects-developers

userzimmermann:
Refining https://github.com/coala/coala/pull/4111#issuecomment-294610911 related to https://github.com/coala/coala/pull/4111#issuecomment-294360035 : 

But of course #4109 is indirectly related to this, we only don't have to wait for it to finish this PR :)

@MariosPanag

coalib/bearlib/aspects/collections.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, seq=()):
...
         for item in self:
             if issubaspect(aspect, item):
```
userzimmermann: One further remark: You should move the whole language match logic to here, so you know first if the `aspect` is applicable at all for the bear. And you implicitly know that the incoming `aspect` object is either an aspect class or instance and can simplify `if isinstance(aspect, aspectbase)` to just `if not isclass(aspect)`

tests/bearlib/aspects/CollectionsTest.py
pratyushprakash: Maybe we do need it there but I don't think we should have any dependencies on other test files. We could have it in both places

Techievena: Actually whatever you define in `conftest.py` is used as settings for other test files in the repository or directory you can say because we have two `conftest files one for coala repo  and one for aspects.

pratyushprakash:
So now since `bear.languages` and `aspect.language` both contain `Language` instances, we have to wait for https://github.com/coala/coala/pull/4109 since there is no method to compare `Language` instances as of now.

coalib/bearlib/aspects/collections.py
```diff
@@ -1,15 +1,23 @@
...
-    def __init__(self, seq=()):
+    def __init__(self, seq=(), bear=None):
         super().__init__(map(assert_aspect, seq))
+        self.bear = bear
 
     def __contains__(self, aspect):
+        if self.bear is not None and not isclass(aspect):
+            for language in self.bear.languages:
+                if type(aspect.language) is type(language):
+                    break
```
pratyushprakash: This can be changed to
`if aspect.language not in self.bear.languages:
     return False`
after https://github.com/coala/coala/pull/4109/ is merged

userzimmermann: `Languages.__contains__` should also be sufficient and doesn't depend on #4109

coalib/bears/meta.py
```diff
@@ -14,16 +15,22 @@ class bearclass(type):
...
         """
         type.__init__(cls, clsname, bases, clsattrs, *varargs)
         if aspects is not None:
             cls.aspects = defaultdict(
-                lambda: aspectlist([]),
-                ((k, aspectlist(v)) for (k, v) in dict(aspects).items()))
+                lambda: aspectlist([], cls),
+                ((k, aspectlist(v, cls)) for (k, v) in dict(aspects).items()))
+            for aspect_list in cls.aspects.values():
+                aspect_list.bear = cls
```
userzimmermann: This `for` block is not needed anymore

coalib/bears/meta.py
```diff
@@ -14,16 +15,22 @@ class bearclass(type):
...
     aspects = defaultdict(lambda: aspectlist([]))
 
-    def __new__(mcs, clsname, bases, clsattrs, *varargs, aspects=None):
+    def __new__(mcs, clsname, bases, clsattrs, *varargs, aspects=None,
+                languages=Languages([])):
         return type.__new__(mcs, clsname, bases, clsattrs, *varargs)
 
-    def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None):
+    def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None,
+                 languages=Languages([])):
```
userzimmermann: I think `languages=None` would be more consistent here

userzimmermann:
@coala/aspects-developers I also think it would be clearer to split this PR into 2 commits, the first extending `bearclass` and closing https://github.com/coala/coala/issues/4147 , and the second only adding the `aspectlist` <--> bear linking and closing https://github.com/coala/coala/issues/4094

coalib/bears/meta.py
```diff
@@ -14,16 +15,19 @@ class bearclass(type):
...
-        given to the keyword-only `aspects` argument.
+        Initializes the ``.aspects`` dict and ``.languages`` array on new
+        bear classes from the mapping given to the keyword-only `aspects`
+        and `languages` arguments.
         """
         type.__init__(cls, clsname, bases, clsattrs, *varargs)
         if aspects is not None:
             cls.aspects = defaultdict(
-                lambda: aspectlist([]),
-                ((k, aspectlist(v)) for (k, v) in dict(aspects).items()))
```
userzimmermann: 😲 What happened here? I know you fixed that in the 2nd commit https://github.com/coala/coala/pull/4111/commits/7512b0486c6af1d48ed9ebaf407c8672d251c1aa but please make every commit work on its own ;)

tests/bearlib/aspects/conftest.py
```diff
@@ -76,3 +78,13 @@ class SubAspect:
         sour = SubAspect_tastes['sour']
 
     return SubAspect
+
+
+@pytest.fixture
+def aspectsTestBear():
+    class testBear(LocalBear, aspects={
```
userzimmermann: Can+Should also be named `aspectsTestBear`. Hmm... I wonder if you could directly decorate the `class` definition with `@pytest.fixture` in that case...

tests/bearlib/aspects/conftest.py
```diff
@@ -76,3 +78,13 @@ class SubAspect:
...
+        'detect': [Metadata.CommitMessage.Shortlog],
+        'fix': [Metadata.CommitMessage.Shortlog.TrailingPeriod]},
+                      languages=['Python']):
```
userzimmermann: And we should agree on some common syntax style here. Personally, I prefer:

```python
class aspectsTestBear(LocalBear, aspects={
        'detect': [Metadata.CommitMessage.Shortlog],
        'fix': [Metadata.CommitMessage.Shortlog.TrailingPeriod],
}, languages=['Python']):
    ...
```

But all @coala/aspects-developers have to make a common decision about this! :)

coalib/bears/meta.py
```diff
@@ -14,16 +15,19 @@ class bearclass(type):
...
 
-    def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None):
+    def __init__(cls, clsname, bases, clsattrs, *varargs, aspects=None,
+                 languages=None):
         """
-        Initializes the ``.aspects`` dict on new bear classes from the mapping
-        given to the keyword-only `aspects` argument.
+        Initializes the ``.aspects`` dict and ``.languages`` array on new
+        bear classes from the mapping given to the keyword-only `aspects`
+        and `languages` arguments.
```
userzimmermann: `languages` is no mapping. Needs some change like:

```
... from the mapping and the sequence given to the keyword-only `aspects`
and `languages` arguments, respectively.
```

tests/bearlib/aspects/conftest.py
Techievena: @pratyushprakash I think we should go with our team leader :wink:

coalib/bears/meta.py
```diff
@@ -14,16 +15,19 @@ class bearclass(type):
...
         """
         type.__init__(cls, clsname, bases, clsattrs, *varargs)
         if aspects is not None:
             cls.aspects = defaultdict(
-                lambda: aspectlist([]),
-                ((k, aspectlist(v)) for (k, v) in dict(aspects).items()))
+        if languages is not None:
```
userzimmermann: @sims1253 https://www.python.org/dev/peps/pep-0008/#programming-recommendations 

Depedending on its type, `languages.__bool__()` could return `False` but `languages` could be nevertheless a sequence of more than 0 elements

tests/bearlib/aspects/conftest.py
userzimmermann: @Techievena You make me very proud ;) but we should all provide our opinions on that :)

@coala/aspects-developers

userzimmermann:
@adhikasp Please take this over for your GSoC. I created https://gitlab.com/coala/GSoC-2017/issues/361 for this

@jayvdb

Issue #4010, opened 2017-03-28
Add `next_action` behaviour in Result Action


satwikkansal:
It should be possible for a `ResultAction` instance to determine the next action(s) for the user to choose at the runtime.

See https://github.com/coala/coala/issues/3650#issuecomment-289486520

sils:
hey @satwikkansal we have a PR open by @mrtes related to this as well, please check his commit https://github.com/coala/coala/pull/4179/commits/40fd9e6085fdbca0891c98f8b0b121e97fac630d , I do like his approach - do you have an opinion on that?

satwikkansal:
If I understand correctly, 40fd9e6 is about "not prompting the user to apply an action if it has already been applied once" . The approach of maintaining `successful_actions` (or previous actions) seems sane to me.

And `next_actions` are about "If the user applies X action, what are the next actions he may want to execute" (the `GitCommitBear` example in the end of https://github.com/coala/coala/issues/3650#issuecomment-289465422 may explain it more clearly)

There will be conflicts with https://github.com/coala/coala/pull/4112/commits/868991213aebce8e7bb8aa33d6287d7cc2bd94a8 but I feel like we can make both of the features work.

satwikkansal:
Also https://github.com/coala/coala/pull/4112#discussion_r115557903 seems relevant and needs attention

mrtes:
The question is if the behaviour I have implemented is still necessary, after your changes are completed.
If the next_actions are chosen properly, to avoid duplication (except where it makes sense, of course), the current issue with ignore comments will most likely be resolved as well.

However, I'm not sure I understand exactly _how_ the actions are chosen, so maybe I might be missing something :)
In the meantime, I will change my implementation to really return the action, instead of the action_name and fix the remaining issues. 
I guess we can decide how and if it should be merged afterwards!

satwikkansal:
> The question is if the behaviour I have implemented is still necessary, after your changes are completed.
If the next_actions are chosen properly, to avoid duplication (except where it makes sense, of course), the current issue with ignore comments will most likely be resolved as well.

Simply removing the while loop https://github.com/mrtes/coala/blob/40fd9e6085fdbca0891c98f8b0b121e97fac630d/coalib/output/ConsoleInteraction.py#L140 will solve the problem of ignore comments imo. However, this will lead to loss of the behavior "Asking users to chose same actions after executing an action", so the CLI will simply display next result after executing an action on the result. 
[...]

Issue #3650, opened 2017-01-29
Allow bears to pass their own applicable actions


Makman2:
I would suggest to introduce a new system for results so bears can easily add possible fixes / choices for the user.

For example for the `GitCommitBear`: When a commit message does not meet the standards, the user should have the choice to amend the current commit (see https://github.com/coala/coala-bears/issues/1360).

To allow such functionality, I would propose a field inside `Result` which is a list of `ResultAction`-instances.
```
Result.actions = XYZAction(...)
```
And the bear would generate a result with the possible actions. It's possible to set even multiple ones:
```
my_result = Result('Something went wrong')
my_result.actions += ApplyPatchAction(...)
my_result.actions += EmailSomeone(...)
```
This may also solve https://github.com/coala/coala/issues/1021, as you can easily add different patches:
```
my_result.actions += ApplyPatchAction(...) # <-- patch variant 1
my_result.actions += ApplyPatchAction(...) # <-- patch variant 2
```
Some actions are implicitly added by coala, like the `IgnoreResultAction`.

CC @sils @jayvdb

jayvdb:
Yes. We really need to slurp up pywikibot console functionality, which already has all this.

satwikkansal:
@Makman2 
Just wanted to discuss an approach that I was thinking.

The issue involves allowing the bears to add custom actions apart from those provided by coala by default (`ApplyPatchAction`, `PrintMoreInfoAction`, `ShowPatchAction`, `IgnoreResultAction`)

The suggested approach involves
```sh
   Problem detected
        /  |   \  \
      /    |    \   \
    /      |     |    \ 
   A      B     C    D ...
```
Where A, B, C, D .. are Result.action instances.

I was wondering if instead of just collecting the actions in a list, we can also allow actions to be cascaded or combined based on logical if else conditions. So the bear can define an `ActionSequence` as 

```sh
   A    B       C       H        i
   |    |      |        |         |
   D    F     exit     exit     exit
   |   / \
exit  E   G
      |    |
    exit   exit
```
So users are initially given an option

```sh
|    | [INFO] SomeBear:
|    | Some problems in the code
|    | The following actions are applicable to this result:
|    |  0: Apply no further actions. (equivalent to exit)
|    |  1:   Action A
|    |  2:  Action B
|    |  3:  Action C
|    |  4:  Action H
|    |  5:  Action i
|    | Please enter the number of the action you want to execute.
```
[...]

PR #4112, opened 2017-04-16
Add Result.actions and ResultAction.next_action fields

satwikkansal:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


satwikkansal:
@Makman2 I've tried to implement basic functionalities related to #3650 and #4010 in this iteration of the PR. There were few design decisions which I've added as FIXME commands in the code.

coalib/output/ConsoleInteraction.py
```diff
@@ -250,7 +252,9 @@ def print_result(console_printer,
     console_printer.print(format_lines(result.message))
 
     if interactive:
-        cli_actions = CLI_ACTIONS
+        # FIXME: Handle duplicate actions of same type across
+        # NATIVE_ACTIONS and result.actions
+        cli_actions = NATIVE_ACTIONS + tuple(result.actions)
```
Makman2: don't we need to define this field as an empty list/collection inside `Result`?

coalib/results/result_actions/ResultAction.py
```diff
@@ -12,6 +12,9 @@ class ResultAction:
 
     SUCCESS_MESSAGE = 'The action was executed successfully.'
 
+    def __init__(self):
+        self.next_actions = []
+
     @staticmethod
     def is_applicable(result, original_file_dict, file_diff_dict):
```
Makman2: We need to consider for the new design if we still need this or in what degree. I currently imagine that `ResultAction` classes with the new design will perform necessary checks of runtime-values a bear provides at instantiation. Just you are aware ;)

coalib/output/ConsoleInteraction.py
```diff
@@ -609,6 +609,51 @@ def print_actions(console_printer, section, actions, failed_actions):
...
+                            file_dict,
+                            file_diff_dict,
+                            console_printer,
+                            failed_actions):
+    """
+    Applies the given action.
+
+    :param action:          The action to the applied.
+    :param section:         Currently active section.
+    :param action_name:     The name of the action.
```
Makman2: can't you use the class name of `action`?

coalib/output/ConsoleInteraction.py
```diff
@@ -609,6 +609,51 @@ def print_actions(console_printer, section, actions, failed_actions):
...
+    :param result:          Result corresponding to the actions.
+    :param file_diff_dict:  If it is an action which applies a patch, this
+                            contains the diff of the patch to be applied to
+                            the file with filename as keys.
+    :param file_dict:       Dictionary with filename as keys and its contents
+                            as values.
+    :param console_printer: Object to print messages on the console.
```
Makman2: please use logging ;) Only the old functions do use `printer` params, but those are deprecated / are going to be deprecated, the more API we use without passing printers around, the better ;)

coalib/output/ConsoleInteraction.py
satwikkansal: Ah, do you mean the native actions are also the parts of the `result.actions` field implicitly and the `cli_actions` just become equivalent to  `result.actions`?

satwikkansal: Also, I'm confused regarding handling these `NATIVE_ACTIONS` for this PR's use case.
1. Should we allow users to pass native actions like `OpenEditorAction` in the `actions` field.
2. If yes, then how do we handle multiple occurrences of such actions in `result.actions`
3. Should the users be allowed to define `next_actions` for these Native actions, and if yes, how are they going to do so.

This issue becomes complex when we combine 1,2 and 3 points. So I think we'll need to design it properly.

satwikkansal: btw the strategy I had in mind for dealing with it was to remove duplicate actions of same class type. In case of  duplicacy, the action having non-empty `next_actions` field is preferred. If the duplicity still presists,  the most recent action is given preference.

coalib/output/ConsoleInteraction.py
satwikkansal: Yeah, actually the `ask_for_action_and_apply` already has `action_name` defined via the `print_action` function ( this [line](https://github.com/coala/coala/pull/4112/files/ebef35bd716b9bdb26e0a64543509d9e1ab1e1a3..0e2b15b19ed0fb39d4242c517611b617df7d573f#diff-5be6f2c646fedbab1fd2aea142f54303R686)), so I thought just to propagate it through.

satwikkansal: But this would make the function less usable, so I think better to remove it :+1:

coalib/output/ConsoleInteraction.py
satwikkansal: copied this ^^ discussion to https://github.com/coala/coala/issues/3650#issuecomment-294363647

coalib/output/ConsoleInteraction.py
satwikkansal: Okay, will make a separate commit for this transition from `printer` to `logging` in the module.

coalib/output/ConsoleInteraction.py
Makman2: > Ah, do you mean the native actions are also the parts of the result.actions field implicitly and the cli_actions just become equivalent to result.actions?

Ah no sorry I just meant we have to do inside `Result.__init__`
```
self.results = []
```
that's all ;)

coalib/output/ConsoleInteraction.py
Makman2: why not using logging from the beginning?

coalib/output/ConsoleInteraction.py
satwikkansal: Just confirming
`self.results = []` 
or 
`self.actions = NATIVE_ACTIONS` inside `Result.__init__`

coalib/output/ConsoleInteraction.py
satwikkansal: I meant the existing `printer` params, but nvm, I misinterpreted your first comment :sweat_smile:  I'll use logging for the new code and existing usage of `printer` should be changed to logging in separate issue :+1:

coalib/results/result_actions/ResultAction.py
satwikkansal: hmm, we can check the applicability of the ResultAction at the time of instantiation, but then it'd need to be initialized only after the Result is available.

coalib/output/ConsoleInteraction.py
satwikkansal: @Makman2 I've confusion between `log_printer` and `console_printer`, from the codebase and currently filed issues it looks like `log_printer` is deprecated. Here I'm using `console_printer` to print the success message after applying the action, and it also involves formatting and colors. Do you want me to use `logging.info` instead to display the success message and use `assertLogs` in the tests?

satwikkansal:
Phew, handling too many commits is pain. I'll add more tests in the next iteration once the design is agreed :)

coalib/output/ConsoleInteraction.py
```diff
@@ -131,36 +130,61 @@ def acquire_actions_and_apply(console_printer,
...
-                actions.append(action)
+        actions = result.get_filtered_actions(file_dict, file_diff_dict)
 
         if actions == []:
             return
 
         action_dict = {}
         metadata_list = []
+        # FIXME: If we want to allow users to add multiple actions of same
+        # type, ``metadata.name`` should not be used as a key for action_dict
```
satwikkansal: Also, we'll have design a way to let the user know how are the multiple actions of same class different, because currently if we allow multiple actions, all of the actions would have the same description leading to confusion for the user.

Makman2: good point. So I think easiest is to let the action take another parameter if necessary, like `tag` or `name`, and the description is built with that tag. Like
```
my_actions += ApplyPatchAction(patch, 'rename')
```
or we allow to modify the description completely.
```
my_actions += ApplyPatchAction(patch, 'Rename the variable to "hello_world"')
```

Makman2: (I like alternative 2 more, as we can suggest better descriptions)

coalib/output/ConsoleInteraction.py
```diff
@@ -131,36 +130,61 @@ def acquire_actions_and_apply(console_printer,
...
-                                        result,
-                                        file_diff_dict,
-                                        file_dict):
+        if not applied_action:
             break
+        else:
+            next_actions = applied_action.next_actions
+
+            # FIXME: currently the code doensn't check for
+            # circular dependencies among the next_actions
```
sils: actions aren't dependencies right? And the user decides which to execute next, they may be circular because there's always a way out for the user right?

coalib/output/ConsoleInteraction.py
```diff
@@ -131,36 +130,61 @@ def acquire_actions_and_apply(console_printer,
...
+            while next_actions:
+                if len(next_actions) == 1:
```
sils: this will never be the case, either there are no actions or there's one possible action plus the option to do nothing

Makman2: you sure? The idea is to chain actions together so they get executed automatically in some cases^^ Though maybe we should always insert a "do nothing" action. Not sure...

satwikkansal: Ah yep, the basic idea was to allow Bears to implicitly execute actions in a chain, though I'm also confused whether we should always have a "do nothing" action or not.

coalib/output/ConsoleInteraction.py
satwikkansal: The reason for this is https://github.com/coala/coala/pull/4112#discussion_r115533561 but if we have a "do nothing" action always, then there won't be any such case

coalib/output/ConsoleInteraction.py
```diff
@@ -147,16 +147,44 @@ def acquire_actions_and_apply(console_printer,
...
+                                    if applied_action else [])
+
+                if len(next_actions) > 1:
+                    result.actions = next_actions
+                    acquire_actions_and_apply(console_printer,
+                                              section,
+                                              file_diff_dict,
+                                              result,
+                                              file_dict)
+                    break
```
satwikkansal: There is one unintuitive behavior currently, 
Due to the outer `while True:` loop and the inner `while next_actions:` loop, the user is prompted for a particular set of `next_actions` twice (because a "do nothing" action breaks the inner while loop and the outer while loop still persists) which seems unintuitive to me as a chained action behaviour. Not sure if that should be the desired behavior. 

imo the "do nothing" should break out of both the loops or we may have to combine these two while loops somehow.

satwikkansal: Or maybe we omit the outer `while` loop. I can't figure any other use out the use of that "while" loop except 
"Asking the users to choose the same actions again after applying an action, until and unless user chooses to "do nothing".

Makman2: You will figure out something ;D

mrtes: See my suggestion in #4010 (last comment).

sils:
marking this blocked by https://github.com/coala/coala/pull/4179

@satwikkansal can you rebase it after that one is merged and do the needed additions or complain if I'm misjudging this? :)

PR #4133, opened 2017-04-24
Makman2/nextgen3-6

Makman2:


coalib/core/DependencyTracker.py
```diff
@@ -97,6 +97,65 @@ def get_dependencies(self, dependant):
...
+            A set of dependants.
+        """
+        dependants = set()
+
+        def append_to_dependants(prev, nxt):
+            dependants.add(nxt)
+
+        traverse_graph(
+            [dependency],
+            lambda node: self._dependency_dict.get(node, frozenset()),
```
Nosferatul: is there any difference between frozenset and set except set has mutating methods added?

Nosferatul:
ack feda9df

coalib/core/DependencyTracker.py
Makman2: no that's the only difference :+1:
Actually I thought this gives a slight performance improvement as frozenset doesn't require an extra allocation, though the same would apply to a normal set, as it is reserved once before `traverse_graph` is called for the passed lambda. If you want I change it to set, it doesn't really matter^^

Makman2: Okay there still a little improvement: If somebody has used an empty frozenset before in the program, it's reused here. But this is such a super super minor mini improvement^^

Nosferatul: it's fine 👍

Nosferatul:
ack 6adec40

Makman2:
reack 8e3c280, 8dbd5e6

Makman2:
@rultor merge

Issue #4163, opened 2017-05-03
Add argcomplete to test-requirements


jayvdb:
Currently we do not have any testing to ensure that our `argcomplete` support continues to work, with any specific version of `argcomplete`.

PR #4164, opened 2017-05-03
test-requirements.txt: Add argcomplete

jayvdb:
Adding argcomplete to test-requirements.txt allows simple
compatibility testing with argcomplete.

Closes https://github.com/coala/coala/issues/4163

### DIFF
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -230,11 +230,12 @@ To run coala without user interaction, run the `coala --non-interactive`,
         '-n', '--no-orig', const=True, action='store_const',
         help="don't create .orig backup files before patching")
 
-    try:  # pragma: no cover
+    try:
         # Auto completion should be optional, because of somewhat complicated
         # setup.
         import argcomplete
         argcomplete.autocomplete(arg_parser)
-    except ImportError:
+    except ImportError:  # pragma: no cover
         pass
+
     return arg_parser
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
+argcomplete~=1.8
 coverage~=4.3.4
 codecov~=2.0.5
 pytest~=3.0


all3fox:
I see nothing wrong with bundling `argcomplete` as a dependency. Yet I do not really see the point: the user still has to read the documentation on which lines to add to their `.bashrc` or `.zshrc` or, alternatively, they need to have `argcomplete` enabled globally.

I tried to test the global completion scenario. As far as I understand, this PR does not completely enable it because it does not add `PYTHON_ARGCOMPLETE_OK` comments to `coala` script. However, I was not able to either confirm or deny my suspicion: [bash on archlinux][1]

all3fox:
Also, I tried installing a separate autocompletion package for bash on archlinux:

```
sudo pacman -S bash-completion
```

Then, I logged-out-then-back-in to the shell:

```
bash > logout
zsh > bash
bash > activate-global-python-argcomplete
```

which fails, so I looked at where the `bash-completion` package puts its stuff:

```
pacman -Ql bash-completion
```

and then tried installing with a custom directory:

```
sudo activate-global-python-argcomplete --dest=/usr/share/bash-completion/completions
```

Only after that the global autocompletion mode kicked in (becaise Arch packages bash and its completion scripts separately and the scripts go into a non-default directory).

Now everything that begins with `python` autcompletes, like `python setup.pu -<TAB>` shows possible switches. The `coala` still does not autocomplete though, even if I add the `PYTHON_ARGCOMPLETE_OK` comment onto the second line of that script and restart the shell. Sad for now.

jayvdb:
This PR isnt about argcomplete global mode (`PYTHON_ARGCOMPLETE_OK`).  It is about tracking the optional dependency that already exists as a test dependency, to allow us to become aware of any breakages.

Once this PR is merged, the next step is a hard one: somehow testing the custom `.completer` attribute: https://github.com/coala/coala/issues/4165 .

sils:
ack https://github.com/coala/coala/pull/4164/commits/bdae61aa5771727addb5a6d6684aeecaaf44cf5b

sils:
@all3fox hey if the autocomplete doesn't work for you it's probably best to discuss this in an issue,there might already be one

jayvdb:
reack d2d8769

jayvdb:
@rultor merge

Issue #4166, opened 2017-05-03
Remove pragma: no cover


jayvdb:
`pragma: no cover` is being over used to avoid writing test cases, and slipping through review.
It should be prevented, requiring more specific pragma's that are less likely to skip through review without being questioned.

jayvdb:
See also https://github.com/coala/coala-bears/issues/1618 for the same problem in bears

jayvdb:
Two recent merged PRs which helped in this regard: https://github.com/coala/coala/pull/4160 and https://github.com/coala/coala/pull/4162

jayvdb:
A pending PR which increases the ability to whitelist a specific scenario where testing is 'hard': https://github.com/coala/coala/pull/4164

jayvdb:
Another pending PR to reduce the usage of pragma : https://github.com/coala/coala/pull/4167

PR #4169, opened 2017-05-04
Remove unspecific pragma no cover

jayvdb:
The assault on https://github.com/coala/coala/issues/4166 continues...


coalib/collecting/Importers.py
```diff
@@ -21,12 +21,17 @@ def _import_module(file_path):
...
         for cased_file_path in os.listdir(module_dir):
             cased_module_name = os.path.splitext(cased_file_path)[0]
             if cased_module_name.lower() == module_name.lower():
                 module_name = cased_module_name
                 break
+        else:  # pragma nt: no cover
+            # There should never be a file_path which exists, passing through
+            # the check at the top of the function, and fails to be a case
+            # insensitive match of an item in the directory module_dir.
+            raise RuntimeError
```
jayvdb: codecov is complaining about this not being covered.  I've reported it on their gitter support.
The only way I can see to avoid this problem, is to remove it.  Which is OK, because it should be unreachable.
But it is a safety net.

coalib/collecting/Importers.py
```diff
@@ -21,12 +21,17 @@ def _import_module(file_path):
...
         for cased_file_path in os.listdir(module_dir):
             cased_module_name = os.path.splitext(cased_file_path)[0]
             if cased_module_name.lower() == module_name.lower():
                 module_name = cased_module_name
                 break
+        else:  # pragma nt: no cover
+            # There should never be a file_path which exists, passing through
+            # the check at the top of the function, and fails to be a case
+            # insensitive match of an item in the directory module_dir.
+            raise RuntimeError('Unreachable code')
```
userzimmermann: The whole `_import_module` function should be rewritten according to https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

That implicitly works case-insensitive under Windows, so there wouldn't be any `# pragma` need at all anymore...

jayvdb: Nice. I felt this was somehow unnecessary.

setup.cfg
```diff
@@ -57,7 +57,10 @@ sort = Miss
 include =
     coalib/*
 exclude_lines =
-    pragma: no ?cover
     pragma ${PLATFORM_SYSTEM}: no cover
     pragma ${OS_NAME}: no cover
     pragma Python [0-9.,]*${PYTHON_VERSION}[0-9.,]*: no cover
+    raise RuntimeError.*Unreachable code
```
Makman2: maybe we should put those two changes into a separate commit? So first we try to improve current pragmas, and in the end we remove the generic one?

Makman2: Up to you :+1:

Makman2:
ack e36871c

jayvdb:
Split off https://github.com/coala/coala/pull/5680 and https://github.com/coala/coala/pull/5682

jayvdb:
reack https://github.com/coala/coala/pull/4169/commits/737a8cf40f1b98d191606f6dbe8db330daf27ede

coalib/core/Core.py
```diff
@@ -241,8 +241,8 @@ def _schedule_bears(self, bears):
         bears_without_tasks = []
 
         for bear in bears:
-            if self.dependency_tracker.get_dependencies(
-                    bear):  # pragma: no cover
+            always_false = bool(self.dependency_tracker.get_dependencies(bear))
```
jayvdb: I'm not very happy about this solution, so happy to hear of alternatives.

setup.cfg
```diff
@@ -65,10 +65,15 @@ show_missing = True
...
 exclude_lines =
-  pragma: no ?cover
   pragma ${PLATFORM_SYSTEM}: no cover
   pragma ${OS_NAME}: no cover
   pragma Python [0-9.,]*${PYTHON_VERSION}[0-9.,]*: no cover
   sys.exit\(main\(\)\)
+  raise RuntimeError.*Unreachable code
+  if always_false is True:
+  if get_running_processes\(processes\) < 2:
+partial_branches =
```
jayvdb: this needs to be put in the template

coalib/output/ConsoleInteraction.py
```diff
@@ -9,7 +9,7 @@
 try:
     # This import has side effects and is needed to make input() behave nicely
     import readline  # pylint: disable=unused-import
-except ImportError:  # pragma: no cover
+except ImportError:  # pragma posix: no cover
```
jayvdb: This is causing problems on Windows, as we use coala-utils, and it includes pyreadline on Windows, so this never fails.

https://gitlab.com/coala/coala-utils/commit/43c81871476e14ce0d777129bff8cdfc557be342

possibly we need tests like were written for https://github.com/coala/coala/pull/5645

jayvdb:
https://codecov.io/gh/coala/coala/pull/4169/diff showing coverage of these changes; unfortunately AppVeyor isnt posting results atm, so the coverage of Windows specific bits isnt shown.

coalib/collecting/Importers.py
jayvdb: https://github.com/coala/coala/pull/4210 was a PR to solve that, but it is abandoned it seems.

@refeed 's https://github.com/coala/coala/pull/4725 is also related

PR #4171, opened 2017-05-04
Setting.__path__: Use Setting.value instead of str(Setting)

jayvdb:
str(Setting) will call unescape, which apparently causes problems on Windows. e.g. https://github.com/coala/coala-bears/pull/1425
c.f. https://gitlab.com/coala/coala-utils/issues/27

This has been tested on all bears using the docker: https://travis-ci.org/jayvdb/coala/builds/228653730

### DIFF
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -163,7 +163,11 @@ class Setting(StringConverter):
         :raises ValueError:        If no origin is specified in the setting
                                    nor the given origin parameter.
         """
-        strrep = str(self).strip()
+        if hasattr(self, 'value'):
+            strrep = self.value.strip()
+        else:
+            strrep = str(self).strip()
+
         if os.path.isabs(strrep):
             return strrep
 
--- a/tests/coalaDeleteOrigTest.py
+++ b/tests/coalaDeleteOrigTest.py
@@ -1,7 +1,6 @@
 import tempfile
 import unittest
 import os
-import re
 
 from coalib import coala_delete_orig
 from coala_utils.ContextManagers import retrieve_stderr
@@ -45,7 +44,7 @@ class coalaDeleteOrigTest(unittest.TestCase):
             temporary = tempfile.mkstemp(suffix='.orig', dir=directory)
             os.close(temporary[0])
             section = Section('')
-            section.append(Setting('project_dir', re.escape(directory)))
+            section.append(Setting('project_dir', directory))
             retval = coala_delete_orig.main(section=section)
             self.assertEqual(retval, 0)
             self.assertFalse(os.path.isfile(temporary[1]))
--- a/tests/settings/SettingTest.py
+++ b/tests/settings/SettingTest.py
@@ -1,5 +1,4 @@
 import os
-import re
 import unittest
 from collections import OrderedDict
 
@@ -24,7 +23,7 @@ class SettingTest(unittest.TestCase):
                          os.path.abspath(os.path.join('.', '22')))
 
         abspath = os.path.abspath('.')
-        self.uut = Setting('key', re.escape(abspath))
+        self.uut = Setting('key', abspath)
         self.assertEqual(path(self.uut), abspath)
 
         self.uut = Setting('key', ' 22', '')


coalib/settings/Setting.py
```diff
@@ -163,7 +163,11 @@ def __path__(self, origin=None, glob_escape_origin=False):
         :raises ValueError:        If no origin is specified in the setting
                                    nor the given origin parameter.
         """
-        strrep = str(self).strip()
+        if hasattr(self, 'value'):
+            strrep = self.value.strip()
```
adtac: `str(self)` basically does [`unescape(self.value)`](https://gitlab.com/coala/coala-utils/blob/master/coala_utils/string_processing/StringConverter.py#L38). [`unescape`](https://gitlab.com/coala/coala-utils/blob/master/coala_utils/string_processing/Core.py#L409) basically strips away all backslashes. It looks like Lasse added the `str(self)` thing in the [first commit](https://github.com/coala/coala/commit/fbe61543788c20c0806755e27662ba48d4cf5d03#diff-2663e4c825081e483f6ed8d673b78545R26) that added `Settings.py`. I'm not entirely sure why we have that.

Anyway, I think the solution should be to override the `__str__` of `StringConvertor` here (by adding a `def __str__(string):`.

adtac: Because I think `self` will always have `value` and the `else` path will never be taken.

jayvdb: > I think self will always have value and the else path will never be taken.

I thought the same too, but ... we have `str` objects being passed in as self, as crazy as that sounds.

If the `else` path wasnt being taken, the 100% coverage would fail :P 

I've tested adding a `Setting.__str__`, and that fails oh so many tests in `coala`, and PyLintBear fails also.  Nice idea though; worth a shot.

I can replace `if hasattr(self, 'value'):` with `if isinstance(self, Setting):` , which makes the bizarre situation a bit more obvious.  I've got no idea what madness has created this situation; I am sure when I figure it out, I will be hell bent on eradicating it.

adtac: `str` objects are used as `Section` objects? WTF?!

>I can replace `if hasattr(self, 'value'):` with `if isinstance(self, Setting):`

That should work yeah. But the fact that we need it is baffling!

jayvdb:
Good news.  This patch does fix the problems with XMLLint on Windows.
https://ci.appveyor.com/project/jayvdb17618/coala-bears/build/1.0.119/job/anorujalhbnjixcy#L550

Bad news. `GitCommitBearTest.test_different_path` breaks because `GitCommitBearTest` adds an explicit `escape`.
https://ci.appveyor.com/project/jayvdb17618/coala-bears/build/1.0.119/job/anorujalhbnjixcy#L719
The simple fix would be to remove that `escape` after this fix is merged, but that approach will mean there are 'some' red builds after this is merged until the bears fix is merged.
The yucky approach is to update the bears test to succeed if the string is escaped or not, then merge this, then restrict the bears test to only accept unescaped.

adtac:
+1 for the temporary failures. That's the more long-term solution IMO.

adtac:
But I'm actually surprised that GitCommitBear is the only one that's failing.

Techievena:
Other bears skip on Appveyor. There are more. :wink:

jayvdb:
@Techievena , could you clarify what you mean?

https://ci.appveyor.com/project/coala/coala-bears/build/1.0.6307/job/crs6da8fyv934fny has
> = 315 passed, 273 skipped, 27 deselected, 1 pytest-warnings in 140.56 seconds =

https://ci.appveyor.com/project/jayvdb17618/coala-bears/build/1.0.119/job/anorujalhbnjixcy (my build) has
> 1 failed, 323 passed, 261 skipped, 27 deselected, 1 pytest-warnings in 139.81 seconds 

In my build, more are passing, and less are being skipped.  Obviously I am not attempting to solve all skips.  Patches like https://github.com/coala/coala-bears/pull/1312 will achieve that.

coalib/settings/Setting.py
jayvdb: I have raised that as a separate issue : https://github.com/coala/coala/issues/4201

jayvdb:
An important part of creating a clean fix for this requires that `Setting.origin` is a `str` and is not an `int`, which happens to be semi-legal.  Please review https://github.com/coala/coala/pull/4203 rather urgently.

sils:
#4203 is merged

coalib/settings/Setting.py
sils: I think it'd be good to have a comment with a link to the issue here otherwise someone will come and remove this, figure out it doesn't work, file another issue, ...

jayvdb:
Thx.  I can proceed with https://github.com/coala/coala/issues/4201 , with a clear and simple fix.

But the more I dig under the covers, the less happy I am with this fix.
The escaping is very hard wired into Setting. It should need to be unescaped before use with `os.path`. Thus I cant explain why this fix is needed, and I doubt it is correct.

sils:
given https://github.com/coala/coala/issues/4201 is this change needed?

sils:
If I'm judging this correctly this can be closed, if I'm not you can reopen it :)

PR #4174, opened 2017-05-04
core: Add group function

Makman2:
Allows to group sequences using a key. Supports out-of-order-processing
of elements. That's why it's worse to use Python's `itertools.groupby`,
because it does not support out-of-order-processing.

### DIFF
--- a/coalib/core/Core.py
+++ b/coalib/core/Core.py
@@ -4,6 +4,61 @@ import functools
 import logging
 
 
+def group(iterable, key=lambda x: x):
+    """
+    Groups elements (out-of-order) together in the given iterable.
+
+    Supports non-hashable keys by comparing keys with ``==``.
+
+    Accessing the groups is supported using the iterator as follows:
+
+    >>> for key, elements in group([1, 3, 7, 1, 2, 1, 2]):
+    ...     print(key, list(elements))
+    1 [1, 1, 1]
+    3 [3]
+    7 [7]
+    2 [2, 2]
+
+    You can control how elements are grouped by using the ``key`` parameter. It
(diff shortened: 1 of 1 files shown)


coalib/core/Core.py
```diff
@@ -4,6 +4,61 @@
...
+            position = keys.index(k)
+            element_list = elements[position]
+        except ValueError:
+            keys.append(k)
+            element_list = []
+            elements.append(element_list)
+
+        element_list.append(element)
+
+    return zip(keys, elements)
```
Udayan12167: """
def group(iterable, key=lambda x: x):
    group_dict = defaultdict(list)
    for element in iterable:
        k = key(element)
        group_dict[k].append(element)
    return group_dict.items()
"""

Udayan12167: This better? Much less lines of code.

Makman2: the key might not be hashable (and we use an unhashable key in our core^^)

Makman2: There's another shorter possibility to do this, but it has another disadvantage: The elements have to support ordering (and in our case for the core the elements (bears) don't support ordering).
```
from itertools import groupby

return groupby.sorted(iterable, key=key)
```

userzimmermann:
ack https://github.com/coala/coala/pull/4174/commits/61cf38e6c33dacfe4cf869f77a2b286b56454a73

Makman2:
Yeah I was disappointed too that Python has no existing way to meet those requirements... :/

Makman2:
reack 2cf2416

Makman2:
@rultor merge

Issue #3375, opened 2016-12-23
After adding an ignore comment, the option should not be available any more.


Udayan12167:

Udayan12167:
See https://github.com/coala/coala/issues/3366

adtac:
@mrtes From gitter:

> how about this: have a successful_actions set similar to failed_actions that will be appended the name to every time an action is successfully. Then allow IgnoreResultAction.is_applicable to take another parameter: successful_actions. If successful_name is in the list, make is_applicable return False.

>the difficulty is just a label - anyone is welcome to tackle any issue (provided you can)
you can solve the coala/coala#3377 with the other method too
you might need to tweak things a little bit
the reason I prefer the method I mentioned: this way we make our changes mostly on the *Action.py files instead of the core itself

Issue #3377, opened 2016-12-23
When applying a patch, don't ask for something else, continue with next result.


Udayan12167:

Udayan12167:
See https://github.com/coala/coala/issues/3366

PR #4179, opened 2017-05-05
Ignore Action: Remove action after comment added

mrtes:
This avoids asking the users if they want to add an ignore comment,
AFTER they already applied one.
Includes a new parameter for Action.is_applicable, which can be
checked for earlier actions.
Change in behaviour for IgnoreResultAction:
If this action was already used before, Action.is_applicable will
always return false, even on other failure cases.

Closes https://github.com/coala/coala/issues/3375

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -165,45 +165,37 @@ def acquire_actions_and_apply(console_printer,
     cli_actions = CLI_ACTIONS if cli_actions is None else cli_actions
     failed_actions = set()
     applied_actions = {}
+
     while True:
-        actions = []
-        for action in cli_actions:
-            if action.is_applicable(result, file_dict, file_diff_dict) is True:
-                actions.append(action)
-
-        if actions == []:
-            return
-
         action_dict = {}
         metadata_list = []
-        for action in actions:
-            metadata = action.get_metadata()
-            action_dict[metadata.name] = action
-            metadata_list.append(metadata)
+
+        for action in cli_actions:
+            if action.is_applicable(result,
+                                    file_dict,
+                                    file_diff_dict,
+                                    tuple(applied_actions.keys())) is True:
+                metadata = action.get_metadata()
+                action_dict[metadata.name] = action
+                metadata_list.append(metadata)
+
+        if not metadata_list:
+            return
 
         # User can always choose no action which is guaranteed to succeed
-        if apply_single:
-            ask_for_action_and_apply(console_printer,
-                                     section,
-                                     metadata_list,
-                                     action_dict,
-                                     failed_actions,
-                                     result,
-                                     file_diff_dict,
-                                     file_dict,
-                                     applied_actions,
-                                     apply_single=apply_single)
-            break
-        elif not ask_for_action_and_apply(console_printer,
-                                          section,
-                                          metadata_list,
-                                          action_dict,
-                                          failed_actions,
-                                          result,
-                                          file_diff_dict,
-                                          file_dict,
-                                          applied_actions,
-                                          apply_single=apply_single):
+        continue_interaction = ask_for_action_and_apply(
+            console_printer,
+            section,
+            metadata_list,
+            action_dict,
+            failed_actions,
+            result,
+            file_diff_dict,
+            file_dict,
+            applied_actions,
+            apply_single=apply_single
+        )
+        if not continue_interaction:
             break
 
 
@@ -778,6 +770,7 @@ def ask_for_action_and_apply(console_printer,
     :return:                Returns a boolean value. True will be returned, if
                             it makes sense that the user may choose to execute
                             another action, False otherwise.
+                            If apply_single ist set, always return False.
     """
     actions_desc, actions_name = choose_action(console_printer, metadata_list,
                                                apply_single)
@@ -801,6 +794,7 @@ def ask_for_action_and_apply(console_printer,
                                     file_diff_dict,
                                     file_dict,
                                     applied_actions)
+        return False
     else:
         for action_choice, action_choice_name in zip(actions_desc,
                                                      actions_name):
--- a/coalib/results/result_actions/IgnoreResultAction.py
+++ b/coalib/results/result_actions/IgnoreResultAction.py
@@ -18,11 +18,18 @@ class IgnoreResultAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         """
         For being applicable, the result has to point to a number of files
         that have to exist i.e. have not been previously deleted.
+        Additionally, the action should not have been applied to the current
+        result before.
         """
+        if IgnoreResultAction.__name__ in applied_actions:
+            return 'An ignore comment was already added for this result.'
 
         if len(result.affected_code) == 0:
             return 'The result is not associated with any source code.'
--- a/coalib/results/result_actions/OpenEditorAction.py
+++ b/coalib/results/result_actions/OpenEditorAction.py
@@ -101,7 +101,10 @@ class OpenEditorAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         """
         For being applicable, the result has to point to a number of files
         that have to exist i.e. have not been previously deleted.
--- a/coalib/results/result_actions/PrintAspectAction.py
+++ b/coalib/results/result_actions/PrintAspectAction.py
@@ -8,7 +8,10 @@ class PrintAspectAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         if result.aspect is None:
             return 'There is no aspect associated with the result.'
         return True
--- a/coalib/results/result_actions/PrintDebugMessageAction.py
+++ b/coalib/results/result_actions/PrintDebugMessageAction.py
@@ -8,7 +8,10 @@ class PrintDebugMessageAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         if result.debug_msg != '':
             return True
         return 'There is no debug message.'
--- a/coalib/results/result_actions/PrintMoreInfoAction.py
+++ b/coalib/results/result_actions/PrintMoreInfoAction.py
@@ -8,7 +8,10 @@ class PrintMoreInfoAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         if result.additional_info != '':
             return True
         return 'There is no additional info.'
--- a/coalib/results/result_actions/ResultAction.py
+++ b/coalib/results/result_actions/ResultAction.py
@@ -13,7 +13,10 @@ class ResultAction:
     SUCCESS_MESSAGE = 'The action was executed successfully.'
 
     @staticmethod
-    def is_applicable(result, original_file_dict, file_diff_dict):
+    def is_applicable(result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
         """
         Checks whether the Action is valid for the result type.
 
@@ -28,6 +31,9 @@ class ResultAction:
                                    original_file_dict to the current state.
                                    This dict will be altered so you do not
                                    need to use the return value.
+        :applied_actions:          List of actions names that have already been
+                                   applied for the current result. Action names
+                                   are stored in order of application.
         """
         return True
 
--- a/coalib/results/result_actions/ShowPatchAction.py
+++ b/coalib/results/result_actions/ShowPatchAction.py
@@ -58,7 +58,10 @@ class ShowPatchAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=()):
 
         if not result.diffs:
             return 'This result has no patch attached.'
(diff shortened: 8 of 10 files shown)


mrtes:
I implemented this according to your suggestions @adtac.
Most changes were test fixes and additions as well as signature fixes to accomodate the new parameter.
Its kinda weird to have this new parameter in every ResultAction, while only using it once (currently).

Of course this new parameter could be changed by #3377, to remove all actions after an applied patch.
However, a clean exit like that should probably be handled like the "do_nothing" operation, i.e. with a immediate break of the loop (instead of checking every action).

Anyway, what are your thoughts about the current implementation?

mrtes:
On the other hand, if we want a consistent behaviour for all actions, i.e. they should all not be displayed after being applied once, the new parameter will find more uses and will also allow for mutually exclusive decisions (if that should become an issue at some point in time).

.gitignore
```diff
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+.vscode/
```
mrtes: OK. So I put in in a separate commit, but in the same pull request?
Should I also reference the issue in this second commit?

coalib/output/ConsoleInteraction.py
```diff
@@ -135,32 +135,37 @@ def acquire_actions_and_apply(console_printer,
...
     failed_actions = set()
-    while True:
-        actions = []
-        for action in cli_actions:
-            if action.is_applicable(result, file_dict, file_diff_dict) is True:
-                actions.append(action)
-
-        if actions == []:
-            return
+    successful_actions = set()
```
mrtes: Yeah, i guess using a set was only to be consistent with the failed_action variable, which is also a set :)
Shouldn't be a problem to use a list instead.

sils:
Separate commit, same PR is fine - we don't care much which commits are I.
What pr; it has nothing to do with the issue

mrtes:
Applied the changes we discussed (list instead of set) and moved the .gitignore to a new commit.
Got some issues with newlines in the files along the way and wanted stuff to be consistent (i.e. no changes to the .gitignore, not even new lines, in the first commit).

.gitignore
```diff
@@ -36,3 +36,4 @@ node_modules
 .vendor
 Goopfile.lock
 venv/
+.vscode
```
sils: missing newline before EOF

sils:
btw. this stuff is great, there might be some potential conflicts with a PR from @satwikkansal : https://github.com/coala/coala/pull/4112/ ; checking...

.gitignore
mrtes: Fixed.
I also changed the return type from action_names to the action itself (see #4010 (last comment))

mrtes:
@sils: Check the discussion in #4010.

mrtes:
Rebased and ready for review. 
Thanks for the reminder.

sils:
I'm not sure I've seen anyone thanking our bot for reminding to pick up a PR, usually people are quite annoyed ^^ we should use this as PR for gitmate.io :D

mrtes:
I was waiting for a review anyway, so if the bot complains I don't have to ;)
Btw. @sils care to review?

sils:
@mrtes I planned to and then two other projects got on fire ^^

mrtes:
Sooo, someone gonna review that? 
I can rebase it :)

sils:
@mrtes reviewing now :)

.gitignore
sils: https://github.com/coala/coala/pull/4179/commits/2752e510bad6fe38161473502d59b94b5693203a still shows no newline before EOF

coalib/output/ConsoleInteraction.py
```diff
@@ -135,32 +135,37 @@ def acquire_actions_and_apply(console_printer,
...
+                                                 metadata_list,
+                                                 action_dict,
+                                                 failed_actions,
+                                                 result,
+                                                 file_diff_dict,
+                                                 file_dict)
+        if not chosen_action:
             break
+        else:
+            successful_actions.append(chosen_action)
```
sils: successful_actions is misleading since this also includes actions that had exceptions right?

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -16,11 +16,18 @@ class IgnoreResultAction(ResultAction):
 
     @staticmethod
     @enforce_signature
-    def is_applicable(result: Result, original_file_dict, file_diff_dict):
+    def is_applicable(result: Result,
+                      original_file_dict,
+                      file_diff_dict,
+                      applied_actions=[]):
```
sils: can we have an empty tuple here? Mutable default args are dangerous.

sils: we should have pylint on gitmate enabled... ^^

sils:
unack 2752e51 21260ad

coalib/results/result_actions/IgnoreResultAction.py
mrtes: Sure, I can change that. Should I also change the non-default parameter type to a tuple, to be consistent?
In that case should I use a tuple for the `successful_actions` variable, or just convert to tuple when it is passed to the function?

coalib/output/ConsoleInteraction.py
mrtes: Yes, all action (names) that are chosen by the user are stored there. So if an error happens they will be included too.
I can rename the variable, or I can try to handle it differently, i.e. failed actions are not included in successful actions, which would you prefer?

I just noticed that I didn't actually change the return type of `ask_for_action_and_apply` from action_name to the action, which confuses me, because I thought I had. 
I will do that now :)

coalib/results/result_actions/IgnoreResultAction.py
sils: type annotation would be an iterable as long as the code doesn't rely on mutability; applied_actions is a continuously changing thing so a mutable arg makes sense as long as it's just only mutable within one fuction and not through many (because that adds complexity)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -16,11 +16,18 @@ class IgnoreResultAction(ResultAction):
...
+                      file_diff_dict,
+                      applied_actions=()):
         """
         For being applicable, the result has to point to a number of files
         that have to exist i.e. have not been previously deleted.
+        Additionally, the action should not have been applied to the current
+        result before.
         """
+        if any(isinstance(x, IgnoreResultAction) for x in applied_actions):
+            return False
```
sils: docs say you should only return True or a string, this is breaking that contract. I.e. there should always be a string message with a reason, also in this case

coalib/output/ConsoleInteraction.py
sils: this issue is still open?

sils:
unack https://github.com/coala/coala/pull/4179/commits/b1336b0ac8f3bd224695a1f285d0bb6d649f1504 getting better I think, some open things still

sils:
and in https://github.com/coala/coala/pull/4179/commits/f73762d7e31d64b4d162d33b27b9c648cf39b2fb still the newline is missing; unack

sils:
/me sees forward to this being merged

mrtes:
Seems there were quite a lot of changes in this region in the last two weeks, resulting in some conflicts which i will have to look into. But i'm on it now :)

mrtes:
With the work done by @Nosferatul in the meantime some of the issues, like the naming of the new variable (was "successful_actions") are already solved.

I now hijacked his "applied_action" dictionary and use the keys of the actions that are stored there for checking if `IgnoreResultAction` has already been applied before.

Some of the stuff discussed in #4010 is obsolete too (like returning the action, instead of the action_name) because the `applied_actions` dict is filled in a different way.

I changed the return value of `IgnoreResultAction.is_applicable` to a string, to fit the docs, as you suggested.

I'm a bit confused about the `ChainPatchAction`, which is an empty stub class, whose functionality seems to happen in a `ConsoleInteraction.py`-function. Maybe @Nosferatul  can comment on this.

shreyans800755:
ack 86c6d2a

jayvdb:
One timeout problem it seems; consistent on Circle & AppVeyor.  Very close... ;-)

mrtes:
Alright, I forgot to test after the rebase, and have therefore not noticed the problems (I had no rebase conflicts so I assumed no changes were made in this area - my mistake).

After looking into it for a bit, I noticed some conflicts with @Nosferatul's implementation of action chaining. 
Since multiple actions can now be applied at once, is #3375 still an issue?
I.e. should I continue to look into this?

.gitignore
```diff
@@ -37,3 +37,4 @@ node_modules
 Goopfile.lock
 venv/
 rultor_secrets.sh
+.vscode
```
jayvdb: Move this out to a separate PR, or create a newcomer issue about it (for every repository...)?

jayvdb:
> is #3375 still an issue?

I dont know this area very well, especially after the changes in GSoC, so you'll need to ping people on gitter about that.

mrtes:
Ok the problem was the following:

According to description the `apply_single` parameter can only be False, or some Action Description.
However, some unittests set this parameter to True.
This value is accepted by the functions that reference it and will be interpreted as a `DoNothingAction`.  With my changes in `acquire_action_and_apply`, this leads to an endless loop, explaining the test timeout.

I now query for this parameter to be true and exit after one cycle, if that is the case.  
Maybe the docs should be tweaked too, to document this behaviour.

In any case, the code that queries user input for actions seems quite complicated and could use a refactoring. What do you think?

@jayvdb I removed the .gitignore commit - Will open a newcomer issue for this.
@sils (or someone else): Do you think this PR is still relevant?

userzimmermann:
@mrtes I must revise the 2nd part of https://github.com/coala/coala/pull/4179#pullrequestreview-75956793 :) They are actually all necessary! But I still need to take that deeper look...

coalib/output/ConsoleInteraction.py
```diff
@@ -165,45 +165,35 @@ def acquire_actions_and_apply(console_printer,
...
+                                                 section,
+                                                 metadata_list,
+                                                 action_dict,
+                                                 failed_actions,
+                                                 result,
+                                                 file_diff_dict,
+                                                 file_dict,
+                                                 applied_actions,
+                                                 apply_single=apply_single)
+        if not chosen_action or apply_single is True:
```
userzimmermann: @mrtes `apply_single` can never be `True`! Only `None`, `False`, or an action string, so remove `is True`

There need to be tests for this `apply_single` stuff...

userzimmermann: @mrtes Sorry! Just read the reason for `apply_single is True` in https://github.com/coala/coala/pull/4179#issuecomment-341969865 :)

mrtes: Yeah, that's what I thought, too.
However, apply_single is set to true in some of the tests for this function (which results in an endless loop when combined with my changes).

Therefore, I thought it would be sensible to catch that possibility.
I can remove it, but will have to change these tests as well.

The fact that this parameter can be a magic string or a boolean or none (in a language without union types) is questionable design and will likely lead to further bugs in the future.

userzimmermann: But it's still a bit strange... Before your changes, the `break` was always done `if apply_single`. Now `if apply_single != 'Do (N)othing'` (and not just `True`) then `chosen_action` will not be `False` and there will be no `break`. This is different behavior now

mrtes: You are right.
Possible fix: Always return false in the `if apply_single` branch of `ask_for_action_and_apply`.
In the current code this return value is not used `if apply_single`, so it is no change in behaviour.
Additionally, it would remove the `àpply_single is True` check.

userzimmermann: @mrtes Look at https://travis-ci.org/coala/coala/builds/300932914

I removed the `is True` and all tests are also passing. There need to be more specific tests for different `apply_single` states... and a better understanding of this former general `if apply_single: ... break`

@Nosferatul HELP! :D

userzimmermann: @mrtes Hadn't read your https://github.com/coala/coala/pull/4179#discussion_r150409810 before writing https://github.com/coala/coala/pull/4179#discussion_r150409980 :) But it seems we are getting to some common level of thinking about that weird stuff ;)

userzimmermann: @mrtes BTW: `chosen_action` is a misleading name for the result of `ask_for_action_and_apply`, since the value determines if there will be more choices afterwards

userzimmermann: @mrtes It's getting worse and worse :P I just realized that by setting `apply_single=True` the whole `test_acquire_actions_and_apply` becomes useless, because the simulated console inputs are never used. By `apply_single=True`, the whole console input is just bypassed. You can set `with simulate_console_inputs('to', 'whatever')` and the test passes

userzimmermann: @mrtes OMG :D `test_acquire_actions_and_apply` is defined a second time for testing `apply_single=True`, but without the same method name, so just replacing the original test

Let's quickly merge https://github.com/coala/coala/pull/4860 to get proper testing for this PR herre

mrtes: Woops, lots of stuff I didn't even think about.
And I also didn't really look into the existing tests besides getting them to run. 
The name `chosen_action` is a relict I forgot to adapt.
I would propose a refactoring of the `apply_single` stuff, after we're done here :)

userzimmermann: @mrtes Created the first issues for that refactoring as you can see below... So let's decide what to do before and what to do after merging this PR

mrtes:
Renamed the "chosen_action" to "continue_interaction". This should capture the intent a bit better.
In `ask_for_action_and_apply` the path with apply_single now always returns False, breaking the loop.
This is consistent with the behaviour (the return value was not queried) before this PR and removes the necessity to check if 'apply_single' is true.

userzimmermann:
@mrtes But I doubt always stopping interaction was the actual intention behind `apply_single`. What `if apply_single != 'Do (N)othing'`? Might result in further interaction...

We need @Nosferatul to tell us the truth ;) Or one of his GSoC mentors @Adrianzatreanu and @yukiisbored

userzimmermann:
@mrtes BTW: You did a great job here! Are you also interested in dealing with #4850 #4851 #4852 or #4854

mrtes:
Well, it effectively worked the same way.
In `acquire_action...` the code checks for `apply_single` and goes into a branch, calling `ask_for_action...` and **ignores** the result, breaking immediately after the call.

The same happens now.
`ask_for_action...` now returns False, if it is in the `apply_single` branch, and True otherwise (with the exception of DoNothingAction.
We query the result and break if it is False.

So -> same behavior, but a more consistent return value for `ask_for_action...`

userzimmermann:
@mrtes I know! In the end it's like before... But is that good? ;)

Let's merge it and create another issue for that!

userzimmermann:
ack https://github.com/coala/coala/pull/4179/commits/9d70b497b31a833cde10c7bbf38ce43d18b4a10a

userzimmermann:
@rultor merge

userzimmermann:
@mrtes rebase :/

mrtes:
@userzimmermann Done

userzimmermann:
reack https://github.com/coala/coala/pull/4179/commits/52d4eef3010585be49268cebedd502701aa403e2

userzimmermann:
@rultor merge

PR #4192, opened 2017-05-07
Makman2/nextgen3-9

Makman2:
Now we come to the hard part: Dependencies. Feel free to ask me questions on how things do work ;) (Dependency support is not described in the [nextgen-design cEP](https://github.com/coala/cEPs/blob/master/cEP-0002.md))

### DIFF
--- a/coalib/core/Core.py
+++ b/coalib/core/Core.py
@@ -63,32 +63,62 @@ def group(iterable, key=lambda x: x):
 
 
 def cleanup_bear(bear,
+                 result_callback,
+                 dependency_tracker,
                  running_tasks,
-                 event_loop):
+                 event_loop,
+                 executor):
     """
     Cleans up state of an ongoing run for a bear.
 
-    - If the given bear has no running tasks left, it removes the bear from
-      the ``running_tasks`` dict.
+    - If the given bear has no running tasks left:
+      - Resolves its dependencies.
+      - Schedules dependant bears.
+      - Removes the bear from the ``running_tasks`` dict.
     - Checks whether there are any remaining tasks, and quits the event loop
       accordingly if none are left.
 
     :param bear:
         The bear to clean up state for.
+    :param result_callback:
+        The result-callback handling results from bears.
+    :param dependency_tracker:
+        The dependency-tracker holding all relations of bears.
     :param running_tasks:
         The dict of running-tasks.
     :param event_loop:
         The event-loop tasks are scheduled on.
+    :param executor:
+        The executor tasks are executed on.
     """
     if not running_tasks[bear]:
+        resolved_bears = dependency_tracker.resolve(bear)
+
+        if resolved_bears:
+            schedule_bears(resolved_bears, result_callback,
+                           dependency_tracker, event_loop, running_tasks,
+                           executor)
+
         del running_tasks[bear]
 
     if not running_tasks:
+        # Check the DependencyTracker additionally for remaining
+        # dependencies.
+        resolved = dependency_tracker.are_dependencies_resolved
+        if not resolved:  # pragma: no cover
+            logging.warning(
+                'Core finished with run, but it seems some dependencies '
+                'were unresolved: {}. Ignoring them, but this is a bug, '
+                'please report it to the developers.'.format(', '.join(
+                    repr(dependant) + ' depends on ' + repr(dependency)
+                    for dependency, dependant in dependency_tracker)))
+
         event_loop.stop()
 
 
 def schedule_bears(bears,
                    result_callback,
+                   dependency_tracker,
                    event_loop,
                    running_tasks,
                    executor):
@@ -100,6 +130,8 @@ def schedule_bears(bears,
         A list of bear instances to be scheduled onto the process pool.
     :param result_callback:
         A callback function which is called when results are available.
+    :param dependency_tracker:
+        The object that keeps track of dependencies.
     :param event_loop:
         The asyncio event loop to schedule bear tasks on.
     :param running_tasks:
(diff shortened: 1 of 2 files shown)


coalib/core/Core.py
```diff
@@ -63,32 +63,62 @@ def group(iterable, key=lambda x: x):
...
     if not running_tasks:
+        # Check the DependencyTracker additionally for remaining
+        # dependencies.
+        resolved = dependency_tracker.are_dependencies_resolved
+        if not resolved:  # pragma: no cover
+            logging.warning(
+                'Core finished with run, but it seems some dependencies '
+                'were unresolved: {}. Ignoring them, but this is a bug, '
+                'please report it to the developers.'.format(', '.join(
+                    repr(dependant) + ' dependent on ' + repr(dependency)
```
Makman2: dependent on -> depends on

coalib/core/Core.py
```diff
@@ -183,6 +186,91 @@ def finish_task(bear,
...
+    # Pre-collect bears in a set as we use them more than once. Especially
+    # remove duplicate instances.
+    bears = set(bears)
+
+    dependency_tracker = DependencyTracker()
+
+    # For a consumer-based system, we have a situation which can be visualized
+    # with a graph:
+    #
+    # (section1, file_dict1) (section1, file_dict2) (section2, file_dict2)
```
Udayan12167: When can a bear have the same section but different filedicts?

coalib/core/Core.py
```diff
@@ -183,6 +186,91 @@ def finish_task(bear,
...
+    This function acquires, processes and registers bear dependencies
+    accordingly using a consumer-based system, where each dependency bear has
+    only a single instance per section and file-dictionary.
+
+    The bears set returned accounts for bears that have dependencies and
+    excludes them accordingly. Dependency bears that have themselves no further
+    dependencies are included so the dependency chain can be processed
+    correctly.
+
+    :param bears:
```
Udayan12167: Are these bear instances or bear_type?

coalib/core/Core.py
```diff
@@ -183,6 +186,91 @@ def finish_task(bear,
...
+    # (section1, file_dict1) (section1, file_dict2) (section2, file_dict2)
+    #       |       |                  |                      |
+    #       V       V                  V                      V
+    #     bear1   bear2              bear3                  bear4
+    #       |       |                  |                      |
+    #       V       V                  |                      |
+    #  BearType1  BearType2            -----------------------|
+    #       |       |                                         |
+    #       |       |                                         V
+    #       ---------------------------------------------> BearType3
```
Udayan12167: BearType 1 ---> BearType2 mean 1 dependent on 2 or the opposite.

coalib/core/Core.py
Makman2: As sections and filedicts are feeded in separately this case is theoretically possible (if someone invokes the core directly).
But actually we could merge filedicts that refer to the same section. What do you think @Udayan12167 ?

coalib/core/Core.py
Makman2: 1 dependent on 2 :)

Makman2: ah I understand the confusion because of the arrow directions... you think making them reverse is better?

coalib/core/Core.py
Makman2: bear instances. Shall I update the docstring to
```
The set of instantiated bears to run that serve as an entry-point.
```
?

coalib/core/Core.py
Makman2: Ah but merging makes problems if a file with the same name has different content....
```python
filedictA = {'afile.txt': ['line1\n']}
filedictB = {'afile.txt': ['this line is definitely different\n']}
```

coalib/core/Core.py
Udayan12167: Yes.

coalib/core/Core.py
Udayan12167: Just mentioning what the arrow direction means above the diagram should suffice.

coalib/core/Core.py
Udayan12167: Yeah this is good. I wasn't thinking of independent core execution. 👍

coalib/core/Core.py
```diff
@@ -183,6 +186,93 @@ def finish_task(bear,
...
+    #
+    # We need to traverse this graph and instantiate dependency bears
+    # accordingly, one per section.
+
+    # Group bears by sections and file-dictionaries. These will serve as
+    # entry-points for the dependency-instantiation-graph.
+    grouping = group(bears, key=lambda bear: (bear.section, bear.file_dict))
+    for (section, file_dict), bears_per_section in grouping:
```
Udayan12167: Can we separate this out in a separate method and make it more generic. Like here instead of looking for a 2-tuple we just do `for group, items_per_group in grouping`.

coalib/core/Core.py
```diff
@@ -183,6 +186,93 @@ def finish_task(bear,
...
+        # accordingly.
+        type_to_instance_map = {}
+        for bear in bears_per_section:
+            type_to_instance_map[bear] = bear
+            type_to_instance_map[type(bear)] = bear
+
+        def instantiate_and_track(prev_bear_type, next_bear_type):
+            if next_bear_type not in type_to_instance_map:
+                type_to_instance_map[next_bear_type] = (
+                    next_bear_type(section, file_dict))
```
Udayan12167: And then to instantiate here we can do `*group` as the arguments to instantiate.

coalib/core/Core.py
Udayan12167: All of this would then fit in a more generic dependency API.

Udayan12167:
ack https://github.com/coala/coala/pull/4192/commits/bdaa67f95f1b1d4033c4150803bf5fc8c3ad9186

coalib/core/Core.py
Makman2: What do you want to move into a separate method? Could you give me an example?
\+
Unpacking those values here and doing `Bear(arg1, arg2)` is actually faster than unpacking them on each call of `Bear(*args)` below (about 30% faster).

Udayan12167: Oh wow did not know that unpacking took so much extra time.

Makman2: This is just compared directly (always unpacking vs unpacking once before), not inside this function ;)

Udayan12167:
ack https://github.com/coala/coala/pull/4192/commits/c01c1ecb493a63293b781cf73d2b3a6e99341098

Udayan12167:
@rultor merge

jayvdb:
I'm a bit worried this is causing https://github.com/coala/docker-coala-base/issues/224

Issue #4191, opened 2017-05-07
Merge `coalib.collecting.Dependencies.CircularDependencyError` and `coalib.core.CircularDependencyError`


Makman2:
Those error classes are very similar and do serve the same purpose.

See https://github.com/thisch/coala/commit/e492cfd7f9e1dce221ce753ac8a919e0e9e2fb39 for a possible implementation.

Makman2:
@thisch assigned you for now as you have already an implementation ;) feel free to continue or give this issue away ;)

jayvdb:
I suspect the original intention was that `coalib.core` was not supposed to include any 'bears' type code, so it stays generic, so a `for_bears` would need to be in `coalib.collecting`.
Maybe someone just forgot to make `collecting.CircularDependencyError` a subclass of `core.CircularDependencyError`.

Issue #4180, opened 2017-05-07
can't build docs due to multiple targets for cross-reference


thisch:
I can't successfully build the docs using `python setup.py docs` with Sphinx 1.5.5 due to the following error

```
Warning, treated as error:
/home/thomas/gitrepos/coala/coalib/core/DependencyTracker.py:docstring of coalib.core.DependencyTracker.DependencyTracker.check_circular_dependencies:None: WARNING: more than one target found for cross-reference 'CircularDependencyError': coalib.collecting.Dependencies.CircularDependencyError, coalib.core.CircularDependencyError.CircularDependencyError
```

I had a look at the implementation of the two CircularDependenyError classes and they are quite similar, can't they be merged?

thisch:
This might be related to the sphinx-1.5.5, since the docs-requirements.txt 1.4.9 is hardcoded.

Makman2:
Did you try to compile the docs with sphinx 1.4.9?

Makman2:
Indeed those classes seem to be very similar. We should try to merge them :+1:

thisch:
Should I create a PR out of my commit mentioned in https://github.com/coala/coala/pull/4183#issuecomment-299701304, so that we keep sight of my 'merge'-commit?

Makman2:
Let's make a new issue ;)

Makman2:
--> https://github.com/coala/coala/issues/4191. Mentioned your commit there :+1:

PR #4199, opened 2017-05-08
Merge CicularDependencyError classes

thisch:
Closes https://github.com/coala/coala/issues/4191

### DIFF
--- a/coalib/collecting/Dependencies.py
+++ b/coalib/collecting/Dependencies.py
@@ -1,14 +1,4 @@
-class CircularDependencyError(Exception):
-
-    @classmethod
-    def for_bears(cls, bears):
-        """
-        Creates the CircularDependencyError with a helpful message about the
-        dependency.
-        """
-        bear_names = [bear.name for bear in bears]
-
-        return cls('Circular dependency detected: ' + ' -> '.join(bear_names))
+from coalib.core.CircularDependencyError import CircularDependencyError
 
 
 def _resolve(bears, resolved_bears, seen):
@@ -23,7 +13,7 @@ def _resolve(bears, resolved_bears, seen):
 
         if bear in seen:
             seen.append(bear)
-            raise CircularDependencyError.for_bears(seen)
+            raise CircularDependencyError(s.name for s in seen)
 
         seen.append(bear)
         resolved_bears = _resolve(missing, resolved_bears, seen)
--- a/coalib/core/CircularDependencyError.py
+++ b/coalib/core/CircularDependencyError.py
@@ -1,19 +1,19 @@
-class CircularDependencyError(Exception):
+class CircularDependencyError(RuntimeError):
     """
     An error identifying a circular dependency.
     """
 
-    def __init__(self, node=None):
+    def __init__(self, names=None):
         """
         Creates the CircularDependencyError with a helpful message about the
         dependency.
 
-        :param node:
-            The node that was encountered twice and closes the dependency
-            circle.
+        :param names:
+            The names of the nodes that form a dependency circle.
         """
-        message = 'Circular dependency detected.'
-        if node is not None:
-            message += ' {0!r} -> ... -> {0!r}'.format(node)
-
-        Exception.__init__(self, message)
+        if names:
+            msg = 'Circular dependency detected: {names}'.format(
+                names=' -> '.join(names))
+        else:
+            msg = 'Circular dependency detected.'
+        super().__init__(msg)
--- a/coalib/core/DependencyTracker.py
+++ b/coalib/core/DependencyTracker.py
@@ -293,7 +293,7 @@ class DependencyTracker:
          ...
         coalib.core.CircularDependencyError.CircularDependencyError: ...
 
-        :raises coalib.core.CircularDependencyError.CircularDependencyError:
+        :raises CircularDependencyError:
             Raised on circular dependency conflicts.
         """
         traverse_graph(
--- a/coalib/core/Graphs.py
+++ b/coalib/core/Graphs.py
@@ -34,7 +34,7 @@ def traverse_graph(start_nodes, get_successive_nodes,
         A callable that is run on each edge during traversing. Takes in two
         parameters, the previous- and next-node which form an edge. The default
         is an empty function.
-    :raises coalib.core.CircularDependencyError.CircularDependencyError:
+    :raises CircularDependencyError:
         Raised when the graph is cyclic.
     """
     path = set()
@@ -49,8 +49,8 @@ def traverse_graph(start_nodes, get_successive_nodes,
                 run_on_edge(node, subnode)
 
                 if subnode in path:
-                    raise CircularDependencyError(subnode)
-
+                    raise CircularDependencyError([repr(subnode), '...',
+                                                   repr(subnode)])
                 visit(subnode)
 
             path.remove(node)
--- a/tests/core/CircularDependencyErrorTest.py
+++ b/tests/core/CircularDependencyErrorTest.py
@@ -7,14 +7,15 @@ class CircularDependencyErrorTest(unittest.TestCase):
 
     def test_default_message(self):
         with self.assertRaises(CircularDependencyError) as cm:
+            # test the default case (names is None)
             raise CircularDependencyError
 
         self.assertEqual(str(cm.exception), 'Circular dependency detected.')
 
-    def test_message_with_causing_node(self):
+    def test_message_with_dependency_circle(self):
+
         with self.assertRaises(CircularDependencyError) as cm:
-            # "3" is the causing node.
-            raise CircularDependencyError(3)
+            raise CircularDependencyError(['A', 'B', 'C'])
 
         self.assertEqual(str(cm.exception),
-                         'Circular dependency detected. 3 -> ... -> 3')
+                         'Circular dependency detected: A -> B -> C')
--- a/tests/collecting/DependenciesTest.py
+++ b/tests/collecting/DependenciesTest.py
@@ -2,6 +2,7 @@ import unittest
 
 from coalib.bears.Bear import Bear
 from coalib.collecting import Dependencies
+from coalib.core.CircularDependencyError import CircularDependencyError
 
 
 class ResolvableBear1(Bear):
@@ -51,6 +52,6 @@ class DependenciesTest(unittest.TestCase):
 
     def test_unresolvable_deps(self):
         self.assertRaises(
-            Dependencies.CircularDependencyError,
+            CircularDependencyError,
             Dependencies.resolve,
             [UnresolvableBear1])


tests/core/CircularDependencyErrorTest.py
```diff
@@ -5,16 +5,29 @@
...
         with self.assertRaises(CircularDependencyError) as cm:
             # "3" is the causing node.
-            raise CircularDependencyError(3)
+            raise CircularDependencyError.for_node(3)
+
+        self.assertEqual(str(cm.exception),
+                         'Circular dependency detected: 3 -> ... -> 3')
+
+    def test_message_with_causing_bears(self):
+        class FakeBear(object):
```
thisch: applied

coalib/core/CircularDependencyError.py
```diff
@@ -3,7 +3,10 @@ class CircularDependencyError(Exception):
     An error identifying a circular dependency.
     """
```
userzimmermann: I think it should be better based on `RuntimeError`

coalib/core/CircularDependencyError.py
```diff
@@ -3,7 +3,10 @@ class CircularDependencyError(Exception):
...
 
-    def __init__(self, node=None):
+    prefix = 'Circular dependency detected'
```
userzimmermann: Is `prefix` intended to be changeable? If not, there is no need to store it in an attribute.

And in any case, I would prepend that prefix in `.__str__` instead of in every `.for_...` method

coalib/core/DependencyTracker.py
```diff
@@ -293,7 +293,7 @@ def check_circular_dependencies(self):
          ...
         coalib.core.CircularDependencyError.CircularDependencyError: ...
 
-        :raises coalib.core.CircularDependencyError.CircularDependencyError:
+        :raises CircularDependencyError:
```
userzimmermann: Better keep it fully qualified

coalib/core/Graphs.py
```diff
@@ -34,7 +34,7 @@ def traverse_graph(start_nodes, get_successive_nodes,
         A callable that is run on each edge during traversing. Takes in two
         parameters, the previous- and next-node which form an edge. The default
         is an empty function.
-    :raises coalib.core.CircularDependencyError.CircularDependencyError:
+    :raises CircularDependencyError:
```
userzimmermann: Again, better keep it fully qualified

sils:
CC @Makman2 does this destroy your stuff?

@thisch did you talk with @Makman2 about this?

sils:
I guess it's a good thing though starting to migrate to some of the new classes ahead of time if possible :+1:

coalib/core/CircularDependencyError.py
thisch: No prefix is not intended to be changeable.

Do you mean overwriting RuntimeError.__str__ ?

coalib/core/DependencyTracker.py
thisch: I made it fully qualified, due to an issue with sphinx 1.5.5 (see #4180) a few days ago. With this PR we no longer need the 'workaround' from #4180. Do you really want me to make it the name fully qualified again?

sils:
do not merge until @Makman2 looks at this, it looks like you're migrating old functionality that we won't need in the future to the new class which was probably written to get rid of that old stuff

coalib/core/CircularDependencyError.py
```diff
@@ -12,8 +13,24 @@ def __init__(self, node=None):
...
+        Creates the CircularDependencyError with a helpful message about the
+        dependency.
+
+        :param bears:
+            sequence of bears, for which a circular dependency was detected.
+        """
+        bear_names = [bear.name for bear in bears]
+
+        return cls('Circular dependency detected: {names}'.format(
+            names=' -> '.join(bear_names)))
```
Makman2: What about having only a constructor like before, but use this code instead? And the core which currently wants to print `node1 -> ... -> node2` passes `bear1object, '...', bear2object`.
-->
```python
def __init__(nodes):
        prefix = 'Circular dependency detected'
        if nodes is not None:
            message = '{}: {}'.format(prefix, ' -> '.join(str(node) for node in nodes)
        else:
            message = prefix + '.'

        Exception.__init__(self, message)
```
The collection stuff just uses the constructor normally by feeding in the bear names `(bear.name for bear in bears)` and the core would feed in this: `[repr(node), '...', repr(node)]`

coalib/core/DependencyTracker.py
Makman2: I don't think full qualification is necessary. Let's discuss full qualification for references in another issue^^

coalib/core/CircularDependencyError.py
userzimmermann: @thisch Exactly :)

coalib/core/CircularDependencyError.py
userzimmermann: @Makman2 Then you unnecessarily store the whole message including prefix although the message is not necessarily fetched from the exception object depending on use case. That's why I would prefer the solution from https://github.com/coala/coala/pull/4199#discussion_r115360889

cc @thisch

coalib/core/DependencyTracker.py
userzimmermann: @Makman2 Do we use sphinx autodoc for generating API docs? Then I would first take a look if the link to the exception class doc is correctly generated if we use the short form here...

cc @thisch

Makman2: Yeah I think it's autodoc :+1:

coalib/core/CircularDependencyError.py
Makman2: This looks like some performance quirk which doesn't hit. If someone uses an exception, he's likely going to raise it immediately. And this is just saving a little string, imo this doesn't impact performance at all.

Sadly I can't see the comment you're linking to, though if we wanna cache it, we should override `__str__` with `@lru_cache(1)`.

Makman2: Sounds like a general design strategy for exceptions to override `__str__` when using a custom message to save performance. Maybe this should go into a new issue :+1:

Makman2: Btw: if we construct the message later, we need to store a list of the nodes anyway, doesn't really make a difference^^

userzimmermann: @Makman2 I just mean when you catch that exception later, you are not necessarily interested in the message... But you are right! There are advantages and disadvantages... Maybe we should do some performance comparison ;P

thisch: @zimmerman probably meant this comment https://github.com/coala/coala/pull/4199#discussion_r115360889 

@Makman2, I don't see any benefit (except beeing a bit more compact) in merging the two `for_*` clsmethods into `__init__`.  In https://github.com/coala/coala/pull/4199#discussion_r115496791 you move the logic to the client of the `CircularDependencyError` class, whereas in my code, the logic is part of the `CircularDependencyError` class.

I don't get why we need/want to optimize this code, since the exception class is only instantiated ecause (and immediately raised) in two places (Graphs.py, Dependencies.py) in the codebase (excluding tests).

Makman2: Imo it's a matter of concept. The `CircularDependencyError` is responsible for signalling that a such an error happened. It provides a helpful message which shall show you which objects do chain up. How those objects shall be displayed is now moved to the user of the exception in my approach. Your approach would require to code new classmethods for each usecase that may arise in the future.
\+
Less code duplication because the only difference is the different style of printing the objects in the message between ` -> `.

Zimmerman:
Guess you mean @userzimmermann and not @Zimmerman

thisch:
yes, I meant him

coalib/core/CircularDependencyError.py
thisch: Yes, it's a matter of taste. Regarding the code duplication: This is not really an issue IMO, since in the design of the CircularDependencyError class in this PR, the two class methods could both call a common implementation. For the usecases that may arise in the future, this class probably has to be redesigned anyway.

coalib/core/DependencyTracker.py
thisch: @userzimmermann, yes the linkg to the exception class is correctly generated if we use the short form

coalib/core/CircularDependencyError.py
Makman2: > [...] the two class methods could both call a common implementation.

Then what is the need for having special implementations only used at a single place? More flexible functions + less code to maintain :+1:

> For the usecases that may arise in the future, this class probably has to be redesigned anyway.

Why?

coalib/core/CircularDependencyError.py
```diff
@@ -1,9 +1,19 @@
-class CircularDependencyError(Exception):
+class CircularDependencyError(RuntimeError):
     """
     An error identifying a circular dependency.
     """
 
-    def __init__(self, node=None):
+    @classmethod
+    def _init(cls, names):
```
Makman2: why not `__init__` with `names=None` as default?

coalib/core/CircularDependencyError.py
thisch: > Then what is the need for having special implementations only used at a single place? More flexible functions + less code to maintain ￼

That's a good point. Since it's only used at a single place, I've changed it as suggested by you.

coalib/collecting/Dependencies.py
```diff
@@ -23,7 +13,7 @@ def _resolve(bears, resolved_bears, seen):
 
         if bear in seen:
             seen.append(bear)
-            raise CircularDependencyError.for_bears(seen)
+            raise CircularDependencyError([s.name for s in seen])
```
Makman2: I think you don't have to collect the generator in a list before ;) so just
```python
raise CircularDependencyError(s.name for s in seen)
```
should suffice ;)

coalib/core/CircularDependencyError.py
```diff
@@ -1,19 +1,11 @@
...
-            The node that was encountered twice and closes the dependency
-            circle.
-        """
-        message = 'Circular dependency detected.'
-        if node is not None:
-            message += ' {0!r} -> ... -> {0!r}'.format(node)
-
-        Exception.__init__(self, message)
+    def __init__(self, names=None):
+        msg = 'Circular dependency detected.'
```
Makman2: wanna pack this into an `else` clause below? Then you save the performance if `names != None` :)

tests/core/CircularDependencyErrorTest.py
```diff
@@ -5,16 +5,16 @@
...
         with self.assertRaises(CircularDependencyError) as cm:
-            raise CircularDependencyError
+            # test the default case (node is None)
+            raise CircularDependencyError()
 
-        self.assertEqual(str(cm.exception), 'Circular dependency detected.')
+        self.assertEqual(str(cm.exception),
+                         'Circular dependency detected.')
 
-    def test_message_with_causing_node(self):
```
Makman2: why not keeping two separate test cases?

tests/core/CircularDependencyErrorTest.py
```diff
@@ -5,16 +5,16 @@
 
 class CircularDependencyErrorTest(unittest.TestCase):
 
-    def test_default_message(self):
+    def test_message(self):
         with self.assertRaises(CircularDependencyError) as cm:
-            raise CircularDependencyError
+            # test the default case (node is None)
+            raise CircularDependencyError()
```
Makman2: You can keep `raise CircularDependencyError`, this automatically invokes `__init__` without / with default arguments.

coalib/core/CircularDependencyError.py
```diff
@@ -1,19 +1,11 @@
...
-        :param node:
-            The node that was encountered twice and closes the dependency
-            circle.
-        """
-        message = 'Circular dependency detected.'
-        if node is not None:
-            message += ' {0!r} -> ... -> {0!r}'.format(node)
-
-        Exception.__init__(self, message)
+    def __init__(self, names=None):
```
Makman2: Please provide a docstring ;) You can use as much as possible from the old one :+1:

coalib/core/CircularDependencyError.py
```diff
@@ -1,19 +1,11 @@
-class CircularDependencyError(Exception):
+class CircularDependencyError(RuntimeError):
```
Makman2: maybe separate commit? up to you :) :+1:

coalib/core/CircularDependencyError.py
thisch: done

Makman2:
ack 5913bb8

Makman2:
@rultor merge

Issue #4190, opened 2017-05-07
Rewrite coalib.collecting.Importers._import_module


userzimmermann:
Realized in https://github.com/coala/coala/pull/4169#pullrequestreview-36672803

`_import_module` should do its job following https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

That avoids manipulating `sys.modules` and implicitly solves that Windows case insensitivity issue for which the current function implementation has an explicit workaround

@jayvdb

newbazz:
Can someone explain this issue, I would like to work on it if no ones working,

Makman2:
It's taken by @userzimmermann . Also I highly doubt it's `difficulty/low`...

userzimmermann:
@Makman2 @newbazz https://github.com/coala/coala/pull/4210 is done and rebased. Please review again, ack, merge :)

PR #4210, opened 2017-05-09
Rewrite Importers._import_module using importlib

userzimmermann:
Heavily simplifying the `_import_module` function and avoiding manipulation of `sys.modules` and special **windows** handling

Closes https://github.com/coala/coala/issues/4190

Unblocks https://github.com/coala/coala/pull/4169

@jayvdb @Makman2 @sils


userzimmermann:
Hmm... Something doesn't work as expected here 😕

jayvdb:
Half of the ImportersTest.py are failing , so I think that would be the first place to look.

coalib/collecting/Importers.py
```diff
@@ -1,34 +1,30 @@
...
     if not os.path.exists(file_path):
-        raise ImportError
+        raise ImportError("can's import non-existing file %s"
+                          % repr(file_path))
 
     module_name = os.path.splitext(os.path.basename(file_path))[0]
-    module_dir = os.path.dirname(file_path)
-
-    if module_dir not in sys.path:
-        sys.path.insert(0, module_dir)
```
jayvdb: You still need to do this, otherwise `spec.loader.exec_module` can't load the module.

coalib/collecting/Importers.py
```diff
@@ -1,34 +1,30 @@
...
-
-    return __import__(module_name)
+    spec = spec_from_file_location(module_name, file_path)
+    if not spec:
+        raise ImportError("%s doesn't seem to be a python module"
+                          % repr(file_path))
+
+    module = module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
```
jayvdb: also need `sys.modules[module_name] = module` here for reasons that will become clear after https://github.com/coala/coala/pull/4612 .

Full fix at https://github.com/jayvdb/coala/commit/c72508fa3

userzimmermann: @jayvdb Thanks for those hints! I see now that the `Importers._is_defined_in` check is the actual problem. We need to change this verification mechanism. Doing `sys.path.insert` and adding every imported module to `sys.modules` is a bad strategy. When all bears are discovered, we recursively load every `*.py` file in the search path, not only the actual bear modules. All those get added to `sys.modules`, even if we don't need them, even the `__init__.py` modules. This is clutter. In the end there is `sys.modules['__init__']`, referencing the last one loaded during the process:

```python
>>> sys.modules['__init__']
<module '__init__' from 'c:\\users\\zimmermann\\projects\\coala\\bears\\bears\\__init__.py'>
```

An having the modules' directories in `sys.path` also makes all non-bear `*.py` files directly `import`able even if not in `sys.modules`, also `import __init__` then always works

coalib/collecting/Importers.py
userzimmermann: @jayvdb The module is loaded without the `sys.path` entry. That's the actual advantage of this module loading mechanism. The problem is the `Importers._is_defined_in` check done with all module members. I further explained everything below in https://github.com/coala/coala/pull/4210#discussion_r131373415

coalib/collecting/Importers.py
jayvdb: @userzimmermann , that is different problems ; problems that already exist in the current implementation.  Please do not keep blocking https://github.com/coala/coala/pull/4169 because of some new improvement that ends up stalled for months!  Go create bugs for these other problems.

Bugs are created almost every time someone uses `pragma: no cover`.

userzimmermann: @jayvdb I wonder if we will ever get rid of those existing implementation problems this way ;) When there is no pressure, they will never get touched... They will only become more and more... But you are actually right! What can we do? We don't have the time... I will resemble the old behavior :)

jayvdb:
Please test locally, or at least mark you commit with WIP if you havent tested it locally.
And as previously indicated, I suggest that you first get https://github.com/coala/coala/pull/4612 merged, as it will help you find the answer.

userzimmermann:
@jayvdb Thanks :) Did more local testing now! But I already know the answer... Why those workarounds are necessary to make our current implementation work

coalib/collecting/Importers.py
```diff
@@ -7,28 +9,35 @@
...
+    if not spec:
+        raise ImportError("{!r} doesn't seem to be a python module"
+                          .format(file_path))
+
+    # load module, and add it to sys.modules for making inspect.getfile work
+    # with module members
+    module = sys.modules[module_name] = module_from_spec(spec)
+    spec.loader.exec_module(module)
+    # and extend sys.path for making pickling work
+    sys.path.insert(0, os.path.dirname(os.path.realpath(file_path)))
```
jayvdb: Why are you doing this here, instead of keeping the existing code at https://github.com/coala/coala/pull/4210/files#r131307918

userzimmermann: @jayvdb Good point! :D Restarting from the original equivalent... But making the containment check a little more precise ;)

userzimmermann:
@jayvdb This PR doesn't fix any bugs. It only improves the code of one function, including simplification of a Windows specific workaround inside. That's all. I don't see any reasonable split here

newbazz:
sorry to be late ;)

userzimmermann:
@Makman2 Followed your suggestion for moving the `import` line

userzimmermann:
@Makman2 DONE!!! ;)

coalib/collecting/Importers.py
```diff
@@ -7,28 +9,40 @@
...
+# in Python > 3.4 this can be imported from importlib.utils
+from coalib.misc.Compatibility import module_from_spec
+
 
 def _import_module(file_path):
+    file_path = os.path.realpath(file_path)
     if not os.path.exists(file_path):
-        raise ImportError
+        raise ImportError("can't import non-existing file {!r}"
+                          .format(file_path))
```
Makman2: maybe put this change into a new commit :+1:

coalib/collecting/Importers.py
```diff
@@ -7,28 +9,40 @@
...
 
+    # little HACK to get correct filename case on Windows (if there is at
+    # least on [character] matching group, glob will return correct case)
+    file_path = glob(file_path[:-1] + '[' + file_path[-1] + ']')[0]
```
Makman2: optional: maybe use `str.format`?
```
file_path = glob('{}[{}]'.format(file_path[:-1], file_path[-1]))[0]
```

coalib/collecting/Importers.py
```diff
@@ -7,28 +9,40 @@
...
-
-    if module_dir not in sys.path:
+    spec = spec_from_file_location(module_name, file_path)
+    if not spec:
+        raise ImportError("{!r} doesn't seem to be a python module"
+                          .format(file_path))
+
+    # don't replace any already imported modules
+    if module_name in sys.modules:
+        return sys.modules[module_name]
```
Makman2: shouldn't we do this before getting the spec for performance?

Makman2:
Are there alternatives to this huge chunk of code? https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path

userzimmermann:
@Makman2 I am actually doing what is proposed in https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path ... And my original idea was to avoid unnecessary manipulation of `sys.path` and `sys.modules` when loading bears from arbitrary locations... But I had to put those manipulations back in for keeping the whole bear loading and running mechanisms around working... To keep `inspect.getfile` working for reading the module contents... To keep pickling working for `multiprocessing`... So actually this makes only sense with changing all that other stuff around, too...

At least we could use the simplification of the correct Windows case determination from this PR

userzimmermann:
@Makman2 I think the best approach for improvement in this whole complex would be some `sys.meta_path` plugin for bear loading...

coalib/collecting/Importers.py
userzimmermann: @Makman2 Then you'll never know if something is wrong with the file itself...

Makman2:
> I think the best approach for real improvement in this whole complex would be some sys.meta_path plugin for bear loading...

So you mean our own bear-path variable that we maintain for bear imports?

But okay just wanted to confirm that the bear method is doing something different than normal importing :+1:

coalib/collecting/Importers.py
Makman2: Hm but if the module is already cached, why checking the file again? Python does it the same way, you have to use `importlib.reload` if you want to refresh a module, otherwise it's always "loaded" from cache again. Even if the file was removed, Python still picks it from the cache.

tests/collecting/ImportersTest.py
```diff
@@ -2,7 +2,35 @@
...
+        current_dir = os.path.dirname(os.path.relpath(__file__, '.'))
+        self.non_existing_path = os.path.join(current_dir,
+                                              'importers_test_dir',
+                                              'non_existing.py')
+        self.non_python_path = os.path.join(current_dir,
+                                            'importers_test_dir',
+                                            'test.nonpy')
+
+    def test_non_existing(self):
+        with self.assertRaisesRegex(ImportError, (
```
Makman2: I think the brace at end of line is not necessary^^

(same below)

userzimmermann: @Makman2 It saves us one line, because otherwise I would have to put `ImportError` on the next line for PEP8 consistency ;)

userzimmermann: @Makman2 Everything I do has some very special reason ;P

Makman2: ... then just put it onto the next line 😅 don't do tricks with codestyles, just ensure it's nice to read :3

jayvdb:
Guys, please review https://github.com/coala/coala/pull/4612 ; I will get extremely annoyed if this goes in before the tests are merged.

userzimmermann:
@jayvdb Done :)

jayvdb:
Conflicts need manual resolution.

PR #4217, opened 2017-05-11
Diff.py: Remove has_changes and fix __bool__

### LINKED ISSUE #4178: Diff: Remove has_changes field
The better alternative to `has_changes` is using `__bool__`.
Task is to remove `has_changes` again by reverting commit https://github.com/coala/coala/commit/671e6a2e9fe1f5ab8bab5370c1848f00b3472030.

Sorry @lshobith, but your work is not in vain, now we know also thanks to you how to do it better ;)

See also https://github.com/coala/coala/issues/3790

mrtes:
Removes the `self.has_changes` property, since its functionality can be
accessed from the bool conversion.
Fixes inconstency of `__bool__` that results from looking at
`self._changes`:
removing one line, then adding the same content again resulted in
`bool(diff) == True`, instead of False.

`__bool__` now uses the mechanism that was employed by `has_changes`, to
fix this bug.

Closes #4178

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -26,7 +26,7 @@ class Diff:
         :param delete:    True if file is set to be deleted.
         """
         self._changes = {}
-        self._file = file_list
+        self._file = list(file_list)
         self.rename = rename
         self.delete = delete
 
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -278,14 +278,6 @@ class Diff:
         return result
 
     @property
-    def has_changes(self):
-        """
-        True if the modified file is different than the original file,
-        else False.
-        """
-        return self.modified != self._file
-
-    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
@@ -440,7 +432,7 @@ class Diff:
         """
         return (self.rename is not False or
                 self.delete is True or
-                len(self._changes) > 0)
+                self.modified != self._file)
 
     def delete_line(self, line_nr):
         """
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -148,12 +148,21 @@ class DiffTest(unittest.TestCase):
         del result_file[2]
         self.assertEqual(self.uut.modified, result_file)
 
-    def test_has_changes(self):
-        self.assertFalse(self.uut.has_changes)
+    def test_bool(self):
+        self.assertFalse(self.uut)
         self.uut.add_line(4, '4')
-        self.assertTrue(self.uut.has_changes)
+        self.assertTrue(self.uut)
         self.uut.delete_line(4)
-        self.assertFalse(self.uut.has_changes)
+        self.assertFalse(self.uut)
+
+        # test if it works with tuples.
+        uutuple = Diff(('1', '2', '3', '4'))
+
+        self.assertFalse(uutuple)
+        uutuple.add_line(4, '4')
+        self.assertTrue(uutuple)
+        uutuple.delete_line(4)
+        self.assertFalse(uutuple)
 
     def test_addition(self):
         self.assertRaises(TypeError, self.uut.__add__, 5)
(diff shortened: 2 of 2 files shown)


Makman2:
See gitmate comment, looks good otherwise :+1:

mrtes:
Does someone have an idea why both TravisCI and CircleCI are failing for this commit?

SanketDG:
this creates some sort of regression in `Diff.from_string_arrays()`

mrtes:
Yeah, I finally managed to get my debugger running.

The problem is, that the comparison `self.modified != self._file` tries to compare a list with a tuple.
The file handed to the bear (both UnusedCodeBear and Pep8Bear) is actually in tuple format for some reason. That breaks split_diff on:
```
if not self: 
    return
``` 
[Diff.py](https://github.com/coala/coala/blob/671e6a2e9fe1f5ab8bab5370c1848f00b3472030/coalib/results/Diff.py#L356)

A possible solution is to convert self._file to a list, but I don't know if that issue (tuple instead of list) could have effect at other parts of the program.

mrtes:
Actually it seems all file contents are read into tuples (probably for immutability?) and passed like that to all matching bears.

However going by documentation and test cases, the Diff.from_string_arrays methods seems to expect a list instead of a tuple. 
Everything works right now, but changes like the current one break the behaviour.

The current issue could be fixed by simply converting `list(self._file)` in  `__bool__`, but the issue could resurface in other ways.
Another possibility would be convert the tuple to a list, already on creation of the diff.
However, this breaks `test_equality` of the Diff, probably because the test depends on mutability of the reference that is passed to the diff ([see here](https://github.com/coala/coala/blob/671e6a2e9fe1f5ab8bab5370c1848f00b3472030/tests/results/DiffTest.py#L558-L565)).
I'm not sure if that mutability is beneficial or even wanted (since in most instances the diff is created from tuples anyway).

What do you think would be the better choice?

Makman2:
Ah because it only checks the delta, not the file versions. So it seems the docs do defy what was actually intended.
Two possibilities:
1. We consider `__bool__` to be broken because of the docs and we fix it.
2. We don't change `__bool__` and change its docs instead and keep `has_changes`.

If we are going with 1.:
> However, this breaks test_equality of the Diff, probably because the test depends on mutability of the reference that is passed to the diff (see here).

I think this is a bad design currently and also for the test to rely on that, and we should change this then^^

mrtes:
Even if we keep the `has_changes` function, the issue with tuple/list comparison will remain.
So we will have to decide how to avoid it.

mrtes:
Not sure I understand your first sentence. Can you explain?

Actually, even if we change it back to `has_changes`, the tuple/list comparison issue will persist.
The tests just don't catch it, because they use lists as the original "file", while in reality its mostly a tuple (could be I missed some cases, where an actual list is used by coala).

So we need to make a decision on how to avoid this.

Makman2:
> Not sure I understand your first sentence. Can you explain?

I mean we test on purpose that the change-list is checked, not the actual `diff.modified` and `diff.original`. For example a `Diff` with the same changes (the changes are called "delta", as this normally denotes some different in mathematics) but different files operated on, is considered the same by the tests.

> So we need to make a decision on how to avoid this.

I'm for changing the interior functionality of `Diff` then to make this more concise. Meaning we maintain a list if it solves the issue instead of tuples :+1:

mrtes:
Alright, I made the change.
I went with option 1: fixing ` __bool__` and split it into two commits.
The first one just converts the incoming file into a list and updates breaking tests (test_equality).

The second commit contains the actual changes to `__bool__`.

What do you think?

Makman2:
looks good :+1:

mrtes:
Rebased for mergeability. 
Can you review @Makman2 ?

Makman2:
ack a9b8d6a a3bf0b6

Makman2:
@rultor merge

Issue #4185, opened 2017-05-07
--no-color is ignored for loglines


Makman2:
Migrated from https://github.com/coala/coala-bears/issues/1714.

From @thisch:

> The string '[WARNING]' is colored even if the --no-color (-N) flag is used.
>
> ```
> > coala -N --bears=VultureBear --files \*\*/\*.py --save
> [WARNING][11:57:36] 'cli' is an internally reserved section name. It may have been generated into your coafile while running coala with `--save`. The settings in that section will inherit implicitly to all sections as defaults just like CLI args do.Please change the name of that section in your coafile to avoid any unexpected behavior.
> Executing section cli...
> ```

Makman2:
From @Makman2:

> Confirmed. Happens with latest version `0.12.0.dev20170507091731`.

thisch:
Sry, I accidentally posted this issue in coala-bears

shreyans800755:
I would like to work this issue.
I tried to replicate the scenario in my local environment. And I get colored output for the name of logging level. I'm not able to figure out, how only logging level is colored and not the whole logging message.

I tried to go through various files like LOG_LEVEL, LogPrinter, LogMessage, ConsoleInteraction, etc. But couldn't find the exact location, where color is set for log level name.

Can someone please guide me ?

Makman2:
> I'm not able to figure out, how only logging level is colored and not the whole logging message.

We don't color the message itself on purpose, just the log level tag.

> I tried to go through various files like LOG_LEVEL, LogPrinter, LogMessage, ConsoleInteraction, etc. But couldn't find the exact location, where color is set for log level name.

Actually I don't know where we set colors, should be somewhere where we initialize logging I believe. See https://docs.python.org/3/library/logging.config.html. I think we use `dictConfig()`.

mrtes:
The console_printer, which is configured with args.no_color, is currently only used in the ConsoleInteraction,py file, which explains why the argument does not do anything regarding logging colors.

For logging, the console_printer is not used at all. It is handed to the log_printer, but that uses the "logging" package internally. Additionally, the "logging" package is used directly in some cases in "ConfigurationGathering".

Both are configured with the dict you mentioned.

I guess there is a reason for separating those two interaction forms, but why is the console_printer passed to the log_printer and not used at all?

Makman2:
Those log- and console-printers are deprecated. We haven't removed them completely yet, that's why you encounter those sometimes.

PR #4224, opened 2017-05-12
no_color: Fix coloring bug in --no-color option

shreyans800755:
Fixes https://github.com/coala/coala/issues/4185

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/output/Logging.py
+++ b/coalib/output/Logging.py
@@ -32,7 +32,7 @@ class CounterHandler(logging.Handler):
         return cls._call_counter[level]
 
 
-def configure_logging():
+def configure_logging(color=True):
     """
     Configures the logging with hard coded dictionary.
     """
@@ -46,7 +46,7 @@ def configure_logging():
         'handlers': {
             'colored': {
                 'class': 'logging.StreamHandler',
-                'formatter': 'colored',
+                'formatter': 'colored' if color else 'plain',
                 'stream': sys.stderr
             },
             'counter': {
@@ -69,6 +69,10 @@ def configure_logging():
                     'INFO': 'blue',
                     'DEBUG': 'green'
                 }
+            },
+            'plain': {
+                'format': '[%(levelname)s][%(asctime)s] %(message)s',
+                'datefmt': '%X',
             }
         }
     })
--- a/tests/coalaTest.py
+++ b/tests/coalaTest.py
@@ -179,3 +179,30 @@ class coalaTest(unittest.TestCase):
             )
 
             assert exitcode == 1
+
+    def test_coala_with_color(self):
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala')
+            errors = filter(bool, stderr.split('\n'))
+            # Every error message must start with characters
+            # used for coloring.
+            for err in errors:
+                self.assertNotRegex(err, '^\[WARNING\]')
+            self.assertEqual(
+                retval, 0, 'coala must return zero when there are no errors')
+
+    def test_coala_without_color(self):
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+
+            retval, stdout, stderr = execute_coala(
+                             coala.main, 'coala', '-N')
+            errors = filter(bool, stderr.split('\n'))
+            # Any error message must not start with characters
+            # used for coloring.
+            for err in errors:
+                self.assertRegex(err, '^\[WARNING\]')
+            self.assertEqual(
+                retval, 0, 'coala must return zero when there are no errors')
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -36,6 +36,7 @@ def main():
             show_bears, show_language_bears_capabilities)
 
         console_printer = ConsolePrinter(print_colored=not args.no_color)
+        configure_logging(not args.no_color)
 
         if args.json:  # needs to be checked in order to display bears in json
             return mode_json(args)


shreyans800755:
I'm not really sure how to check the color of the log being printed in stdout. I've added a test case to get the coverage to 100%.

shreyans800755:
@Makman2 @thisch 
Can someone please review this ?

coalib/output/Logging.py
```diff
@@ -41,37 +41,64 @@ def configure_logging():
...
+                    'class': 'coalib.output.Logging.CounterHandler'
+                }
+            },
+            'root': {
+                'level': 'DEBUG',
+                'handlers': ['plain', 'counter']
+            },
+            'formatters': {
+                'plain': {
+                    '()': 'colorlog.ColoredFormatter',
```
Makman2: is there no plain formatter?

coalib/output/Logging.py
```diff
@@ -41,37 +41,64 @@ def configure_logging():
...
     CounterHandler.reset()
 
-    logging.config.dictConfig({
-        'version': 1,
-        'handlers': {
-            'colored': {
-                'class': 'logging.StreamHandler',
-                'formatter': 'colored',
-                'stream': sys.stderr
+    if not no_color:
```
Makman2: this looks too complicated to build up two completely different dicts just because of this param. Can't you modify only the entries of the current dict that would be affected by `no-color = True`? As far as I see we have to add the `plain` handler and formatter to the config and just use it for `root.handlers` instead of the `colored`-handler.

If you really don't want to create the color formatters if `no-color` is given, you should try to build upon a dict with common settings. In this case only the version is common, but at least we don't duplicate information:
```
config = {'version': 1}

if not_color:
    # I don't know a better name now for this variable...
    print_config = ...
else:
    print_config = ...

config.update(print_config)
logging.config.dictConfig(config)
```

coalib/output/Logging.py
```diff
@@ -32,7 +32,7 @@ def get_num_calls_for_level(cls, level):
         return cls._call_counter[level]
 
 
-def configure_logging():
+def configure_logging(no_color=False):
```
Makman2: why not `color` instead of `no_color`? This looks like doing things the other way around ;D

Makman2: Although coala uses a `no-color` argument, imo the internal API doesn't have to follow the nomenclature 1:1 ;)

shreyans800755:
@Makman2 Rebased.

coalib/output/Logging.py
```diff
@@ -41,37 +41,46 @@ def configure_logging():
...
-    logging.config.dictConfig({
-        'version': 1,
-        'handlers': {
-            'colored': {
-                'class': 'logging.StreamHandler',
-                'formatter': 'colored',
-                'stream': sys.stderr
+    formatter = 'colored' if color else 'plain'
+
+    if color is True:
```
Makman2: and if `color == False`? Then you don't configure logging at all :O

shreyans800755: Ouch. Forgot to remove the condition.

coalib/output/Logging.py
```diff
@@ -60,15 +62,21 @@ def configure_logging():
...
                 'datefmt': '%X',
                 'log_colors': {
                     'ERROR': 'red',
                     'WARNING': 'yellow',
                     'INFO': 'blue',
                     'DEBUG': 'green'
                 }
+            },
+            'plain': {
+                '()': 'colorlog.ColoredFormatter',
```
mrtes: I believe you can remove this line, and it will use the default formatter of the logging package.
The default should be able to handle the format string as well, no need to use a custom one in this case.

shreyans800755: Makes sense. Updated.

shreyans800755:
LLVM page is not accessible by appveyor, and not even in my local for that matter. It is causing appveyor build failure.

coalib/output/Logging.py
```diff
@@ -41,12 +41,14 @@ def configure_logging():
     # reset counter handler
     CounterHandler.reset()
 
+    formatter = 'colored' if color else 'plain'
```
Makman2: if you want you can inline that into the dict, but up to you ;)

tests/coalaTest.py
```diff
@@ -179,3 +179,19 @@ def test_logged_error_causes_non_zero_exitcode(self):
...
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                             coala.main,
+                             'coala', '-c', os.devnull,
+                             '-f', re.escape(filename),
+                             '-b', 'LineCountTestBear')
+            self.assertIn('This file has 1 lines.',
+                          stdout,
+                          'The output should report count as 1 lines')
+            self.assertIn('During execution of coala', stderr)
```
Makman2: Why does an error happen?

Makman2:
Even I can't access the download of LLVM... Filing an issue. This needs to be fixed before merge :/

Makman2:
Filed here: https://github.com/coala/coala/issues/4246

coalib/output/Logging.py
```diff
@@ -60,15 +62,19 @@ def configure_logging():
         'formatters': {
             'colored': {
                 '()': 'colorlog.ColoredFormatter',
-                'format': '%(log_color)s[%(levelname)s]%(reset)s[%(asctime)s]'
-                          ' %(message)s',
+                'format': '%(log_color)s[%(levelname)s]%(reset)s'
+                          '[%(asctime)s] %(message)s',
```
Makman2: hmm if possible remove the change or make a different commit, it's a layout change ;) Also the previous layout is valid, so it's more personal preference ;)

shreyans800755: This is just to make line length <80. I didn't intend to change the layout.
Will change it back to how it was.

tests/coalaTest.py
shreyans800755: Not sure. I've just copied test case `test_coala` from https://github.com/coala/coala/blob/master/tests/coalaTest.py#L26. And made changes related to what is needed to be tested for the changes I made in source.

Makman2: Ah okay then it's alright. I just wondered why something is reported as an error (or maybe warning) :)

tests/coalaTest.py
```diff
@@ -179,3 +179,19 @@ def test_logged_error_causes_non_zero_exitcode(self):
...
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                             coala.main,
+                             'coala', '-c', os.devnull,
+                             '-f', re.escape(filename),
+                             '-b', 'LineCountTestBear')
+            self.assertIn('This file has 1 lines.',
+                          stdout,
+                          'The output should report count as 1 lines')
```
Makman2: Up to you:
Maybe we want to be a bit more restrictive here? We know that the non-colored output has no special characters so we may assert for the complete message. But I would allow this too^^

tests/coalaTest.py
mrtes: All the logging stuff is configured to be logged to stderr.

tests/coalaTest.py
shreyans800755: I'm making stderr check more restrictive. I think timestamps are also there which can't be checked easily. IMO stdout check is good enough.

Makman2:
Gaaaah and now language-check... anyway going to ack :+1:

Makman2:
ack 7cf63bb

Makman2:
Btw filed an issue for the unknown source of the stderr error message: https://github.com/coala/coala/issues/4249

shreyans800755:
> Btw filed an issue for the unknown source of the stderr error message: #4249
That looks good.

Makman2:
ack 123424c

Makman2:
@rultor merge

Makman2:
Let's retry...
@rultor merge

Makman2:
@rultor merge

Makman2:
Still not working... let's wait a while, your PR is approved anyway ;)
(hopefully no conflicts arise meanwhile :3)

shreyans800755:
Ah! Same error again. Isn't there any other way to merge PR ?
@Makman2 
We shouldn't rely on a single system for such an importance task, I guess.

Makman2:
Merging is a single process, so we have a single system for it ;) rultor is sometimes not so reliable all the time. But don't worry we don't lose much time, we cherry pick yours and other approved PRs into a master PR as soon as merging flows again.

Maybe we get someday our own merge bot... with more features :3 (and no damn security holes...)

tests/coalaTest.py
```diff
@@ -179,3 +179,23 @@ def test_logged_error_causes_non_zero_exitcode(self):
...
+            self.assertIn('An unknown error occurred. This is a bug. '
```
jayvdb: Why is this allowing stderr with `This is a bug. ... This should never happen.`
Why is there an unknown error?  Why is that a good thing?
If it really is a good thing, it definitely needs a comment to explain why this should be in the output.

shreyans800755: I've copied this test case from https://github.com/coala/coala/blob/master/tests/coalaTest.py#L26 and made changes into it related to issue.

jayvdb: Found the problem, and it is fixable by improving the test logic that you copied.  See https://github.com/coala/coala/pull/4266

jayvdb:
unack 123424c

shreyans800755:
@Makman2 @jayvdb Updated test case to check coloring functionality properly.
Not sure why `circleCI` is failing.
As per logs, pytest is timed out.

> ========== 1 failed, 723 passed, 1 pytest-warnings in 111.75 seconds ===========

And as per https://github.com/coala/coala/blob/master/circle.yml#L23, timeout is 900 sec.
Why is test failing with timeout before 900 sec ? Or 900 sec is for overall CI action ?

Surprisingly rest of the CI are passing fine.

tests/coalaTest.py
```diff
@@ -179,3 +179,24 @@ def test_logged_error_causes_non_zero_exitcode(self):
...
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala')
+            errors = filter(bool, stderr.split('\n'))
+            # Every error message must start with characters
+            # used for coloring.
+            for err in errors:
+                self.assertNotRegex(err, '^\[WARNING\]')
+            self.assertEqual(
+                retval, 0, 'coala must return zero when there are no errors')
```
Makman2: could you move the color testing into an own testing function `test_coala_color`?
(Hm maybe we should call then both functions `test_coala_log_with_colors` and `test_coala_log_without_colors` as `test_coala_color` sounds kind of bad :D)

Makman2:
Let's just retry for CircleCI, maybe it's a random fail^^

jayvdb:
The commit message prefix "no_color" is rather unusual.  I would pick "Logging.py" as that is the main part of the codebase that is being modified.

shreyans800755:
> Let's just retry for CircleCI, maybe it's a random fail^^
@Makman2 You're right.
@jayvdb Updated.

Makman2:
ack ff52cab

Makman2:
oops needs a rebase again :3

shreyans800755:
@Makman2 Rebased.

Makman2:
reack e05c816

Makman2:
@rultor merge

Issue #2509, opened 2016-07-22
coalib/parsing/DefaultArgParser: Add more options in --filter-by


mr-karan:
I think we should have more options such as 
- `--show-bears --filter-by-can_fix`
- `--show-bears --filter-by-can_detect`
- `--show-bears --filter-by-use_stdin`
  The point being, why filter by only one parameter when we can do so for all other Bear variables? 

If this makes sense, I would like to work on it.

mr-karan:
@sils1297 @Makman2 @AbdealiJK  your thoughts on this?

sils1297:
More importantly: we can't filter by bear name at all ATM.

Also don't mix underscores and dashes.

AbdealiJK:
I don't understand. This seems like you're trying to provide a complex UI in CLI - which is not a good idea, because CLI is a bad UI.

Would it not be better to launch a web browser with a nice searchable table of bears ? with every property you could ever want ? (I think this was discussed at some point)

sils1297:
@AbdealiJK I totally agree. There's two points though:

- some people *want* to do it on the CLI
- You *need* the CLI if you're looking for e.g. own bears that are not official

AbdealiJK:
@sils1297 

 - Some people *want* to do it with CLI
  - I agree, makes sense. But it may be wise to get some data on number of people (possibly a poll? Preferably a recommendation from an org that uses coala/agrees to use coala soon)
  - But yeah, having both is a good way to cater to both if time allows.

 - Own bears that are not official
  - What I was talking about is a locally created web page like the coala-html stuff which just generated nice information about bears and gives it to a html page which shows it with eye-candy.
  - The "official" bear webpage is an extension to this idea which is a hosted version of this generated page made from the officially supported bears (So, UI is same and hence easy to use)

Makman2:
I think we should still provide some CLI interface, it doesn't harm.

Makman2:
but maybe we should introduce some kind of "search-language" instead of having an own flag for each kind of field^^

mr-karan:
I agree with the point that *it doesn't harm*. More options never harm, it just increases your help texts and available number of commands ;)

I like the idea of launching a web browser with a searchable table too. I think we first need to implement the functionality in our API and only then think of which UI to go for? :)

Makman2:
so let's make a new issue for implementing some web UI?

mr-karan:
https://github.com/coala-analyzer/coala/issues/2564

[...]

Issue #3472, opened 2017-01-04
ConsoleInteraction.py: Replace log_printers with built-in logging.


tulphoon:

aptrishu:
This would end up removing log_printer from some other places too. are we deprecating it ? @sils

sils:
LogPrinter is already deprecated, it should be removed whereever possible

Sincerely,

Lasse Schuirmann

lasse@schuirmann.net
http://coala.io/ - http://viperdev.io/ - http://gitmate.com/

shreyans800755:
@sils Do we need ListLogPrinter ?
Is it needed by any bears or any other functionality ?
If so, we'll have to build it using logger instead of LogPrinterMixin.

thisch:
@shreyans800755 I would remove ListLogPrinter, since it is only used in the tests

PR #4240, opened 2017-05-17
DefaultArgParser: Add generic filter-by option

shreyans800755:
Closes https://github.com/coala/coala/issues/2509

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -18,6 +18,7 @@ from pyprint.ConsolePrinter import ConsolePrinter
 
 from dependency_management.requirements.PipRequirement import PipRequirement
 
+from coalib.parsing.FilterHelper import FilterHelper, InvalidFilterException
 from coalib.output.Logging import configure_logging
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.parsing.DefaultArgParser import default_arg_parser
@@ -59,11 +60,29 @@ def main(debug=False):
 
         if args.show_bears:
             from coalib.settings.ConfigurationGathering import (
-                get_filtered_bears)
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                log_printer.warn(
+                    "'--filter-by-language ...' is deprecated. "
+                    "Use '--filter-by language ...' instead.")
+                if args.filter_by is None:
+                    args.filter_by = []
+                args.filter_by.append(['language'] + args.filter_by_language)
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x, *filter_args = filter_by
+                    try:
+                        filtered_bears = FilterHelper.get_filtered_bears(
+                            filter_by_x, filter_args, filtered_bears)
+                    except InvalidFilterException as ex:
+                        # If filter is not available
+                        console_printer.print(ex)
+                        return 2
 
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, log_printer)
-
+            local_bears, global_bears = filtered_bears
             show_bears(local_bears,
                        global_bears,
                        args.show_description or args.show_details,
@@ -74,11 +93,8 @@ def main(debug=False):
         elif args.show_capabilities:
             from coalib.collecting.Collectors import (
                 filter_capabilities_by_languages)
-            from coalib.settings.ConfigurationGathering import (
-                get_filtered_bears)
-
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, log_printer)
+            local_bears, _ = FilterHelper.get_filtered_bears(
+                'language', args.show_capabilities)
             capabilities = filter_capabilities_by_languages(
                 local_bears, args.show_capabilities)
             show_language_bears_capabilities(capabilities, console_printer)
--- a/coalib/coala_modes.py
+++ b/coalib/coala_modes.py
@@ -50,8 +50,6 @@ def mode_json(args, debug=False):
     from coalib.misc.Exceptions import get_exitcode
     from coalib.output.Logging import configure_json_logging
     from coalib.output.JSONEncoder import create_json_encoder
-    from coalib.output.printers.LogPrinter import LogPrinter
-    from coalib.settings.ConfigurationGathering import get_filtered_bears
 
     if args.log_json:
         log_stream = configure_json_logging()
@@ -61,8 +59,10 @@ def mode_json(args, debug=False):
 
     if args.show_bears:
         try:
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, LogPrinter())
+            from coalib.parsing.FilterHelper import FilterHelper
+
+            local_bears, global_bears = FilterHelper.get_filtered_bears(
+                'language', args.filter_by_language)
             bears = inverse_dicts(local_bears, global_bears)
             for bear, _ in sorted(bears.items(),
                                   key=lambda bear_tuple:
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -2,6 +2,7 @@ import argparse
 
 from coalib.misc import Constants
 from coalib.collecting.Collectors import get_all_bears_names
+from coalib.parsing.FilterHelper import FilterHelper
 
 
 class CustomFormatter(argparse.RawDescriptionHelpFormatter):
@@ -185,6 +186,12 @@ To run coala without user interaction, run the `coala --non-interactive`,
     outputs_group.add_argument(
         '-l', '--filter-by-language', nargs='+', metavar='LANG',
         help='filters `--show-bears` by the given languages')
+
+    outputs_group.add_argument(
+        '--filter-by', action='append', nargs='+',
+        help='filters `--show-bears` by the filter given as argument. '
+             'Available filters: {}'.format(
+                 FilterHelper.get_all_filters_str()))
 
     outputs_group.add_argument(
         '-p', '--show-capabilities', nargs='+', metavar='LANG',
--- a/coalib/parsing/FilterHelper.py
+++ b/coalib/parsing/FilterHelper.py
@@ -1,0 +1,35 @@
+from coalib.output.printers.LogPrinter import LogPrinter
+
+from coalib.parsing.InvalidFilterException import InvalidFilterException
+
+from coalib.parsing.filters.LanguageFilter import language_filter
+from coalib.parsing.filters.CanDetectFilter import can_detect_filter
+from coalib.parsing.filters.CanFixFilter import can_fix_filter
+
+
+class FilterHelper:
+    available_filters = {'language': language_filter,
+                         'can_detect': can_detect_filter,
+                         'can_fix': can_fix_filter}
+
+    @classmethod
+    def get_all_filters_str(cls, sep=', '):
+        return sep.join(sorted(cls.available_filters))
+
+    @classmethod
+    def is_valid_filter(cls, filter):
+        return filter in cls.available_filters
+
+    @classmethod
+    def get_filtered_bears(cls, filter, args, all_bears=None):
+        if all_bears is None:
+            from coalib.settings.ConfigurationGathering import (
+                get_all_bears)
+            all_bears = get_all_bears(LogPrinter())
+        if not cls.is_valid_filter(filter):
+            raise InvalidFilterException('{!r} is an invalid filter. '
+                                         'Available filters: {}'.format(
+                                             filter, cls.get_all_filters_str()))
+        if not args or len(args) == 0:
+            return all_bears
+        return cls.available_filters[filter](all_bears, args)
--- a/coalib/parsing/filters/CanDetectFilter.py
+++ b/coalib/parsing/filters/CanDetectFilter.py
@@ -1,0 +1,15 @@
+from coalib.parsing.filters.Filter import filter
+
+
+@filter
+def can_detect_filter(bear, args):
+    """
+    Filters the bears by ``CAN_DETECT``.
+
+    :param bear: Bear object.
+    :param args: Set of detectable issue types on which ``bear`` is to be
+                 filtered.
+    :return:     ``True`` if this bear matches the criteria inside args,
+                 ``False`` otherwise.
+    """
+    return bool({detect.lower() for detect in bear.CAN_DETECT} & args)
--- a/coalib/parsing/filters/CanFixFilter.py
+++ b/coalib/parsing/filters/CanFixFilter.py
@@ -1,0 +1,15 @@
+from coalib.parsing.filters.Filter import filter
+
+
+@filter
+def can_fix_filter(bear, args):
+    """
+    Filters the bears by ``CAN_FIX``.
+
+    :param bear: Bear object.
+    :param args: Set of fixable issue types on which ``bear`` is to be
+                 filtered.
+    :return:     ``True`` if this bear matches the criteria inside args,
+                 ``False`` otherwise.
+    """
+    return bool({fix.lower() for fix in bear.CAN_FIX} & args)
--- a/coalib/parsing/filters/Filter.py
+++ b/coalib/parsing/filters/Filter.py
@@ -1,0 +1,19 @@
+from functools import wraps
+
+
+def filter(filter_function):
+    def filter_section_bears(bears, args):
+        return {section:
+                tuple(bear for bear in bears[section]
+                      if filter_function(bear, args))
+                for section in bears}
+
+    @wraps(filter_function)
+    def filter_wrapper(section_to_bears_dict, args):
+        args = {arg.lower() for arg in args}
+        local_bears, global_bears = section_to_bears_dict
+        local_bears = filter_section_bears(local_bears, args)
+        global_bears = filter_section_bears(global_bears, args)
+        return local_bears, global_bears
+
+    return filter_wrapper
--- a/coalib/parsing/filters/LanguageFilter.py
+++ b/coalib/parsing/filters/LanguageFilter.py
@@ -1,0 +1,14 @@
+from coalib.parsing.filters.Filter import filter
+
+
+@filter
+def language_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
+
+    :param bear: Bear object.
+    :param args: Set of languages on which ``bear`` is to be filtered.
+    :return:     ``True`` if this bear matches the criteria inside args,
+                 ``False`` otherwise.
+    """
+    return bool({lang.lower() for lang in bear.LANGUAGES} & (args | {'all'}))
--- a/coalib/parsing/filters/__init__.py
+++ b/coalib/parsing/filters/__init__.py
@@ -1,0 +1,4 @@
+"""
+This package holds filter functions. Filter objects are used to
+get a list of bears with specific properties.
+"""
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -3,8 +3,7 @@ import re
 import sys
 import logging
 
-from coalib.collecting.Collectors import (
-    collect_all_bears_from_sections, filter_section_bears_by_languages)
+from coalib.collecting.Collectors import collect_all_bears_from_sections
 from coalib.misc import Constants
 from coalib.output.ConfWriter import ConfWriter
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
@@ -322,11 +321,8 @@ def get_config_directory(section):
     return config if os.path.isdir(config) else os.path.dirname(config)
 
 
-def get_filtered_bears(languages, log_printer, arg_parser=None):
+def get_all_bears(log_printer, arg_parser=None):
     """
-    Fetch bears and filter them based on given list of languages.
-
-    :param languages:   List of languages.
     :param log_printer: The log_printer to handle logging.
     :param arg_parser:  An ``ArgParser`` object.
     :return:            Tuple containing dictionaries of local bears
@@ -337,11 +333,6 @@ def get_filtered_bears(languages, log_printer, arg_parser=None):
                                      arg_parser=arg_parser)
     local_bears, global_bears = collect_all_bears_from_sections(
         sections, log_printer)
-    if languages:
-        local_bears = filter_section_bears_by_languages(
-            local_bears, languages)
-        global_bears = filter_section_bears_by_languages(
-            global_bears, languages)
     return local_bears, global_bears
 
 
--- a/tests/coalaTest.py
+++ b/tests/coalaTest.py
@@ -164,7 +164,8 @@ class coalaTest(unittest.TestCase):
             self.assertEqual(retval, 0)
             # 2 bears plus 1 line holding the closing colour escape sequence.
             self.assertEqual(len(stdout.splitlines()), 3)
-            self.assertFalse(stderr)
+            self.assertIn(
+                "'--filter-by-language ...' is deprecated", stderr)
 
     def test_show_language_bears_debug(self):
         self.test_show_language_bears(debug=True)
--- a/tests/parsing/FilterTest.py
+++ b/tests/parsing/FilterTest.py
@@ -1,0 +1,121 @@
+import unittest
+
+from coalib import coala
+from coalib.parsing.FilterHelper import FilterHelper, InvalidFilterException
+from tests.TestUtilities import execute_coala, bear_test_module
+
+
+class FilterTest(unittest.TestCase):
+
+    def test_filter_by_language_c(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by', 'language', 'c')
+            self.assertEqual(retval, 0)
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_by_language_java_can_fix_syntax(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B',
+                '--filter-by', 'language', 'java',
+                '--filter-by', 'can_fix', 'syntax')
+            self.assertEqual(retval, 0)
+            # 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 1)
+            self.assertIn('No bears to show.', stdout)
+
+    def test_filter_by_language_java_can_detect_formatting(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B',
+                '--filter-by', 'language', 'java',
+                '--filter-by', 'can_detect', 'formatting')
+            self.assertEqual(retval, 0)
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_bylanguage_java_can_detect_syntax(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B',
+                '--filter-by-language', 'java',
+                '--filter-by', 'can_detect', 'formatting')
+            self.assertEqual(retval, 0)
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_by_can_detect_syntax(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by',
+                'can_detect', 'syntax')
+            self.assertEqual(retval, 0)
+            # 2 bears plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 3)
+
+    def test_filter_by_can_detect_security(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by',
+                'can_detect', 'security')
+            self.assertEqual(retval, 0)
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_by_can_detect_spelling(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by',
+                'can_detect', 'spelling')
+            self.assertEqual(retval, 0)
+            # 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 1)
+            self.assertIn('No bears to show.', stdout)
+
+    def test_filter_by_can_fix_syntax(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by',
+                'can_fix', 'syntax')
+            self.assertEqual(retval, 0)
+            # 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 1)
+            self.assertIn('No bears to show.', stdout)
+
+    def test_filter_by_can_fix_redundancy(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by',
+                'can_fix', 'redundancy')
+            self.assertEqual(retval, 0)
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_by_unknown(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by', 'unknown', 'arg1')
+            self.assertEqual(retval, 2)
+            self.assertRaisesRegex(InvalidFilterException,
+                                   '{!r} is an invalid filter. Available '
+                                   'filters: {}'.format(
+                                       filter,
+                                       FilterHelper.get_all_filters_str()))
+
+    def test_filter_by_can_fix_null(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by', 'can_fix')
+            self.assertEqual(retval, 0)
+            # 8 bears plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 14)
+
+    def test_filter_by_can_detect_null(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by', 'can_detect')
+            self.assertEqual(retval, 0)
+            # 8 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 14)
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -184,26 +184,6 @@ def collect_bears(bear_dirs, bear_globs, kinds, log_printer,
     return bears_found
 
 
-def filter_section_bears_by_languages(bears, languages):
-    """
-    Filters the bears by languages.
-
-    :param bears:       The dictionary of the sections as keys and list of
-                        bears as values.
-    :param languages:   Languages that bears are being filtered on.
-    :return:            New dictionary with filtered out bears that don't match
-                        any language from languages.
-    """
-    new_bears = {}
-    # All bears with "all" languages supported shall be shown
-    languages = set(language.lower() for language in languages) | {'all'}
-    for section in bears.keys():
-        new_bears[section] = tuple(
-            bear for bear in bears[section]
-            if {language.lower() for language in bear.LANGUAGES} & languages)
-    return new_bears
-
-
 def filter_capabilities_by_languages(bears, languages):
     """
     Filters the bears capabilities by languages.
--- a/coalib/parsing/InvalidFilterException.py
+++ b/coalib/parsing/InvalidFilterException.py
@@ -1,0 +1,2 @@
+class InvalidFilterException(Exception):
+    pass
--- a/tests/collecting/CollectorsTest.py
+++ b/tests/collecting/CollectorsTest.py
@@ -6,9 +6,9 @@ from pyprint.ConsolePrinter import ConsolePrinter
 
 from coalib.bears.Bear import Bear
 from coalib.collecting.Collectors import (
-    collect_all_bears_from_sections, collect_bears, collect_dirs, collect_files,
-    collect_registered_bears_dirs, filter_section_bears_by_languages,
-    get_all_bears, get_all_bears_names)
+    collect_all_bears_from_sections, collect_bears, collect_dirs,
+    collect_files, collect_registered_bears_dirs, get_all_bears,
+    get_all_bears_names)
 from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.output.printers.ListLogPrinter import ListLogPrinter
 from coalib.settings.Section import Section
@@ -283,24 +283,6 @@ class CollectorsTests(unittest.TestCase):
         self.collectors_test_dir = os.path.join(current_dir,
                                                 'collectors_test_dir')
         self.log_printer = LogPrinter(ConsolePrinter())
-
-    def test_filter_section_bears_by_languages(self):
-        test_section = Section('test_section')
-        test_section.bear_dirs = lambda: os.path.join(self.collectors_test_dir,
-                                                      'bears_local_global',
-                                                      '**')
-        local_bears, global_bears = collect_all_bears_from_sections(
-            {'test_section': test_section},
-            self.log_printer)
-        local_bears = filter_section_bears_by_languages(local_bears, ['C'])
-        self.assertEqual(len(local_bears['test_section']), 1)
-        self.assertEqual(str(local_bears['test_section'][0]),
-                         "<class 'bears2.Test2LocalBear'>")
-
-        global_bears = filter_section_bears_by_languages(global_bears, ['Java'])
-        self.assertEqual(len(global_bears['test_section']), 1)
-        self.assertEqual(str(global_bears['test_section'][0]),
-                         "<class 'bears1.Test1GlobalBear'>")
 
     def test_get_all_bears(self):
         with bear_test_module():
--- a/tests/test_bears/JavaTestBear.py
+++ b/tests/test_bears/JavaTestBear.py
@@ -4,6 +4,8 @@ from coalib.bears.LocalBear import LocalBear
 class JavaTestBear(LocalBear):
     LANGUAGES = {'java'}
     LICENSE = 'AGPL-3.0'
+    CAN_DETECT = {'Syntax', 'Formatting'}
+    CAN_FIX = {'Security'}
 
     def run(self, filename, file, config: str=''):
         """
--- a/tests/test_bears/LinterBear.py
+++ b/tests/test_bears/LinterBear.py
@@ -5,6 +5,8 @@ from coalib.bearlib.abstractions.Linter import linter
         output_format='regex',
         output_regex=r'.+:(?P<line>\d+):(?P<message>.*)')
 class EchoBear:
+    CAN_DETECT = {'Syntax', 'Security'}
+    CAN_FIX = {'Redundancy'}
     """
     A simple bear to test that collectors are importing also bears that are
     defined in another file *but* have baseclasses in the right file.


shreyans800755:
This PR needs work regarding adding filters related to can_detect, can_fix, etc.
I've created PR just to show the approach before moving ahead.

Here, we can add filtering logic inside FilterImplementation.py to add more options like can_detect, can_fix, etc. Currently I've only added one option for language since its logic was already available in coala.py.
@Makman2 
Please review the approach.

FYI: One more approach I could think was to create abstract class with method to get bears and implement all the filters as different class inheriting from the abstract class. But it would require creation of a new file every time we need to add an filter. So, I did not choose this approach.

coalib/coala.py
```diff
@@ -41,18 +41,35 @@ def main():
...
-                args.filter_by_language, log_printer)
-
-            show_bears(local_bears,
-                       global_bears,
-                       args.show_description or args.show_details,
-                       args.show_details,
-                       console_printer)
-
+            # filter_by_language option should be deprecated soon.
+            # Since it is already covered by filter_by option.
```
Makman2: log a warning ;)

coalib/parsing/DefaultArgParser.py
```diff
@@ -187,6 +187,10 @@ def default_arg_parser(formatter_class=None):
         help='filters `--show-bears` by the given languages')
 
     outputs_group.add_argument(
+        '--filter-by', nargs='+',
```
Makman2: normally I would say that this would be good and intuitive, but I'm not sure how it plays well with other arguments. coala takes quite many currently, and I have the feeling that this doesn't work any more:
```
# The args here don't make sense together, but just for what I mean:
coala --filter-by filter1 filter2 --non-interactive --save
```
But I recall that argparse might still handle that properly... could you test that? :)

coalib/parsing/FilterImplementation.py
```diff
@@ -0,0 +1,7 @@
+def LanguageFilter(languages, log_printer):
```
Makman2: so we could still do one filter one file, can't we? Actually the more you modularize, the better, maintaining multiple files is not a problem, it's even better than maintaining a single file ;)

coalib/parsing/Filter.py
```diff
@@ -0,0 +1,19 @@
...
+    all_filters = {'language': LanguageFilter}
+
+    def __init__(self):
+        pass
+
+    def is_valid_filter(filter):
+        return filter in Filter.all_filters
+
+    @staticmethod
+    def get_filtered_bears(filter, args, log_printer):
```
Makman2: don't use `log_printer`, use `logging` instead. `log_printer` is inofficially deprecated and we are going to remove it step by step in future. The current CLI is just quite complicated and we are all to lazy to do that properly :/

coalib/parsing/Filter.py
```diff
@@ -0,0 +1,19 @@
...
+        if Filter.is_valid_filter(filter):
+            return Filter.all_filters[filter](args, log_printer)
```
Makman2: what about passing each filter the list of bears? :) Then the filters don't have to look for them themselves, but we have one function collecting them all, and then each filter can process them :)

Makman2: (together with the filter options)

Makman2:
Idea and direction looks good, you distribute different types of filter across functions/classes/other entities.

> FYI: One more approach I could think was to create abstract class with method to get bears and implement all the filters as different class inheriting from the abstract class. But it would require creation of a new file every time we need to add an filter. So, I did not choose this approach.

The more files, the better to maintain :) As this speaks for a high modularization of your functionality, which is always good. But it doesn't have to be classes, you can also use functions (like you already do with `LanguageFilter`, weird naming convention btw for functions ;D).

coalib/parsing/DefaultArgParser.py
shreyans800755: It works fine. Although  need to add `-B` before `--filter-by`.
argparse takes care of that.

coalib/parsing/Filter.py
shreyans800755: Sounds good.

coalib/parsing/Filter.py
shreyans800755: Is there any issue taking care of it ? If not, I can file it and also start working on it. IMO, it is going to be time consuming to remove dependency from `log_printer`.

Makman2: could you search for one? If not then feel free to file one :+1:

shreyans800755: Already there, https://github.com/coala/coala/issues/3472.

shreyans800755:
@Makman2 
This is wip, but I want to confirm the approach. I've done LanguageFilter part. I'll add other filters soon.
Testing is still remaining(which is causing the CI builds failure).

I'm not removing log_printer as of now, as there's separate issue for that, which will take care of it, I believe.(coz I'm working on that issue :)).

Makman2:
Regarding `log_printers`: Just don't use them, saves code and is easier ;)

coalib/settings/ConfigurationGathering.py
```diff
@@ -329,11 +326,6 @@ def get_filtered_bears(languages, log_printer, arg_parser=None):
                                      arg_parser=arg_parser)
     local_bears, global_bears = collect_all_bears_from_sections(
         sections, log_printer)
-    if languages:
-        local_bears = filter_section_bears_by_languages(
-            local_bears, languages)
-        global_bears = filter_section_bears_by_languages(
-            global_bears, languages)
```
Makman2: perfect :+1:

coalib/parsing/Filter.py
```diff
@@ -0,0 +1,22 @@
+from coalib.parsing.filters.LanguageFilter import language_filter
+
+
+class Filter:
+    all_filters = {'language': language_filter}
```
Makman2: I think to modularize things more, you should provide a function `registerFilter` where someone can register his own filter object/class/function/etc. Then we can dynamically built a filter with different sets of functionality (though I cannot think of a useful case right now).

This would work for example like https://docs.python.org/3/library/unittest.html#unittest.TestCase.addTypeEqualityFunc

coalib/coala.py
```diff
@@ -42,18 +42,39 @@ def main():
...
-            show_bears(local_bears,
-                       global_bears,
-                       args.show_description or args.show_details,
-                       args.show_details,
-                       console_printer)
-
+            if args.filter_by:
+                from coalib.parsing.Filter import Filter
+                filter = args.filter_by[0]
+                filter_args = args.filter_by[1:]
```
Makman2: what about multiple filters? Like
```
--filter-by language cpp --filter-by aspect License
```
or
```
--filter-by language cpp&haskell
```

Makman2: The syntax can be different, but just as an example for more complex filters^^

coalib/parsing/filters/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+This package holds filter objects. Filter objects are used to
+get the list of bears with specific properties.
+
+Example:
+    LanguageFilter: To get the list of bears for provided language/
```
Makman2: I wouldn't provide an example, I think people know how to handle the package roughly once they see its content ;)

coalib/parsing/filters/LanguageFilter.py
```diff
@@ -0,0 +1,27 @@
+def language_filter(bears, languages, log_printer):
+    local_bears, global_bears = bears
```
Makman2: do we want to separate local and global bears? I wouldn't do that, especially as in the next-gen core we won't make a distinction any more in narrower sense, as their difference is only the degree in parallelization. The current core has special code for both of them, but the next-gen core won't. The bears do specify themselves the degree of parallelization^^

Makman2: ah but we need that for `show_bears`... ugh that sucks...

Makman2: why the heck do we need to display bears differently depending on whether they are a `GlobalBear` or `LocalBear`... maybe we wanna change that @shreyans800755 ? But that's quite a task, so you don't have to ;)

coalib/parsing/filters/LanguageFilter.py
```diff
@@ -0,0 +1,27 @@
...
+    if languages:
+        local_bears = filter_section_bears_by_languages(local_bears, languages)
+        global_bears = filter_section_bears_by_languages(
+            global_bears, languages)
+    return local_bears, global_bears
```
Makman2: I imagine currently that every filter will contain this code, maybe you invoke in the "filter-core" the filter function twice? Once for local and once for global bears? Then people only have to write the filter^^

coalib/coala.py
Makman2: Ah and what about aliases? So I can do
```
--filter-by lang cpp
# equivalent to
# --filter-by language cpp
```

coalib/parsing/Filter.py
shreyans800755: I think I'll first complete rest, and then work on this one. This sounds like an interesting feature. But may require a bit more design.

coalib/coala.py
shreyans800755: I had thought of this, but the issue comes with arg parser. 

> --filter-by language cpp --filter-by aspect License

In case of multiple filter-by available, it only yields results for the later filter. I'll try to fix that.

> --filter-by language cpp&haskell

Currently `And` is not available as operator, but if space separated list is provided, it gives bears with `or` operator.
So, for the following:

> coala -B --filter-by language c python

It gives all the bears that can be used of either C or python or both.

As far as `and` operator is concerned, fixing of multiple `--filter-by` will help.

I've not thought of aliases, at the moment. But we just need to add an entry in the map for that. That's fairly easy to do.

coalib/parsing/filters/__init__.py
shreyans800755: Cool

coalib/parsing/filters/LanguageFilter.py
shreyans800755: Should I add warning log in show bears regarding parameters deprecation ?

coalib/parsing/filters/UseStdinFilter.py
```diff
@@ -0,0 +1,31 @@
...
+                        should use stdin
+    :return:            New dictionary with filtered out bears with
+                        use_stdin value specified in param.
+    """
+    new_bears = {}
+    stdin_values = set(stdin.lower() for stdin in use_stdin)
+    for section in bears.keys():
+        new_bears[section] = tuple(
+            bear for bear in bears[section]
+            if {stdin.lower() for stdin in bear.use_stdin} & stdin_values)
```
shreyans800755: @Makman2 
With `coala -B --filter-by use_stdin True`,
I'm getting the following error:

> type object 'VerilogLintBear' has no attribute 'use_stdin'

Bear object doesn't have use_stdin attribute. I'm not sure how to access a value from decorator's object.
Do you have any clues ?

coalib/parsing/Filter.py
Makman2: shouldn't be that hard though ;)

coalib/coala.py
Makman2: > >  --filter-by language cpp --filter-by aspect License
>
> In case of multiple filter-by available, it only yields results for the later filter. I'll try to fix that.

That would be good to have definitely. But actually we can merge the more primitive version first and improve from that. Step by step is always good ;)

okay an "and" operator is anyway not really useful, but it's good that "or" is supported :+1: just an impression that we may want to support multiple operators.

> I've not thought of aliases, at the moment. But we just need to add an entry in the map for that. That's fairly easy to do.

:+1:

coalib/parsing/filters/UseStdinFilter.py
Makman2: you shouldn't filter for `use_stdin`, this is an implementation detail of the bears. `use_stdin` is stored inside a dict of the bear decorated by `@linter`, but I'm not sure if it's scoped and thus not accessible. But like said, I wouldn't implement a filter for this^^ Maybe for the executable of a linter this might be useful, but not all bears are linter bears ;)

Makman2: Just look at `Linter.py`, although it's not so easy code ;)

shreyans800755: Makes sense. I don't think it would be much useful to end users since most of them won't even know what is `use_stdin` in the context.

Makman2: exactly :)

shreyans800755:
@Makman2 
I've added the tests.
Currently the only function, I think, not working is when more than one `--filter-by` available. In that case, only later one is considered. I'll try to see if it can be fixed easily.

coalib/coala.py
shreyans800755: `or` operator is allowed only within a single filter.
It means we can write `filter-by C or python`but not`filter-by C or filter-by can_fix syntax`. Although I'm not sure later is that useful. It is mainly useful with`and`operators, which I'm a working on to implement.

shreyans800755:
@Makman2 
I've added `and` operator and couple tests in `FilterTest.py`.

Note: we'll have to wait for https://github.com/coala/coala/issues/4293 to get merged. Due to some issue, urllib3 is not working and causing Travis and appveyor build failure. (Surprisingly, circle CI is passing successfully)

Currently working register filter functionality.
What kind of interface are we looking for ?
Are we looking for end users functionality to make shortcuts for filter by registering it like the following example:
`coala -B --register-filter pylang language python`
`coala -B --filter-by pylang` or simply `coala -B pylang`

or am I misunderstanding the use case ?

userzimmermann:
@shreyans800755 The waiting for https://github.com/coala/coala/issues/4293 is over! https://github.com/coala/coala/pull/4294 has been merged :)

shreyans800755:
@userzimmermann Thanks for the quick work. I'll just rebase.

Makman2:
> Are we looking for end users functionality to make shortcuts for filter by registering it like the following example:
> coala -B --register-filter pylang language python
> coala -B --filter-by pylang or simply coala -B pylang
>
> or am I misunderstanding the use case ?

I meant registering filters from code side. The user shall not be able to register new filters (anyway way too complicated)

```
class LanguageFilter:
    ...

Filter.add_filter('language', LanguageFilter) # <-- example signature, we may want to move the specifier (here "language") to the filter class/callable/object
# or if LanguageFilter would contain the 'language' in itself, just
Filter.add_filter(LanguageFilter)

# we may built an instance of a filter with a certain feature-set, not sure if it makes sense in our case:
coala_filter = Filter()
coala_filter.add_filter(...)
coala_filter.add_filter(...)
# or maybe we support something like
coala_filter = Filter(LanguageFilter, AnotherFilter, AndAgainAnotherFilter, ...)
```

shreyans800755:
@Makman2 
I'm still not able to find any good use case or intention behind this. Filter functionality is just to show the bears with specific functionality to users. I don't see how building a filter object will help us.
Can you please mention a use case where this might be useful ? And is it really in scope of this issue ? If not, may be we can file an issue and work can be done with respect to new issue.

Makman2:
I don't have one, it was just a design intention. Feel free to ignore, I just wanted to give some input ;)

Makman2:
Though we could actually try to automatically find filters in the same package, then you don't have to register it manually with your dict you currently use.

shreyans800755:
Can you please elaborate a bit more ?

Makman2:
Every filter we put into the same package where you define your main filter class handling everything can be autodetected maybe. Just an example:

```
# don't use star imports! I just do it because of simplicity!
from coalib.parsing.filters import *


def find_filters():
    for cls in dir():
        # Just an idea how we can identify filters using classes (I'm always using classes it seems...)
        if issubclass(cls, Filter):
            yield cls

class Filter:
    all_filters = set(find_filters) # <-- I know this needs to be a dict, but I hope you get the idea.
```

Actually when I'm thinking of it, this is quite difficult, nvm.

shreyans800755:
I'll try to implement it.

Makman2:
@shreyans800755 no really let's make it afterwards with a separate issue. This takes too much time in the first iteration.

coalib/coala.py
```diff
@@ -59,26 +59,56 @@ def main(debug=False):
...
+                for filter_by in args.filter_by:
+                    from coalib.parsing.Filter import Filter
+                    filter = filter_by[0]
+                    filter_args = filter_by[1:]
+                    try:
+                        all_bears = Filter.get_filtered_bears(
+                            filter, filter_args, all_bears)
+                    except Exception as ex:
+                        console_printer.print(ex)
+                        return 255
```
shreyans800755: @Makman2 I'm returning 255 as error code in case the first argument of filter-by (type of filter) is not in the list. I'm not sure if there is a meaning of each return code in coala conventions. I tried to find if there's a list which describe the meaning of return codes, or list of constants/enums assigning a constant variable with return code, but didn't find it.
Can you please confirm if this return code is valid or any list of constant return code variable is available ?
Or if any other improvements are required, or is good to merge ?

Makman2: Yes each code has a meaning, and we have a list here: https://docs.coala.io/en/latest/Users/Exit_Codes.html

We still have some slots left, though maybe we should keep 255. Not sure^^

shreyans800755: > 2 - Invalid arguments were passed to coala in the command line.

error code 2 seems more suitable to me.
I'll update it.

shreyans800755:
@Makman2 Are we looking for any other changes ? Or is this good to merge ?

coalib/coala.py
```diff
@@ -59,26 +59,56 @@ def main(debug=False):
...
+            all_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead")
+                if args.filter_by is None:
+                    args.filter_by = []
+                filter = ['language']
```
Makman2: could you maybe alter the variable name a bit? `filter` is a built-in function, we shouldn't override it^^

coalib/coala.py
```diff
@@ -59,26 +59,56 @@ def main(debug=False):
...
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    from coalib.parsing.Filter import Filter
+                    filter = filter_by[0]
+                    filter_args = filter_by[1:]
+                    try:
+                        all_bears = Filter.get_filtered_bears(
+                            filter, filter_args, all_bears)
```
Makman2: hm I dislike the name `all_bears`, as it's not all of them ;) Maybe `filtered_bears` fits better :+1:

coalib/coala.py
```diff
@@ -59,26 +59,56 @@ def main(debug=False):
...
+                    except Exception as ex:
```
Makman2: This is too broad. Which exceptions can happen explicitly?

coalib/coala.py
```diff
@@ -59,26 +59,56 @@ def main(debug=False):
...
+                           args.show_details,
+                           console_printer)
+            else:
+                # To print all the bears if no filters provided
+                local_bears, global_bears = all_bears
+                show_bears(local_bears,
+                           global_bears,
+                           args.show_description or args.show_details,
+                           args.show_details,
+                           console_printer)
```
Makman2: the code in the else branch looks the same like in the normal "if" branch. The loop won't enter if there are no filters, so imo you can safely remove the `if`

coalib/coala_modes.py
```diff
@@ -61,8 +59,10 @@ def mode_json(args, debug=False):
 
     if args.show_bears:
         try:
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, LogPrinter())
+            from coalib.parsing.Filter import Filter
```
Makman2: why not importing at top?

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,31 @@
+def can_detect_filter(bears, detects):
+    if detects:
+        bears = filter_bears_by_can_detects(bears, detects)
+    return bears
+
+
+def filter_bears_by_can_detects(bears, detects):
+    local_bears, global_bears = bears
+    local_bears = filter_section_bears_by_can_detects(local_bears, detects)
+    global_bears = filter_section_bears_by_can_detects(
```
Makman2: what about my suggestion to put this code into the filter class? Otherwise every filter needs to do this^^

shreyans800755: How about creating decorator for this ? And it will take care of this repetition.

coalib/coala_modes.py
shreyans800755: Updated. Surprisingly its not able to resolve the reference of `Filter`. Logged at https://ci.appveyor.com/project/coala/coala/build/1.0.8117/job/78fcbqt9qp6atmyj#L507.

coalib/parsing/filters/CanDetectFilter.py
shreyans800755: Figured out, scope is not proper for Filter.

coalib/coala_modes.py
shreyans800755: Figured out, scope is not appropriate for `Filter`.

coalib/parsing/filters/CanDetectFilter.py
shreyans800755: @Makman2  I've added implementation with class. Now, to add new filter, new class has to be derived from `FilterBase` and implement an abstract method.

shreyans800755:
@Makman2 Can you please review this ?

coalib/parsing/filters/FilterBase.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABC, abstractmethod
+
+
+def overrides(interface_class):
+    def overrider(method):
+        assert(method.__name__ in dir(interface_class))
+        return method
+    return overrider
```
Makman2: Hm that makes me remember java :D Usually not bad this kind of contract programming, but I think we should discuss that in a separate issue if we really want that. Until then I wouldn't use this as this is inconsistent with the rest of the codebase^^

shreyans800755: This is just to impose more restrictions to reduce errors and make the code a bit more self-documented.

We can remove this too, if not inconsistent. I've tried removing it in local and everything works fine. I can push changes if you want.

shreyans800755: Or may be this `overrides` decorator can be kept in some utility class/function so that whole codebase can use it easily.

coalib/coala.py
```diff
@@ -59,26 +60,49 @@ def main(debug=False):
...
+                    args.filter_by = []
+                filter_by_x = ['language']
+                filter_by_x.extend(args.filter_by_language)
+                args.filter_by.append(filter_by_x)
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x = filter_by[0]
+                    filter_args = filter_by[1:]
```
userzimmermann: You can also do just `filter_by_x, *filter_args = filter_by`

coalib/coala.py
```diff
@@ -59,26 +60,49 @@ def main(debug=False):
...
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead")
+                if args.filter_by is None:
+                    args.filter_by = []
+                filter_by_x = ['language']
+                filter_by_x.extend(args.filter_by_language)
+                args.filter_by.append(filter_by_x)
```
userzimmermann: You can simplify this to:

```python
                args.filter_by = args.filter_by or []
                args.filter_by += ['language'] + args.filter_by_language
```

coalib/parsing/filters/FilterBase.py
shreyans800755: Its not about `ABC` and `@abstractmethod` that @Makman2 is discussing about.
Its mainly custom `@overrides` decorator.

coalib/coala.py
shreyans800755: Have to keep `append` instead of `+=` as it is args and being used in the else part also. 
Rest of the suggestion is applied.

shreyans800755:
@userzimmermann Pushed changes. Let me know if anything else to be changed.

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,21 @@
...
+
+class CanDetectFilter(FilterBase):
+
+    def filter_section_bears(self, bears, args):
+        """
+        Filters the bears by can_detect.
+
+        :param bears:       The dictionary of the sections as keys and list of
+                            bears as values.
+        :param args:        detects that bears are being filtered on.
```
Makman2: detects -> Detects :)

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,21 @@
+from coalib.parsing.filters.FilterBase import FilterBase
+
+
+class CanDetectFilter(FilterBase):
+
+    def filter_section_bears(self, bears, args):
+        """
+        Filters the bears by can_detect.
```
Makman2: can_detect -->
```
``CAN_DETECT``
```

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,21 @@
+from coalib.parsing.filters.FilterBase import FilterBase
+
+
+class CanFixFilter(FilterBase):
+
+    def filter_section_bears(self, bears, args):
+        """
+        Filters the bears by can_detect.
```
Makman2: copy'n'paste gremlin sneaked in :P
-->
```
``CAN_FIX``
```

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,21 @@
...
+
+        :param bears:       The dictionary of the sections as keys and list of
+                            bears as values.
+        :param args:        fixes that bears are being filtered on.
```
Makman2: fixes -> Fixes

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,31 @@
...
+        return filter in FilterHelper.all_filters
+
+    @staticmethod
+    def get_filtered_bears(filter, args, all_bears=None):
+        if all_bears is None:
+            from coalib.settings.ConfigurationGathering import (
+                get_all_bears)
+            all_bears = get_all_bears(LogPrinter())
+        if FilterHelper.is_valid_filter(filter):
+            filter_object = FilterHelper.all_filters[filter]()
```
Makman2: Use rather `@classmethod` instead of `@staticmethod` and do `cls.all_filters[filter]()`

(same for `is_valid_filter` above)

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,21 @@
...
+        :param args:        detects that bears are being filtered on.
+        :return:            New dictionary with filtered out bears that matches
+                            any detect from detects.
+        """
+        detects = {detect.lower() for detect in args}
+        return {section:
+                tuple(bear for bear in bears[section]
+                      if {detect.lower()
+                          for detect in bear.CAN_DETECT} & detects)
+                for section in bears.keys()}
```
Makman2: Do you need the section actually for filters?

Makman2: Btw: -->
```python
{section:
 tuple(bear for bear in bs
       if {detect.lower() for detect in bear.CAN_DETECT} & detects)
 for bs in bears.values()}
```
1. Iterating over `bears.keys()` is the same as iterating over just `bears`.
2. You just use the values, so just use `bears.values()` :)

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,21 @@
+from coalib.parsing.filters.FilterBase import FilterBase
+
+
+class CanDetectFilter(FilterBase):
+
+    def filter_section_bears(self, bears, args):
```
Makman2: don't have a problem to keep it like this currently, but do we need `self`? Then you don't need to instantiate a filter object^^

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,21 @@
...
+        :return:            New dictionary with filtered out bears that matches
+                            any fix from fixes.
+        """
+        fixes = {fix.lower() for fix in args}
+        return {section:
+                tuple(bear for bear in bears[section]
+                      if {fix.lower()
+                          for fix in bear.CAN_FIX} & fixes)
+                for section in bears.keys()}
```
Makman2: --> `bears.values()`, see above

coalib/parsing/filters/CanDetectFilter.py
Makman2: Nvm because you use base types^^

coalib/parsing/filters/LanguageFilter.py
```diff
@@ -0,0 +1,20 @@
...
+                            bears as values.
+        :param args:        languages that bears are being filtered on.
+        :return:            New dictionary with filtered out bears that matches
+                            any language from languages.
+        """
+        languages = {lang.lower() for lang in args} | {'all'}
+        return {section:
+                tuple(bear for bear in bears[section]
+                      if {lang.lower() for lang in bear.LANGUAGES} & languages)
+                for section in bears.keys()}
```
Makman2: --> `bears.values()`, see above

coalib/parsing/filters/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+This package holds filter objects. Filter objects are used to
+get the list of bears with specific properties.
+
```
Makman2: redundant newline

coalib/parsing/filters/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+This package holds filter objects. Filter objects are used to
+get the list of bears with specific properties.
```
Makman2: `the` -> `a`, sounds a bit nicer ;)

coalib/settings/ConfigurationGathering.py
```diff
@@ -4,7 +4,7 @@
 import logging
 
 from coalib.collecting.Collectors import (
-    collect_all_bears_from_sections, filter_section_bears_by_languages)
+    collect_all_bears_from_sections)
```
Makman2: this does fit into a single line now :+1:

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,31 @@
...
+            return filter_object.get_filtered_bears(all_bears, args)
+        raise AttributeError(FilterHelper.get_help_message(filter))
+
+    @staticmethod
+    def get_help_message(filter):
+        return '{} is invalid filter.\nAvailable filters: {}'.format(
```
Makman2: --> is **an** invalid filter.

Makman2: I would rather avoid `\n`s in tracebacks, they usually distort the output layout. But up to you^^

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,31 @@
...
+            filter, ', '.join(sorted(FilterHelper.all_filters.keys())))
```
Makman2: you don't need `.keys()` :)

coalib/parsing/FilterHelper.py
Makman2: Maybe you want to use `{!r}` so the filter string gets enclosed inside quotes?

Makman2:
I'm still a bit unsure about using filter objects as we don't store any additional information except the filter function itself currently. The base class could be inlined into `FilterHelper`, so we don't need to override anything any more and we can use pure functions.

But the design is okay for now imo :)

coalib/parsing/filters/CanDetectFilter.py
shreyans800755: > Do you need the section actually for filters?

This is how `ConsoleInteraction.show_bears` takes parameters (with sections as keys of dict).

coalib/parsing/filters/CanDetectFilter.py
shreyans800755: Exactly. I had thought of making classmethod/staticmethod, but it didn't work with the obvious reason you specified.

coalib/parsing/filters/LanguageFilter.py
shreyans800755: It will be `bears` and not `bears.values()` as I'm going through keys and not values for all three filters.

shreyans800755:
@Makman2 

> The base class could be inlined into FilterHelper, so we don't need to override anything any more and we can use pure functions.

I think, this way, we've got more restrictions and proper structure. If someone wants to add a filter, one just have to inherit from FilterBase and implement abstract method.
This way, I've tried to follow strategy design pattern, since we've got different implementation of filters and can choose one at a time to get the output.
IMO, by inlining FilterBase into FilterHelper, it would be highly coupled, and if any changes come into filters structure, it might become more difficult to incorporate new structure. IMO, losely coupled approach is better and more maintainable.

Makman2:
> and if any changes come into filters structure

You have something in mind that might change in future? :D
What I only have in mind is that with the next-gen core there will be no distinction between Local and GlobalBears any more^^

Anyway, sounds fine :D

coalib/parsing/DefaultArgParser.py
```diff
@@ -187,6 +187,10 @@ def default_arg_parser(formatter_class=None):
         help='filters `--show-bears` by the given languages')
 
     outputs_group.add_argument(
+        '--filter-by', action='append', nargs='+',
+        help='filter --show-bears by the filter given as argument.')
```
Makman2: Maybe we should list available filters here or somewhere else and refer?

coalib/parsing/filters/CanDetectFilter.py
Makman2: ugh ... can we abstract the dict-building away into the `FilterBase` so we just we have to filter pure bears and don't need to build the dict? I think we are likely duplicating code for new filters with this^^
```python
def filter_bears(self, bears, args):
        local_bears, global_bears = bears
        filtered_local_bears = {section: self.filter_section_bears(bears) for section, bears in local_bears)
        filtered_global_bears = {section: self.filter_section_bears(bears) for section, bears in global_bears)
        return filtered_local_bears, filtered_global_bears
```

Makman2: And could you rename the `bears` parameter, to something like `sections_and_bears` or so? Just `bears` indicates that it's a pure collection of bears only :)

coalib/parsing/filters/FilterBase.py
```diff
@@ -0,0 +1,20 @@
+from abc import ABC, abstractmethod
+
+
+class FilterBase(ABC):
+
+    def get_filtered_bears(self, bears, args):
+        if args:
+            bears = self.filter_bears(bears, args)
+        return bears
```
Makman2: up to you: I think it's not really necessary to have a function `get_filtered_bears` which just checks for the presence of arguments and then invokes another function in that case. I think it's simple enough to inline it into `filter_bears`. External calls look also a bit nicer with `filtered = filter_bears(...)` instead of `filtered = get_filtered_bear(...)` (which indicates a bit for me that the bears were already filtered and stored somewhere :D)

coalib/coala.py
```diff
@@ -59,26 +60,45 @@ def main(debug=False):
...
+
+            local_bears, global_bears = filtered_bears
             show_bears(local_bears,
                        global_bears,
                        args.show_description or args.show_details,
                        args.show_details,
                        console_printer)
-
             return 0
+
```
Makman2: unnecessary change :)

coalib/coala.py
```diff
@@ -59,26 +60,45 @@ def main(debug=False):
...
+                        console_printer.print(ex)
+                        return 2
+
+            local_bears, global_bears = filtered_bears
             show_bears(local_bears,
                        global_bears,
                        args.show_description or args.show_details,
                        args.show_details,
                        console_printer)
-
```
Makman2: unnecessary change :)

coalib/coala.py
```diff
@@ -59,26 +60,45 @@ def main(debug=False):
...
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead")
+                args.filter_by = args.filter_by or []
+                args.filter_by.append(['language'] + args.filter_by_language)
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x, *filter_args = filter_by
```
Makman2: wahhhh this is possible? I haven't known that there's such a syntax! :O

coalib/coala.py
```diff
@@ -59,26 +60,45 @@ def main(debug=False):
...
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead")
+                args.filter_by = args.filter_by or []
+                args.filter_by.append(['language'] + args.filter_by_language)
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
```
Makman2: what if `filter-by` is empty? I think it would crash below because of an unpacking error^^

coalib/settings/ConfigurationGathering.py
```diff
@@ -3,8 +3,7 @@
 import sys
 import logging
 
-from coalib.collecting.Collectors import (
-    collect_all_bears_from_sections, filter_section_bears_by_languages)
+from coalib.collecting.Collectors import (collect_all_bears_from_sections)
```
Makman2: no braces necessary around import :D

shreyans800755:
> You have something in mind that might change in future?

I don't have anything in mind at the moment. But with the changes in core design through aspects and all, I though there be some change in future for filters too. Just thinking out loud, and not much consolidated idea.

coalib/parsing/DefaultArgParser.py
shreyans800755: Write now,
`coala -B --filter-by` gives default error of `Usage : all the possible options...`.
But `coala -B --filter-by unknown` gives the following error:
`'unknown' is an invalid filter. Available filters: can_detect, can_fix, language`

IMO, this is sufficient. We don't need to show which filters are available in former query.

coalib/coala.py
shreyans800755: Kudos to @userzimmermann . He suggested this at https://github.com/coala/coala/pull/4240#discussion_r122594079.

coalib/coala.py
shreyans800755: Arg parser takes care of this.
`coala -B --filter-by` gives error thrown by arg parser with `Usage` message.

coalib/parsing/DefaultArgParser.py
Makman2: Though sometimes people don't use the feature when they don't know whether typing something wrong gives a better message (and I would count to those people).

Makman2: If I get immediate help, then I immediately use that :D

shreyans800755: @Makman2 That makes sense. Currently there's no direct way of knowing which filters are available unless user tries to filter bears with some wrong filter name.
One solution I can think of is to add one more option called `--show-filters` in `--show-bears` which will display all the available filters.
I'm confused b/w two of the following options:
`coala -B --show-filters`
or
`coala -B show-filters`

What's your opinion ?

Makman2: why not embedding that into the argparser help? You can just fetch available filters from `FilterHelper` and construct a message :)

shreyans800755:
@Makman2 
Added filters in help message.

One analysis (a bit out of context for this PR):
I pushed my changes here, and got few errors from gitmate-bot regarding PEP8 styling (Sadly I forgot to do it in local before pushing). But my travis-ci job also failed due to this reason. Build and tests succeeded, but the status of build was **red**. That's because of the command `coala --non-interactive` run in travis. I'm curious if its really required to do the same thing multiple times at different places. IMO, job of travis should be only limited to test the code. (Appveyor passed successfully though, as it doesn't have this command)

Makman2:
>  I'm curious if its really required to do the same thing multiple times at different places.

Yeah you are right, if that's the case, we should encourage to optimize our build times. Feel free to file an issue :+1:

> Apart from that, Circle CI is giving time out. I've faced this situation multiple times. Although it passes on rebuild. Is there any issue filed for this ?

Do you have a link to a timeouted build?

coalib/coala.py
```diff
@@ -59,11 +60,35 @@ def main(debug=False):
...
-                args.filter_by_language, log_printer)
-
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead")
```
Makman2: missing period at end of sentence :D

coalib/coala.py
```diff
@@ -59,11 +60,35 @@ def main(debug=False):
...
+                args.filter_by = args.filter_by or []
```
Makman2: isn't this one always a list because of the argument parser?
Maybe you can omit this statement?

coalib/coala.py
```diff
@@ -59,11 +60,35 @@ def main(debug=False):
...
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x, *filter_args = filter_by
+                    if len(filter_args) == 0:
+                        continue
+                    try:
+                        filtered_bears = FilterHelper.get_filtered_bears(
+                            filter_by_x, filter_args, filtered_bears)
```
Makman2: This is just a design consideration, you don't have to implement it (it's actually not so performant, just for the concept).

Mathematically speaking, because we apply a filter on a set of bears, and especially in a sequence, the invocation order of filters may define different results. A filter usually applies to a single bear, so we don't need to pass a filter the complete set of bears to filter. To ensure that the order of invocation and actual set of bears is independent from the result, we could invoke a single function that just says "yeah this bear shall be taken into the results -> True" or "no the bear should be filtered out -> False".
This kind of principle btw exists in many languages (also in Python with list comprehensinos), they usually take in a lambda that defines when an object shall be kept or not with a boolean value.

coalib/coala.py
```diff
@@ -59,11 +60,35 @@ def main(debug=False):
...
+                    except NotImplementedError:
+                        logging.exception(filter_by_x +
+                                          ' filter is not implemented')
```
Makman2: Missing period :)

coalib/parsing/DefaultArgParser.py
```diff
@@ -187,6 +188,12 @@ def default_arg_parser(formatter_class=None):
         help='filters `--show-bears` by the given languages')
 
     outputs_group.add_argument(
+        '--filter-by', action='append', nargs='+',
+        help='filters `--show-bears` by the filter given as argument.'
+             'Available filters: {}'.format(
```
Makman2: missing space in string after `argument.` / before `Available` :)
Renders as
```
... by the filter given as argument.Available filter: ...
```
currently

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class FilterHelper:
+    all_filters = {'language': LanguageFilter,
+                   'can_detect': CanDetectFilter,
+                   'can_fix': CanFixFilter}
+
+    @classmethod
+    def get_all_filters_str(cls, sep=', '):
+        return sep.join(sorted(cls.all_filters))
```
Makman2: I think you should inline this into the `argparse` code above, imo this is usually application specific code, having a function here is not so modular imo :+1:

coalib/coala.py
```diff
@@ -59,11 +60,35 @@ def main(debug=False):
...
+                    "Use '--filter-by language ...' option instead")
+                args.filter_by = args.filter_by or []
+                args.filter_by.append(['language'] + args.filter_by_language)
+            if args.filter_by:
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x, *filter_args = filter_by
+                    if len(filter_args) == 0:
+                        continue
```
Makman2: you already handle this case in `FilterHelper`, I think you can safely omit it here :)

coalib/coala.py
shreyans800755: If args.filter_by is None, then it will be empty list after this statement.

This is to cover the following case:
`coala -B --filter-by-language java`

In this case, we first build `args.filter_by` from `args.filter_by_language`. And then process it, to make it simple. And it will also make it easier to remove `if args.filter_by_language:` in future as it will be deprecated after adding this generic filter processor.

coalib/coala.py
shreyans800755: Here, the invocation order of filters won't give different results.
If I understood your comment correctly,
In your opinion:
`coala -B --filter-by language c --filter-by can_fix syntax`
and
`coala -B --filter-by can_fix syntax --filter-by language c`
can give different results.

But, as per the implementation, the result is nothing but a set of bears containing intersection of all the bears filtered out from each filters individually. However, implementation is not directly taking intersection, but it doing something like folding. But the essence is same. We just want to show the intersection of all the filters.

Correct me if I'm wrong.

coalib/parsing/FilterHelper.py
shreyans800755: Its also being used in `FilterHelper.get_help_message` function. So, I think it makes sense not to repeat it at bother places.

shreyans800755:
@Makman2 

> Do you have a link to a timeouted build?

https://circleci.com/gh/coala/coala/13182

coalib/coala.py
Makman2: Can't you set a default for `--filter-by` to use an empty list?

coalib/parsing/FilterHelper.py
Makman2: Hmm. Okay that makes sense.

Although I imagine when I would now use your Filter stuff, I actually wouldn't look into your class when I need to build some help strings ;) Especially I wouldn't understand what they are exactly producing as output and where they are especially useful for, and then I'm usually faster writing some quick expression myself (thank you Python :D), using `all_filters`.

Makman2:
Ahhh... `language-check`... We have an issue for that (can't find it right now). It was always a pain in our builds ;D

coalib/coala.py
shreyans800755: I can set default value while adding it in `DefaultArgParser.py`. But this is causing test failures in the following tests:
`tests/output/ConfWriterTest.py::ConfWriterTest::test_append`
`tests/output/ConfWriterTest.py::ConfWriterTest::test_write`

Because it assumes that config doesn't have `filter_by` option as it is not specified in this test.

Similarly, following two are also failing:
tests/settings/ConfigurationGatheringTest.py::ConfigurationGatheringTest::test_back_saving
tests/settings/ConfigurationGatheringTest.py::ConfigurationGatheringTest::test_gather_configuration

shreyans800755: We can fix it by adding `filter_by` attribute though.
IMO, its better to keep it this way, since this statement is anyways part of condition which is to be deprecated.

Makman2: Yeah in this case let's keep using `None` :+1:

Makman2: Ah maybe to make it more clear -->
```
args.filter_by = [] if args.filter_by is None else args.filter_by
```
Up to you, I'm also fine with the short-circuitance-using expression :)

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,38 @@
+from coalib.output.printers.LogPrinter import LogPrinter
+
+from coalib.parsing.filters.LanguageFilter import LanguageFilter
+from coalib.parsing.filters.CanDetectFilter import CanDetectFilter
+from coalib.parsing.filters.CanFixFilter import CanFixFilter
+
+
+class FilterHelper:
+    all_filters = {'language': LanguageFilter,
```
Makman2: hm just thinking, maybe `available_filters` is a better name? What do you think?

coalib/coala.py
shreyans800755: Good enough. Makes it more readable.

coalib/coala.py
userzimmermann: @Makman2 That's Python 3 ;)

coalib/coala.py
```diff
@@ -59,11 +60,34 @@ def main(debug=False):
...
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead.")
+                args.filter_by = [] if args.filter_by is None else \
+                    args.filter_by
```
userzimmermann: Continuation of https://github.com/coala/coala/pull/4240#discussion_r122601529

In this solution the `else` part is obsolete. Then it would be better to keep:

```python
if args.filter_by is None:
    args.filter_by = []
```

But IMO it's still sufficient enough to just do:

```python
args.filter_by = args.filter_by or []
```

If it's already an empty `list` it doesnt' really hurt ;)

shreyans800755: `args.filter_by = args.filter_by or []` is definitely sufficient. Main concern is readability. I think, as you mentioned,
`if args.filter_by is None:
    args.filter_by = []`
is better option in terms of readability.

Makman2: hm the two-liner if is also fine with me, actually it's slightly more performant^^ But up to you @shreyans800755

coalib/coala.py
```diff
@@ -59,11 +60,34 @@ def main(debug=False):
...
+                args.filter_by.append(['language'] + args.filter_by_language)
```
Makman2: shouldn't it be `extend`/`+=`? `append` would place the list itself as an item into `args.filter_by`, so in the end you have
`[['language', 'python']]`

Makman2: I'm not sure at least that `argparse` does it like this^^

coalib/coala.py
```diff
@@ -59,11 +60,34 @@ def main(debug=False):
...
+                    filter_by_x, *filter_args = filter_by
+                    try:
+                        filtered_bears = FilterHelper.get_filtered_bears(
+                            filter_by_x, filter_args, filtered_bears)
+                    except NotImplementedError:
+                        logging.exception(filter_by_x +
+                                          ' filter is not implemented.')
+                        return 255
+                    except AttributeError as ex:
+                        console_printer.print(ex)
```
Makman2: use `logging.exception` instead

Makman2: When did this error occur again? I already forgot... maybe you should place a comment :/

coalib/coala_modes.py
```diff
@@ -61,8 +59,10 @@ def mode_json(args, debug=False):
 
     if args.show_bears:
         try:
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, LogPrinter())
+            from coalib.parsing.FilterHelper import FilterHelper
+
+            local_bears, global_bears = FilterHelper.get_filtered_bears(
+                'language', args.filter_by_language)
```
Makman2: also here you are using the old setting again^^

coalib/coala.py
Makman2: ahhh you raise an `AttributeError` for non-existing filters. I would consider implementing a new exception, like `InvalidFilterException`. An attribute error is so general that it might occur everywhere, and possible bugs would slip through.

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,38 @@
...
+        if cls.is_valid_filter(filter):
+            filter_object = cls.available_filters[filter]()
+            return filter_object.filter_bears(all_bears, args)
+
+        raise AttributeError(cls.get_help_message(filter))
+
+    @classmethod
+    def get_help_message(cls, filter):
+        return '{!r} is an invalid filter. Available filters: {}'.format(
+            filter, cls.get_all_filters_str())
```
Makman2: as you only use this message inside the error, I would rather inline it. Nobody would call `get_help_message` again ;)

coalib/coala.py
shreyans800755: If we use `extend`, it would be:
`['language', 'python']` and not `[['language', 'python']]`. `extend` essentially appends all the elements present in the list given as argument. But, we want to append it as list and not as individual elements.

Regarding, `+=`, it actually creates new object. Since, `args.filter_by` is not our own variable, and its created by `argparse`, its not advised to change its reference. It might to lead to unexpected behavior. And there's no advantage of it.

tests/parsing/FilterTest.py
```diff
@@ -0,0 +1,128 @@
...
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_filter_by_unknown(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--filter-by', 'unknown', 'arg1')
+            self.assertEqual(retval, 2)
+            self.assertEqual(
+                stdout, FilterHelper.get_help_message('unknown') + '\n')
```
Makman2: I would rather write the messages out, testing by expected output, not by expected algorithm behaviour ;)

(applies everywhere in this test ;) )

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_filtered_bears(cls, filter, args, all_bears=None):
+        if all_bears is None:
+            from coalib.settings.ConfigurationGathering import (
+                get_all_bears)
+            all_bears = get_all_bears(LogPrinter())
+        if not args or len(args) == 0:
+            return all_bears
+        if cls.is_valid_filter(filter):
+            filter_object = cls.available_filters[filter]()
+            return filter_object.filter_bears(all_bears, args)
```
Makman2: have you read my comment before? To execute filters for each section, so the users don't need to handle them?
For filter writers it would be really confusing, e.g. I would really ask myself "why do I need this?", and especially "what shall I do with it?"

coalib/coala.py
Makman2: > Regarding, `+=`, it actually creates new object.

Nope it doesn't, it's in-place :)

```
>>> a=[1,2,3]
>>> id(a)
140015517382472
>>> a+=[4,5]
>>> id(a)
140015517382472
```

But okay so having a twice-nested list is intended. Does argparse create the same, e.g. when doing
```
--filter-by language python C --filter-by can_fix all
```
Does it create `[['language', 'python', 'C'], ['can_fix', 'all']]`?

Makman2:
Some relevant discussion in gitter, just for reference:
> @shreyans800755 hm, regarding filtering, I've been thinking again about using a function that returns `True`/`False` instead of the completely filtered set
and I'm quite tempted to it, because for example the CAN_FIX filter, would decrease from 5 lines to 1!
> ```
> return any(fix.lower() in bear.CAN_FIX for fix in args)
> ```
>
> instead of
> ```
>        fixes = {fix.lower() for fix in args}
>        return {section:
>                tuple(bear for bear in bears[section]
>                      if {fix.lower()
>                          for fix in bear.CAN_FIX} & fixes)
>                for section in bears}
> ```
> (oh it's even 6 lines --> 1) :3
> or to keep your intersection behaviour:
> ```
> return bool({fix.lower() for fix in args} & bear.CAN_FIX)
> ```
> (trivia: both 1-liner versions perform differently regarding performance, depending on the number of arguments passed and whether they are in bear.CAN_FIX or not, just tested it :wink: )
> @shreyans800755 using those functions would make those filter also way easier to test. Then you just need to make a few little tests invoking coala, so you can see whether filter functionality is correctly executed, and then you can battle-test the filters separately
> \+ I'm still not very convinced using a FilterBase class :confused: filters don't hold a state (at least not now), so classes would be not the right design choice imo. I would agree on keeping them now, but I would raise an issue discussing that again :wink: (fyi :smiley:)
> especially there's a division-of-work problem, what does the FilterHelper, FilterBase and our coala implementation do? What are there duties?
when I would write now a filter, I would really wonder why I need to use base class in python for that. Usually I would expect to write a simple lambda or so^^
instead of classes you could maybe use decorators, just as an idea for later :thumbsup:

coalib/coala.py
```diff
@@ -74,11 +98,8 @@ def main(debug=False):
         elif args.show_capabilities:
             from coalib.collecting.Collectors import (
                 filter_capabilities_by_languages)
-            from coalib.settings.ConfigurationGathering import (
-                get_filtered_bears)
-
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, log_printer)
+            local_bears, global_bears = FilterHelper.get_filtered_bears(
+                'language', args.filter_by_language)
```
shreyans800755: Analyzing it properly, it should ideally be `args.show_capabilities` since languages are given as the arguments of option `show_capabilities` as per tests.
So, `coala -p java` works fine. No need of `coala -p -l java`.

coalib/parsing/FilterHelper.py
shreyans800755: I've already removed a unnecessary function from `FilterBase`. Anyways, this is not something filter writers have to worry about. They just gotta implement `filter_section_bears` function for the new filter. This is part of helper class and not base class of filters.

shreyans800755:
@Makman2 
Regarding usage of decorator, As far as I can remember, I had suggested this alternative, but moved further with this approach. But definitely worth giving a though. Like you said, may be we should discuss it over some other issue.
Apart from that, there are multiple parameters involved in here, like local/global bears, sections, etc. Some of which are to be deprecated in future, and will be faded away. Due to these parameters, design is quite complex at the moment as these require more and more processing at each and every step.

shreyans800755:
@Makman2 I had not really thought of using boolean values for filters. That's actually a good idea. I wasn't aware of the function `any()` :smile:

coalib/coala_modes.py
shreyans800755: I think we need a flag to get the list of languages apart from `filter_by_language` (`-l`). May be `--language`. Or we gotta change the meaning of `-l` to language, as filter_by_language doesn't always make sense.

This condition is not making sense to me (`if args.show_bears`). We are calling the function `mode_json` from coala if `args.json` is true. IMO, we should do processing using normally, and at the time of dumping data to stdout (using console_printer), we should check if the output format is chosen by user to some specific format(`json` in this case).
This seems like repeatation of code at the moment.

shreyans800755: At the moment, there's no other way of providing languages apart from `filter_by_language`. So, to use `filter_by` over here, we need to repeat the same logic as present in coala.py

coalib/coala.py
shreyans800755: Yeah, argparse create the list as you mentioned.

shreyans800755:
From gitter at https://gitter.im/coala/coala?at=5955421c11755ab05636c3d0:

> @Makman2 May be a separate issue can be filed, which can make these objects light weight, and only contain necessary info like name, description, etc. It will also improve the speed. A function similar to get_all_bears, but has only necessary info, and we can make it independent of local/global bears as that's not a concern in most functionalities like filter or just showing the bear's info.
Apart from that, may be one more filter can be added to filter bears by name. One use case is that, after getting name from web or somewhere else, user should be able to get description on cli. Something like coala -B --show-description --filter-by name cppcheck would give the description of cppcheckbear. Just an idea. Not sure whether it should be part of this PR or not.

shreyans800755:
@Makman2 
Removed inheritance approach, and added decorator approach. Also shifted main logic in the decorator, so that it need not be repeated whenever we want to add a new filter.
Just one concern I can see here:
Right now, filter function looks like this: `xxx_filter(bears, args)`. I couldn't enforce the return type to be `bool` for any filter. This will not cause any big issues as it will be the job of the developer writing new filters, and it will never reach to end user. But its just something good to have IMO. (Or may be I'm just using too much of static typed languages(C++) :smile: )
Let me know if you've any suggestions for this.

Makman2:
It's Python, we are all adults here, don't enforce types ;) (if it's really not necessary ;D)

Actually I also like type safety, in many cases it's - yeah- just safe :D

coalib/coala_modes.py
Makman2: hmm kk. Let's file an issue for that after merge to find some neat idea without duplicating code :+1:

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,37 @@
...
+            raise InvalidFilterException('{!r} is an invalid filter. '
+                                         'Available filters: {}'.format(
+                                             filter, cls.get_all_filters_str()))
+        if not args or len(args) == 0:
+            return all_bears
+        return cls.available_filters[filter](all_bears, args)
+
+
+class InvalidFilterException(Exception):
+    pass
```
Makman2: We should put the exception into an own file ;)

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,14 @@
...
+def can_detect_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
+
+    :param bear:        Bear object
+    :param args:        Set of detects on which bear is to be filtered
+    :return:            True if this bear is to be displayed for given
+                        args, False otherwise
+    """
+    return {detect.lower() for detect in bear.CAN_DETECT} & args
```
Makman2: not sure, maybe we should explicitly "boolify" the value?
```
return bool({detect.lower() for detect in bear.CAN_DETECT} & args)
```
I don't know. Up to you. There's actually no real reason to do this, except that it's maybe slightly better readable.

shreyans800755: Its not really required, as this function is used as if condition. So, it is by default converted to bool. Still I added this, just to make it more readable as you mentioned.

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+@filter
+def can_fix_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
+
+    :param bear:        Bear object
+    :param args:        Set of fixes on which bear is to be filtered
+    :return:            True if this bear is to be displayed for given
+                        args, False otherwise
```
Makman2: missing periods at end of sentences :)
\+
Enclose True and False into double backticks, as these are code terms ;)
(applies for all filter docstrings)

coalib/parsing/filters/Filter.py
```diff
@@ -0,0 +1,15 @@
+def filter(filter_applier):
+    def filter_wrapper(bears, args):
+        args = {arg.lower() for arg in args}
+        local_bears, global_bears = bears
+        local_bears = filter_section_bears(local_bears, args)
+        global_bears = filter_section_bears(global_bears, args)
+        return local_bears, global_bears
+
+    def filter_section_bears(bears, args):
```
Makman2: maybe put `filter_section_bears` before `filter_wrapper`? As you use it above "before it's defined" regarding reading order. Up to you :+1:

coalib/parsing/filters/Filter.py
```diff
@@ -0,0 +1,15 @@
+def filter(filter_applier):
+    def filter_wrapper(bears, args):
```
Makman2: use `functools.wraps` to make the old function signature appear for debugging purposes

shreyans800755:
@Makman2 Pushed changes as requested.

coalib/parsing/FilterHelper.py
```diff
@@ -0,0 +1,35 @@
+from coalib.output.printers.LogPrinter import LogPrinter
+
+from coalib.parsing.InvalidFilterException import InvalidFilterException
+
+from coalib.parsing.filters.LanguageFilter import language_filter
+from coalib.parsing.filters.CanDetectFilter import can_detect_filter
+from coalib.parsing.filters.CanFixFilter import can_fix_filter
+
+
+class FilterHelper:
```
Makman2: hm kinda interesting: You never use an instance of this class :D (makes also no sense as there are no class members :D)

Up to you: Maybe moving all functions to module scope^^

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+@filter
+def can_detect_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
+
+    :param bear:        Bear object.
+    :param args:        Set of detects on which bear is to be filtered.
+    :return:            ``True`` if this bear is to be displayed for given
+                        args, ``False`` otherwise.
```
Makman2: please align docs without leaving so much space :+1:

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+@filter
+def can_fix_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
+
+    :param bear:        Bear object.
+    :param args:        Set of fixes on which bear is to be filtered.
+    :return:            ``True`` if this bear is to be displayed for given
+                        args, ``False`` otherwise.
```
Makman2: Same here with alignment

coalib/parsing/FilterHelper.py
shreyans800755: I think this way it looks neater.

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,14 @@
+from coalib.parsing.filters.Filter import filter
+
+
+@filter
+def can_fix_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
```
Makman2: copy'n'paste worm :D

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,14 @@
+from coalib.parsing.filters.Filter import filter
+
+
+@filter
+def can_detect_filter(bear, args):
+    """
+    Filters the bears by ``LANGUAGES``.
```
Makman2: here also, copy'n'paste worm :D

coalib/parsing/filters/CanFixFilter.py
shreyans800755: Ugh...

coalib/coala.py
```diff
@@ -59,11 +60,31 @@ def main(debug=False):
...
-                args.filter_by_language, log_printer)
-
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
+                    "Use '--filter-by language ...' option instead.")
```
Makman2: Use '--filter-by language ...' ~~option~~ instead.

coalib/coala.py
```diff
@@ -59,11 +60,31 @@ def main(debug=False):
...
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, log_printer)
-
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
+                log_printer.warn(
+                    "'--filter-by-language ...' option is deprecated. "
```
Makman2: '--filter-by-language ...' ~~option~~ is deprecated.

coalib/coala.py
```diff
@@ -59,11 +60,31 @@ def main(debug=False):
...
+                # Each iteration of the following loop applies
+                # filters one by one provided as arguments
+                for filter_by in args.filter_by:
+                    filter_by_x, *filter_args = filter_by
+                    try:
+                        filtered_bears = FilterHelper.get_filtered_bears(
+                            filter_by_x, filter_args, filtered_bears)
+                    except InvalidFilterException as ex:
+                        # If filter is not available
+                        logging.exception(ex)
```
Makman2: Does this show a traceback to the user? As it's common that a user might mistype a filter name, we only should display the exception message.

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+    :param bear:        Bear object.
+    :param args:        Set of detects on which bear is to be filtered.
```
Makman2: `bear` inside double backticks, as you refer to the parameter.

coalib/parsing/filters/CanDetectFilter.py
Makman2: > ``True`` if this bear is to be displayed for given args, ``False`` otherwise.

Maybe rephrase this into something more general^^

> ``True`` if this bear matches the criteria inside ``args``, ``False`` otherwise.

Up to you :+1:

Makman2: Applies to all filters :+1:

coalib/parsing/filters/CanDetectFilter.py
Makman2: Maybe `Set of detectable issue types`? Up to you :+1:

coalib/parsing/filters/CanFixFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+    :param bear:        Bear object.
+    :param args:        Set of fixes on which bear is to be filtered.
```
Makman2: `bear` inside double backticks, as you refer to the parameter.

Makman2: Maybe `Set of fixable issue types`? Up to you again :+1:

coalib/parsing/filters/Filter.py
```diff
@@ -0,0 +1,19 @@
+from functools import wraps
+
+
+def filter(filter_applier):
```
Makman2: `filter_applier` -> `filter_function`

coalib/parsing/filters/Filter.py
```diff
@@ -0,0 +1,19 @@
...
+    def filter_section_bears(bears, args):
```
Makman2: you should rename the `bears` parameter into something that does fit more. It contains also section names as keys^^

coalib/parsing/filters/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+This package holds filter objects. Filter objects are used to
+get a list of bears with specific properties.
```
Makman2: maybe filter objects --> filter functions?

coalib/coala.py
shreyans800755: Yeah, it actually prints traceback to the user. I'll just use `console_printer.print` to print the message.

coalib/parsing/filters/Filter.py
shreyans800755: How about `section_to_bears_dict` or may be just `section_to_bears` ?

coalib/parsing/filters/CanDetectFilter.py
```diff
@@ -0,0 +1,15 @@
...
+@filter
+def can_detect_filter(bear, args):
+    """
+    Filters the bears by ``CAN_DETECT``.
+
+    :param bear:    Bear object.
+    :param args:    Set of detectable issue types on which ``bear`` is to be
+                    filtered.
+    :return:        ``True`` if this bear matches the criteria inside args,
+                    ``False`` otherwise.
```
Makman2: still indentation:
```
:param bear: Bear object.
:param args: Set of...
:return:     ``True`` if this bear...

Comment (Makman2) [inline coalib/parsing/filters/CanDetectFilter.py]: Same applies below

File: coalib/coala.py [hunk 79]
```diff
@@ -59,11 +60,31 @@ def main(debug=False):
...
-                get_filtered_bears)
-
-            local_bears, global_bears = get_filtered_bears(
-                args.filter_by_language, log_printer)
-
+                get_all_bears)
+            filtered_bears = get_all_bears(log_printer)
+            if args.filter_by_language:
+                # filter_by_language option should be deprecated soon.
+                # Since it is already covered by filter_by option.
```
sils: this comment doesn't really add info to what's in the warning below so IMO it can just be removed

tests/coalaTest.py
```diff
@@ -185,7 +186,8 @@ def test_show_capabilities_with_supported_language_debug(self):
     def test_version_conflict_in_collecting_bears(self, import_fn, _):
         with bear_test_module():
             import_fn.side_effect = VersionConflict('msg1', 'msg2')
-            retval, stdout, stderr = execute_coala(coala.main, 'coala', '-B')
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '-l ')
```
Makman2: why the space after `-l`?

Makman2: btw what does `-l` mean?

shreyans800755: In fact, `-l`  is not required. Can't recall why I added it. Removing.

Makman2:
ack 498c731

Makman2:
Let's do it! :D
@rultor merge

Makman2:
@rultor merge

Issue #2006, opened 2016-04-13
Diff: Highlight missing \n correctly


Makman2:
This is what someone gets when no \n at EOF is provided (patch from the SpaceConsistencyBear):
![tmp](https://cloud.githubusercontent.com/assets/6023916/14480942/468e1432-012e-11e6-8224-7eb3c02606ac.png)

We need to adjust the diff output so it displays the unified diff correctly (maybe using a special character, or like git does, what's best :D).

sils1297:
see also http://bugs.python.org/issue2142

rmariano:
As agreed with the team and @sils1297 we should handle the edge case for the diff when there are no `\n` at EOF, by adding a message displaying `\ No newline at end of file`

```diff
diff --git a/coalib/bearlib/abstractions/Linter.py b/coalib/bearlib/abstractions/Linter.py
index 9be433c..d26fc5b 100644
--- a/coalib/bearlib/abstractions/Linter.py
+++ b/coalib/bearlib/abstractions/Linter.py
@@ -692,4 +692,4 @@ def linter(executable: str,

     _prepare_options(options)

-    return partial(_create_linter, options=options)
+    return partial(_create_linter, options=options)
\ No newline at end of file
```

Therefore, the idea would be to show both lines (the `-` and `+` equally), plus adding the message at the end.

Adrianzatreanu:
ping @rmariano

rmariano:
pong @Adrianzatreanu 
I was looking at it. Should I add an exception in coalib (I was thinking in the Diff objecct), or maybe only when it's shown?

Makman2:
What exception? :O

rmariano:
By "exception" I meant a particular rule for that edge case (not a Python exception you usually raise :-), a way to distinguish the diff, and how to show it.

Makman2:
I would say so

sils1297:
@rmariano any progress?

sils1297:
Unassigning @rmariano due to inactivity. If he's still working on this he can ping us later anyway and we can reassign.

jayvdb:
I dont believe this is difficulty/low . I would put it at medium at least.

aptrishu:
This seems to be a gitmate related issue. As it produced no error while testing it locally with no newline at the end of a file.

![spacebear](https://cloud.githubusercontent.com/assets/22702242/19883340/ce24a88e-a039-11e6-88fe-46738bb6549b.png)

[...]

PR #4244, opened 2017-05-20
Diff.py: Add newlines to outward representation

### LINKED ISSUE #3834: diff: diff.add_lines doesn't append newline to the lines passed as a list
example:
```
diff = Diff([])
diff.add_lines(0, ["1", "2", "3"])
print(diff.unified_diff)
  --- 
  +++ 
  @@ -0,0 +1,3 @@
  +1+2+3

```
while, for `add_lines(line_br, list)` current implementation requires user to append new lines, we should check if the lines in the list endswith a `\n` if not append one implicitly

Opened via [gitter](https://gitter.im/coala/coala/?at=58b8338f1465c46a56d78e16) by @meetmangukiya

mrtes:
Closes #3834 

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -62,8 +62,7 @@ class Diff:
                     result.add_lines(a_index_1,
                                      file_array_2[b_index_1:b_index_2])
                 elif tag == 'replace':
-                    result.change_line(a_index_1+1,
-                                       file_array_1[a_index_1],
+                    result.modify_line(a_index_1+1,
                                        file_array_2[b_index_1])
                     result.add_lines(a_index_1+1,
                                      file_array_2[b_index_1+1:b_index_2])
@@ -424,7 +423,7 @@ class Diff:
             if change.add_after is not False:
                 result.add_lines(line_nr, change.add_after)
             if change.change is not False:
-                result.change_line(line_nr, change.change[0], change.change[1])
+                result.modify_line(line_nr, change.change[1])
 
         return result
 
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -292,13 +292,21 @@ class Diff:
         """
         Generates a unified diff corresponding to this patch.
 
+        Each change will be displayed on its own line. Additionally, the
+        unified diff preserves the EOF-state of the original file. This
+        means that the ``Diff`` will only have a linebreak on the last line,
+        if that was also present in the original file.
+
         Note that the unified diff is not deterministic and thus not suitable
         for equality comparison.
         """
-        return ''.join(difflib.unified_diff(
+
+        list_unified_diff = list(difflib.unified_diff(
             self._file,
             self._raw_modified(),
             tofile=self.rename if isinstance(self.rename, str) else ''))
+
+        return ''.join(self._generate_linebreaks(list_unified_diff))
 
     def __json__(self):
         """
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -616,3 +616,32 @@ class Diff:
         :param range: The range to delete.
         """
         self.replace(range, '')
+
+    @staticmethod
+    def _add_linebreaks(lines):
+        """
+        Validate that each line in lines ends with a
+        newline character and appends one if that is not the case.
+
+        :param lines: A list of strings, representing lines.
+        """
+
+        return [line
+                if line.endswith('\n')
+                else line + '\n'
+                for line in lines]
+
+    @staticmethod
+    def _generate_linebreaks(lines):
+        """
+        Validate that each line in lines ends with a
+        newline character and appends one if that is not the case.
+        Exception is the last line in the list.
+
+        :param lines: A list of strings, representing lines.
+        """
+
+        if lines == []:
+            return []
+
+        return Diff._add_linebreaks(lines[:-1]) + [lines[-1]]
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -27,6 +27,7 @@ class Diff:
         """
         self._changes = {}
         self._file = list(file_list)
+        self._original = self._generate_linebreaks(self._file)
         self.rename = rename
         self.delete = delete
 
@@ -244,10 +245,9 @@ class Diff:
         """
         Retrieves the original file.
         """
-        return self._file
+        return self._original
 
-    @property
-    def modified(self):
+    def _raw_modified(self):
         """
         Calculates the modified file, after applying the Diff to the original.
         """
@@ -278,6 +278,17 @@ class Diff:
         return result
 
     @property
+    def modified(self):
+        """
+        Calculates the modified file, after applying the Diff to the original.
+
+        This property also adds linebreaks at the end of each line.
+        If no newline was present at the end of file before, this state will
+        be preserved, except if the last line is deleted.
+        """
+        return self._generate_linebreaks(self._raw_modified())
+
+    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
@@ -286,8 +297,8 @@ class Diff:
         for equality comparison.
         """
         return ''.join(difflib.unified_diff(
-            self.original,
-            self.modified,
+            self._file,
+            self._raw_modified(),
             tofile=self.rename if isinstance(self.rename, str) else ''))
 
     def __json__(self):
@@ -432,7 +443,7 @@ class Diff:
         """
         return (self.rename is not False or
                 self.delete is True or
-                self.modified != self._file)
+                self.modified != self.original)
 
     def delete_line(self, line_nr):
         """
@@ -524,11 +535,12 @@ class Diff:
             if len(replacement) == len(linediff.change[1]) == 1:
                 raise ConflictError('Cannot merge the given line changes.')
 
+            # The following diffs are created from strings, instead of lists.
             orig_diff = Diff.from_string_arrays(linediff.change[0],
                                                 linediff.change[1])
             new_diff = Diff.from_string_arrays(linediff.change[0],
                                                replacement)
-            replacement = ''.join((orig_diff + new_diff).modified)
+            replacement = ''.join((orig_diff + new_diff)._raw_modified())
 
         linediff.change = (self._file[line_nr-1], replacement)
         self._changes[line_nr] = linediff
@@ -565,8 +577,8 @@ class Diff:
         """
         # Remaining parts of the lines not affected by the replace.
         first_part = (
-            self.original[range.start.line - 1][:range.start.column - 1])
-        last_part = self.original[range.end.line - 1][range.end.column - 1:]
+            self._file[range.start.line - 1][:range.start.column - 1])
+        last_part = self._file[range.end.line - 1][range.end.column - 1:]
 
         self.delete_lines(range.start.line, range.end.line)
         self.add_lines(range.start.line - 1,
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -23,7 +23,7 @@ class DiffTest(unittest.TestCase):
     def test_add_line(self):
         self.uut.add_line(0, 't')
         self.assertRaises(ConflictError, self.uut.add_line, 0, 't')
-        self.assertEqual(self.uut.modified, ['t', '1', '2', '3', '4'])
+        self.assertEqual(self.uut.modified, ['t\n', '1\n', '2\n', '3\n', '4'])
 
     def test_double_addition(self):
         self.uut.add_lines(0, ['t'])
@@ -125,11 +125,11 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.stats(), (4, 2))
 
     def test_modified(self):
-        result_file = ['0.1',
-                       '0.2',
-                       '1',
-                       '1.1',
-                       '3.changed',
+        result_file = ['0.1\n',
+                       '0.2\n',
+                       '1\n',
+                       '1.1\n',
+                       '3.changed\n',
                        '4']
 
         self.uut.delete_line(2)
@@ -138,10 +138,10 @@ class DiffTest(unittest.TestCase):
         self.uut.change_line(3, '3', '3.changed')
 
         self.assertEqual(self.uut.modified, result_file)
-        self.assertEqual(self.uut.original, self.file)
 
         self.uut.delete_line(len(self.file))
-        del result_file[len(result_file) - 1]
+        del result_file[-1]
+        result_file[-1] = result_file[-1].rstrip('\n')
         self.assertEqual(self.uut.modified, result_file)
 
         self.uut.delete_line(1)
@@ -154,6 +154,8 @@ class DiffTest(unittest.TestCase):
         self.assertTrue(self.uut)
         self.uut.delete_line(4)
         self.assertFalse(self.uut)
+        self.uut.modify_line(1, '1\n')
+        self.assertFalse(self.uut)
 
         # test if it works with tuples.
         uutuple = Diff(('1', '2', '3', '4'))
@@ -163,12 +165,14 @@ class DiffTest(unittest.TestCase):
         self.assertTrue(uutuple)
         uutuple.delete_line(4)
         self.assertFalse(uutuple)
+        uutuple.modify_line(1, '1\n')
+        self.assertFalse(uutuple)
 
     def test_addition(self):
         self.assertRaises(TypeError, self.uut.__add__, 5)
 
-        result_file = ['1',
-                       '2',
+        result_file = ['1\n',
+                       '2\n',
                        '2']
 
         other = Diff(self.file)
@@ -200,34 +204,34 @@ class DiffTest(unittest.TestCase):
         self.assertRaises(ConflictError, other.__add__, uut)
 
     def test_from_string_arrays(self):
-        a = ['q', 'a', 'b', 'x', 'c', 'd']
-        b = ['a', 'b', 'y', 'c', 'd', 'f']
+        a = ['q\n', 'a\n', 'b\n', 'x\n', 'c\n', 'd\n']
+        b = ['a\n', 'b\n', 'y\n', 'c\n', 'd\n', 'f\n']
         self.uut = Diff.from_string_arrays(a, b)
         self.assertEqual(self.uut.modified, b)
 
-        a = ['first', 'fourth']
-        b = ['first', 'second', 'third', 'fourth']
+        a = ['first\n', 'fourth\n']
+        b = ['first\n', 'second\n', 'third\n', 'fourth\n']
         self.uut = Diff.from_string_arrays(a, b)
         self.assertEqual(self.uut.modified, b)
 
-        a = ['first', 'fourth']
-        b = ['first_changed', 'second', 'third', 'fourth']
+        a = ['first\n', 'fourth\n']
+        b = ['first_changed\n', 'second\n', 'third\n', 'fourth\n']
         self.uut = Diff.from_string_arrays(a, b)
         self.assertEqual(self.uut.modified, b)
 
-        a = ['first', 'second', 'third', 'fourth']
-        b = ['first', 'fourth']
+        a = ['first\n', 'second\n', 'third\n', 'fourth\n']
+        b = ['first\n', 'fourth\n']
         self.uut = Diff.from_string_arrays(a, b)
         self.assertEqual(self.uut.modified, b)
 
-        a = ['first', 'second', 'third', 'fourth']
-        b = ['first_changed', 'second_changed', 'fourth']
+        a = ['first\n', 'second\n', 'third\n', 'fourth\n']
+        b = ['first_changed\n', 'second_changed\n', 'fourth\n']
         self.uut = Diff.from_string_arrays(a, b)
         self.assertEqual(self.uut.modified, b)
 
     def test_from_unified_diff_single_addition(self):
         source = ['single line']
-        target = ['single line', 'another line added']
+        target = ['single line\n', 'another line added']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1 +1,2 @@',
@@ -239,8 +243,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.modified, target)
 
     def test_from_unified_diff_single_deletion(self):
-        source = ['two lines', 'to be removed']
-        target = ['two lines']
+        source = ['two lines\n', 'to be removed']
+        target = ['two lines\n']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1,2 +1 @@',
@@ -252,8 +256,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.modified, target)
 
     def test_from_unified_diff_single_modification(self):
-        source = ['first', 'second']
-        target = ['only_first_changed', 'second']
+        source = ['first\n', 'second']
+        target = ['only_first_changed\n', 'second']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1,2 +1,2 @@',
@@ -266,8 +270,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.modified, target)
 
     def test_from_unified_diff_multiple_additions_different_orderings(self):
-        source = ['A', 'B', 'C']
-        target = ['A', 'Y', 'Z', 'B', 'C']
+        source = ['A\n', 'B\n', 'C']
+        target = ['A\n', 'Y\n', 'Z\n', 'B\n', 'C']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1,3 +1,5 @@',
@@ -281,8 +285,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.original, source)
         self.assertEqual(self.uut.modified, target)
 
-        source = ['A', 'B', 'C']
-        target = ['A', 'Y', 'Z', 'C']
+        source = ['A\n', 'B\n', 'C']
+        target = ['A\n', 'Y\n', 'Z\n', 'C']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1,3 +1,5 @@',
@@ -296,8 +300,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.original, source)
         self.assertEqual(self.uut.modified, target)
 
-        source = ['A', 'B', 'C']
-        target = ['Y', 'Z', 'C']
+        source = ['A\n', 'B\n', 'C']
+        target = ['Y\n', 'Z\n', 'C']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -1,3 +1,5 @@',
@@ -311,8 +315,8 @@ class DiffTest(unittest.TestCase):
         self.assertEqual(self.uut.original, source)
         self.assertEqual(self.uut.modified, target)
 
-        source = ['A', 'B', 'C']
-        target = ['A', 'B', 'C', 'Y', 'Z']
+        source = ['A\n', 'B\n', 'C']
+        target = ['A\n', 'B\n', 'C\n', 'Y\n', 'Z']
         diff = ['--- a/testfile',
                 '+++ b/testfile',
                 '@@ -3 +3,3 @@',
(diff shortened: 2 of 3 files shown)


coalib/results/Diff.py
```diff
@@ -286,9 +286,9 @@ def unified_diff(self):
         for equality comparison.
         """
         return ''.join(difflib.unified_diff(
-            self.original,
-            self.modified,
-            tofile=self.rename if isinstance(self.rename, str) else ''))
+                self.original,
+                self.modified,
+                tofile=self.rename if isinstance(self.rename, str) else ''))
```
Makman2: is this indentation necessary for this commit? -> New commit.
As it's a codestyle change only, rather omit this change completely as the previous version is also valid.

coalib/results/Diff.py
```diff
@@ -467,7 +467,6 @@ def add_lines(self, line_nr_before, lines):
         """
         if lines == []:
             return  # No action
-
```
Makman2: unnecessary codestyle change (and imo makes code less readable too ;D)

coalib/results/Diff.py
```diff
@@ -616,3 +615,30 @@ def remove(self, range):
         :param range: The range to delete.
         """
         self.replace(range, '')
+
+    def _add_linebreaks(self, lines):
+        """
+        Validate that each line in lines end with a
```
Makman2: end -> ends

coalib/results/Diff.py
```diff
@@ -616,3 +615,30 @@ def remove(self, range):
...
+        """
+
+        return [line
+                if line.endswith('\n')
+                else line + '\n'
+                for line in lines]
+
+    def _generate_linebreaks(self, lines):
+        """
+        Validate that each line in lines end with a
```
Makman2: end -> ends

coalib/results/Diff.py
```diff
@@ -244,10 +243,9 @@ def original(self):
         """
         Retrieves the original file.
         """
-        return self._file
+        return self._generate_linebreaks(self._file)
```
Makman2: definitely cache this! This is quite cost intensive if a user needs it several times. Either use `lru_cache(1)` or save it directly on instantiation inside a variable :+1:

coalib/results/Diff.py
```diff
@@ -278,18 +276,37 @@ def modified(self):
...
+        """
+        eof_linebreak = False
+
+        if self._file and self._file[-1].endswith('\n'):
+            eof_linebreak = True
+
+        file = [line.rstrip('\n') for line in self._modified()]
+
+        if eof_linebreak:
+            return self._add_linebreaks(file)
```
Makman2: why not using `generate_linebreaks` without r-stripping? `_add_linebreaks` seems to check itself when it has to append a newline^^

coalib/results/Diff.py
```diff
@@ -244,10 +243,9 @@ def original(self):
         """
         Retrieves the original file.
         """
-        return self._file
+        return self._generate_linebreaks(self._file)
 
-    @property
-    def modified(self):
+    def _modified(self):
```
Makman2: maybe call it `_raw_modified` or so? It's hard to tell where the difference is between this one and `modified` if you don't know about this `'\n'` conversion thing^^

coalib/results/Diff.py
```diff
@@ -278,18 +276,37 @@ def modified(self):
...
 
         Note that the unified diff is not deterministic and thus not suitable
         for equality comparison.
         """
-        return ''.join(difflib.unified_diff(
-                self.original,
-                self.modified,
+        list_unified_diff = list(difflib.unified_diff(
+                self._file,
+                self._modified(),
```
Makman2: why not using `self.original` and `self.modified`? Then you don't have to do `_generate_linebreaks`^^

coalib/results/Diff.py
```diff
@@ -413,7 +430,7 @@ def __add__(self, other):
             if change.add_after is not False:
                 result.add_lines(line_nr, change.add_after)
             if change.change is not False:
-                result.change_line(line_nr, change.change[0], change.change[1])
+                result.modify_line(line_nr, change.change[1])
```
Makman2: new commit

coalib/results/Diff.py
```diff
@@ -616,3 +615,30 @@ def remove(self, range):
         :param range: The range to delete.
         """
         self.replace(range, '')
+
+    def _add_linebreaks(self, lines):
```
Makman2: can be made a staticmethod

coalib/results/Diff.py
```diff
@@ -616,3 +615,30 @@ def remove(self, range):
...
+
+        :param lines: A list of strings, representing lines.
+        """
+
+        return [line
+                if line.endswith('\n')
+                else line + '\n'
+                for line in lines]
+
+    def _generate_linebreaks(self, lines):
```
Makman2: can be made a staticmethod

coalib/results/Diff.py
mrtes: No, I probably forgot to remove it after trying something. Will remove.

coalib/results/Diff.py
mrtes: Probably also something I forgot. Normally, I'm all for blank lines :)

coalib/results/Diff.py
mrtes: You are right. I didn't know about the `lru_cache` thing, seems handy. 
Which approach do you recommend?

coalib/results/Diff.py
mrtes: Thats a weird edge case with deletions.

Assume a file like this: `file = ['a\n', 'b']`
We delete the last line, resulting in `file = ['a\n']`. 

Since the "eof-state" (i.e. no new line at the end of the file) should be kept, the `\n` now be deleted.
This would require functionality my current functions do not have, and would likely be more complicated to implement:
lots of checks like where did the deletion happen and did we have a `\n` at the eof or not, and probably also some other edge cases. 
Stripping all new lines and then just choosing the correct function, based on the original "eof-state" is far easier. 

Actually, all these edge cases between add, modify and delete led me to just changing the files (or diffs) we hand out from this function, after all changes have been applied.

coalib/results/Diff.py
mrtes: Should I move all my modifications, that replace `change_line` with `modify_line` into the new commit, or just in functions that i have not altered in other ways?

coalib/results/Diff.py
mrtes: In a case with "eof-state" without `\n`, this would result in the last line of the original/changed file being on the same line in the diff, something like:
`original = ['1','2','3']`
`target = ['1','2','4']`
`diff = ['+++\n', '---\n', '@@ xxxxx @@\n', '1\n', '2\n', '-3+4']`

I found this weird, because each change should be on a new line.
So I have to run either `_generate_linebreaks` or `_add_linebreaks` anyway to avoid this problem.
(However, the latter is unnecessary, because it doesn't matter if the last line of the diff has a '\n' or not - the representation seems equal).

By using the un-linebreaked files for unified_diff creation and add linebreaks later, I avoid unnecessary calculations.

coalib/results/Diff.py
Makman2: Up to you ;)

coalib/results/Diff.py
Makman2: > Since the "eof-state" (i.e. no new line at the end of the file) should be kept, the \n now be deleted.

That's the question, do we wanna keep that? I'm for abandoning the EOF-state, the user needs to change the new last line if he wants to keep the old no-\n-at-EOF-state. As it's in fact a change of that line.

coalib/results/Diff.py
Makman2: ahhhh I think we still haven't fixed that for unified-diff that no \n at EOF gets handled correctly. There's somewhere an issue and even a PR lying around already for longer time... Could you place an explanatory comment? And maybe a link to the issue if you like :+1:

coalib/results/Diff.py
Makman2: > Should I move all my modifications, that replace change_line with modify_line into the new commit

:+1:

coalib/results/Diff.py
mrtes: Well, that's what I understood from the discussion in #3834.

If we do not need to keep the EOF-state, this change will be far less complex because we can just append linebreaks to everything.

coalib/results/Diff.py
mrtes: Anyway, the `Diff.unified_diff` functions is used nowhere in the core except for the json output.
The printing in `ShowPatchAction` is done by using the difflib version directly. 

Interestingly, I don't see the issue you describe in #2006, even with the unmodified coala version.
Probably due to how printing the diff is handled in `ShowPatchAction.apply`, or rather in the printer, that adds a linebreak anyway ;)

In the end the question is, if the newline stuff in the diff is necessary, or for which usecase.

Makman2: if someone wants to check if files do end with \n, we have to support that. Then we have a diff changing the last line from `XYZ` to `XYZ\n`.

mrtes:
Meh, messed up the commits somehow.

mrtes:
I implemented your suggestions @Makman2.
I have kept the EoF-State in modified/original currently, but we can still discuss how to deal with it.

mrtes:
@Makman2  you wanted to be pinged.

coalib/results/Diff.py
Makman2: > Assume a file like this: file = ['a\n', 'b']
> We delete the last line, resulting in file = ['a\n'].

I just meant this case. When a user supplies a text initially (i.e. via `__init__`) we shall preserve the the EOF-state. Only when we remove that line, we don't preserve state for "new last lines".

coalib/results/Diff.py
```diff
@@ -244,10 +245,9 @@ def original(self):
         """
         Retrieves the original file.
         """
-        return self._file
+        return self._original
 
-    @property
-    def modified(self):
+    def _raw_modified(self):
```
Makman2: actually do we still need this property? Maybe we should initially append newlines with `_generate_linebreaks` and consider this the `_original`/`_file` then?

Makman2: ugh nvm I thought total nonsense...

Makman2: Ah but maybe use in this function instead of `_file` -> `_original`? So then we don't need the real original `_file` any more^^

coalib/results/Diff.py
```diff
@@ -278,18 +278,42 @@ def modified(self):
...
+        file = [line.rstrip('\n') for line in self._raw_modified()]
+
+        if eof_linebreak:
+            return self._add_linebreaks(file)
+        return self._generate_linebreaks(file)
+
+    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
```
Makman2: leave a newline here, as the docs below are detailed documentation :+1:

coalib/results/Diff.py
Makman2: So we would copy your new `modified` code to `__init__`, initially linebreak-ify the originally and all patches are referring to this version then. We would need to patch `add_line`, `change_line` and so on to properly convert inputs then^^
What do you think?

mrtes: I thought keeping `_file` would be a good idea if a un-linebreak-ified version of the file is required somewhere. However, it seems it is not called anywhere outside of Diff, so we could ditch it.

Not sure I understand the other suggestion (or rather, the advantage of it, assuming we keep EOF-state). 
Initially adding the linebreaks will complicate the modification functions, because we have to check for addition/modification/delection locations to preserve the EOF-state.
However, if we ditch that too, and apply linebreaks everywhere (like you suggested earlier) it could be feasible.

We would essentially trade the linebreak-calculations in public data-revealing functions against a one-off linebreak addition on every change.

Makman2: Hmm... and now I'm lost on what way to take :D
Let's take yours then, it's definitely a valid approach and your work is then not thrown away just by changing the design ;)

btw: `s/originally/original` (damn sometimes I do some grammatical mistakes... that comes from always changing a sentence :3)

coalib/results/Diff.py
mrtes: Where should I add the newline?

mrtes: Nvm got it.

tests/results/DiffTest.py
```diff
@@ -626,3 +626,21 @@ def test_delete(self):
...
         self.assertEqual(self.uut.modified, [])
         self.uut.delete = False
+
+    def test_linebreaks(self):
+        diff = Diff(['1'])
+        lines = ['1', '2', '3']
+        appended = diff._add_linebreaks(lines)
+        skip_eof = diff._generate_linebreaks(lines)
+        empty = diff._add_linebreaks([])
+        empty_skip = diff._generate_linebreaks([])
```
Makman2: you don't need the diff instance, access the staticmethods from `Diff` instead of `diff` :+1:

tests/results/DiffTest.py
```diff
@@ -626,3 +626,21 @@ def test_delete(self):
...
+
+        for line in skip_eof[:-1]:
+            self.assertRegex(line, '\d\n')
+        self.assertRegex(skip_eof[-1], '\d(?!\n)')
+
+        for line in appended:
+            self.assertRegex(line, '\d\n')
+
+        self.assertEqual(empty, [])
+        self.assertEqual(empty_skip, [])
```
Makman2: you could maybe simplify this a bit by inlining the call made at definition of `empty`. Then those things are closer together which actually belong together ;) That makes it especially easier to follow (as you don't have to think about 4 test variables at the same time)
```python
self.assertEqual(Diff._add_linebreaks([]), [])
self.assertEqual(Diff._add_linebreaks([]), Diff._generate_linebreaks[])
```

tests/results/DiffTest.py
```diff
@@ -626,3 +626,21 @@ def test_delete(self):
         self.uut.delete = True
         self.assertEqual(self.uut.modified, [])
         self.uut.delete = False
+
+    def test_linebreaks(self):
```
Makman2: maybe you want two different test functions for two staticmethods? up to you^^

tests/results/DiffTest.py
```diff
@@ -626,3 +626,21 @@ def test_delete(self):
...
+        diff = Diff(['1'])
+        lines = ['1', '2', '3']
+        appended = diff._add_linebreaks(lines)
+        skip_eof = diff._generate_linebreaks(lines)
+        empty = diff._add_linebreaks([])
+        empty_skip = diff._generate_linebreaks([])
+
+        for line in skip_eof[:-1]:
+            self.assertRegex(line, '\d\n')
```
Makman2: why not asserting against a list? easier code + easier to read + less error prone I think
```
self.assertEqual(Diff._add_linebreaks(lines), ['1\n', '2\n', '3\n'])
```

coalib/results/Diff.py
Makman2: ping, that still looks like that this will do:
```
return self._add_linebreaks(self._raw_modified)
```

tests/results/DiffTest.py
mrtes: Hm, I like to separate Variable initialization from actual use, but I see your point. 

I would still avoid multiple `Diff._add_linebreaks([])` in favor of a variable empty, but maybe directly above the asserts?

tests/results/DiffTest.py
mrtes: Great idea, that didn't occur to me until now :+1:

tests/results/DiffTest.py
mrtes: Nvm, with the other changes it makes more sense this way.

coalib/results/Diff.py
```diff
@@ -278,18 +278,37 @@ def modified(self):
...
+        return self._generate_linebreaks(self._raw_modified())
+
+    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
 
+        Each change will be displayed on its own line. Additionally, the
+        unified_diff preserves the EOF-state of the original file. This
+        means that the Diff will only have a linebreak on the last line,
```
Makman2: Diff in double backticks so it gets rendered as code ;) Usually code references like these should be enclosed in double-backticks so people know that this is a legal identifier :+1:

coalib/results/Diff.py
```diff
@@ -278,18 +278,37 @@ def modified(self):
...
+        """
+        return self._generate_linebreaks(self._raw_modified())
+
+    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
 
+        Each change will be displayed on its own line. Additionally, the
+        unified_diff preserves the EOF-state of the original file. This
```
Makman2: I think `_` -> `-` is better, unified-diff is a valid term imo :)

coalib/results/Diff.py
```diff
@@ -432,7 +451,7 @@ def __bool__(self):
         """
         return (self.rename is not False or
                 self.delete is True or
-                self.modified != self._file)
+                self._raw_modified() != self._file)
```
Makman2: consider this:
```
d = Diff(['1', '2', '3'])
d.modify_line(1, '1\n')
```
This would effectively render no difference in `d.modified` before and after the change, but internally we say that something changed because we handle things without a \n.

coalib/results/Diff.py
```diff
@@ -523,12 +542,12 @@ def modify_line(self, line_nr, replacement):
         if linediff.change is not False and linediff.change[1] != replacement:
             if len(replacement) == len(linediff.change[1]) == 1:
                 raise ConflictError('Cannot merge the given line changes.')
-
+            # strings are used here instead of lists.
```
Makman2: don't understand this comment :)

mrtes: When a line is modified twice, it is possible for the contents to be merged.
If that is the case, in the lines below 545 new diffs are generated from the changes lines. Since these diffs are not generated from a list, but instead from a string, the elements of these diffs will be chars. 
`diff = Diff('hello')`
 => `diff._file = ['h','e','l','l','o']`
These will be added and formed to a new replacement line. 
Its quite interesting actually, but a bit confusing if you don't know these diffs are created from strings, so I thought I add a comment.

coalib/results/Diff.py
mrtes: Right.
I changed it to compare `self.modified ` with `self.original`.
Also added a test for this case.

coalib/results/Diff.py
mrtes: I removed the dash, because it is written "unified diff" in the summary line.

coalib/results/Diff.py
mrtes: I tried to clarify the comment in the current iteration.

coalib/results/Diff.py
```diff
@@ -523,12 +542,12 @@ def modify_line(self, line_nr, replacement):
         if linediff.change is not False and linediff.change[1] != replacement:
             if len(replacement) == len(linediff.change[1]) == 1:
                 raise ConflictError('Cannot merge the given line changes.')
-
```
Makman2: minor and up to you: wanna keep this blank line? I think it improves readability^^

coalib/results/Diff.py
```diff
@@ -278,18 +278,37 @@ def modified(self):
...
         Note that the unified diff is not deterministic and thus not suitable
         for equality comparison.
         """
-        return ''.join(difflib.unified_diff(
-            self.original,
-            self.modified,
+
+        list_unified_diff = list(difflib.unified_diff(
+            self._file,
+            self._raw_modified(),
```
Makman2: `_raw_modified` is a property.

coalib/results/Diff.py
```diff
@@ -278,18 +278,37 @@ def modified(self):
...
             tofile=self.rename if isinstance(self.rename, str) else ''))
 
+        return ''.join(self._generate_linebreaks(list_unified_diff))
```
Makman2: do we have to do `generate_linebreaks`? As you are using `self._file` and `self._raw_modified` the behaviour should be the same as before, shouldn't it?

coalib/results/Diff.py
mrtes: Currently `_raw_modified` is implemented as a function.
`modified`, on the other hand is a property.

Anyway, what would be the advantage of using it as a property? 
In this case it should not be used outside the class anyway and does not need to be stored safely as a field, because it has to be recalculated based on possible changes.

coalib/results/Diff.py
mrtes: Yes, but in case of a file without linebreaks , this would print all changes in one line.
`diff = Diff(['1','2'])`
`diff.modify_line(1,'2')`
`diff.modify_line(2,'1')`
`diff.unified_diff => '--- \n+++ \n@@ -1,2 +1,2 @@\n+2 1-2'`

In a case where only the last line does not have a linebreak it would look even weirder:

`diff = Diff(['1\n','2'])
diff.modify_line(1,'2')
diff.modify_line(2,'1')
diff.unified_diff => '--- \n+++ \n@@ -1,2 +1,2 @@\n-1\n 2+1'`

(I guess the ordering change is due to the unified diff being non-deterministic?)

mrtes: The usage of `Diff.from_unified_diff` for a unified diff that is created from an input file without linebreaks is currently broken, too.
(It is just never used)

Makman2: Ah so you actually repaired it? Maybe a separate commit for that, explaining the old broken behaviour. You could also file yourself an issue and fix it together :+1:

Makman2: (so the variable-change needs to be still here in this commit, but the `return ''.join(self._generate_linebreaks(...)` can be moved out into a new one imo)

coalib/results/Diff.py
Makman2: oh my bad have misread the diff. Nvm then, a function is totally fine :+1:

Actually there are no differences between functions in properties regarding behaviour, properties just impose different syntax features:
```
a.get_value()
# as property:
a.value

# using setters can be more idiomatic with properties:
a.set_value(3)
a.value = 3
```

coalib/results/Diff.py
mrtes: Nah, it isn't fixed...yet.
The linebreaks are a first step, but the diff still needs filenames for it to work D:
It would be best if we could access the filename of the original, although a placeholder would work too.
I guess that should really be in a new issue, though.

Still not sure the linebreak addition should be moved though, cause that would require me to change all related tests again ;(

Makman2: hmmm... is it that hard? :/
would be nice to have it in a separate commit :/
(`git reset HEAD~1` and `git add -p` are your friends :3)

mrtes: No actually it wasn't. Got it mixed up with `from_unified_diff`, which had lots of changes to tests.

tests/results/DiffTest.py
```diff
@@ -626,3 +626,17 @@ def test_delete(self):
...
+            ['1\n', '2\n', '3\n'])
+
+        self.assertEqual(Diff._add_linebreaks([]), [])
+
+    def test_generate_linebreaks(self):
+        self.assertEqual(
+            Diff._generate_linebreaks(['1', '2', '3']),
+            ['1\n', '2\n', '3'])
+
+        self.assertEqual(Diff._generate_linebreaks([]), [])
```
Makman2: ah actually could you add tests when `\n`s are already there? :3

Makman2: so something like `['1\n', '2\n', '3\n']

Makman2: and same for `test_add_linebreaks` :)

Makman2: oh and you may want to test a mixed sample:
```
['1\n', '2', '3\n']
```
or similar

coalib/results/Diff.py
```diff
@@ -278,16 +278,28 @@ def modified(self):
...
         Generates a unified diff corresponding to this patch.
 
         Note that the unified diff is not deterministic and thus not suitable
         for equality comparison.
         """
-        return ''.join(difflib.unified_diff(
-            self.original,
-            self.modified,
+
+        list_unified_diff = list(difflib.unified_diff(
```
Makman2: 1. you don't join
2. you don't return it

Tests do pass because they are executed PR wide, not on each commit :/ (would take to long anyway often)

coalib/results/Diff.py
```diff
@@ -432,7 +444,7 @@ def __bool__(self):
         """
         return (self.rename is not False or
                 self.delete is True or
-                self.modified != self._file)
+                self.modified != self.original)
```
Makman2: the tests did pass before when comparing `self._raw_modified()` and `self._file`, right? Could you please add a test testing the edge-case I mentioned last time? When modifying a line suddenly with `\n` characters that hadn't some before^^

tests/results/ResultTest.py
```diff
@@ -182,8 +182,13 @@ def test_json_diff(self):
...
+                                                         '-2\n'
+                                                         '-3\n'
+                                                         '+3_changed')
+
         JSONEncoder = create_json_encoder(use_relpath=True)
         json_dump = json.dumps(diff, cls=JSONEncoder, sort_keys=True)
         self.assertEqual(
-            json_dump, '"--- \\n+++ \\n@@ -1,3 +1,2 @@\\n 1-2-3+3_changed"')
+            json_dump,
+            '"--- \\n+++ \\n@@ -1,3 +1,2 @@\\n 1\\n-2\\n-3\\n+3_changed"')
```
Makman2: layout change -> separate commit or remove :)

Makman2: rather remove, the old one is fine imo ;)

Makman2:
You may like to create an issue for dd305f0, but up to you. I think it would be good to track that this is actually a bug that you discovered and fix now :+1:

tests/results/ResultTest.py
mrtes: Actually it is a content change (multiple \n), which renders the line longer than allowed.

Makman2: oh my bad :3
(why the heck I haven't seen this... /me is going to buy new eyes...)

coalib/results/Diff.py
Makman2: ping

Makman2: oops no bump, change was made below.

coalib/results/Diff.py
```diff
@@ -278,18 +278,32 @@ def modified(self):
...
-        return ''.join(difflib.unified_diff(
-            self.original,
-            self.modified,
+
+        list_unified_diff = list(difflib.unified_diff(
+            self._file,
+            self._raw_modified(),
             tofile=self.rename if isinstance(self.rename, str) else ''))
 
+        return ''.join(list_unified_diff)
```
Makman2: I would prefer if you keep the old code structure and just change the variables. Easier to understand for foreigners + you currently create a list of the diff, which is not necessary. Keeping the old structure automatically avoids this :+1:

tests/results/DiffTest.py
```diff
@@ -628,15 +630,30 @@ def test_delete(self):
...
 
         self.assertEqual(Diff._generate_linebreaks([]), [])
+
+        self.assertEqual(
+            Diff._generate_linebreaks(no_lb_eof),
+            no_lb_eof)
+
+        self.assertEqual(
+            Diff._generate_linebreaks(lb_eof),
+            lb_eof)
```
Makman2: I think those changes should go into the other commit introducing `generate_linebreaks` and `add_linebreaks`

(GitHub does display the same comments in different commits if they touch the same parts of code, sadly... not really fitting our review style^^)

coalib/results/Diff.py
```diff
@@ -278,16 +278,28 @@ def modified(self):
...
+
+    @property
     def unified_diff(self):
         """
         Generates a unified diff corresponding to this patch.
 
         Note that the unified diff is not deterministic and thus not suitable
         for equality comparison.
         """
+
```
Makman2: hmm let's be conform with everything else ;) so don't remove the newline.

Main problem is that those style changes are likely reverted again by someone else who's modifying this code. So we get many commits adding a newline, removing it again, adding it, and so on and so on.

mrtes: So should I remove it, or not?

Makman2: ah egh sorry I meant to remove it :3

Makman2: nvm :)

tests/results/DiffTest.py
```diff
@@ -138,10 +138,10 @@ def test_modified(self):
         self.uut.change_line(3, '3', '3.changed')
 
         self.assertEqual(self.uut.modified, result_file)
-        self.assertEqual(self.uut.original, self.file)
 
         self.uut.delete_line(len(self.file))
-        del result_file[len(result_file) - 1]
+        del result_file[-1]
+        result_file[-1] = result_file[-1].rstrip('\n')
```
Makman2: ah this is quite an edge case ... if the old version has no \n's and we delete the last line... though not really important for now, we should maybe document that case more properly later or change behaviour, don't know ;)

Makman2:
ack 0e8b8d5 34a4520 16b3938

Makman2:
@rultor merge

Issue #4249, opened 2017-05-21
coalaTest: Unexpected assert for an actual error


Makman2:
See https://github.com/coala/coala/blob/master/tests/coalaTest.py#L37

We assert that there's some error output which seem to be some kind of traceback which is bad.

Task:
1. Check what's actually going wrong
-> If it's reasonable, place a comment above to reason why this is done so nobody is going to be confused and raises an issue again (like me ;D)
-> If it can/should be fixed, fix it and remove assertion for stderr.

jayvdb:
Ok, there are also three in `tests/misc/CachingTest.py`.  They were all introduced in cef20b6a7 (https://github.com/coala/coala/pull/3638), which at least confirms they are not a 'very recent' regression, but that commit was still within the 0.11 cycle, so it could still be a 0.11 regression that caused it.

jayvdb:
The exception is

> OSError: reading from stdin while output is captured

It is raised from the pytest capture handler: https://github.com/pytest-dev/pytest/blob/d86c89e/_pytest/capture.py#L446

This is unlikely to be a regression.  The easy fix is to add `--ci` to all of these commands, so the tests dont ask the user for a value.

PR #4266, opened 2017-05-23
tests: Avoid exception in test logic

jayvdb:
Three tests allowed coala to have an internal exception occur,
which was then captured and reported as a success in the test
assertions.

These invocations of coala where seeking user input, which
failed under pytest as tests should not depend on user input.

Fixes https://github.com/coala/coala/issues/4249

### DIFF
--- a/tests/misc/CachingTest.py
+++ b/tests/misc/CachingTest.py
@@ -122,20 +122,28 @@ class CachingTest(unittest.TestCase):
             retval, stdout, stderr = execute_coala(
                 coala.main,
                 'coala',
+                '--non-interactive', '--no-color',
                 '-c', os.devnull,
                 '-f', re.escape(filename),
                 '-b', 'LineCountTestBear')
             self.assertIn('This file has', stdout)
-            self.assertIn(' During execution of coala', stderr)
+            self.assertEqual(1, len(stderr.splitlines()))
+            self.assertIn(
+                'LineCountTestBear: This result has no patch attached.',
+                stderr)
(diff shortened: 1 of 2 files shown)


tests/misc/CachingTest.py
```diff
@@ -122,20 +122,28 @@ def test_caching_results(self):
             retval, stdout, stderr = execute_coala(
                 coala.main,
                 'coala',
+                '--non-interactive', '--no-color',
```
mrtes: Is the `--no-color` flag necessary here (and at the other positions)?

jayvdb: Not strictly necessary, but it will make it easier to read any logging output on the CI log, at least it will when https://github.com/coala/coala/pull/4224 is fixed.

mrtes: But text with colors should be _more_ readable (if used in moderation), no?

jayvdb: No.  Escape sequences are printed in raw bytes when there is an error in a test, and as pytest trims long strings, all the color information is shown while the useful information is not show.  See
https://travis-ci.org/jayvdb/coala/jobs/235260955#L813

mrtes: Ah ok, I didn't know that.

mrtes:
ack 22ff7ff

jayvdb:
reack 8c41320

jayvdb:
@rultor merge

PR #4270, opened 2017-05-23
cli: Allow substring matching of sections

netman92:
Coala now runs subsections of sections split by `.`.

Closes: https://github.com/coala/coala/issues/4259


coalib/settings/ConfigurationGathering.py
```diff
@@ -105,8 +107,12 @@ def warn_nonexistent_targets(targets, sections, log_printer):
     :param sections:    The sections to search. (Dict.)
     :param log_printer: The log printer to warn to.
     """
+
+    sections_substrings = list(itertools.chain(
+        *[sections_name.split('.') for sections_name in sections]))
```
underyx: Could you change this to

```python
sections_substrings = {
    substring
    for sections_name in sections
    for substring in sections_name.split('.')
}
```

?

I feel like it's waaay cleaner. And `set` is a more appropriate data type here than `list`.

coalib/settings/ConfigurationGathering.py
```diff
@@ -105,8 +107,12 @@ def warn_nonexistent_targets(targets, sections, log_printer):
...
     for target in targets:
-        if target not in sections:
+        if (target not in sections_substrings and
+                target not in sections):
```
underyx: I'm pretty sure you can remove `target not in sections` here, no? `sections_substrings` should be a superset of `sections`.

coalib/settings/ConfigurationGathering.py
```diff
@@ -105,8 +107,12 @@ def warn_nonexistent_targets(targets, sections, log_printer):
...
             log_printer.warn(
                 "The requested section '{section}' is not existent. "
                 'Thus it cannot be executed.'.format(section=target))
```
underyx: I think we should update the error message, this one makes it sound like only exact matches were tried. How about `No matching section found for '{section}'.`?

coalib/settings/ConfigurationGathering.py
```diff
@@ -386,9 +392,28 @@ def gather_configuration(acquire_settings,
...
 
     return (sections,
             local_bears,
             global_bears,
             targets)
+
+
+def add_substring_targets(targets, sections):
+    if not targets:
+        return
```
underyx: Can you try the code without this? I think it will work without it as well cause `for target in targets` etc. will just iterate through an empty list.

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -222,6 +224,53 @@ def test_targets(self):
...
+                log_printer,
+                arg_list=['-c', 'inherit_coafile', 'ja'])
+            self.assertEqual(targets, ['ja'])
+
+            log_messages = [_l.message for _l in log_printer.logs]
+            msg = (
+                "The requested section 'ja' is not existent.",
+                'Thus it cannot be executed.'
+            )
+            self.assertNotIn(msg, log_messages)
```
underyx: Shouldn't this be `assertIn`?

coalib/settings/ConfigurationGathering.py
netman92: nope ... If you have e.g. `python.all.somethig` and sections_substrings are `['python', 'all', 'somethig']

underyx:
unack https://github.com/coala/coala/pull/4270/commits/806f52cf15553d37d69eeec4518a75d27806c1a0

coalib/settings/ConfigurationGathering.py
underyx: Ah, yes, that's right. In that case I'd recommend `if target not in (sections_substrings 
 + set(sections)):`

coalib/settings/ConfigurationGathering.py
netman92: 👍

coalib/settings/ConfigurationGathering.py
netman92: 👍

tests/settings/ConfigurationGatheringTest.py
netman92: Nope, `ja` is not a section neither a subsection.

coalib/settings/ConfigurationGathering.py
netman92: 👍

tests/settings/ConfigurationGatheringTest.py
netman92: 👍

coalib/settings/ConfigurationGathering.py
```diff
@@ -386,9 +394,25 @@ def gather_configuration(acquire_settings,
...
+    new_targets = {
+        section_name
+        for target in targets
+        for section_name in section_names
+        if target in section_name.split('.')
+    }
+
+    new_targets = new_targets.difference(targets)
+
+    targets.extend(sorted(new_targets, key=lambda x: section_names.index(x)))
```
underyx:
```python
targets.extend(sorted(new_targets, key=section_names.index))
```

I just realized this does the same 😮

coalib/settings/ConfigurationGathering.py
```diff
@@ -105,11 +105,19 @@ def warn_nonexistent_targets(targets, sections, log_printer):
...
+
+    sections_substrings = {
+        substring
+        for sections_name in sections
+        for substring in sections_name.split('.')
+    }
     for target in targets:
-        if target not in sections:
+        if (target not in sections_substrings and
+                target not in sections):
```
underyx: I would suggest
```python
if target not in sections_substrings + set(sections):
```
here.

coalib/settings/ConfigurationGathering.py
underyx: And maybe doing
```python
new_targets.sort(key=section_names.index)
```
on the line above is nicer.

coalib/settings/ConfigurationGathering.py
```diff
@@ -386,9 +393,26 @@ def gather_configuration(acquire_settings,
...
+    section_names = list(sections.keys())
+
+    new_targets = {
+        section_name
+        for target in targets
+        for section_name in section_names
+        if target in section_name.split('.')
+    }
+
+    new_targets = new_targets.difference(targets)
```
underyx:
```python
new_targets -= targets```

Comment (netman92) [inline coalib/settings/ConfigurationGathering.py]: Not possible, `list` and `set`

**Comment by netman92** [comment]
@underyx done 🙃

File: coalib/settings/ConfigurationGathering.py (hunk 8 above)
Comment (underyx) [inline coalib/settings/ConfigurationGathering.py]: Oooooooh.

**Comment by underyx** [comment]
That's all I got! Thanks for *bearing* with me throughout this learning session for the both of us 🙃

**Comment by underyx** [comment]
ack https://github.com/coala/coala/pull/4270/commits/6b5b463937052a8295a508f69912f8ead60cea92

**Comment by supergr35** [comment]
unack 6b5b463

**Comment by supergr35** [comment]
No colon after `Closes` in the commit message please, also you can remove the period at the end of the commit body for clarity ;)

Not quite sure about the actual code, would need someone else more suitable to look over that

File: coalib/settings/ConfigurationGathering.py [hunk 9]
```diff
@@ -386,9 +393,25 @@ def gather_configuration(acquire_settings,
...
+
+
+def add_substring_targets(targets, sections):
+    section_names = list(sections.keys())
+
+    new_targets = {
+        section_name
+        for target in targets
+        for section_name in section_names
+        if target in section_name.split('.')
```
jayvdb: I suspect this means that cli arg `all.python` wont match `all.python.foo`

netman92: 👍 fixed

netman92:
@underyx Can you re-run the CI ?

underyx:
@netman92 probably easier if you `git commit --amend --no-edit && git push --force-with-lease` (I don't know if I have access to all 3 CI services.)

netman92:
@underyx Done.

netman92:
```pkg_resources.DistributionNotFound: The 'chardet<3.1.0,>=3.0.2' distribution was not found and is required by requests

coala --non-interactive returned exit code 1```

**Comment by jayvdb** [comment]
@netman92 please rebase in order to get past the problems of https://github.com/coala/coala/issues/4277

**Comment by netman92** [comment]
Yes, @jayvdb so should I change something?

**Comment by netman92** [comment]
@jayvdb @underyx Should I change something? For what are we waiting here?

File: tests/settings/ConfigurationGatheringTest.py [hunk 10]
```diff
@@ -222,6 +224,53 @@ def test_targets(self):
...
+        current_dir = os.path.abspath(os.path.dirname(__file__))
+        test_dir = os.path.join(current_dir, 'section_manager_test_files')
+        with change_directory(test_dir):
+            log_printer = ListLogPrinter()
+            _, _, _, targets = gather_configuration(
+                lambda *args: True,
+                log_printer,
+                arg_list=['-c', 'inherit_coafile', 'python'])
+            self.assertEqual(
+                targets, ['python', 'all.python', 'all.python.codestyle'])
```
jayvdb: I am curious why we want `python` to be a target, when it isnt a value section of the `.coafile`.

It feels like that might cause other problems. cc @Udayan12167 who will know the ramifications of this better.

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -222,6 +224,53 @@ def test_targets(self):
...
+
+            log_messages = [_l.message for _l in log_printer.logs]
+            msg = "No matching section found for 'python'."
+            self.assertNotIn(msg, log_messages)
```
jayvdb: This test is very brittle, as someone might modify the message `"No matching section found for 'python'."`, and then this test will continue to pass even if it is broken.

A better way to do this is assert that there are no warnings in `log_messages`.

jayvdb:
@netman92 , ah, I see the `python` test is done.  I think the `c` test mentioned at https://github.com/coala/coala/pull/4270#pullrequestreview-40771734 is helpful, to make sure partial matches within a subsection name are not matched.

tests/settings/ConfigurationGatheringTest.py
netman92: This is not possible because there are already warning like `No bears matching 'LineLengthBear' were found. Make sure you have coala-bears installed or you have typed the name correctly.`

tests/settings/ConfigurationGatheringTest.py
netman92: We want to check if coala will run all `*.python.*` sections.

netman92:
@jayvdb  Can you please do CR again?

sils:
commit message: Coala -> coala

netman92:
@sils Done

sils:
sorry got distracted, doing full review now :)

sils:
Design question:

Given the following coafile:

```ini
[all]
stuff
[all.python]
stuff
[all.python.style]
stuff
```

If I run `coala all` I expect the all section to be run, but not the others. If I run `python` I expect `all.python` to be run but not the others. My opinion, intuitively. Though I find this a bit strage as a feature.

OTOH with our scoping it could make more sense to have the use invoke `coala all.python` and doing a startswith, i.e. running all sections starting with the name and not all containing the name.

Containing feels wrong and weird :/ CC @underyx and @jayvdb

underyx:
I like contain cause then you can do:
```ini
[python.style]
...

[js.style]
...

[css.style]
...
```

and run all with just `coala style`. Not just that, but the entire reason is because when I have
```
[all.python.pep8]
...
```
I didn't want to type out that entire thing, I wanted to just type `coala pep8` like on older versions of coala.

underyx:
PS: @netman92 had to tell me IRL that you tagged me since I'm just ignoring all coala GitHub notifications now due to the gitmate spam.

coalib/settings/ConfigurationGathering.py
```diff
@@ -109,11 +109,18 @@ def warn_nonexistent_targets(targets, sections, log_printer):
...
     """
+
+    sections_substrings = {
+        substring
+        for sections_name in sections
+        for substring in sections_name.split('.')
+    }
     for target in targets:
-        if target not in sections:
+        if target not in sections_substrings | set(sections):
```
sils: if I have sections

```
[all.python]
[all.python.style]
[other.python]
```

and I run `coala all.python` according to this new logic I'd expect the first two sections to be run right?

Maybe something like:

```

sections_name = ['.' + section + '.' for section in sections_name]
for target in targets:
    target = '.' + target + '.'
    if any(target in secname for secname in sections_name):
       ...
```

sils: I'd rather have the logic of the get_substring_targets somehow in here than below adding artificial targets, this should change the interpretation of the targets and not define new ones IMO

coalib/settings/ConfigurationGathering.py
```diff
@@ -398,9 +405,33 @@ def gather_configuration(acquire_settings,
...
 
     return (sections,
             local_bears,
             global_bears,
             targets)
+
+
+def get_substring_targets(target):
+    for i in range(len(target)):
+        if not target[i] == '.':
```
sils: str.find?

sils:
Feels like this can be simplified, if it's simplified, reviewed and acked I'm fine with this

coalib/settings/ConfigurationGathering.py
netman92: Hmm, but if I will run `coala pyt` it should tell me this 
```No matching section found for 'pyt'.```

In your case, it will not.  I am not sure if it's good or not.

Comment (sils) [inline coalib/settings/ConfigurationGathering.py]: it should definitely tell that, we can do `if any ... : else: logging.warn(...)` or so maybe?

File: coalib/settings/ConfigurationGathering.py (hunk 13 above)
Comment (netman92) [inline coalib/settings/ConfigurationGathering.py]: I need to find all occurrences. Used `re`.

File: coalib/settings/ConfigurationGathering.py (hunk 12 above)
Comment (sils) [inline coalib/settings/ConfigurationGathering.py]: also that code was just pseudo code, the most important thing to me is to keep the logic in that place and not generating new artificial targets as you do below IIUC

File: coalib/settings/ConfigurationGathering.py [hunk 14]
```diff
@@ -398,9 +405,31 @@ def gather_configuration(acquire_settings,
                                               acquire_settings,
                                               log_printer)
     save_sections(sections)
+    add_substring_targets(targets, sections)
```
sils: instead of adding artificial targets maybe we should change warn_nonexistent_targets to interpret the targets the right way?

sils:
@netman92 use `cobot mark pending` on gitter to mark pending once the discussions are resolved, I think leaving this in the review queue would just waste people's attention as long as we're discussing this also as per our slack conversation which isn't recorded here yet

PR #4309, opened 2017-06-01
ResultFilter: Allow relative file paths

fneu:


shreyans800755:
@fneu Can you please mention which issue is this PR fixing in the commit message itself ?
Have a look at http://api.coala.io/en/latest/Developers/Writing_Good_Commits.html

fneu:
@shreyans800755 I am afraid that there is no corresponding issue being close by this change.

The crash mentioned in the commit message (key error caused by the wrong path being requested from the file dict) does not appear in coala, since absolute file paths are being passed to this function at all times.

I, however, was using this module out of context in another project and noticed the needless limitation of this function, which I incidentally authored a long time ago. I guess I could file an issue tomorrow and add it when rebasing.

Monal5031:
@fneu Please always create an issue first and get it assigned before working on anything or sending a PR.
Create an issue for this and link it in your commit message, also discuss with the developers if this change is needed or not.

Issue #4302, opened 2017-06-01
Testing API doesn't yield useful results


yash-nisar:
The current testing API doesn’t provide useful results as a traceback if a given test fails. It becomes a strenuous and time consuming job to follow the traceback searching for errors.

Eg. For the check_results(...) method, http://paste.ubuntu.com/24269550/ is the output that I received when I had a minor difference of line=5 instead of line=6 in the given code snippet :
```
[Result.from_values('ElmLintBear',
                     message=result_message_bad_function,
                     file=fname,
                     line=5,
                     severity=RESULT_SEVERITY.NORMAL)]
```
The given output below appear to be strange as both the Result objects contain the same number of characters (632) due to which testing transforms to a challenging task. Hit and trial methods or the usage of a debugger are the only possible ways to overcome these errors.
```
E           AssertionError: Lists differ: [<Result object(id=0xce0383d576a84f34848d66d0ddc87479, origin=[632 chars]710>] != [<Result object(id=0x7e58527926a345f8a2bb36305c58f9f6, origin=[632 chars]400>]
E           
E           First differing element 0:
E           <Result object(id=0xce0383d576a84f34848d66d0ddc87479, origin=[631 chars]0710>
E           <Result object(id=0x7e58527926a345f8a2bb36305c58f9f6, origin=[631 chars]0400>
E           
E           Diff is 2804 characters long. Set self.maxDiff to None to see it. : The local bear 'ElmLintBear' doesn't yield the right results. Or the order may be wrong.
E           Running bear ElmLintBear...
E           Running 'elm-format /tmp/tmpzu0n4474 --yes'
```
difficulty/medium
area/tests
type/test

jayvdb:
@yash-nisar , but if you set `self.maxDiff = None`, the output becomes much more usable, but the pytest approach of showing all of the backtrace results in a horrible maze of information.

We actually have a few `self.maxDiff = None` littered around the tests.

Makman2:
It's still quite some output, so we make it a bit more precise :+1:

jayvdb:
The design at https://yash-nisar.github.io/blog/testing-api-coala/ doesnt indicate how you will improve on the pytest output, which is mostly ugly and unusable because of the depth of the stack, and volume of information with self.maxDiff is disabled, and I dont see that being addressed in the proposal.

Makman2:
[...]

Issue #1436, opened 2016-02-12
python setup.py test throws VersionConflict(dist, req) errors


sudheesh001:
If there are already packages installed in the virtualenv or the system requirements, any changes in the version stop the test run and raise a VersionConflict error.

Expected behaviour: it runs pip upgrade of that particular requirement.

sils1297:
Any idea on how that can be fixed?

Makman2:
problem is what when it automatically downgrades? At least it should ask the user^^

AbdealiJK:
This should be fixed by the better requirements managing right ?
The reason the VersionConflict happens is because setuptools checks the dependencies in `setup.py test`

sils1297:
so is this fixed? Requirements management is better now after all

AbdealiJK:
If the virtualenv has let's sat setuptools v18.0 and we need >=19.0 I think the error will still be there.
But overall it should occur less frequently and only if we *cannot* use that version.

sils1297:
and we have a way better error message now, can we close?

AbdealiJK:
Closing this.
If others have a different opinion, please reopen

PR #4310, opened 2017-06-02
LocalBearTestHelper: Add ``assertObjectsEqual``

yash-nisar:
Add ``assertObjectsEqual(..)`` method that compares individual
fields of the Result object and yields better, easy to understand
messages in case of an attribute mismatch.

Closes https://github.com/coala/coala/issues/4302

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/testing/LocalBearTestHelper.py
+++ b/coalib/testing/LocalBearTestHelper.py
@@ -8,6 +8,7 @@ import pytest
 from coalib.bearlib.abstractions.LinterClass import LinterClass
 from coalib.testing.BearTestHelper import generate_skip_decorator
 from coalib.bears.LocalBear import LocalBear
+from coala_utils.Comparable import Comparable
 from coala_utils.ContextManagers import prepare_file
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
@@ -109,6 +110,28 @@ class LocalBearTestHelper(unittest.TestCase):
 
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
+
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        if len(observed_result) == len(expected_result):
+            messages = ''
+            for observed, expected in zip(observed_result, expected_result):
+                if (isinstance(observed, Comparable)
+                    and isinstance(expected, Comparable)) and (
+                            type(observed) is type(expected)):
+                    for attribute in type(observed).__compare_fields__:
+                        try:
+                            self.assertEqual(
+                                getattr(observed, attribute),
+                                getattr(expected, attribute),
+                                msg='{} mismatch.'.format(attribute))
+                        except AssertionError as ex:
+                            messages += (str(ex) + '\n\n')
+                else:
+                    self.assertEqual(observed_result, expected_result)
+            if messages:
+                raise AssertionError(messages)
+        else:
+            self.assertEqual(observed_result, expected_result)
 
     def check_validity(self,
                        local_bear,
@@ -227,16 +250,6 @@ class LocalBearTestHelper(unittest.TestCase):
                               list,
                               msg='The given results are not a list.')
 
-        if results in [[], ()]:
-            msg = ("The local bear '{}' yields a result although it "
-                   "shouldn't.".format(local_bear.__class__.__name__))
-            check_order = True
-        else:
-            msg = ("The local bear '{}' doesn't yield the right results."
-                   .format(local_bear.__class__.__name__))
-            if check_order:
-                msg += ' Or the order may be wrong.'
-
         bear_output = get_results(local_bear, lines,
                                   filename=filename,
                                   force_linebreaks=force_linebreaks,
@@ -244,9 +257,10 @@ class LocalBearTestHelper(unittest.TestCase):
                                   tempfile_kwargs=tempfile_kwargs,
                                   settings=settings)
         if not check_order:
-            self.assertEqual(sorted(bear_output), sorted(results), msg=msg)
+            self.assertComparableObjectsEqual(
+                sorted(bear_output), sorted(results))
         else:
-            self.assertEqual(bear_output, results, msg=msg)
+            self.assertComparableObjectsEqual(bear_output, results)
 
         return bear_output
 
--- a/tests/testing/LocalBearTestHelperTest.py
+++ b/tests/testing/LocalBearTestHelperTest.py
@@ -7,6 +7,10 @@ from tests.test_bears.TestBearDep import (TestDepBearBDependsA,
                                           TestDepBearCDependsB,
                                           TestDepBearDependsAAndAA)
 from coalib.bearlib.abstractions.Linter import linter
+from tests.test_bears.LineCountTestBear import LineCountTestBear
+from coala_utils.ContextManagers import prepare_file
+from coalib.results.Result import Result
+from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 from coalib.testing.LocalBearTestHelper import verify_local_bear, execute_bear
@@ -38,6 +42,55 @@ class LocalBearCheckResultsTest(Helper):
         with self.assertRaises(AssertionError):
             self.check_results(self.uut, ['a', 'b'], ['b', 'a'],
                                check_order=True)
+
+    def test_result_inequality(self):
+        with self.assertRaises(AssertionError):
+            self.check_results(self.uut, ['a', 'b'], ['a', 'b', None],
+                               check_order=True)
+
+    def test_good_assertComparableObjectsEqual(self):
+        self.uut = LineCountTestBear(Section('name'), Queue())
+        file_content = 'a\nb\nc'
+        with prepare_file(file_content.splitlines(), filename=None,
+                          create_tempfile=True) as (file, fname):
+            self.check_results(self.uut,
+                               file_content.splitlines(),
+                               [Result.from_values(
+                                origin='LineCountTestBear',
+                                message='This file has 3 lines.',
+                                severity=RESULT_SEVERITY.INFO,
+                                file=fname)],
+                               filename=fname,
+                               create_tempfile=False)
+
+    def test_bad_assertComparableObjectsEqual(self):
+        with self.assertRaises(AssertionError) as cm:
+            self.uut = LineCountTestBear(Section('name'), Queue())
+            file_content = 'a\nb\nc'
+            with prepare_file(file_content.splitlines(), filename=None,
+                              create_tempfile=True) as (file, fname):
+                self.check_results(self.uut,
+                                   file_content.splitlines(),
+                                   [Result.from_values(
+                                    origin='LineCountTestBea',
+                                    message='This file has 2 lines.',
+                                    severity=RESULT_SEVERITY.INFO,
+                                    file=fname)],
+                                   filename=fname,
+                                   create_tempfile=False)
+        self.assertEqual('\'LineCountTestBear\' != \'LineCountTestBea\'\n'
+                         '- LineCountTestBear\n'
+                         '?                 -\n'
+                         '+ LineCountTestBea\n'
+                         ' : origin mismatch.\n\n'
+                         '\'This file has 3 lines.\' != \'This file has 2 '
+                         'lines.\'\n'
+                         '- This file has 3 lines.\n'
+                         '?               ^\n'
+                         '+ This file has 2 lines.\n'
+                         '?               ^\n'
+                         ' : message_base mismatch.\n\n',
+                         str(cm.exception))
 
 
 class LocalBearTestCheckLineResultCountTest(Helper):


yash-nisar:
status/blocked on `coala_utils` release which will include https://gitlab.com/coala/coala-utils/merge_requests/42

yash-nisar:
CC @Makman2

yash-nisar:
Note that the tests will fail because we still don't have the `__compare_fields` attribute which will only be added after https://gitlab.com/coala/coala-utils/merge_requests/42 is merged and a new version of `coala_utils` is released(maybe).

coalib/testing/LocalBearTestHelper.py
```diff
@@ -185,10 +201,11 @@ def check_results(self,
                                   create_tempfile=create_tempfile,
                                   tempfile_kwargs=tempfile_kwargs,
                                   settings=settings)
+        self.addTypeEqualityFunc(list, self.assertObjectsEqual)
```
Makman2: 1. This should be run somewhere in the beginning of a base test class or so, otherwise it takes effect only in `check_validity`
2. `list` doesn't use `generate_eq` or `generate_repr`, so this is not really working

yash-nisar: But we're basically comparing list of `Result` objects, so whenever we pass lists to the `assertEqual(...)` for comparison, it will call the `assertObjectsEqual(...)` method IIUC. What I had initially done was put `Result` object instead of `list` since we wanna call that method whenever we're comparing `Result` objects, see below : 
Snippet: `self.addTypeEqualityFunc(list, self.assertObjectsEqual)`
Output: 
```
E   AssertionError: 'mixed spaces and tabs' != 'mixed spaces and tab'
E   - mixed spaces and tabs
E   ?                     -
E   + mixed spaces and tab
E    : message_base mismatch
```
Snippet: `self.addTypeEqualityFunc(Result, self.assertObjectsEqual)`
Output: 
```
E   AssertionError: Lists differ: [<Result object(id=0xf33dc5cf898c48b4b7114cffd86e8172, origin='[490 chars]438>] != [<Result object(id=0xf211aef4abb24ecb8cc36beebbd848a8, origin='[489 chars]d30>]
E   
E   First differing element 0:
E   <Result object(id=0xf33dc5cf898c48b4b7114cffd86e8172, origin='[489 chars]a438>
E   <Result object(id=0xf211aef4abb24ecb8cc36beebbd848a8, origin='[488 chars]2d30>
E   
E   Diff is 2240 characters long. Set self.maxDiff to None to see it.
```
CC @Makman2

yash-nisar: Should I create an `__init__` method inside our `LocalBearTestHelper` class and add it there ?

yash-nisar: Another alternative that I've found is we pass the `Result` object to `assertEqual` which is something like(Error handling of empty lists, and so on will be done..) : 
```
        if not check_order:
            self.assertEqual(sorted(bear_output)[0], sorted(results)[0])
        else:
            self.assertEqual(bear_output[0], results[0])
```
Then `self.addTypeEqualityFunc(Result, self.assertObjectsEqual)` snippet is justified IIUC.

Makman2: Ouch haven't thought about that we pass results with lists... here we could just iterate over all passed results and compare them one by one with `assertEqual`. And we still register `addTypeEqualityFunc` with `Result` globally (or an ABC-type of classes using `generate_eq` and `generate_ordering` which would be the ideal solution).

Hm this is better than nothing, but ideally we would be able to have nicer outputs for collections... don't know something yet, maybe you also get some ideas how to make `self.assertObjectsEqual` working automatically everywhere in our tests, even those ones not testing only bears.

yash-nisar: > Ouch haven't thought about that we pass results with lists... here we could just iterate over all passed results and compare them one by one with assertEqual. And we still register addTypeEqualityFunc with Result globally (or an ABC-type of classes using generate_eq and generate_ordering which would be the ideal solution).

Aren't we doing this already in https://github.com/coala/coala/pull/4310#discussion_r119993949 ?
> Hm this is better than nothing, but ideally we would be able to have nicer outputs for collections... don't know something yet, maybe you also get some ideas how to make self.assertObjectsEqual working automatically everywhere in our tests, even those ones not testing only bears

Our main concern ATM is the not so useful output of bears so IMO we should get this basic functionality merged and those not testing only bears can be looked after later on ? What say ?

Makman2: We could actually use `Iterable` instead of only `list`, so we support any kind of iterable :+1:

coalib/testing/LocalBearTestHelper.py
```diff
@@ -185,10 +201,11 @@ def check_results(self,
...
         if not check_order:
-            self.assertEqual(sorted(bear_output), sorted(results), msg=msg)
+            self.assertEqual(sorted(bear_output), sorted(results))
         else:
-            self.assertEqual(bear_output, results, msg=msg)
+            self.assertEqual(bear_output, results)
```
Makman2: can we delete the `msg` variable above as we don't seem to use it any more?

coalib/testing/LocalBearTestHelper.py
Makman2: Or maybe for this case we should explicitly use `assertObjectsEqual` instead of registering it for tests. Makes it easier to understand^^

yash-nisar:
Will fix things suggested by the review comments in the next iteration and iterate over the list containing more than multiple results.

yash-nisar:
Updated @Makman2

coalib/results/Result.py
```diff
@@ -25,7 +26,7 @@
                    'aspect',
                    'additional_info',
                    'debug_msg')
-class Result:
+class Result(Comparables):
```
Makman2: We want to register `Result` to be a deriving type of `Comparables` inside the `generate_eq`/`generate_ordering` decorators, not here :+1:

Makman2: See https://docs.python.org/3.6/library/abc.html#abc.ABCMeta.register

yash-nisar: Yeah, I just got to know the difference between the 2 ways of subclassing, the real one and the virtual one. I had actually referred to https://pymotw.com/3/abc/ which doesn't explicitly state the difference. My bad . :)

yash-nisar:
How does this look ? @Makman2

yash-nisar:
Strange circle only error. The package seems to be installed but cannot be imported. However, the tests pass for travis and appveyor.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,18 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertObjectsEqual(self, observed_result, expected_result):
+        attributes = Result.__compare_fields__
```
Makman2: instead of using only the fields from `Result`, get it from the compared objects below :)

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,18 @@ class LocalBearTestHelper(unittest.TestCase):
...
+        for observed, expected in zip(observed_result, expected_result):
+            if isinstance(observed, Comparable) and isinstance(
+                    expected, Comparable):
```
Makman2: we should also check if the type is the same (or maybe subclasses from each other)? (I think `typeA is typeB` is easier for now)

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,18 @@ class LocalBearTestHelper(unittest.TestCase):
...
+                for attribute in attributes:
```
Makman2: see above, get the attributes dynamically :)
`type(observed).__compare_fields__`

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,18 @@ class LocalBearTestHelper(unittest.TestCase):
...
+                    self.assertEqual(getattr(observed, attribute),
+                                     getattr(expected, attribute),
+                                     msg='{} mismatch'.format(attribute))
```
Makman2: Would put a period after `mismatch`^^

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,18 @@ class LocalBearTestHelper(unittest.TestCase):
...
+            else:
+                self.assertSequenceEqual(observed_result, expected_result)
```
Makman2: you can use `self.assertEqual` again, we don't register a `TypeEqualityFunc` :)

setup.py
```diff
@@ -65,11 +65,29 @@ def run(self):
               '-b {}'.format(current_version)])
         VERSION = get_version()
 
-with open('requirements.txt') as requirements:
-    required = requirements.read().splitlines()
+dependency_links = []
```
Makman2: you don't use this value any further^^

yash-nisar:
Unable to figure out what's wrong with circle ci. Anyways are the changes good @Makman2 ?

Makman2:
```
ImportError while importing test module '/home/ubuntu/coala/tests/testing/LocalBearTestHelperTest.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/testing/LocalBearTestHelperTest.py:11: in <module>
    from coalib.testing.LocalBearTestHelper import verify_local_bear, execute_bear
coalib/testing/LocalBearTestHelper.py:10: in <module>
    from coala_utils.Comparable import Comparable
E   ImportError: No module named 'coala_utils.Comparable'
```
Could you try to rebuild without cache? Maybe that solves the problem^^

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,19 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        for observed, expected in zip_longest(
+                observed_result, expected_result):
+            if (isinstance(observed, Comparable)
+                and isinstance(expected, Comparable)) and (
+                        type(observed) is type(expected)):
```
Makman2: hmm indentation doesn't look so right, but okay :P

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,19 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        for observed, expected in zip_longest(
+                observed_result, expected_result):
```
Makman2: we definitely need to assert for length of the lists. Consider we put `None` either into `observed_result` or `expected_result`, and one of the iterators is too short, then we can't distinguish this case and would actually pass the assertion!
```
assertComparableObjectsEqual([a, None], [a]) # <-- Oops, this passes!
```
(Please make a testcase for this :D)

tests/testing/LocalBearTestHelperTest.py
```diff
@@ -34,6 +38,21 @@ def test_require_order(self):
             self.check_results(self.uut, ['a', 'b'], ['b', 'a'],
                                check_order=True)
 
+    def test_assertObjectsEqual(self):
```
Makman2: test function name should be updated ;)

tests/testing/LocalBearTestHelperTest.py
```diff
@@ -34,6 +38,21 @@ def test_require_order(self):
...
+                          create_tempfile=True) as (file, fname):
+            self.check_results(self.uut,
+                               file_content.splitlines(),
+                               [Result.from_values(
+                                origin='LineCountTestBear',
+                                message='This file has 3 lines.',
+                                severity=RESULT_SEVERITY.INFO,
+                                file=fname)],
+                               filename=fname,
+                               create_tempfile=False)
```
Makman2: Hm tests could be a bit better :3 But we can do this after merge^^

coalib/testing/LocalBearTestHelper.py
yash-nisar: Nice catch, :O

yash-nisar:
Circle is fixed. Thanks to the legendary @Makman2 :tada: :rocket:

coalib/testing/LocalBearTestHelper.py
yash-nisar: When we have a list like ^^, `self.assertEqual(observed_result, expected_result)` line inside our function gets called. I've tested it manually : 
```
AssertionError: Lists differ: ['a'] != ['a', None]

Second list contains 1 additional elements.
First extra element 1:
None

- ['a']
+ ['a', None]
```


Also, we have taken care of cases where we have `self.assertComparableObjectsEqual([Result object..., None], [Result object...])`
```
self.assertComparableObjectsEqual([Result.from_values('StylintBear', message='a', file='xyz', line=2, column=6), None], [Result.from_values('StylintBear', message='a', file='xyz', line=2, column=6)])
```
for which we get an error message as : 
```
AssertionError: Lists differ: [<Result object(id=0x377292abafc14e5bbac9614d8db12536, origin=[345 chars]None] != [<Result object(id=0x9a39c2e5f3bb4ba78d7d41e4353ff798, origin=[339 chars]2b0>]

First list contains 1 additional elements.
First extra element 1:
None

Diff is 1647 characters long. Set self.maxDiff to None to see it.
```

yash-nisar: Still, we assert for lengths :wink:

yash-nisar: So, our updated message looks like : 
```
E   AssertionError: 4 != 3 : The number of elements in the observed result and expected result differ.
```

yash-nisar:
@Makman2 Updated.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,22 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        self.assertEqual(len(observed_result), len(expected_result),
+                         msg='The number of elements in the observed result '
+                             'and expected result differ.')
```
jayvdb: If they have unequal length, this will be less useful than assertEqual, as there is no details in this message.  It doesnt even give the lengths.  The logs need to provide details, so that we can determine the source of the problem.

The most basic improvement here is to check if the length difference is due to only new items or only deleted items.  That is what assertEqual does, and it has a reasonably short but informative way of showing that.  Where assertEqual error message gets messy is if there is a mixture of additions and deletions and modifications.  That can certainly be improved on.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,22 @@ class LocalBearTestHelper(unittest.TestCase):
...
+        for observed, expected in zip_longest(
+                observed_result, expected_result):
+            if (isinstance(observed, Comparable)
+                and isinstance(expected, Comparable)) and (
+                        type(observed) is type(expected)):
+                for attribute in type(observed).__compare_fields__:
+                    self.assertEqual(getattr(observed, attribute),
+                                     getattr(expected, attribute),
+                                     msg='{} mismatch.'.format(attribute))
```
jayvdb: again, details of the differences are lost.

There is also a problem with iterating and using assertEqual, that the first attribute of the first message to be different will cause the test to fail and no further details being shown.  That isnt helpful if there are more than one difference in the output vs expected, which is extremely common.
Also the order of the attributes compared needs to be considered if the first difference is going to be the only mismatch shown.  The ordering for `Result` is currently:

1. 'affected_code'
2. 'severity'
3. 'confidence'
4. 'origin'
5. 'message_base'
6. 'message_arguments'
7. 'aspect'
8. 'additional_info'
9. 'debug_msg'

The better implementation of something like this is to do all of the comparisons first, and then if there are any differences an attempt is made to show them sanely to the user.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -186,9 +204,10 @@ def check_results(self,
                                   tempfile_kwargs=tempfile_kwargs,
                                   settings=settings)
         if not check_order:
-            self.assertEqual(sorted(bear_output), sorted(results), msg=msg)
+            self.assertComparableObjectsEqual(
+                sorted(bear_output), sorted(results))
         else:
-            self.assertEqual(bear_output, results, msg=msg)
```
jayvdb: It looks like you are no longer using `msg`, in which case you should remove the code above which creates `msg`.

But I think that is very misguided.  This method has a special case `if results in [[], ()]:`, which does not need to use "comparable" `Result`s because it is expected that there are no results, and thus any `Result` has nothing to be compared against.

coalib/testing/LocalBearTestHelper.py
yash-nisar: I'm thinking of doing something like : 
```
    def assertComparableObjectsEqual(self, observed_result, expected_result):
        if len(observed_result) != len(expected_result):
            if len(observed_result) > len(expected_result):
                msg = ('The number of elements in the observed result({}) is '
                       'greater than the number of elements in the expected '
                       'result({}). The lacking elements in the expected '
                       'results are:'.format(len(observed_result),
                                             len(expected_result)))
                for i in range(len(observed_result) - len(expected_result)):
                    msg += str(observed_result[i])
                    raise AssertionError(msg)
            else:
                msg = ('The number of elements in the observed result({}) is '
                       'lesser than the number of elements in the expected'
                       'result({}). The extra elements in the expected '
                       'results are:'.format(len(observed_result),
                                             len(expected_result)))
                for i in range(len(expected_result) - len(observed_result)):
                    msg += str(expected_result[i])
                    raise AssertionError(msg)
        else:
            for observed, expected in zip_longest(
                    observed_result, expected_result):
                if (isinstance(observed, Comparable)
                    and isinstance(expected, Comparable)) and (
                            type(observed) is type(expected)):
                    for attribute in type(observed).__compare_fields__:
                        self.assertEqual(getattr(observed, attribute),
                                         getattr(expected, attribute),
                                         msg='{} mismatch.'.format(attribute))
                else:
                    self.assertEqual(observed_result, expected_result)
```

yash-nisar: I'm thinking of something like : 
```
    def assertComparableObjectsEqual(self, observed_result, expected_result):
        if len(observed_result) != len(expected_result):
            if len(observed_result) > len(expected_result):
                msg = ('The number of elements in the observed result({}) is '
                       'greater than the number of elements in the expected '
                       'result({}). The lacking elements in the expected '
                       'results are:\n'.format(len(observed_result),
                                               len(expected_result)))
                raise AssertionError(msg)
            else:
                msg = ('The number of elements in the observed result({}) is '
                       'lesser than the number of elements in the expected '
                       'result({}). The extra elements in the expected '
                       'results are:\n'.format(len(observed_result),
                                               len(expected_result)))
                raise AssertionError(msg)
        else:
            for observed, expected in zip_longest(
                    observed_result, expected_result):
                if (isinstance(observed, Comparable)
                    and isinstance(expected, Comparable)) and (
                            type(observed) is type(expected)):
                    for attribute in type(observed).__compare_fields__:
                        self.assertEqual(getattr(observed, attribute),
                                         getattr(expected, attribute),
                                         msg='{} mismatch.'.format(attribute))
                else:
                    self.assertEqual(observed_result, expected_result)
```

jayvdb: No, your proposed code here still only looks at length.

Checking length does not do "check if the length difference is due to only new items or only deleted items."

To determine if it is only new or only deleted items, you *must* look at each expected item and find it in the resulting array.  Then you can report exactly which item(s) was new or deleted.  That is what assertEqual already does.

yash-nisar: Sorry, will take care of it. :)

yash-nisar: > To determine if it is only new or only deleted items, you must look at each expected item and find it in the resulting array

This might lead to problems if they don't match (i.e. the user makes a mistake in putting wrong values of the attributes ?)

However, I have a found a workaround that will work when both the lists are sorted (i.e. the `expected_result` and `observed_result`) : (Note that I have skipped the lower part i.e. outermost `else`)

```
    def assertComparableObjectsEqual(self, observed_result, expected_result):
        if len(observed_result) != len(expected_result):
            if len(observed_result) > len(expected_result):
                msg = ('The number of elements in the observed result({}) is '
                       'greater than the number of elements in the expected '
                       'result({}). The lacking elements in the expected '
                       'results are:\n'.format(len(observed_result),
                                               len(expected_result)))
                for i in range(
                     len(observed_result) - 1, len(expected_result) - 1, -1):
                    msg += (str(observed_result[i]) + '\n')
                raise AssertionError(msg)
            else:
                msg = ('The number of elements in the observed result({}) is '
                       'lesser than the number of elements in the expected '
                       'result({}). The extra elements in the expected '
                       'results are:\n'.format(len(observed_result),
                                               len(expected_result)))
                for i in range(
                     len(expected_result) - 1, len(observed_result) - 1, -1):
                    msg += (str(expected_result[i]) + '\n')
                raise AssertionError(msg)
```
When the results are sorted, the extra elements can be found out and reported which we've done here. There is another case when the results are not sorted because `check_order` is `True`. Here, we expect the users to enter the results in the same order. We can use a similar approach here too.

CC @jayvdb

coalib/testing/LocalBearTestHelper.py
Makman2: > Also the order of the attributes compared needs to be considered if the first difference is going to be the only mismatch shown.

Makes sense to display all flaws at once. :+1:

However, the ordering is actually not important imo.

coalib/testing/LocalBearTestHelper.py
Makman2: Actually I would fall back to `assertEqual` when the length mismatches. This is the easiest to get this comparative behaviour. We can improve the output messages later for such cases taking `Comparable`-classes into account.

Makman2: As this is totally non-trivial any more^^

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +54,29 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        if len(observed_result) == len(expected_result):
+            messages = ''
+            for observed, expected in zip_longest(
+                    observed_result, expected_result):
```
Makman2: ah btw you don't need `zip_longest` any more, you can just `zip` :) we have ensured that the length is the same :)

coalib/testing/LocalBearTestHelper.py
```diff
@@ -169,26 +194,17 @@ def check_results(self,
...
 
-        if results in [[], ()]:
-            msg = ("The local bear '{}' yields a result although it "
-                   "shouldn't.".format(local_bear.__class__.__name__))
-            check_order = True
-        else:
-            msg = ("The local bear '{}' doesn't yield the right results."
-                   .format(local_bear.__class__.__name__))
-            if check_order:
-                msg += ' Or the order may be wrong.'
```
Makman2: ahhh or wait, we can maybe move it for now to `assertComparableObjectsEqual` into the first `else` clause, above `self.assertEqual(observed_result, expected_result)`. We can then talk about deleting that again in Phase2. This way we keep the old behaviour :D

coalib/testing/LocalBearTestHelper.py
```diff
@@ -52,6 +53,29 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        if len(observed_result) == len(expected_result):
+            messages = ''
+            for observed, expected in zip(
+                    observed_result, expected_result):
```
Makman2: fits into a single line ;)

coalib/testing/LocalBearTestHelper.py
Makman2: Because it was a bit difficult to implement the messages in a modular way, we shifted that for consideration into Phase 2 of GSoC. We will also discuss the behaviour and improval of the general assertion there, for cases where `len(A) != len(B)` :+1:

Makman2:
ack 7f89dca

yash-nisar:
According to @jayvdb ,
> The problem with the current patch is that we do not want to allow VCS links into releases (we want wheels! #1436), so the solution for this issue should not allow it to happen, otherwise it will happen accidentally, which is essentially a busted release.

We either wait for the release of `coala-utils`(that includes the Comparable class) or the patch that adds wheels. This remains blocked till then.

Makman2:
The commit is approved and is tested already and works. For this to pass the official CI, see https://github.com/coala/coala/pull/4310#issuecomment-310916616 .

jayvdb:
can you get CI to pass please ; restarts required?

jayvdb:
Has the bears tests been tested with this patch?

yash-nisar:
@jayvdb , The CI fails because I've removed the VCS commit. So there's only one commit (adding `assertComparableObjectsEqual(..)` method). For the tests, see https://github.com/yash-nisar/coala/blob/7f89dca753d89649b438d9551014c12427a8e103/tests/testing/LocalBearTestHelperTest.py#L76-L88

meetmangukiya:
@yash-nisar bump

yash-nisar:
@meetmangukiya We wait for the release of `coala-utils`, this remains blocked till then. :)

yash-nisar:
@Makman2 Ready for merge ?

Makman2:
ack 14642d8fc23e8b21e2aee05180938033be15d117

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

PR #4312, opened 2017-06-02
Implement Info and InfoExtractor classes cEP-009

satwikkansal:


satwikkansal:
After discussion with @adtac and @Adrianzatreanu, we decided that as of now it makes more sense to add these modules to quickstart as no module is going to use it in near future.

So closing this and will make a similar PR for coala-quickstart.

Issue #4341, opened 2017-06-07
typed_list/dict/ordered_dict(...) annotations look ugly in API docs


userzimmermann:
Let's take `PEP8Bear.run()` as example. `sphinx` autodoc creates the following confusing signature:

> `run(filename, file, max_line_length: int = 79, indent_size: int = 4, pep_ignore: <function typed_list.<locals>.<lambda> at 0x000001CC99B73378> = (), pep_select: <function typed_list.<locals>.<lambda> at 0x000001CC99B73400> = (), local_pep8_config: bool = False)`

`coalib.settings.Setting.typed_list/dict/ordered_dict()` should better return converters with useful `__repr__` instead of `lambda` functions

@sils @Makman2 @jayvdb

Makman2:
Let's override their `__repr__` then :+1: :)

Makman2:
In the end I think it should look like
```
typed_list(str)
typed_list(int)
```
etc.

userzimmermann:
@Makman2 Those were exactly my ideas :)

PR #4342, opened 2017-06-07
typed_...: Create converter class instead of func

userzimmermann:
For better function signatures in `sphinx`-autodoc-generated API docs when `typed_list/dict/ordered_dict(...)` are used as argument annotation

In case of `PEP8Bear.run()`, instead of current

> `run(filename, file, max_line_length: int = 79, indent_size: int = 4, pep_ignore: <function typed_list.<locals>.<lambda> at 0x000001CC99B73378> = (), pep_select: <function typed_list.<locals>.<lambda> at 0x000001CC99B73400> = (), local_pep8_config: bool = False)`

you get now

> `run(filename, file, max_line_length: int = 79, indent_size: int = 4, pep_ignore: typed_list(str) = (), pep_select: typed_list(str) = (), local_pep8_config: bool = False)`

Closes https://github.com/coala/coala/issues/4341

### DIFF
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -46,47 +46,75 @@ def glob_list(obj, *args, **kwargs):
 
 def typed_list(conversion_func):
     """
-    Creates a function that converts a setting into a list of elements each
+    Creates a class that converts a setting into a list of elements each
     converted with the given conversion function.
 
     :param conversion_func: The conversion function that converts a string into
                             your desired list item object.
-    :return:                A conversion function.
+    :return:                An instance of the created conversion class.
     """
-    return lambda setting: [
-        conversion_func(StringConverter(elem)) for elem in setting]
+
+    class Converter:
+
+        def __call__(self, setting):
+            return [conversion_func(StringConverter(elem))
+                    for elem in setting]
+
+        def __repr__(self):
+            return 'typed_list(%s)' % conversion_func.__name__
+
+    return Converter()
 
 
 def typed_dict(key_type, value_type, default):
     """
-    Creates a function that converts a setting into a dict with the given
-    types.
+    Creates a class that converts a setting into a dict with the given types.
 
     :param key_type:   The type conversion function for the keys.
     :param value_type: The type conversion function for the values.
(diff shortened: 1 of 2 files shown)


coalib/settings/Setting.py
```diff
@@ -53,8 +53,17 @@ def typed_list(conversion_func):
...
-        conversion_func(StringConverter(elem)) for elem in setting]
+
+    class Converter:
+
+        def __call__(self, setting):
+            return [conversion_func(StringConverter(elem))
+                    for elem in setting]
+
+        def __repr__(self):
+            return 'typed_list(%s)' % conversion_func.__name__
```
Makman2: why not just annotating the lambda with `__repr__`?

```
converter = lambda ...
converter.__repr__ = lambda: return 'typed_list(%s)' % conversion_func.__name__
```

Makman2: nvm this doesn't work at all :O

userzimmermann: @Makman2 Just try it ;)

`__repr__` has to defined in class, but `lambda` is a `function` instance, and you can't derive the `function` class:

`TypeError: type 'function' is not an acceptable base type`

userzimmermann:
After this gets merged I will also do the same for the other `typed_...()` functions

userzimmermann:
@sils @Makman2 @jayvdb ACK ACK ACK please ;)

userzimmermann:
@jayvdb The before and after examples in the summary https://github.com/coala/coala/pull/4342#issue-234307578 are actually both copied from successfully sphinx-built html docs :) So that works...

But of course I can already apply those changes to the other `typed_...()` functions as well in this PR, combined with some proper base class abstraction

userzimmermann:
@jayvdb Done :) But no useful reason for any inheritance or other abstraction... Besides a possible common base class with a metaclass that automagically uses `sys._getframe()` to fetch the name of the function in which the subclass is created for use in `__repr__` 😄 

cc @sils @Makman2

sils:
rebased this for you to https://github.com/coala/coala/pull/4430/commits

Issue #4201, opened 2017-05-09
Setting.__path__ allows use with strings


jayvdb:
In https://github.com/coala/coala/pull/4171 , we found that `Setting.__path__` is being used with `self` being an instance of `str` rather than instance of `Setting`.

`Setting.__path__` docstring even says it is OK to use the method with strings.

Cute idea, but no. (c.f. [PEP 20](https://www.python.org/dev/peps/pep-0020/) / `import this`)

https://travis-ci.org/jayvdb/coala/jobs/230175885 is a test run with the following added in `Setting.__path__`:
```python
assert isinstance(self, Setting), "self is %s" % self.__class__.__name__
```

A typical failure is:

```python
___________________ ProcessingTest.test_create_process_group ___________________

self = <tests.processes.ProcessingTest.ProcessingTest testMethod=test_create_process_group>

    def setUp(self):
        config_path = os.path.abspath(os.path.join(
            os.path.dirname(__file__),
            'section_executor_test_files',
            '.coafile'))
        self.testcode_c_path = os.path.join(os.path.dirname(config_path),
                                            'testcode.c')

        self.result_queue = queue.Queue()
        self.queue = queue.Queue()
        self.log_queue = queue.Queue()
        self.console_printer = ConsolePrinter()
        log_printer = LogPrinter(ConsolePrinter())
        self.log_printer = ProcessingTestLogPrinter(self.log_queue)

        (self.sections,
         self.local_bears,
         self.global_bears,
         targets) = gather_configuration(lambda *args: True,
                                         log_printer,
                                         arg_list=['--config',
>                                                  re.escape(config_path)])

tests/processes/ProcessingTest.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

coalib/settings/ConfigurationGathering.py:387: in gather_configuration
    log_printer)
coalib/settings/SectionFilling.py:31: in fill_settings
    bear_dirs = section.bear_dirs()
coalib/settings/Section.py:108: in bear_dirs
    bear_dirs = path_list(self.get('bear_dirs', ''))
coalib/settings/Setting.py:14: in path_list
    return obj.__path_list__(*args, **kwargs)
coalib/settings/Setting.py:209: in __path_list__
    return [Setting.__path__(elem, self.origin) for elem in self]
coalib/settings/Setting.py:209: in <listcomp>
[...]

PR #4357, opened 2017-06-12
Setting: Prevent Setting.__path__ with self as str

jayvdb:
Fixes https://github.com/coala/coala/issues/4201

### DIFF
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -89,6 +89,24 @@ def typed_ordered_dict(key_type, value_type, default):
         for key, value in OrderedDict(setting).items())
 
 
+def _path_worker(path: str, origin: str, glob_escape_origin: bool=True):
+    strrep = path.strip()
+    if os.path.isabs(strrep):
+        return strrep
+
+    if origin is None:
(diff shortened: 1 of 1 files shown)


coalib/settings/Setting.py
```diff
@@ -143,13 +161,12 @@ def __init__(self,
         self.key = key
         self.origin = str(origin)
 
+    # use staticmethod !?
```
sils: was that comment for yourself?

coalib/settings/Setting.py
```diff
@@ -89,6 +89,24 @@ def typed_ordered_dict(key_type, value_type, default):
         for key, value in OrderedDict(setting).items())
 
 
+def _path_worker(path: str, origin: str, glob_escape_origin: bool=True):
```
sils: I find the name `_path_worker` totally confusing and it doesn't tell me what it does

sils:
There's no tests for not allowing strings FWIW

sils:
unack 5c5d297

jayvdb:
c.f. https://gitlab.com/coala/coala-utils/issues/77

PR #4376, opened 2017-06-18
Create check_unsupported.sh

jayvdb:
Fixes https://github.com/coala/coala/issues/4375 , and a related bug in pipe handling exposed during https://github.com/coala/coala/issues/4372 .

It also prepares the ground for https://github.com/coala/coala/issues/4323 , making it easier to move the check around.

### DIFF
--- a/.misc/check_unsupported.sh
+++ b/.misc/check_unsupported.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set +e
+
+# Enable capturing the non-zero exit status of coverage instead of tee
+set -o pipefail
+
+coverage run setup.py install | tee setup.log
+
+retval=$?
+
+# coalib.__init__.py should exit with 4 on unsupported versions of Python
+if [[ $retval != 4 ]]; then
+  echo "Unexpected error code $?"
+
+  # When the exit code is 0, use a non-zero exit code instead
+  if [[ $retval == 0 ]]; then
+    exit 127
+  fi
+  exit $retval
+fi
+
+# error when no lines selected by grep
+set -e
+
+grep -q 'coala supports only python 3.4 or later' setup.log
+
+echo "Unsupported check completed successfully"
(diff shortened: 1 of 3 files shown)


.misc/check_unsupported.sh
```diff
@@ -0,0 +1,23 @@
...
+# Enable capturing the non-zero exit status of coverage instead of tee
+set -o pipefail
+
+coverage run setup.py install | tee setup.log
+
+retval=$?
+
+if [[ $retval != 4 ]]; then
+  echo "Unexpected error code $?"
+  if [[ $retval == 0 ]]; then
```
Makman2: hm every error code except 4, but 0 is forbidden? I assume that's on purpose, even so I'm curious :3

.travis.yml
```diff
@@ -11,8 +11,10 @@ matrix:
     # Restore matrix job entry from d2d67fab to test OSX
     - python: 2.7
       env: UNSUPPORTED=true
+      script: .misc/check_unsupported.sh
     - python: 3.3
       env: UNSUPPORTED=true
```
Makman2: do we still need those? Or shall we keep it so we can see it inside the build overview?

jayvdb: Ya, unnecessary now, but helps people understand results on the build overview page.

.misc/check_unsupported.sh
jayvdb: Sort of.
We expect error code 4 (per https://github.com/coala/coala/blob/master/coalib/__init__.py#L27)
So 4 means the script exists without error (0).

If we see 0, then we cant exit without error (0), so we map 0 to a different value.

yukiisbored:
ack https://github.com/coala/coala/pull/4376/commits/e3f817733d291cfc3134acae445e7307dd7e2bc4

yukiisbored:
ack https://github.com/coala/coala/pull/4376/commits/48f784acf51b8d3ea23b3e12359435d108bbe15b

jayvdb:
@rultor merge

PR #4383, opened 2017-06-21
DocumentationExtraction: Ignore fancy string

damngamerz:
Add regex method to ignore fancy strings.
Closes https://github.com/coala/coala/issues/4291

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/languages/documentation/DocumentationExtraction.py
+++ b/coalib/bearlib/languages/documentation/DocumentationExtraction.py
@@ -174,12 +174,16 @@ def _extract_doc_comment(content, line, column, markers):
 
 def _compile_multi_match_regex(strings):
     """
-    Compiles a regex object that matches each of the given strings.
+    Compiles a regex object that checks for indentation before the starting
+    marker (so as to ignore triple quote string literals) and group matches
+    each of the given strings.
 
     :param strings: The strings to match.
     :return:        A regex object.
     """
-    return re.compile('|'.join(re.escape(s) for s in strings))
+    return re.compile('(?P<indent>^\s*)(?P<marker>' +
+                      ('|'.join(re.escape(s) for s in strings)) +
+                      ')')
 
 
 def _extract_doc_comment_from_line(content, line, column, regex,
@@ -187,15 +191,15 @@ def _extract_doc_comment_from_line(content, line, column, regex,
     cur_line = content[line]
     begin_match = regex.search(cur_line, column)
     if begin_match:
-        indent = cur_line[:begin_match.start()]
         column = begin_match.end()
-        for marker in marker_dict[begin_match.group()]:
+        indent = begin_match.group('indent')
+        for marker in marker_dict[begin_match.group('marker')]:
             doc_comment = _extract_doc_comment(content, line, column, marker)
             if doc_comment is not None:
                 end_line, end_column, documentation = doc_comment
 
                 rng = TextRange.from_values(line + 1,
-                                            begin_match.start() + 1,
+                                            len(indent) + 1,
                                             end_line + 1,
                                             end_column + 1)
                 doc = DocumentationComment(documentation, docstyle_definition,
--- a/tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
+++ b/tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
@@ -64,3 +64,8 @@ def foobar_triangle(side_A, side_B, side_C):
     :return: returns perimeter
     """
     return side_A + side_B + side_C
+
+    # This example of triple quote string literal is ignored.
+    triple_quote_string_literal_test = """
+This is a triple quoted string and is not a valid docstring.
+"""
--- a/tests/bearlib/languages/documentation/DocumentationExtractionTest.py
+++ b/tests/bearlib/languages/documentation/DocumentationExtractionTest.py
@@ -243,3 +243,16 @@ class DocumentationExtractionTest(unittest.TestCase):
                                   docstyle_PYTHON3_doxygen, '',
                                   docstyle_PYTHON3_doxygen.markers[1],
                                   TextRange.from_values(1, 1, 1, 55))])
+
+    def test_extract_documentation_PYTHON3_4(self):
+        data = ['\n', 'triple_quote_string_test = """\n',
+                'This is not a docstring\n', '"""\n']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        # Nothing is yielded as triple quote string literals are being
+        # ignored.
+        self.assertEqual(
+            list(extract_documentation(data, 'PYTHON3', 'default')),
+            [])


NiklasMM:
Where are the tests for this?

damngamerz:
@NiklasMM @Makman2 @SanketDG This is one of the solution. Although Niklas suggested me to change the begin regex. I found begin regex is further bound to `marker[2]` in some cases. Also it caused some serious issues with docstrings starting with `column=0`.

Hence I moved to this solution. Hope this solves it to some extent.
Also I have for now included fancy string test for python, for other languages teach me senpai xD :stuck_out_tongue_closed_eyes:

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -162,14 +162,21 @@ def _extract_doc_comment(content, line, column, markers):
...
-    if markers[1] == '':
-        # Extract and align to start marker.
-        return _extract_doc_comment_simple(content, line, column, markers)
-    elif markers[1] == markers[2]:
-        # Search for the each-line marker until it runs out.
-        return _extract_doc_comment_continuous(content, line, column, markers)
-    else:
-        return _extract_doc_comment_standard(content, line, column, markers)
+    # This ignores fancy docstrings.
+    valid_docstring_regex = '\s*[' + markers[0] + '].*'
```
Makman2: `[]` defines a character class, not a group. Use `()` instead.

Makman2: Actually I don't understand the regex. It will only match `markers[0]` in the worst case.

Makman2: ah okay you use `split()`. Though this looks super complicated, can't you just modify the begin regex here: https://github.com/coala/coala/pull/4383/commits/e44c88d5b00728c12eac417ca37170c4f0a3f50e#diff-df547e1526ff7317adef884f7414cdd8R247  ?

NiklasMM:
> Also I have for now included fancy string test for python, for other languages teach me senpai xD :stuck_out_tongue_closed_eyes:

No, your're the one getting paid for this. Changes without proper testing are not acceptable and we're not here to spoon-feed them to you. Please create explicit test cases in `DocumentationExtractionTest` that make it very clear, what we expect to happen here, just adding this one string variable ist not making anything clear e.g. for someone just looking at documentation extraction.

I don't understand, why you don't make the changes at the first position where the begin marker is found, as I described [here](https://github.com/coala/coala/pull/4383#issuecomment-310141092)

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: @Makman2 
>[] defines a character class, not a group. Use () instead.

yup a group :O this is good.

>Actually I don't understand the regex. It will only match markers[0] in the worst case.

Ok so what im doing is that I find the valid docstrings. If you see `line` var. it contains only the starting line of the docstring. hence with `contain[line]` we get the first raw line. now regex only matches with `     """`. If theres something suppose `   abc="""xyz` on the first line. The regex returns `"""xyz` now I compare it with stripped first line(i.e `content[line]`) and voila as the strings doesn't match it is being ignored. The main reason of using `strip()` is, the regex I don't know why doesn't match the `\n` eg. if first raw line is `    """\n` regex matches `    """`.

Now I tried modifying begin regex but the problem I m facing is [here](https://github.com/damngamerz/coala/blob/e44c88d5b00728c12eac417ca37170c4f0a3f50e/coalib/bearlib/languages/documentation/DocumentationExtraction.py#L199)
regex will get spaces. Hence raising `KeyError` I can't use `strip()` as some markers have spaces in it.

damngamerz:
@NiklasMM I already did my homework :D
Most of the languages. have explicit markers for docstrings like in C,java...
And hence their assignment of strings are always in form `abc="xyz"`. leaving no space for using those explicit markers for assignment. This is the case only in python. Although I was just asking if there's any which i might have missed.

>Please create explicit test cases in DocumentationExtractionTest

Yes I have some more tests in mind. Amending them atm.

About the begin marker let me try again if can find a workaround to the issue which I have mentioned in my [comment](https://github.com/coala/coala/pull/4383#discussion_r123685756) above.

NiklasMM:
> And hence their assignment of strings are always in form abc="xyz". leaving no space for using those explicit markers for assignment. This is the case only in python. Although I was just asking if there's any which i might have missed.

No, this seems right. Python seems to be the weirdo in that regard ;)

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: Ok so it goes like this `marker_dict` is specifically made for `begin_regex` to find the `marker[0]` exactly as it is.
As `marker[0]` is the key of `marker_dict` which will return `markers`. eg`marker_dict` for python is `['"""':['"""','','"""']]`.
Hence, this is the reason why I can't change/tamper `begin_regex`. It would require a major change so as to do it(which I think it should not be removed as it has its importance). look at this [for more info.](https://github.com/damngamerz/coala/blob/690ab278c331ffa2c6d53179e5d010b23c9107f6/coalib/bearlib/languages/documentation/DocumentationExtraction.py#L230)

damngamerz: Further modified regex :tada: got this idea from `begin_regex` no need of group matching `()`. we now add escape sequence before each marker character(This is required as for other languages special character needs escape sequences eg: java `*/**`). So for python the regex is now
`\s*\"|\"|\".*`

damngamerz:
With the help of @NiklasMM I was able to amend the changes through `begin_regex` :tada: 
Hope this solves it.
CC^ @Makman2

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -64,3 +64,7 @@ def foobar_triangle(side_A, side_B, side_C):
     :return: returns perimeter
     """
     return side_A + side_B + side_C
+
+    fancy_string_variable = """
```
SanketDG: can you add a comment this is being ignored

SanketDG:
ack e3056fd

SanketDG:
I'll let Mischa take one final look

also, needs rebase

SanketDG:
Also, let's not call it fancy string, lets call something like triple quote string literals

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -179,23 +179,25 @@ def _compile_multi_match_regex(strings):
     :param strings: The strings to match.
     :return:        A regex object.
     """
-    return re.compile('|'.join(re.escape(s) for s in strings))
+    return re.compile('^\s*(?P<marker>' +
+                      ('|'.join(re.escape(s) for s in strings)) +
+                      ')')
```
damngamerz: I like the parent name `_compile_multi_match_regex` as we are adding escape sequences over each character of the docstring. Let me write a description inside the docstring :+1:

Makman2: Instead of compiling, we maybe just return the regex string? Then this function can still keep it's old pure functionality, and below at `begin_regex = ...` we compile it ourselves, adding the space-matching-stuff. What do you think?

Though like this it's also fine, I mean space matching is still understandable for this function :) Just an idea.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -174,28 +174,32 @@ def _extract_doc_comment(content, line, column, markers):
...
 
 def _extract_doc_comment_from_line(content, line, column, regex,
                                    marker_dict, docstyle_definition):
     cur_line = content[line]
     begin_match = regex.search(cur_line, column)
     if begin_match:
-        indent = cur_line[:begin_match.start()]
         column = begin_match.end()
-        for marker in marker_dict[begin_match.group()]:
+        indent = cur_line[:column-len(begin_match.group('marker'))]
```
Makman2: I would rather match the complete indent and reuse it here :+1:
```
begin_match.group(1)  # <-- would instead contain the full indentation
# we could optionally assert that begin_match.start() == 0
```

damngamerz: yup i like the idea of group matching `indent` too and use it here to extract indent :+1:

Makman2:
ack 89b39dc

SanketDG:
rebase needed

damngamerz:
@SanketDG done :+1:

SanketDG:
reack 0114024

SanketDG:
@rultor merge

Issue #4384, opened 2017-06-21
Add aspectlist.get method


userzimmermann:
```python
>>> aspects = aspectlist([Root.Smell.MethodSmell, Root.Smell.ClassSmell.DataClump()])
```

```python
>>> aspects.get('MethodSmell')
<aspectclass 'Root.Smell.MethodSmell'>
```

```python
>>> aspects.get('ClassSmell.DataClump')
<DataClump object(...) ...>
```

```python
>>> aspects.get('BadSmell')
None
```

Relates to https://gitlab.com/coala/GSoC-2017/issues/301

Issue #4388, opened 2017-06-22
Instanced aspect doesn't instance its children


adhikasp:
Current behaviour
```python
>>> coalib.bearlib.aspects['Metadata']('Python 3.4')
<coalib.bearlib.aspects.Metadata.Root.Metadata object at 0x...>
>>> coalib.bearlib.aspects['Metadata']('Python 3.4').CommitMessage
<aspectclass 'Root.Metadata.CommitMessage'>
```

Expected behaviour
```python
>>> coalib.bearlib.aspects['Metadata']('Python 3.4')
<coalib.bearlib.aspects.Metadata.Root.Metadata object at 0x...>
>>> coalib.bearlib.aspects['Metadata']('Python 3.4').CommitMessage
<coalib.bearlib.aspects.Metadata.Root.Metadata.CommitMessage object at 0x...>
```

area/aspects difficulty/medium

PR #4392, opened 2017-06-23
AspectList: Add ``get()`` method

adhikasp:
Closes https://github.com/coala/coala/issues/4384

Relates to https://gitlab.com/coala/GSoC-2017/issues/301

### DIFF
--- a/coalib/bearlib/aspects/base.py
+++ b/coalib/bearlib/aspects/base.py
@@ -1,6 +1,97 @@
+import functools
+import re
+
 from coalib.bearlib.languages import Language
 
+import coalib.bearlib.aspects
 from .taste import TasteError
+
+
+def get_subaspect(parent, subaspect):
+    """
+    Get a subaspect from an aspectclass or aspectclass instance.
+
+    >>> import coalib.bearlib.aspects as coala_aspects
+    >>> metadata = coala_aspects['Metadata']
+    >>> commit_msg = coala_aspects['CommitMessage']
+    >>> shortlog = coala_aspects['Shortlog']
+
+    We can get direct children.
+
+    >>> get_subaspect(metadata, commit_msg)
+    <aspectclass 'Root.Metadata.CommitMessage'>
+
+    Or even a grandchildren.
+
+    >>> get_subaspect(metadata, shortlog)
+    <aspectclass 'Root.Metadata.CommitMessage.Shortlog'>
+
+    Or with string of aspect name
+
+    >>> get_subaspect(metadata, 'shortlog')
+    <aspectclass 'Root.Metadata.CommitMessage.Shortlog'>
+
+    Trying to access children of aspect instance will raise
+    NotImplementedError because of instanced aspect doesn't instance
+    its children and will cause wrong result.
+    See https://github.com/coala/coala/issues/4388
+
+    >>> get_subaspect(metadata('Python'), commit_msg)
+    Traceback (most recent call last):
+    ...
+    NotImplementedError: Cannot access children of aspect instance.
+
+    But, passing subaspect instance as argument is prohibited, because
+    it doesn't really make sense.
+
+    >>> get_subaspect(metadata('Python'), commit_msg('Java'))
+    Traceback (most recent call last):
+    ...
+    AttributeError: Cannot search an aspect instance using another ...
+
+    :param parent:    The parent aspect that should be searched.
+    :param subaspect: An subaspect that we want to find in an
+                      aspectclass.
+    :return:          An aspectclass. Return None if not found.
+    """
+    # Avoid circular import
+    from .meta import isaspect, issubaspect
+    if not isaspect(subaspect):
+        subaspect = coalib.bearlib.aspects[subaspect]
+    if not issubaspect(subaspect, parent):
+        return None
+    if isinstance(subaspect, aspectbase):
+        raise AttributeError('Cannot search an aspect instance using '
+                             'another aspect instance as argument.')
+    if isinstance(parent, aspectbase):
+        raise NotImplementedError('Cannot access children of aspect '
+                                  'instance.')
+
+    parent_qualname = (type(parent).__qualname__ if isinstance(
+                       parent, aspectbase) else parent.__qualname__)
+    if parent_qualname == subaspect.__qualname__:
+        return parent
+
+    # Trim common parent name
+    aspect_path = re.sub(r'^%s\.' % parent_qualname, '',
+                         subaspect.__qualname__)
+    aspect_path = aspect_path.split('.')
+    child = parent
+    # Traverse through children until we got our subaspect
+    for path in aspect_path:
+        child = child.subaspects[path]
+    return child
+
+
+class SubaspectGetter:
+    """
+    Special "getter" class to implement ``get()`` method in aspectbase that
+    could be accessed from the aspectclass or aspectclass instance.
+    """
+
+    def __get__(self, obj, owner):
+        parent = obj if obj is not None else owner
+        return functools.partial(get_subaspect, parent)
 
 
 class aspectbase:
@@ -12,6 +103,8 @@ class aspectbase:
     This is automatically handled by
     :meth:`coalib.bearlib.aspects.meta.aspectclass.subaspect` decorator.
     """
+
+    get = SubaspectGetter()
 
     def __init__(self, language, **taste_values):
         """
--- a/coalib/bearlib/aspects/collections.py
+++ b/coalib/bearlib/aspects/collections.py
@@ -30,3 +30,19 @@ class AspectList(list):
             if issubaspect(aspect, item):
                 return True
         return False
+
+    def get(self, aspect):
+        """
+        Return first item that match or contain an aspect. See
+        :meth:`coalib.bearlib.aspects.aspectbase.get` for further example.
+
+        :param aspect: An aspectclass OR name of an aspect.
+        :return:       An aspectclass OR aspectclass instance, depend on
+                       AspectList content. Return None if no match found.
+        """
+        if not isaspect(aspect):
+            aspect = coalib.bearlib.aspects[aspect]
+        try:
+            return next(filter(None, (item.get(aspect) for item in self)))
+        except StopIteration:
+            return None
--- a/tests/bearlib/aspects/CollectionsTest.py
+++ b/tests/bearlib/aspects/CollectionsTest.py
@@ -43,3 +43,14 @@ class AspectListTest(unittest.TestCase):
         assert not isaspect(str)
         exc.match("<class 'str'> is not an "
                   'aspectclass or an instance of an aspectclass')
+
+    def test_get(self):
+        list_of_aspect = AspectList(
+            [Metadata.CommitMessage.Shortlog, Metadata.CommitMessage.Body])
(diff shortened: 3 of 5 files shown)


coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,89 @@
+import re
+
 from coalib.bearlib.languages import Language
 
+import coalib.bearlib.aspects
 from .taste import TasteError
 
 
+class subaspectGetter:
```
jayvdb: PEP8 naming for classes.

tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -33,3 +33,25 @@ def test__contains__(self):
         assert not isaspect(str)
         exc.match("<class 'str'> is not an "
                   'aspectclass or an instance of an aspectclass')
+
+    def test_get(self):
+        list_of_aspect = AspectList(
+            [Metadata.CommitMessage.Shortlog, Metadata.CommitMessage.Body])
+        assert list_of_aspect.get(Metadata.CommitMessage.Shortlog) is \
```
adhikasp: Change all assert to unittest `self.assert...`

adhikasp: I can't change it to unittest style assert because the pytest fixture will not injected even after adding `@pytest.mark.usefixtures`

jayvdb: These can be rewritten to not use pytest fixtures, but at this stage it isnt worth it.  So long as you are using normal unittest where possible, using pytest only when needing its advanced features is reasonable.

coalib/bearlib/aspects/collections.py
```diff
@@ -14,3 +15,20 @@ def __contains__(self, aspect):
...
+        :return:       An aspectclass OR aspectclass instance, depend on
+                       AspectList content. Return None if no match found.
+        """
+        if isinstance(aspect, str):
+            aspect = coalib.bearlib.aspects[aspect]
+        try:
+            return next(item.get(aspect) for item in self if
+                        issubaspect(aspect, item))
+        except StopIteration:
+            return None
```
jayvdb: `list.get` doesnt exist, but `dict.get` does, and I feel it would be better to use the [`dict.get` semantics](https://docs.python.org/3/library/stdtypes.html#dict.get), which means it should raise `KeyError` here.
@userzimmermann , your thoughts on this?

adhikasp: `dict.get` return None, so this is fine?

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,92 @@
...
+            >>> metadata.get(commit_msg)
+            <aspectclass 'Root.Metadata.CommitMessage'>
+
+            Or even a grandchildren.
+
+            >>> metadata.get(shortlog)
+            <aspectclass 'Root.Metadata.CommitMessage.Shortlog'>
+
+            Trying to access children of aspect instance will raise
+            NotImplementedError.
```
jayvdb: very briefly, explain why

coalib/bearlib/aspects/collections.py
jayvdb: Sorry ... my bad.

tests/bearlib/aspects/conftest.py
```diff
@@ -76,3 +76,64 @@ class SubAspect:
...
+
+@pytest.fixture
+def SubSubAspect(SubAspect, SubSubAspect_docs, SubAspect_tastes):
+    """
+    An exclusive SubAspect class for unit tests.
+    """
+    @SubAspect.subaspect
+    class SubSubAspect:
+        """
+        Definition
```
jayvdb: trailing period?
(or are they omitted for aspect/taste docstrings?)

adhikasp: Omitted, like SubAspect declaration above it

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+                              aspectclass.
+            :return:          An aspectclass. Return None if not found.
+            """
+            # Avoid circular import
+            from .meta import issubaspect
+            if isinstance(subaspect, str):
+                subaspect = coalib.bearlib.aspects[subaspect]
+            if not issubaspect(subaspect, parent):
+                return None
+            if isinstance(subaspect, aspectbase):
```
userzimmermann: `if not isclass(subaspect):` is enough here since we know that we have an aspect. And it looks more intuitive and understandable. The same applies to check below

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+                raise AttributeError('Cannot search an aspect instance using '
+                                     'another aspect instance as argument.')
+            if isinstance(parent, aspectbase):
+                raise NotImplementedError('Cannot access children of aspect '
+                                          'instance.')
+            if issubaspect(parent, subaspect):
```
userzimmermann: This is effectively the same check as `if not issubaspect(subaspect, parent):` above... So fully obsolete. And just returning `parent` if the `subaspect` argument is some parent of `parent` is anyway not correct ;)

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+                return parent
+            # Trim common parent name
+            aspect_path = re.sub(parent_qualname + '.', '',
+                                 subaspect.__qualname__)
+            aspect_path = aspect_path.split('.')
+            child = parent
+            # Traverse through children until we got our subaspect
+            for path in aspect_path:
+                child = child.subaspects[path]
+            return child
```
userzimmermann: You can do this whole search much simpler! You don't need a tree traversal at all, since every aspectclass has a `.parent` attribute. Just go from the `subaspect` back until you find the `parent` or not :)

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+            parent_qualname = type(obj).__qualname__
+        else:
+            parent = owner
+            parent_qualname = owner.__qualname__
+
+        def get_subaspect(subaspect):
+            """
+            Get a subaspect from an aspectclass or aspectclass instance.
+
+            >>> import coalib.bearlib.aspects as aspects
```
userzimmermann: Let's always `import coalib.bearlib.aspects as coala_aspects` as suggested in https://github.com/coala/coala/issues/4408 . That avoids identifier clashes, since `aspects` is often used as variable, argument, or attribute name

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+            We can get direct children.
+
+            >>> metadata.get(commit_msg)
+            <aspectclass 'Root.Metadata.CommitMessage'>
+
+            Or even a grandchildren.
+
+            >>> metadata.get(shortlog)
+            <aspectclass 'Root.Metadata.CommitMessage.Shortlog'>
+
```
userzimmermann: Example for a `'SubAspect'` string is missing

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+    def __get__(self, obj, owner):
+        if obj is not None:
+            parent = obj
+            parent_qualname = type(obj).__qualname__
+        else:
+            parent = owner
+            parent_qualname = owner.__qualname__
+
+        def get_subaspect(subaspect):
```
userzimmermann: This inner method has become pretty huge, so better move it to global module level, giving it an additional parent aspect attribute

coalib/bearlib/aspects/collections.py
```diff
@@ -30,3 +30,20 @@ def __contains__(self, aspect):
...
+        :meth:`coalib.bearlib.aspects.aspectbase.get` for further example.
+
+        :param aspect: An aspectclass OR name of an aspect.
+        :return:       An aspectclass OR aspectclass instance, depend on
+                       AspectList content. Return None if no match found.
+        """
+        if isinstance(aspect, str):
+            aspect = coalib.bearlib.aspects[aspect]
+        try:
+            return next(item.get(aspect) for item in self if
```
adhikasp: But it will fully iterate the list at least once because of `filter`?

coalib/bearlib/aspects/base.py
adhikasp: Aw...

adhikasp: Wait i dont think we can do it. IF `aspect` is an instance, the `subaspect` is just aspectclass, then traversing up the parent will return an aspectclass too.

Relates to https://github.com/coala/coala/issues/4388

userzimmermann: @adhikasp Going up the `.parent` references is just for checking! You always return the subaspect

adhikasp: The subaspect argument is **always** an aspectclass. If we just return that, we cant do this

```
>>> metadata = coala_aspects['metadata']('python 3.4')
... metadata instance
>>> metadata.get('commitmessage')
... commitmessage

Comment (userzimmermann) [inline coalib/bearlib/aspects/base.py]:
```python
aspect = subaspect
while aspect:
    if aspect is parent:
        return subaspect
    aspect = aspect.parent

return None
```

adhikasp: That will give
```python
>>> metadata.get('commitmessage')
... metadata aspectclass
```

coalib/bearlib/aspects/collections.py
userzimmermann: No. `filter` doesn't create a `list` anymore in python 3. It's an iterator itself now, like `itertools.ifilter` in python 2

coalib/bearlib/aspects/base.py
userzimmermann: @adhikasp Sorry! Yes! Given that future case, your are right :) So leave it that way

coalib/bearlib/aspects/base.py
```diff
@@ -1,8 +1,94 @@
...
+                raise AttributeError('Cannot search an aspect instance using '
+                                     'another aspect instance as argument.')
+            if isinstance(parent, aspectbase):
+                raise NotImplementedError('Cannot access children of aspect '
+                                          'instance.')
+            if issubaspect(parent, subaspect):
+                return parent
+            # Trim common parent name
+            aspect_path = re.sub(parent_qualname + '.', '',
+                                 subaspect.__qualname__)
```
userzimmermann: The regex is not so fully nice ;) `'.'` stands for any character. And better start with `'^'` to show that you only substitute from the beginning:

`r'^%s\.' % parent_qualname`

coalib/bearlib/aspects/base.py
adhikasp: Did you mean `inspect.isclass(subaspect)`? Need to import another module :(

userzimmermann:
ack https://github.com/coala/coala/pull/4392/commits/705d256e8344d676e89c0f82c8a4e0a86934e197 https://github.com/coala/coala/pull/4392/commits/79b18a953d32ecc14b00441fb272717f7d4b7f72 https://github.com/coala/coala/pull/4392/commits/0a67a087f1832c7500b8e5ba362c50546e0bd9fa

userzimmermann:
@rultor merge

Issue #4324, opened 2017-06-05
Add `aspects` key in configuration


adhikasp:
If we have `.coafile` like this
```
[all]
aspects = Clone, UnusedImport
```

Then we expect to access those setting as an `aspectlist` instance
```python
>>> type(sections['all']['aspects'])
<class 'aspectlist'>
>>> sections['all']['aspects']
[Root.Redundancy.Clone, Root.Redundancy.UnusedImport]
```

Currently, coala Setting and Section doesn't support this and read that as basic list of string instead.

area/aspect

PR #4389, opened 2017-06-22
AspectList: Overload __init__ to accept strings

adhikasp:
Overload aspectlist so it could be instanced from a list of string
containing aspects name. The reason of this overloading is to make
it possible to instanced list of desired aspects found in coafile.

```python
>>> aspectlist(['coalaCorrect', 'aspectsYEAH'])
[<aspectclass '...coalaCorrect'>, <aspectclass '...aspectsYEAH'>]
```

Replacement of https://github.com/coala/coala/pull/4379
Closes https://github.com/coala/coala/issues/4382

PR #4397, opened 2017-06-24
Processing: Preprocess and pass aspects to bear

adhikasp:
Enable bear to access aspects that defined in configuration files.

Given a coafile:

```
[all]
files = **.py
bears = coalaBear
aspects = coalaCorrect, shortlog.colonexistence
language = Python 3.6
shortlog_colon = false
```

Then, bears can access those aspects with defining a keyword argument
``aspects`` in ``run()`` method.

```python
class SomeBear(LocalBear):
    def run(self, filename, file, aspects:aspectlist=None):
        print(aspects)
        # [<...coalaCorrect object at 0x...>, <...ColonExistence
        # object at 0x...>]
        print(aspects[1])
        # <ColonExistence object(shortlog_colon=False) at
        #  0x7efcd5bf49e8>
        print(aspects[1].shortlog_colon)
        # False
```

Depends on https://github.com/coala/coala/pull/4389
Relates to https://gitlab.com/coala/GSoC-2017/issues/142
Closes https://github.com/coala/coala/issues/4324

### DIFF
--- a/coalib/bearlib/languages/Language.py
+++ b/coalib/bearlib/languages/Language.py
@@ -416,6 +416,9 @@ class Language(metaclass=LanguageMeta):
         return (type(self) is type(item)
                 and set(item.versions).issubset(set(self.versions)))
 
+    def __reduce__(self):
+        return (Language.__getitem__, (str(self),))
+
     @property
     def attributes(self):
         """
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -9,7 +9,7 @@ from coalib.output.ConfWriter import ConfWriter
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.parsing.CliParsing import parse_cli, check_conflicts
 from coalib.parsing.ConfParser import ConfParser
-from coalib.settings.Section import Section
+from coalib.settings.Section import Section, extract_aspects_from_section
 from coalib.settings.SectionFilling import fill_settings
 from coalib.settings.Setting import Setting, path
 from string import Template
@@ -19,6 +19,19 @@ COAFILE_OUTPUT = Template('$type \'$file\' $found!\n'
                           '* add `--save` to generate a config file with '
                           'your current options\n'
                           '* add `-I` to suppress any use of config files\n')
+
+
+def aspectize_sections(sections):
+    """
+    Search for aspects related setting in a section, initialize it, and then
+    embed the aspects information as AspectList object into the section itself.
+
+    :param sections:  List of section that potentially contain aspects setting.
+    :return:          The new sections.
+    """
+    for _, section in sections.items():
+        section.aspects = extract_aspects_from_section(section)
+    return sections
 
 
 def merge_section_dicts(lower, higher):
@@ -385,6 +398,7 @@ def gather_configuration(acquire_settings,
         arg_list = sys.argv[1:] if arg_list is None else arg_list
     sections, targets = load_configuration(arg_list, log_printer, arg_parser,
                                            args=args)
+    aspectize_sections(sections)
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings,
                                               log_printer)
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -3,6 +3,7 @@ import os
 import sys
 from collections import OrderedDict
 
+from coalib.bearlib.aspects import AspectList
 from coalib.collecting.Collectors import collect_registered_bears_dirs
 from coala_utils.decorators import enforce_signature, generate_repr
 from coalib.misc.DictUtilities import update_ordered_dict_key
@@ -41,6 +42,39 @@ def append_to_sections(sections,
 
     sections[section_name.lower()].append(Setting(
         key, str(value), origin, from_cli=from_cli, to_append=to_append))
+
+
+def extract_aspects_from_section(section):
+    """
+    Extracts aspects and their related settings from a section and create an
+    AspectList from it.
+
+    :param section: Section object.
+    :return:        AspectList containing aspectclass instance with
+                    user-defined tastes.
+    """
+    aspects = section.get('aspects')
+    language = section.get('language')
+
+    # Skip aspects initialization if not configured in section
+    if not len(aspects):
+        return None
+
+    if not len(language):
+        raise AttributeError('Language was not found in configuration file. '
+                             'Usage of aspect-based configuration must '
+                             'include language information.')
+
+    aspect_instances = AspectList(exclude=section.get('excludes'))
+
+    for aspect in AspectList(aspects):
+        # Search all related tastes in section.
+        tastes = {name.split('.')[-1]: value
+                  for name, value in section.contents.items()
+                  if name.lower().startswith(aspect.__name__.lower())}
+        aspect_instances.append(aspect(language, **tastes))
+
+    return aspect_instances
 
 
 @generate_repr()
--- a/tests/bearlib/languages/LanguageTest.py
+++ b/tests/bearlib/languages/LanguageTest.py
@@ -1,3 +1,4 @@
+import pickle
 import unittest
 
 from coalib.bearlib.languages.Language import Language, LanguageMeta
@@ -9,6 +10,12 @@ class LanguageTest(unittest.TestCase):
         assert set(dir(Language)) == {
             l.__name__ for l in LanguageMeta.all
         }.union(type.__dir__(Language))
+
+    def test_pickle_ability(self):
+        cpp = Language['CPP']
+        cpp_str = pickle.dumps(cpp)
+        cpp_unpickled = pickle.loads(cpp_str)
+        self.assertEqual(str(cpp), str(cpp_unpickled))
 
 
 class LanguageAttributeErrorTest(unittest.TestCase):
--- a/tests/settings/SectionTest.py
+++ b/tests/settings/SectionTest.py
@@ -1,8 +1,11 @@
 import unittest
 import os
 
+from coalib.bearlib.aspects import AspectList, Root, get as get_aspect
+from coalib.bearlib.aspects.meta import issubaspect
 from coalib.misc import Constants
-from coalib.settings.Section import Section, Setting, append_to_sections
+from coalib.settings.Section import (
+    Section, Setting, append_to_sections, extract_aspects_from_section)
 from coalib.settings.ConfigurationGathering import get_config_directory
 from coalib.parsing.Globbing import glob_escape
 
@@ -204,3 +207,62 @@ class SectionTest(unittest.TestCase):
         sections['all.c.codestyle'].set_default_section(sections)
         self.assertEqual(sections['all.c.codestyle'].defaults,
                          sections['all'])
+
+    def test_extract_aspects_from_section(self):
+        section = Section('section')
+        section.append(Setting(
+            'aspects',
+            'spelling, commitmessage, methodsmell'))
+        # Custom taste for ColonExistence
+        section.append(Setting('commitmessage.shortlog_colon', 'false'))
+        section.append(Setting('language', 'py 3.4'))
+
+        aspects = extract_aspects_from_section(section)
+        spelling_instance = Root.Spelling('py 3.4')
+        colon_existence_instance = (
+            Root.Metadata.CommitMessage.Shortlog.ColonExistence(
+                'py 3.4', shortlog_colon=False))
+        method_smell_instance = Root.Smell.MethodSmell('py 3.4')
+        trailing_period_instance = (
+            Root.Metadata.CommitMessage.Shortlog.TrailingPeriod('py 3.4'))
+
+        self.assertIsInstance(aspects, AspectList)
+        self.assertEqual(aspects.get('spelling'), spelling_instance)
+        self.assertEqual(aspects.get('colonexistence'),
+                         colon_existence_instance)
+        self.assertEqual(aspects.get('methodsmell'), method_smell_instance)
+        self.assertEqual(aspects.get('TrailingPeriod'),
+                         trailing_period_instance)
+
+    def test_extract_aspects_from_section_with_exclude(self):
+        section = Section('section')
+        section.append(Setting('aspects', 'commitmessage'))
+        section.append(Setting('excludes', 'TrailingPeriod'))
+        section.append(Setting('language', 'py 3.4'))
+
+        aspects = extract_aspects_from_section(section)
+
+        self.assertTrue(issubaspect(get_aspect('trailingperiod'),
+                                    get_aspect('commitmessage')))
+        self.assertIsNone(aspects.get('trailingperiod'))
+
+    def test_extract_aspects_from_section_no_aspects(self):
+        section = Section('section')
+        self.assertIsNone(extract_aspects_from_section(section))
+
+    def test_extract_aspects_from_section_no_language(self):
+        section = Section('section')
+        section.append(Setting('aspects', 'commitmessage'))
+        with self.assertRaisesRegex(
+                AttributeError,
+                'Language was not found in configuration file. '
+                'Usage of aspect-based configuration must include '
+                'language information.'):
+            extract_aspects_from_section(section)
+
+    def test_extract_aspects_from_section_incorrect_language(self):
+        section = Section('section')
+        section.append(Setting('aspects', 'commitmessage'))
+        section.append(Setting('language', 'not a language'))
+        with self.assertRaises(AttributeError):
+            extract_aspects_from_section(section)
--- a/coalib/bearlib/aspects/__init__.py
+++ b/coalib/bearlib/aspects/__init__.py
@@ -9,6 +9,7 @@ from .base import aspectbase
 from .exceptions import (AspectTypeError, AspectNotFoundError,
                          MultipleAspectFoundError)
 from .meta import aspectclass
+from .collections import AspectList
 from .taste import Taste, TasteError
 
 # already import Root here to make it available in submodules that define
@@ -19,7 +20,8 @@ from .root import Root
 
 __all__ = ['Root', 'Taste', 'TasteError',
            'aspectclass', 'aspectbase', 'AspectTypeError',
-           'AspectNotFoundError', 'MultipleAspectFoundError']
+           'AspectNotFoundError', 'MultipleAspectFoundError',
+           'AspectList']
 
 
 class aspectsModule(ModuleType):
--- a/tests/settings/ConfigurationGatheringTest.py
+++ b/tests/settings/ConfigurationGatheringTest.py
@@ -9,12 +9,14 @@ from pyprint.NullPrinter import NullPrinter
 import pytest
 
 from coalib.misc import Constants
+from coalib.settings.Section import Section
 from coala_utils.ContextManagers import (
     make_temp, change_directory, retrieve_stdout)
 from coalib.output.printers.LogPrinter import LogPrinter
 from coala_utils.string_processing import escape
 from coalib.settings.ConfigurationGathering import (
-    find_user_config, gather_configuration, load_configuration)
+    find_user_config, gather_configuration, load_configuration,
+    aspectize_sections)
 
 
 @pytest.mark.usefixtures('disable_bears')
@@ -332,3 +334,10 @@ class ConfigurationGatheringTest(unittest.TestCase):
         with retrieve_stdout() as stdout:
             load_configuration(['--no-config'], self.log_printer)
             self.assertNotIn('WARNING', stdout.getvalue())
+
+    def test_aspectize_sections(self):
+        section = Section('test')
+        sections = {'test': section}
+        aspectize_sections(sections)
+
+        self.assertIsNone(sections['test'].aspects)


adhikasp:
@userzimmermann @Asnelchristian

coalib/processes/BearRunning.py
```diff
@@ -469,7 +488,7 @@ def run_global_bears(message_queue,
                                      global_result_dict))
             bearname = bear.__class__.__name__
             result = run_global_bear(message_queue, timeout, bear, dep_results,
-                                     debug=debug)
+                                     aspects, debug=debug)
```
jayvdb: insert a new line with `aspects,` , so the existing line debug=debug doesnt need to be modified.
New parameters will be added before the debug parameter.

coalib/processes/Processing.py
```diff
@@ -436,7 +438,8 @@ def instantiate_processes(section,
                         'message_queue': message_queue,
                         'control_queue': control_queue,
                         'timeout': 0.1,
-                        'debug': debug}
+                        'debug': debug,
```
jayvdb: dont modify the debug line; add the `aspects` line above it.

coalib/settings/Section.py
```diff
@@ -43,6 +46,53 @@ def append_to_sections(sections,
...
+    """
+    Extract aspects and its related setting from a section and create an
+    aspectlist from it.
+
+    :section: Section object.
+    :return:  aspectlist containing aspectclass instance with user-defined
+              tastes.
+    """
+    if not len(section.get('aspects')):
+        # Off switch for aspect-based configuration
```
jayvdb: this comment is not clear and simple.
And `None` should be described in the docstring.

And add a blank line in the code after this `if` , like you have done after the `raise` below.

coalib/settings/Section.py
```diff
@@ -43,6 +46,53 @@ def append_to_sections(sections,
...
+            try:
+                tastes_to_init[taste] = section[taste]
+            except IndexError:
+                pass
+        aspects_instance.append(aspect(language, **tastes_to_init))
+
+    # TODO exclude aspects
+
+    # We don't need it again, so remove it to avoid name conflict.
+    section.delete_setting('aspects')
```
jayvdb: This should not be in an "extract .." method.

The caller should delete the section.

coalib/processes/BearRunning.py
```diff
@@ -167,6 +167,7 @@ def run_local_bear(message_queue,
                    file_dict,
                    bear_instance,
                    filename,
+                   aspects,
```
jayvdb: This would be a breaking change.  Instead, use `aspects=None`.

coalib/processes/BearRunning.py
```diff
@@ -200,7 +203,8 @@ def run_local_bear(message_queue,
     kwargs = {'dependency_results':
               get_local_dependency_results(local_result_list,
                                            bear_instance),
-              'debug': debug}
+              'debug': debug,
+              'aspects': aspects}
```
jayvdb: add aspects before debug, so that debug line doesnt need to be modified

coalib/processes/BearRunning.py
```diff
@@ -245,7 +252,8 @@ def run_global_bear(message_queue,
         return None
 
     kwargs = {'dependency_results': dependency_results,
-              'debug': debug}
+              'debug': debug,
+              'aspects': aspects}
```
jayvdb: switch debug and aspects

coalib/processes/BearRunning.py
```diff
@@ -393,6 +405,7 @@ def run_local_bears(filename_queue,
                     local_bear_list,
                     local_result_dict,
                     control_queue,
+                    aspects,
```
jayvdb: `=None`

coalib/processes/BearRunning.py
```diff
@@ -259,6 +267,7 @@ def run_local_bears_on_file(message_queue,
                             local_result_dict,
                             control_queue,
                             filename,
+                            aspects,
```
jayvdb: `=None`

coalib/processes/BearRunning.py
```diff
@@ -438,6 +454,7 @@ def run_global_bears(message_queue,
                      global_bear_list,
                      global_result_dict,
                      control_queue,
+                     aspects,
```
jayvdb: `=None`

coalib/processes/Processing.py
```diff
@@ -19,6 +19,7 @@
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.results.SourceRange import SourceRange
 from coalib.settings.Setting import glob_list
+from coalib.settings.Section import extract_aspects_from_section
```
Asnelchristian: you can merge this with the previous line right
```
from coalib.settings.Section import (
    glob_list, extract_aspects_from_section)
```
this is just a preference though :sweat_smile:

coalib/settings/Section.py
```diff
@@ -43,6 +46,53 @@ def append_to_sections(sections,
         key, str(value), origin, from_cli=from_cli, to_append=to_append))
 
 
+def extract_aspects_from_section(section):
+    """
+    Extract aspects and its related setting from a section and create an
+    aspectlist from it.
+
+    :section: Section object.
```
Asnelchristian: `param section`

coalib/settings/Section.py
```diff
@@ -43,6 +46,53 @@ def append_to_sections(sections,
         key, str(value), origin, from_cli=from_cli, to_append=to_append))
 
 
+def extract_aspects_from_section(section):
+    """
+    Extract aspects and its related setting from a section and create an
```
Asnelchristian: Extract`s` .... and `their` ..... `settings`

coalib/settings/Section.py
```diff
@@ -43,6 +46,53 @@ def append_to_sections(sections,
...
+
+    aspects_instance = aspectlist()
+    for aspect in aspects:
+        tastes_to_init = dict()
+        for taste in aspect.tastes:
+            # TODO handle duplicate taste meant for different aspect
+            try:
+                tastes_to_init[taste] = section[taste]
+            except IndexError:
+                pass
```
Asnelchristian: this means the default value available for the taste will be used right?

tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -9,10 +9,18 @@
...
 
     def test__init__(self):
-        with pytest.raises(aspectTypeError) as exc:
+        list_of_aspect = aspectlist(['CommitMessage.Shortlog',
+                                     'CommitMessage.Body'])
+        mix_of_aspect = aspectlist(['CommitMessage.Shortlog',
+                                    Metadata.CommitMessage.Body])
+        assert isinstance(list_of_aspect, aspectlist)
+        assert list_of_aspect[0] is Metadata.CommitMessage.Shortlog
+        assert list_of_aspect[1] is Metadata.CommitMessage.Body
```
Asnelchristian: the doctest done in the aspectlist class definition  should be enough right ?

coalib/processes/Processing.py
adhikasp: It was different modules 😅

coalib/settings/Section.py
adhikasp: Yes 👍

tests/bearlib/aspects/CollectionsTest.py
adhikasp: This part of commit is already implemented in other PR and will be removed in next iteration

coalib/processes/Processing.py
Asnelchristian: my bad :sweat_smile:

tests/processes/BearRunningTest.py
```diff
@@ -123,6 +123,7 @@ def setUp(self):
         self.global_result_dict = manager.dict()
         self.message_queue = queue.Queue()
         self.control_queue = queue.Queue()
+        self.aspects = None
```
adhikasp: Unneeded changes

tests/processes/ProcessingTest.py
```diff
@@ -218,6 +218,20 @@ def test_raw_run(self):
                                              'test message',
                                              self.unreadable_path)])
 
+    def test_run_with_aspect(self):
```
adhikasp: Actually this test objective is to cover https://github.com/coala/coala/pull/4397/files#diff-20ec48fe3730a3eba01621d03049f9dbR431 which I don't sure which function I can most independently do unit test for (or whether https://github.com/coala/coala/pull/4397/files#diff-20ec48fe3730a3eba01621d03049f9dbR431 is the best place to inject those aspects).

If anyone could point me better way to test those line, maybe we can avoid PickleError (or maybe this PickleError is also a real error that will shown up in normal coala run with aspects in the future).

coalib/bearlib/languages/Language.py
```diff
@@ -164,6 +164,7 @@ def __getattr__(self, item):
                       ('versions', 'aliases')))
 
             class Sub(cls, metaclass=SubLanguageMeta):
+                __module__ = 'coalib.bearlib.languages.Language'
```
adhikasp: Temporary workaround for PickleError. Doesn't work anyway :(

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+    """
+    # Skip aspects initialization if not configured in section
+    if not len(section.get('aspects')):
+        return None
+
+    if not len(section.get('language')):
+        raise AttributeError('Language was not found in configuration file. '
+                             'Usage of aspect-based configuration must include'
+                             'language information.')
+        # TODO acquire language from interactive input instead of just raising
```
jayvdb: Noooo ... never mix parsing/extraction and UI.  extraction methods should fail if they cant extract, and let the caller do any UI *if* they want a fallback.

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+    """
+    Extracts aspects and their related settings from a section and create an
+    aspectlist from it.
+
+    :param section: Section object.
+    :return:        aspectlist containing aspectclass instance with
+                    user-defined tastes.
+    """
+    # Skip aspects initialization if not configured in section
+    if not len(section.get('aspects')):
```
jayvdb: later in the function, you use the value of `section.get('aspects')` -- so, save it in a variable.

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+        return None
+
+    if not len(section.get('language')):
```
jayvdb: you have saved `section.get('language')` in a variable later in this function.  Move that variable here, so you dont call `.get` twice.

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+                             'language information.')
+        # TODO acquire language from interactive input instead of just raising
+        # error.
+
+    aspects = AspectList(section.get('aspects'))
+    language = section.get('language')
+    try:
+        Language[language]
+    except AttributeError:
+        raise AttributeError("Language '{}' is not a valid language name or "
```
jayvdb: You catch and raise `AttributeError`.  What is wrong with the original `AttributeError`?  Maybe improve it, rather than creating a new exception?

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+                             'is not recognized by coala.'.format(language))
+
+    aspects_instance = AspectList()
```
jayvdb: I think `aspect_instances` is better.
(a collection variable name should always end in `s`)

tests/settings/SectionTest.py
```diff
@@ -204,3 +206,52 @@ def test_set_default_section(self):
...
+
+    def test_extract_aspects_from_section(self):
+        section = Section('section')
+        # No aspects setting
+        self.assertIsNone(extract_aspects_from_section(section))
+
+        section.append(Setting(
+            'aspects',
+            'spelling, commitmessage, methodsmell'))
+        section.append(Setting(
```
jayvdb: The first test method shouldnt use `excludes`.  A second test method should use `excludes`.

coalib/settings/Section.py
jayvdb: As `language` now has a strict set of values it must be contain, the validation of language should be a separate function, with separate tests.

Quick fix: *iff* you need to raise a different exception than `Language` provides, create a separate function in this commit.

A background task to provide a proper solution: https://github.com/coala/coala/issues/4463

coalib/settings/Section.py
```diff
@@ -43,6 +45,52 @@ def append_to_sections(sections,
...
+        tastes_to_init = dict()
+        for taste in aspect.tastes:
+            # TODO handle duplicate taste meant for different aspect
+            try:
+                # Search for user defined taste value
+                tastes_to_init[taste] = section[taste]
+            except IndexError:
+                # Use default taste value
+                pass
+        aspects_instance.append(aspect(language, **tastes_to_init))
```
jayvdb: `language` here is a `str`??  But earlier in this function you have done `Language(language)`, and discard the Language object.
Can't you use that Language object here?

tests/processes/ProcessingTest.py
adhikasp: Resolved. Make Language pickleable.

coalib/settings/Section.py
adhikasp: This one is outdated. Actually `Language(language)` is to validate the string for this. But then I do the try catch here.

coalib/settings/Section.py
adhikasp: I don't need different exception. It just the original error is very cryptic and doesn't give helpful message to user.

adhikasp:
I found better place to initialize the aspects and avoid duplication/unneeded edit in my next patch. I will submit update later tonight.

tests/bearlib/languages/LanguageTest.py
```diff
@@ -27,3 +27,9 @@ def test_invalid_attribute(self):
     def test_attribute_list_empy(self):
         with self.assertRaisesRegex(AttributeError, 'no available attribute'):
             self.lang_unknown.not_an_attribute
+
+    def test__reduce__(self):
+        kallable, args = self.lang_cpp.__reduce__()
```
jayvdb: This is testing *your* code, but it is not testing pickle-ability.
You want to use `pickle` to testing it can be pickled.
see https://github.com/wikimedia/pywikibot-core/blob/master/tests/page_tests.py#L512

coalib/settings/ConfigurationGathering.py
```diff
@@ -394,13 +394,14 @@ def gather_configuration(acquire_settings,
...
+    local_bears, global_bears, aspects = fill_settings(sections,
+                                                       acquire_settings,
+                                                       log_printer)
     save_sections(sections)
     warn_nonexistent_targets(targets, sections, log_printer)
 
     return (sections,
             local_bears,
             global_bears,
+            aspects,
```
jayvdb: This is a breaking change to the public coala API.
A very large one.  coala-quickstart is broken by this.
I dont recall this being in an earlier patch I reviewed.

coalib/settings/SectionFilling.py
```diff
@@ -43,8 +50,9 @@ def fill_settings(sections, acquire_settings, log_printer):
 
         local_bears[section_name] = section_local_bears
         global_bears[section_name] = section_global_bears
+        aspects[section_name] = section_aspects
 
-    return local_bears, global_bears
+    return local_bears, global_bears, aspects
```
jayvdb: This is a breaking change to the public coala API.
A very large one.
I dont recall this being in an earlier patch I reviewed.

coalib/settings/ConfigurationGathering.py
adhikasp: IDK it wiill be bad :(

This is required for bear collection method and processing.py to have access to the same AspectList instance, but It's not mandatory. I can initialize AspectList 2 times  from `section` and deleting the `section['aspects']` in the last usage.

coalib/settings/SectionFilling.py
adhikasp: Same as above

coalib/settings/ConfigurationGathering.py
jayvdb: a tuple return value for a public API is always bad news for extensibility.

one way to work around that is ...

You can create a new method `fill_settings_new` & `gather_configuration_new` (obviously use a different naming convention) which uses the existing `fill_settings` & `gather_configuration` but also does additional logic and has different return values.

coalib/settings/SectionFilling.py
```diff
@@ -20,14 +21,20 @@ def fill_settings(sections, acquire_settings, log_printer):
...
+                             list.
     """
     local_bears = {}
     global_bears = {}
+    aspects = {}
 
     for section_name, section in sections.items():
+        section_aspects = extract_aspects_from_section(section)
+        if section_aspects is not None:
+            # Not required anymore and avoiding name conflict.
```
adhikasp: It is to avoid conflict when running bear. `Bear::run_bear_from_section` will override `aspects` from here with what it find from `section['aspects']`.

coalib/bearlib/languages/Language.py
```diff
@@ -59,6 +59,18 @@ def parse_lang_str(string):
     return name, versions
 
 
+def parse_language(string):
+    """
+    Wrapper for ``Language[]`` so it can be called like a normal function.
```
userzimmermann: Not necessary :) Just use `Language.__getitem__`

coalib/bearlib/languages/Language.py
```diff
@@ -416,6 +428,9 @@ def __contains__(self, item):
         return (type(self) is type(item)
                 and set(item.versions).issubset(set(self.versions)))
 
+    def __reduce__(self):
+        return (parse_language, (str(self),))
```
userzimmermann: `return type(self).__getitem__, (str(self), )`

coalib/bears/Bear.py
```diff
@@ -298,6 +298,10 @@ def execute(self, *args, debug=False, **kwargs):
                     not self.BEAR_DEPS):
                 del kwargs['dependency_results']
 
+            # Don't pass `aspects` if the bear doesn't support it.
+            if not hasattr(self, 'aspects') and 'aspects' in kwargs:
+                del kwargs['aspects']
```
userzimmermann: With `kwargs.pop('aspects', None)` you don't need to check for `'aspects' in kwargs`

coalib/settings/Section.py
```diff
@@ -43,6 +44,40 @@ def append_to_sections(sections,
...
+    if not len(language):
+        raise AttributeError('Language was not found in configuration file. '
+                             'Usage of aspect-based configuration must '
+                             'include language information.')
+
+    aspect_instances = AspectList(exclude=section.get('excludes'))
+
+    for aspect in AspectList(aspects):
+        tastes_name = [k for k, _ in section.contents.items() if
+                       aspect.__name__.lower() in k.lower()]
```
userzimmermann: `k.lower().startswith(aspect.__name__.lower())`

coalib/settings/Section.py
```diff
@@ -43,6 +44,40 @@ def append_to_sections(sections,
...
+        tastes_to_init = {}
+        for name in tastes_name:
+            tastes_to_init[name.split('.')[-1]] = section[name]
```
userzimmermann:
```python
tastes = {name.split('.')[-1]: value
          for name, value in section.contents.items()
          if name.lower().startswith(aspect.__name__.lower())}
```

coalib/settings/Section.py
```diff
@@ -43,6 +44,40 @@ def append_to_sections(sections,
...
+
+        aspect_instances.append(aspect(language, **tastes_to_init))
+    return aspect_instances
```
userzimmermann:
```python
return AspectList((
    aspect(language, **{
        name.split('.')[-1]: value
        for name, value in section.contents.items()
        if name.lower().startswith(aspect.__name__.lower())
    })
    for aspect in AspectList(aspects)
), exclude=section.get('exclude'))

Comment (adhikasp) [inline coalib/settings/Section.py]: I think this one is far too compact and not readable :/

File: coalib/settings/SectionFilling.py [hunk 36]
```diff
@@ -28,6 +29,7 @@ def fill_settings(sections, acquire_settings, log_printer):
     global_bears = {}
 
     for section_name, section in sections.items():
+        section.aspects = extract_aspects_from_section(section)
```
adhikasp: I am not sure whether to declare `section.aspects` in the class Section `__init__` declaration because all section object is guaranteed to run this `fill_settings` function anyway.

userzimmermann: It's fine like this! Since `Section.__init__` also doesn't fill `.contents`. So it's also a kind of verification that the `Section` instance is finished. Otherwise you'll get an exception on accessing `section.aspects`

coalib/settings/Section.py
userzimmermann: @adhikasp Your decision ;) Only keep in mind that for proper use we need to override `AspectList.append` and all other manipulation methods to check for correct aspect input

userzimmermann: We can then simply `@aspectize` `AspectList.append` after https://github.com/coala/coala/issues/4527 is done

cc @jayvdb

userzimmermann:
ack https://github.com/coala/coala/pull/4397/commits/c9bfea42348032e4e031d5b2bb4b19e4779758b4 https://github.com/coala/coala/pull/4397/commits/3d24704f81773b934b56782ad292023775f76a26 https://github.com/coala/coala/pull/4397/commits/ca72d33101915108e717e789cf88eec412edc7c3 https://github.com/coala/coala/pull/4397/commits/6b18cb5217970fafb304be508883ce4bb85967af

userzimmermann:
@jayvdb @Asnelchristian Should be approvable! :)

@adhikasp Please rebase if necessary

tests/settings/SectionFillingTest.py
```diff
@@ -61,6 +62,19 @@ def test_fill_settings(self):
         self.assertEqual(bool(self.section['use_spaces']), True)
         self.assertEqual(len(self.section.contents), 3)
 
+    def test_fill_settings_aspects(self):
+        sections = {'test': self.section}
+        sections['test'].append(Setting('aspects', 'commitmessage'))
+        sections['test'].append(Setting('language', 'py'))
+
+        _, _ = fill_settings(sections,
```
jayvdb: if you are not using any of the returned values, you dont need to store them in `_`

coalib/settings/Section.py
userzimmermann: @adhikasp @jayvdb We have to raise a better exception in `LanguageMeta.__getitem__`!

jayvdb:
> section.aspect references should not be added in fill_settings

We have settled on this being the best spot for it , at least for now.

jayvdb:
reack 1622f30 82d9b00

jayvdb:
ack 1ce4dd9

jayvdb:
@rultor merge

Issue #4351, opened 2017-06-09
Nextgen Core: Port core functions to class methods


Makman2:
This could be a performance gain as we don't have to pass always the same variables around. Instead we can access them from `self`.

Makman2:
How this could look like can be seen here: https://github.com/coala/coala/tree/Makman2/NEXTGEN-to-classes

shreyans800755:
@Makman2 
I would like to work on this issue.
Just needed clarification before I start working, correct me if I'm wrong:
We are supposed to move global functions core/Core.py to a class, to optimize it. As per my analysis, there should be 2 classes, one for group and one for the rest of the functions.

Makman2:
Just one class is enough. I think we only need this for the functions that always pass some states around, like `running_tasks` or `dependency_tracker`. Everything else that is modular and can work on it's own should be still kept inside the module directly :)

So a list of functions I would consider to move to a class:
- `cleanup_bear`
- `schedule_bears`
- `finish_task`

And we can move maybe some mechanics of `run` into a class method too, so we can easily start a session with an instance easily.

About the name of the class: I don't know a proper name, but I have the feeling that calling it somehow a "session" or so is quite fitting^^

shreyans800755:
@Makman2 I'm not able to find the references of any of the functions `initialize_dependencies` and `run` from `Core.py` (apart from CoreTest). Where are these functions used ?

Makman2:
`initialize_dependencies` is used inside `run`, and `run` shall be used outside. `run` is the main entry-point for running bears. Everything else should not be touched by external users.

shreyans800755:
So, right now, `run` is not used anywhere else ?

Makman2:
No, not yet. It will be used by coala once the last step is completed, the integration :)

shreyans800755:
> initialize_dependencies is used inside run, and run shall be used outside. run is the main entry-point for running bears. Everything else should not be touched by external users.

In that case, why do we have different tests for both `initialize_dependencies` and `run` ? Shouldn't the tests only for `run` should suffice ?

Makman2:
Because `initialize_dependencies` features much functionality, this component is tested on its own. It's easier to test than through `run`.

Issue #4349, opened 2017-06-09
Nextgen Core: Allow to pass different executors


Makman2:
This will be needed for the debugging mode of coala.

`core.run` shall take as a parameter the executor to run. By default (`None`) we create a `ProcessPoolExecutor` without arguments (so it picks up the number of CPU cores of the current machine automatically as the number of processes to launch).

Tests need to be written too.

Also I plan in the far future to run coala on distributed systems, currently I have Apache Mesos / DC/OS in mind.

Makman2:
Unassigned due to inactivity.

PR #4414, opened 2017-06-26
Core: Port core functions to class

shreyans800755:
Port core functions to Session class.

Closes https://github.com/coala/coala/issues/4351

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


coalib/core/Core.py
```diff
@@ -327,34 +147,172 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+            :param bears:
+                The bear instances to run.
+            :param result_callback:
+                A callback function which is called when results are available.
+                Must have following signature::
+
+                    def result_callback(result):
+                        pass
+            """
+        cls.bears = bears
```
Makman2: ahh no make it instance bound, that is way safer and then we can execute multiple sessions at once :D

Makman2: So instantiate a new `Session` class passing those fields :)

tests/core/CoreTest.py
```diff
@@ -317,7 +317,8 @@ def test_correct_reuse_of_implicitly_instantiated_dependency(self):
 
     def test_empty_case(self):
         # Test totally empty case.
-        dependency_tracker, bears_to_schedule = initialize_dependencies(set())
+        dependency_tracker, bears_to_schedule = initialize_dependencies(
+            set())
```
Makman2: unrelated change ;)

tests/core/CoreTest.py
```diff
@@ -4,7 +4,7 @@
 
 from coalib.settings.Section import Section
 from coalib.core.Bear import Bear
-from coalib.core.Core import initialize_dependencies, run
+from coalib.core.Core import Session, initialize_dependencies
```
Makman2: if possible maintain alphabetical order of imports ;)
```
from coalib.core.Core import initialize_dependencies, Session
```

coalib/core/Core.py
```diff
@@ -327,34 +147,172 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+
+            :param bears:
+                The bear instances to run.
+            :param result_callback:
+                A callback function which is called when results are available.
+                Must have following signature::
+
+                    def result_callback(result):
+                        pass
+            """
```
Makman2: misaligned docstring

coalib/core/Core.py
```diff
@@ -327,34 +147,172 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
-    dependency_tracker, bears_to_schedule = initialize_dependencies(bears)
-
-    # Let's go.
-    schedule_bears(bears_to_schedule, result_callback, dependency_tracker,
-                   event_loop, {}, executor)
-    try:
-        event_loop.run_forever()
-    finally:
-        event_loop.close()
+class Session:
```
Makman2: a class docstring would be good :+1:
also with a simple usage example :+1:

coalib/core/Core.py
```diff
@@ -254,8 +76,6 @@ def initialize_dependencies(bears):
     dependencies are included so the dependency chain can be processed
     correctly.
 
-    :param bears:
-        The set of instantiated bears to run that serve as an entry-point.
```
Makman2: why that change? the param is still there and needed :O

shreyans800755: By mistake

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+        # FIXME   systems (for example Mesos).
+
+        # Set up event loop and executor.
+        self.event_loop = asyncio.SelectorEventLoop()
+        self.executor = concurrent.futures.ProcessPoolExecutor()
+        self.running_tasks = {}
+
+        # Initialize dependency tracking.
+        self.dependency_tracker, bears_to_schedule = initialize_dependencies(
+            self.bears)
```
Makman2: maybe moving everything up to this statement into `__init__`? Traditionally member variables are set inside `__init__` to keep everything in one place.

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+        self.schedule_bears(bears_to_schedule)
+        try:
+            self.event_loop.run_forever()
+        finally:
+            self.event_loop.close()
+
+    def schedule_bears(self, bears):
+        """
+        Schedules the tasks of bears to the given executor and runs them on the
+        given event loop.
```
Makman2: adapt docstrings ;) you don't have to pass any parameters except bears now.

-->
```
Schedules the tasks of bears.
```

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+                logging.debug('Scheduled {!r} (tasks: {})'.format(bear,
+                                                                  len(tasks)))
+
+                if not tasks:
+                    # We need to recheck our runtime if something is left to
+                    # process, as when no tasks were offloaded the event-loop
+                    # could hang up otherwise.
+                    self.cleanup_bear(bear)
+
+    def cleanup_bear(self, bear):
```
Makman2: I'm thinking of prepending an underscore to the method names, to mark them as "private".
What do you think?

coalib/core/Core.py
shreyans800755: Good point. What about `executor' ? There comments in the code indicating that we'll allow custom executor in future. In that case, if we move `self.executor = concurrent.futures.ProcessPoolExecutor()` in constructor, then it will add additional cost if custom executor is used. I don't know how much expensive this call is.
If it is not much expensive, we can do it constructor. Otherwise, we can do have an param in `run` function called `executor` and code in the function should look something like `self.executor = executor if executor is not None else concurrent.futures.ProcessPoolExecutor()`.
But that change is not in scope of this PR.
What do you think ?

coalib/core/Core.py
```diff
@@ -256,6 +78,7 @@ def initialize_dependencies(bears):
 
     :param bears:
         The set of instantiated bears to run that serve as an entry-point.
+
```
Makman2: unnecessary change ;)

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
 
 
-def run(bears, result_callback):
+class Session:
     """
-    Runs a coala session.
+    Maintains Session for a coala execution. For each session, there are set
+    of bears along with callback function, which is run when results are
+    available. On running a session, first its dependencies are initalized and
+    all the bears are scheduled to run in appropriate order.
```
Makman2: a bit grammatical polishing ;)
-->
```
Maintains a session for a coala execution. For each session, there are set bears to run along with a callback function, which is run when results are available.
```
About this part:
```
On running a session, first its dependencies are initalized and
all the bears are scheduled to run in appropriate order.
```
I would remove that altogether, as 1. it's implementation detail 2. it's expected from the user that this works, so imo it requires no mention. And `all the bears are scheduled to run in appropriate order` is technically not completely correct ;) Bears that depend on another bear aren't scheduled until the bear dependency has finished. So in the beginning not all bears may be scheduled :)

tests/core/CoreTest.py
```diff
@@ -383,7 +383,8 @@ def execute_run(bears):
         def on_result(result):
             results.append(result)
 
-        run(bears, on_result)
+        session = Session(bears, on_result)
+        session.run()
```
Makman2: what do you think about providing a shortcut? So we place a `run` function into the core module, that actually does this:
```
def run(bears, on_result):
    session = Session(bears, on_result)
    session.run()
```
For users this means one statement less (avoiding the `Session` instantiation) and we can maybe add more functions later (if that makes sense of course).

Makman2: If you want to do this (I think we should do this :D), then mention this inside the commit message :+1:

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
 
-    :param bears:
-        The bear instances to run.
-    :param result_callback:
-        A callback function which is called when results are available. Must
-        have following signature::
+    Ex: If BearA requires BearB as dependencies, then, on running BearA,
+        first BearB will be executed, and after that, BearA will start running.
```
shreyans800755: Its actually `Example`. I think you're right. I'll just remove it.

coalib/core/Core.py
```diff
@@ -327,34 +150,177 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+    def __init__(self, bears, result_callback):
+        """
+        :param bears:
+            The bear instances to run.
+        :param result_callback:
+            A callback function which is called when results are available.
+            Must have following signature::
 
             def result_callback(result):
                 pass
```
Makman2: I think you need to indent this example too by 4 spaces :+1:

tests/core/CoreTest.py
shreyans800755: I would rather add an extra parameter in `execute_run()` function related to `on_result`. So, that it can be reused instead of having different functions.

coalib/core/Core.py
Makman2: Like you said, I think we will implement it by just using an additional parameter `executor`.
But separate PR / separate commit, as this is another issue :) (https://github.com/coala/coala/issues/4349)

coalib/core/Core.py
```diff
@@ -327,34 +149,176 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+
+    def run(self):
+        """
+        Runs a coala session.
+        """
+        # FIXME Allow to pass different executors nicely, for example to execute
+        # FIXME   coala with less cores, or to schedule jobs on distributed
+        # FIXME   systems (for example Mesos).
+
+        # Let's go.
```
Makman2: I think this comment is not needed any more :P

coalib/core/Core.py
```diff
@@ -327,34 +149,176 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+        self.result_callback = result_callback
+
+        # Set up event loop and executor.
+        self.event_loop = asyncio.SelectorEventLoop()
+        self.executor = concurrent.futures.ProcessPoolExecutor()
+        self.running_tasks = {}
+
+        # Initialize dependency tracking.
+        self.dependency_tracker, self.bears_to_schedule =\
+            initialize_dependencies(self.bears)
```
Makman2: please use braces instead of `\` :+1:

tests/core/CoreTest.py
```diff
@@ -377,13 +377,17 @@ def setUp(self):
...
 
         def on_result(result):
             results.append(result)
 
-        run(bears, on_result)
+        if result_callback is None:
+            result_callback = on_result
+
+        session = Session(bears, result_callback)
+        session.run()
```
Makman2: have you read my comment about providing a convenience function?

coalib/core/Core.py
```diff
@@ -327,34 +149,176 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+
+    def run(self):
+        """
+        Runs a coala session.
```
Makman2: maybe `a` --> `the`?

coalib/core/Core.py
```diff
@@ -327,34 +149,176 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
     """
-    Runs a coala session.
-
-    :param bears:
-        The bear instances to run.
-    :param result_callback:
-        A callback function which is called when results are available. Must
-        have following signature::
+    Maintains a session for a coala execution. For each session, there are set
+    of bears to run along with a callback function, which is run when results
```
Makman2: ah, we should substitute this:

"which is ~~run~~ called when results are available."

Avoiding using `run` twice.

tests/core/CoreTest.py
shreyans800755: Oh I misunderstood your comment.

tests/core/CoreTest.py
shreyans800755: Misunderstood it on first go. Corrected. Additionally, I'm also returning `Session` object, just in case, if its needed further in code using `Session`.

Asnelchristian:
As far as i am concern this is good, just address the few things i pointed out and we will merge this :tada: , marking WIP in the meantime :smile:

coalib/core/Core.py
```diff
@@ -327,34 +149,192 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+                    # FIXME   complete event-loop traceback.
+                    logging.error(
+                        'An exception was thrown during result-handling.',
+                        exc_info=ex)
+
+
 def run(bears, result_callback):
     """
-    Runs a coala session.
+    It initiates a session with the given parameters and runs it.
```
Makman2: ~~It initiates~~ Initiates a session with ...

:smile:

tests/core/CoreTest.py
```diff
@@ -514,7 +517,7 @@ def test_run_result_handler_exception(self):
         on_result = unittest.mock.Mock(side_effect=ValueError)
 
         with self.assertLogs(logging.getLogger()) as cm:
-            run({bear}, on_result)
+            self.execute_run({bear}, on_result)
```
Makman2: is that change necessary?

Makman2: (also the above ones)

coalib/core/Core.py
```diff
@@ -327,34 +149,192 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
     return dependency_tracker, bears
 
 
+class Session:
+    """
+    Maintains a session for a coala execution. For each session, there are set
+    of bears to run along with a callback function, which is called when
+    results are available.
+
+    If BearA requires BearB as dependencies, then, on running BearA,
```
Makman2: as **dependency**

Makman2: you maybe want to prepend some general statement before the example, like
```
Dependencies of bears (provided via ``bear.BEAR_DEPS``) are automatically handled.
If BearA requires ...
```

Makman2: you maybe want to prepend some general statement before the example, like

Dependencies of bears (provided via ``bear.BEAR_DEPS``) are automatically handled.
If BearA requires ...

coalib/core/Core.py
```diff
@@ -327,34 +149,192 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+
+
 def run(bears, result_callback):
     """
-    Runs a coala session.
+    It initiates a session with the given parameters and runs it.
 
     :param bears:
-        The bear instances to run.
+            The bear instances to run.
```
Makman2: it's even unnecessary, the additional indentation is not needed^^ (and also inconsistent)

tests/core/CoreTest.py
shreyans800755: Not really. This is just to make consistent usage of execute_run every time.

shreyans800755: So, that every time, we just have to call `execute_run` even if the definition of run change in future.

Makman2: actually then this is a good test we should keep :)

Makman2: but if you want to supply a custom callback, really use `run` directly. `execute_run` was written to be able to retrieve the results for assertion. Putting a custom callback into `execute_run` turns this into a simple `run(bear, custom_callback)` call (which is done here)

Makman2: However if you still want it, please file an issue :)

shreyans800755: Nah, I think its better to remove it. Doesn't make sense to put an unnecessary overhead.

Makman2:
https://github.com/coala/coala/pull/4525 takes over.

Issue #4438, opened 2017-07-02
Add AspectList.exclude


adhikasp:
Extend AspectList class to have `excludes` attribute that hold list of aspect that marked as exception of the list. Trying to access aspect or subaspect of item in `excludes` will always give False/None.

```.python
>>> Metadata not in AspectList([Root], excludes=[Metadata])
True
>>> AspectList([Root], excludes=[Metadata]).get('metadata')
None
```

This feature will be used to excluding an aspect (to "deactivate") that was defined in user's coala configuration.

area/aspects

userzimmermann:
@adhikasp @jayvdb Like mentioned in https://github.com/coala/coala/pull/4439#pullrequestreview-48442715 : Better `exclude` instead of `excludes`

PR #4427, opened 2017-06-30
aspectbase: Recursively instance aspect children

adhikasp:
Modify aspectbase constructor to recursively init their children
subaspects.

Note that accessing child instance MUST be done through `.subaspects`
dict.

```python
>>> commit_msg = coala_aspects['CommitMessage']('py')
>>> commit_msg
<...Metadata.CommitMessage object at 0x...>
>>> commit_msg.subaspects
{'Emptiness': <...CommitMessage.Emptiness object at 0x...>,
'Shortlog': <...CommitMessage.Shortlog object at 0x...>, ...}
>>> commit_msg.subaspects['Emptiness'] # Got instance
<..CommitMessage.Emptiness object at 0x...>
>>> commit_msg.Emptiness # Got aspectclass instead
<aspectclass 'Root.Metadata.CommitMessage.Emptiness'>
```

We can also set children taste through its parent.

```python
>>> commit_msg = coala_aspects['CommitMessage'] ('py', shortlog_tense='past')
>>> commit_msg.subaspects['Shortlog'].subaspects['Tense']
<Tense object(shortlog_tense='past') at 0x...>
```

Fixes https://github.com/coala/coala/issues/4388

area/aspects @userzimmermann

PR #4439, opened 2017-07-02
AspectList: Add ``exclude`` attribute

adhikasp:
Extend AspectList class to have `excludes` attribute that hold list of
aspect that marked as exception of the list. Trying to access aspect
or subaspect of item in `excludes` will always give False/None.

```.python
>>> Metadata not in AspectList([Root], excludes=[Metadata])
True
>>> AspectList([Root], excludes=[Metadata]).get('metadata')
None
```

This feature will be used to excluding ("deactivate") an aspect that
was defined in user's coala configuration.

Closes https://github.com/coala/coala/issues/4438

Note that this PR is dependent (and blocked) by https://github.com/coala/coala/pull/4427 and contain commit from that PR too.

### DIFF
--- a/coalib/bearlib/aspects/collections.py
+++ b/coalib/bearlib/aspects/collections.py
@@ -7,7 +7,7 @@ class AspectList(list):
     List-derived container to hold aspects.
     """
 
-    def __init__(self, seq=()):
+    def __init__(self, seq=(), exclude=None):
         """
         Initialize a new AspectList.
 
@@ -21,14 +21,18 @@ class AspectList(list):
 
         :param seq: A sequence containing either aspectclass, aspectclass
                     instance, or string of partial/full qualified aspect name.
+        :param exclude: A sequence of either aspectclass or string of aspect
+                        name that marked as excluded from the list.
         """
         super().__init__((item if isaspect(item) else
                           coalib.bearlib.aspects[item] for item in seq))
 
+        self.exclude = AspectList(exclude) if exclude is not None else []
+
     def __contains__(self, aspect):
         for item in self:
             if issubaspect(aspect, item):
-                return True
+                return aspect not in self.exclude
         return False
 
     def get(self, aspect):
@@ -42,6 +46,8 @@ class AspectList(list):
         """
         if not isaspect(aspect):
             aspect = coalib.bearlib.aspects[aspect]
+        if aspect in self.exclude:
+            return None
         try:
             return next(filter(None, (item.get(aspect) for item in self)))
         except StopIteration:
--- a/tests/bearlib/aspects/CollectionsTest.py
+++ b/tests/bearlib/aspects/CollectionsTest.py
@@ -10,6 +10,18 @@ from coalib.bearlib.aspects.Metadata import Metadata
 
 class AspectListTest(unittest.TestCase):
 
+    def setUp(self):
+        self.aspectlist_excludes = AspectList(
+            seq=[Metadata.CommitMessage.Shortlog,
+                 Metadata.CommitMessage.Body],
+            exclude=[Metadata.CommitMessage.Shortlog.TrailingPeriod,
+                     Metadata.CommitMessage.Body.Existence])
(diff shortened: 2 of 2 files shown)


tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -10,6 +10,17 @@
 
 class AspectListTest(unittest.TestCase):
 
+    def setUp(self):
+        self.aspectlist_excludes = AspectList(
+            [Metadata.CommitMessage.Shortlog, Metadata.CommitMessage.Body],
+            [Metadata.CommitMessage.Shortlog.TrailingPeriod,
```
jayvdb: use `excludes=` so it is more readable.

tests/bearlib/aspects/CollectionsTest.py
```diff
@@ -10,6 +10,17 @@
...
+             Metadata.CommitMessage.Body.Existence])
+        self.instancelist_excludes = AspectList(
+            [Metadata.CommitMessage.Shortlog('py'),
+             Metadata.CommitMessage.Body('py')],
+            [Metadata.CommitMessage.Shortlog.TrailingPeriod,
```
jayvdb: use `excludes=` so it is more readable.

coalib/bearlib/aspects/collections.py
```diff
@@ -7,7 +7,7 @@ class AspectList(list):
     List-derived container to hold aspects.
     """
 
-    def __init__(self, seq=()):
+    def __init__(self, seq=(), excludes=()):
```
userzimmermann: Change argument defaults to `None`. Otherwise you always do checks like `if excludes ...`. But there can be sequence that are `False` but can anyway contain elements. Best is always `seq=None, excludes=None` and then check `if excludes is None ...`. It's also written in some python docs...

userzimmermann:
@adhikasp @jayvdb Just saw that the commit with changes to **coalib/bearlib/aspects/base.py** is actually taken from https://github.com/coala/coala/pull/4427 ... Sorry for complaining! But I marked this extra **[WIP]** ;)

coalib/bearlib/aspects/collections.py
adhikasp: Agree for the `exclude=None`. But if we change `seq=None` then it could give an error for declaring empty AspectList.

```python
>>> from coalib.bearlib.aspects.collections import AspectList
>>> AspectList()
    ...
TypeError: 'NoneType' object is not iterable

**Comment by jayvdb** [comment]
@userzimmermann , the commit for #4427 is a dependency for #4438 , and it is sensible to have one PR that allows them to be reviewed together, provided that the commits are nice and cleanly separated.

File: tests/bearlib/aspects/CollectionsTest.py [hunk 4]
```diff
@@ -54,3 +74,20 @@ def test_get(self):
...
+                      Metadata.CommitMessage.Shortlog.ColonExistence),
+                      Metadata.CommitMessage.Shortlog.ColonExistence)
+        self.assertIsNone(self.aspectlist_excludes.get(
+                          Metadata.CommitMessage.Shortlog.TrailingPeriod))
+        self.assertIsNone(self.aspectlist_excludes.get(
+                          Metadata.CommitMessage.Body.Existence))
+
+        self.assertEqual(self.instancelist_excludes.get(
+                      Metadata.CommitMessage.Shortlog.ColonExistence),
+                      Metadata.CommitMessage.Shortlog.ColonExistence('py'))
```
jayvdb: Put this computed value in a variable before using assertEqual.
assert invocations shouldnt have complex calls in them.

adhikasp: I don't think this is a complex call because it just a method call with 1 arg (albeit with very long names). Also every other test have the same pattern.

But I'll try shorten the aspect name to make it easier to read.

jayvdb:
ack 25bea08

jayvdb:
@rultor merge

PR #4449, opened 2017-07-05
gsoc: Reprompt possible actions

Nosferatul:


Nosferatul:
This will only works in CLI section

coalib/output/ConsoleInteraction.py
```diff
@@ -599,13 +615,29 @@ def print_actions(console_printer, section, actions, failed_actions):
...
                             choose to do nothing, return (None, None).
     """
-    choice = choose_action(console_printer, actions)
-
+    if section.name == 'cli':
+        if section.get('reprompt').value == 'True' or
+        section.get('reprompt').value == 'False' or
+        section.get('reprompt').value == '':
+            if section.get('reprompt').value == 'True'
+            or section.get('reprompt').value == '':
```
yukiisbored: You can use `literal_eval` from ast

yukiisbored: or see how does coala interpret boolean values when asking the user for a missing setting.

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +153,33 @@ def acquire_actions_and_apply(console_printer,
...
+                                            section,
+                                            metadata_list,
+                                            action_dict,
+                                            failed_actions,
+                                            result,
+                                            file_diff_dict,
+                                            file_dict):
+                break
+            question = format_lines(
+                'Reprompt available actions (True/ False)? ', symbol='!')
```
yukiisbored: Remove the space before False

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +153,33 @@ def acquire_actions_and_apply(console_printer,
...
+            input_q = input(question)
+            if input_q == 'True' or input_q == 'False':
```
yukiisbored: Is there a better way to do this? I know that something like SpaceConsistencyBear can accept Yes, No, Yeah, etc

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +153,36 @@ def acquire_actions_and_apply(console_printer,
...
+                                            failed_actions,
+                                            result,
+                                            file_diff_dict,
+                                            file_dict):
+                break
+            question = format_lines(
+                'Reprompt available actions (Y/N)? ', symbol='!')
+            input_q = input(question)
+            input_q = str(input_q).lower()
+            if input_q in ['yes', 'yea', 'y']:
```
Adrianzatreanu: I dont like the hardcoding. Isnt there a "constants" file somewhere, or just define a a global constant list if not? But therr should be one.

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +153,36 @@ def acquire_actions_and_apply(console_printer,
...
+                section.append(Setting('reprompt', 'True'))
+            elif input_q in ['no', 'n']:
+                section.append(Setting('reprompt', 'False'))
+            else:
+                console_printer.print('Bad input. Do (N)othing applied.')
```
Adrianzatreanu: "Do nothing" should be enclosed somehow, or else it looks like bad grammar :p

coalib/output/ConsoleInteraction.py
```diff
@@ -599,12 +622,26 @@ def print_actions(console_printer, section, actions, failed_actions):
...
-
-    return get_action_info(section, actions[choice - 1], failed_actions)
+    if section.name == 'cli':
+        if section.get('reprompt').value == '':
+            choice = choose_action(console_printer, actions)
+            if choice == 0:
+                return None, None
+            return get_action_info(section, actions[choice - 1],
+                                   failed_actions)
+        elif ast.literal_eval(section.get('reprompt').value):
```
Adrianzatreanu: Why do you have an else if u do exactly the same thing in the if and elif?

coalib/output/ConsoleInteraction.py
```diff
@@ -599,12 +623,21 @@ def print_actions(console_printer, section, actions, failed_actions):
...
     """
-    choice = choose_action(console_printer, actions)
-
-    if choice == 0:
-        return None, None
-
-    return get_action_info(section, actions[choice - 1], failed_actions)
+    if section.name == 'cli':
+        if (section.get('reprompt').value == ''
+                or ast.literal_eval(section.get('reprompt').value)):
```
yukiisbored: Hey, `coala_utils` has it's own function that returns for booleans from string, please use it over ast.
Check it out here: https://gitlab.com/coala/coala-utils/blob/master/coala_utils/string_processing/StringConverter.py#L41

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +154,36 @@ def acquire_actions_and_apply(console_printer,
...
+                                            failed_actions,
+                                            result,
+                                            file_diff_dict,
+                                            file_dict):
+                break
+            question = format_lines(
+                'Reprompt available actions (Y/N)? ', symbol='!')
+            input_q = input(question)
+            input_q = str(input_q).lower()
+            if input_q in Constants.TRUE_STRINGS:
```
yukiisbored: Euhm, why not use `StringConverter`?

Nosferatul: Hmm, I don't know why do I need it here. I'm using instead of `ast`. The rest it's the same. Here we should check if it's a `true` string or not

Nosferatul: And the problem if I want to introduce the `True/False` as a bool instead of string `True/False` is that it doesn't accept it.

coalib/output/ConsoleInteraction.py
```diff
@@ -152,14 +154,36 @@ def acquire_actions_and_apply(console_printer,
...
-                                        section,
-                                        metadata_list,
-                                        action_dict,
-                                        failed_actions,
-                                        result,
-                                        file_diff_dict,
-                                        file_dict):
+        if section.name == 'cli':
+            if not ask_for_action_and_apply(console_printer,
+                                            section,
```
Adrianzatreanu: Looks weird how u are calling this twice, u sure u cant simplify it?

coalib/output/ConsoleInteraction.py
```diff
@@ -599,12 +623,21 @@ def print_actions(console_printer, section, actions, failed_actions):
                             values for the action. If the user did
                             choose to do nothing, return (None, None).
     """
-    choice = choose_action(console_printer, actions)
-
-    if choice == 0:
-        return None, None
-
-    return get_action_info(section, actions[choice - 1], failed_actions)
+    if section.name == 'cli':
```
Adrianzatreanu: This is still horribly written.
Cant u just simplify it kind like it was before, with the choice first?

Nosferatul: The choice is 1st anyway. I do this because when I do the next line, if the `reprompt` value doesn't exists it will return by default an empty string. Also this was the same as always, i just changed the next line with `StringConverter(section.get('reprompt').value)).__bool__():`

coalib/output/ConsoleInteraction.py
```diff
@@ -599,12 +614,21 @@ def print_actions(console_printer, section, actions, failed_actions):
...
+        if (section.get('reprompt').value == ''
```
yukiisbored: Can you put `section.get('reprompt').value` into it's own variable? Just to make it shorter and more practical.

yukiisbored: Also, https://www.python.org/dev/peps/pep-0008/#programming-recommendations
says the following:
> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
> Yes:
> ``` python
> if not seq:
>    if seq:
> ```python
> No: 
> ```
> if len(seq):
>    if not len(seq):
> ```

coalib/output/ConsoleInteraction.py
```diff
@@ -599,12 +614,22 @@ def print_actions(console_printer, section, actions, failed_actions):
...
+                return None, None
+            return get_action_info(section, actions[choice - 1],
+                                   failed_actions)
+        else:
+            return None, None
+    else:
+        choice = choose_action(console_printer, actions)
+        if choice == 0:
+            return None, None
+        return get_action_info(section, actions[choice - 1], failed_actions)
```
Adrianzatreanu: the way i see it this could be refactored as this:
https://pastebin.com/NqYPJz5G

i might be wrong, please double check it, but i think it looks slightly better

Nosferatul: the problem with what you wrote is that, whenever I enter `no` when it's asked to reprompt the actions, it will ask for another input and after it will not prompt again.
http://imgur.com/a/IJ1vi

coalib/output/ConsoleInteraction.py
adhikasp: Isn't `StringConverter(input_q).__bool__()` will give the same result?

yukiisbored:
Gonna restart the Appveyor build, dumb error.

coalib/output/ConsoleInteraction.py
Adrianzatreanu: Ah i see

Adrianzatreanu:
Ack 5f0931a

sils:
I don't think I like this design wise, having https://github.com/coala/coala/pull/4179 or something like that which is already in the works would be much smarter and I had to really read through the code what exactly I'd be asked for so as a user it's not clear what this does for me.

sils:
looking at the screenshot, it's kinda cumbersume to have to answer twice. I really don't think this is a wise thing to happen.

sils:
I'm closing this for this reason, @Nosferatul if you disagree with me simply reopen this and we'll discuss it ok?

Nosferatul:
Ok @sils. I understand the reason :+1

PR #4471, opened 2017-07-15
Bear.check_prerequisites: Fallback to True

sils:
CC @NiklasMM @jayvdb 

This is WIP but finishing it today.


sils:
@userzimmermann can you test if this works on windows? I.e. install this coala and run the coala bears tests.

coalib/core/Bear.py
```diff
@@ -331,9 +331,18 @@ def check_prerequisites(cls):
         :return: True if prerequisites are satisfied, else False or a string
                  that serves a more detailed description of what's missing.
         """
+        def is_installed(requirement):
```
userzimmermann: Why not define on module level instead of creating a new function object for every call of `.check_prerequisites`?

sils:
This is blocked by https://gitlab.com/coala/package_manager/merge_requests/102

PR #4517, opened 2017-07-22
BearTest: Remove urllib3 ProtocolError

jayvdb:
When requests was undergoing frequent breakages, a `ProtocolError`

was being raise unexpectedly when provoking a broken read scenario.
This exception was added as an alternative as it was at least
a predictable error occurring, allowing the remainder of the
test method to verify the resulting behaviour of the method,
and there were far bigger problems occurring at the time
due to `requests` being broken several time a day.

Depending on urllib3 for this ProtocolError is problematic,
as requests has its own version requirements for urllib3,
and those must take precedence over coala testing needs.
A conflict between requests and urllib3 has broken the
docker image.

It seems ProtocolError is not being thrown any longer,
and removing it bypasses the urllib3 conflict.
Should a ProtocolError occur again, it should be
investigated.

Fixes coala#4512

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ matrix:
       env: UNSUPPORTED=true
       script: .misc/check_unsupported.sh
 
+env:
+  global:
+    - PATH="$PATH:$TRAVIS_BUILD_DIR/node_modules/.bin"
+
 cache:
   pip: true
   directories:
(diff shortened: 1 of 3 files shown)


yash-nisar:
ack dee8a5a09adc2d887e117389e229d56e406911e8

jayvdb:
@rultor merge

PR #4525, opened 2017-07-22
Core: Port core functions to class

Makman2:

PR #4526, opened 2017-07-22
Allow custom executors for NextGen-Core

Makman2:


coalib/core/Core.py
```diff
@@ -62,184 +62,6 @@ def group(iterable, key=lambda x: x):
     return zip(keys, elements)
 
 
-def cleanup_bear(bear,
```
jayvdb: This is being removed?
Leave a dummy function here to avoid breakages, please?

coalib/core/Core.py
```diff
@@ -327,9 +149,190 @@ def instantiate_and_track(prev_bear_type, next_bear_type):
...
+            if resolved_bears:
+                self._schedule_bears(resolved_bears)
+
+            del self.running_tasks[bear]
+
+        if not self.running_tasks:
+            # Check the DependencyTracker additionally for remaining
+            # dependencies.
+            resolved = self.dependency_tracker.are_dependencies_resolved
+            if not resolved:  # pragma: no cover
```
jayvdb: Could you create a special pragma in setup.cfg for this.  We're quite close to removing the default 'no cover' pragma, replacing it with specific reasons.
I am pretty sure that we have a few other pragma's which exist because the code is apparently unreachable, and the message is very clearly "you've hit a bug; please report it".

coalib/core/Core.py
```diff
@@ -340,21 +337,5 @@ def run(bears, result_callback):
...
-
-    # Let's go.
-    schedule_bears(bears_to_schedule, result_callback, dependency_tracker,
-                   event_loop, {}, executor)
-    try:
-        event_loop.run_forever()
-    finally:
-        event_loop.close()
+    session = Session(bears, result_callback)
+    session.run()
```
sils: simplify

coalib/core/Core.py
sils: it's moved to the class

coalib/core/Core.py
sils: I filed https://github.com/coala/coala/issues/4529 for this, this pragma here is only moved into the class and I think the whole branch should be removed

coalib/core/Core.py
Makman2: also it's not official API yet, so until the official release we can move stuff around like wanted.

sils:
ack https://github.com/coala/coala/pull/4526/commits/c873c346ac95ce350bff744e25d13c6cbee8d738

sils:
ack https://github.com/coala/coala/pull/4526/commits/732f92d946db8a9b91399f9e5386c881e97e520a

sils:
ack https://github.com/coala/coala/pull/4525/commits/45a49faa269a66d62866cd6164a9278027c4406b from me, I don't see that move affecting anything, these core routines aren't meant for usage outside and I think someone woul d have a super hard time using it for other purposes than ours. I don't see that we need to deprecate this API.

sils:
ack 45a49fa

sils:
(as for the pragma, getting rid of that is not task of this commit/pr and an issue was filed)

Makman2:
reack 39af06d 26b10b0 30c21ad

Makman2:
@rultor merge

Issue #4200, opened 2017-05-08
DocStyleBear should also check for <blank line> before and after the DocString.


damngamerz:
The current implementation DocStyleBear violates PEP-257. There’s no function to check a blank line either or before the docstrings. Implement a functionality which basically checks for a blank line before and after a docstring, and if it doesn’t comply to it, generate a proper fix.

PR #4549, opened 2017-07-24
DocumentationComment.py: Add blankline padding

damngamerz:
`assemble()` now adds a blank line before or after
docstring based on top_padding and bottom_padding.

Closes https://github.com/coala/coala/issues/4200

**For short term contributors:** we understand that getting your commits well
defined like we require is a hard task and takes some learning. If you
look to help without wanting to contribute long term there's no need
for you to learn this. Just drop us a message and we'll take care of brushing
up your stuff for merge!

### Checklist

- [x] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [x] I ran coala over my code locally. (*All commits have to pass
      individually.* It is not sufficient to have "fixup commits" on your PR,
      our bot will still report the issues for the previous commit.) You will
      likely receive a lot of bot comments and build failures if coala does not
      pass on every single commit!

After you submit your pull request, **DO NOT click the 'Update Branch' button.**
When asked for a rebase, consult [coala.io/rebase](https://coala.io/rebase)
instead.

Please consider helping us by reviewing other peoples pull requests as well:

- pick up any PR at <https://coala.io/review>
- review it (check <https://coala.io/reviewing> for more info)
- if you are sure that it needs work, use `cobot mark wip <URL>` to get it out
  of the review queue.

The more you review, the more your score will grow at coala.io and we will
review your PRs faster!

PR #4557, opened 2017-07-25
DocumentationAPI: Add `padding`

damngamerz:
Add `padding` in DocstyleDefinition which
are now acquired from `.coalang`. Add
`top_padding` and `bottom_padding` which
automatically instantiate from
`instantiate_padding` in DocBaseClass. Add
supporting test cases.

Related to https://github.com/coala/coala/issues/4200

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/languages/documentation/DocstyleDefinition.py
+++ b/coalib/bearlib/languages/documentation/DocstyleDefinition.py
@@ -18,10 +18,19 @@ class DocstyleDefinition:
     Metadata = namedtuple('Metadata', ('param_start', 'param_end',
                                        'exception_start', 'exception_end',
                                        'return_sep'))
+    ClassPadding = namedtuple('ClassPadding',
+                              ('top_padding', 'bottom_padding'))
+    FunctionPadding = namedtuple('FunctionPadding',
+                                 ('top_padding', 'bottom_padding'))
+    DocstringTypeRegex = namedtuple('DocstringTypeRegex',
+                                    ('class_sign', 'func_sign'))
 
     @enforce_signature
     def __init__(self, language: str, docstyle: str, markers: (Iterable, str),
-                 metadata: Metadata):
+                 metadata: Metadata, class_padding: ClassPadding,
+                 function_padding: FunctionPadding,
+                 docstring_type_regex: DocstringTypeRegex,
+                 docstring_position: str):
         """
         Instantiates a new DocstyleDefinition.
 
@@ -39,6 +48,18 @@ class DocstyleDefinition:
                          e.g. ``param_start`` defining the start symbol of
                          the parameter fields and ``param_end`` defining the
                          end.
+        :param class_padding: A namedtuple consisting of values about
+                         blank lines before and after the documentation of
+                         ``docstring_type`` class.
+        :param function_padding: A namedtuple consisting of values about
+                         blank lines before and after the documentation of
+                         ``docstring_type`` function.
+        :param docstring_type_regex: A namedtuple consisting of regex
+                         about ``class`` and ``function`` of a language, which
+                         is used to determine ``docstring_type`` of
+                         DocumentationComment.
+        :param docstring_position: Defines the position where the regex of
+                         docstring type is present(i.e. ``top`` or ``bottom``).
         """
         self._language = language.lower()
         self._docstyle = docstyle.lower()
@@ -58,6 +79,10 @@ class DocstyleDefinition:
                                  'actually {}).'.format(length))
 
         self._metadata = metadata
+        self._class_padding = class_padding
+        self._function_padding = function_padding
+        self._docstring_type_regex = docstring_type_regex
+        self._docstring_position = docstring_position
 
     @property
     def language(self):
@@ -127,6 +152,48 @@ class DocstyleDefinition:
         """
         return self._metadata
 
+    @property
+    def class_padding(self):
+        """
+        A namedtuple ``ClassPadding`` consisting of values about blank lines
+        before and after the documentation of ``docstring_type`` class.
+
+        These values are official standard of following blank lines before and
+        after the documentation of ``docstring_type`` class.
+        """
+        return self._class_padding
+
+    @property
+    def function_padding(self):
+        """
+        A namedtuple ``FunctionPadding`` consisting of values about blank
+        lines before and after the documentation of ``docstring_type``
+        function.
+
+        These values are official standard of following blank lines before and
+        after the documentation of ``docstring_type`` function.
+        """
+        return self._function_padding
+
+    @property
+    def docstring_type_regex(self):
+        """
+        A namedtuple ``DocstringTypeRegex`` consisting of regex about ``class``
+        and ``function`` of a language, which is used to determine
+        ``docstring_type`` of DocumentationComment.
+        """
+        return self._docstring_type_regex
+
+    @property
+    def docstring_position(self):
+        """
+        Defines the position, where the regex of docstring type is present.
+        Depending on different languages the docstrings are present below or
+        above the defined class or function. This expicitly defines where the
+        class regex or function regex is present(i.e. ``top`` or ``bottom``).
+        """
+        return self._docstring_position
+
     @classmethod
     @enforce_signature
     def load(cls, language: str, docstyle: str, coalang_dir=None):
@@ -186,13 +253,44 @@ class DocstyleDefinition:
         metadata = cls.Metadata(*(str(docstyle_settings.get(req_setting, ''))
                                   for req_setting in metadata_settings))
 
+        try:
+            class_padding = cls.ClassPadding(
+                *(int(padding) for padding in tuple(
+                    docstyle_settings['class_padding'])))
+        except IndexError:
+            class_padding = cls.ClassPadding('', '')
+
+        try:
+            function_padding = cls.FunctionPadding(
+                *(int(padding) for padding in tuple(
+                    docstyle_settings['function_padding'])))
+        except IndexError:
+            function_padding = cls.FunctionPadding('', '')
+
+        try:
+            docstring_type_regex = cls.DocstringTypeRegex(
+                *(str(sign) for sign in tuple(
+                    docstyle_settings['docstring_type_regex'])))
+        except IndexError:
+            docstring_type_regex = cls.DocstringTypeRegex('', '')
+
+        try:
+            docstring_position = docstyle_settings['docstring_position'].value
+        except IndexError:
+            docstring_position = ''
+
+        ignore_keys = (('class_padding', 'function_padding',
+                        'docstring_type_regex', 'docstring_position') +
+                       metadata_settings)
+
         marker_sets = (tuple(value)
                        for key, value in
                        docstyle_settings.contents.items()
-                       if key not in metadata_settings and
+                       if key not in ignore_keys and
                        not key.startswith('comment'))
 
-        return cls(language, docstyle, marker_sets, metadata)
+        return cls(language, docstyle, marker_sets, metadata, class_padding,
+                   function_padding, docstring_type_regex, docstring_position)
 
     @staticmethod
     def get_available_definitions():
--- a/coalib/bearlib/languages/documentation/DocumentationExtraction.py
+++ b/coalib/bearlib/languages/documentation/DocumentationExtraction.py
@@ -3,6 +3,7 @@ import re
 from coalib.bearlib.languages.documentation.DocumentationComment import (
     DocumentationComment, MalformedComment)
 from coalib.results.TextPosition import TextPosition
+from coalib.results.TextRange import TextRange
 from textwrap import dedent
 
 
@@ -266,6 +267,65 @@ def extract_documentation_with_markers(content, docstyle_definition):
             # Starting line of doc_string where marker is present
             start_line = doc.range.start.line - 1
             ignore_string_match = ignore_regex.search(content[start_line])
+
+            # Instantiate padding
+            top_padding = 0
+            bottom_padding = 0
+            # minus 2 because we want to check the line before the marker.
+            start_index = doc.range.start.line - 2
+            end_index = doc.range.end.line
+            while start_index >= 0 and not content[start_index].strip():
+                top_padding += 1
+                start_index -= 1
+            # If the end_index is instantiated above the len(content) i.e.
+            # In case where ending marker of docstring is at the last line.
+            # Then the doc.bottom_padding will be default to 0. This will also
+            # prevent IndexError raised by content[end_index].
+            while end_index < len(content) and not content[end_index].strip():
+                # This condition will take place if theres an inline docstring
+                # following documentation.
+                if ((doc.marker[2]+'\n') != content[end_index-1][-4:]
+                        and bottom_padding == 0):
+                    break
+                bottom_padding += 1
+                end_index += 1
+
+            class_regex = re.compile(
+                doc.docstyle_definition.docstring_type_regex.class_sign)
+            function_regex = re.compile(
+                doc.docstyle_definition.docstring_type_regex.func_sign)
+
+            # End line differs when mid marker and end marker is different
+            if doc.marker[1] == doc.marker[2]:
+                end_index = end_index - 1
+
+            # Check for docstring_position and then check for class regex
+            # and function regex to define the type of docstring.
+            if doc.docstyle_definition.docstring_position == 'top':
+                if class_regex.search(content[start_index]):
+                    doc.docstring_type = 'class'
+                elif function_regex.search(content[start_index]):
+                    doc.docstring_type = 'function'
+            elif doc.docstyle_definition.docstring_position == 'bottom':
+                if (end_index < len(content) and
+                        class_regex.search(content[end_index])):
+                    doc.docstring_type = 'class'
+                elif (end_index < len(content) and
+                        function_regex.search(content[end_index])):
+                    doc.docstring_type = 'function'
+
+            # Disabled automatic padding for docstring_type='others' as this
+            # will cause overlapping of range in consecutive docstrings. Which
+            # diff.replace() is unable to handle.
+            if doc.docstring_type != 'others':
+                doc.top_padding = top_padding
+                doc.bottom_padding = bottom_padding
+
+                doc.range = TextRange.from_values(
+                    start_index + 2,
+                    1 if top_padding > 0 else doc.range.start.column,
+                    end_index,
+                    1 if bottom_padding > 0 else doc.range.end.column)
 
             if ignore_string_match:
                 yield doc
--- a/tests/bearlib/languages/documentation/DocBaseClassTest.py
+++ b/tests/bearlib/languages/documentation/DocBaseClassTest.py
@@ -273,6 +273,151 @@ class DocBaseClassTest(unittest.TestCase):
                                   docstyle_PYTHON3_default.markers[0],
                                   TextPosition(1, 2))])
 
+    def test_DocBaseClass_instantiate_padding_PYTHON3_6(self):
+        data = ['def some_function:\n',
+                '\n',
+                '   """ documentation in single line """\n',
+                '\n',
+                '\n',
+                'print(1)']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual([doc.top_padding, doc.bottom_padding],
+                             [1, 2])
+
+    def test_DocBaseClass_instantiate_padding_PYTHON3_7(self):
+        data = ['class some_class:\n',
+                '\n',
+                '\n',
+                '   """ documentation in single line """\n',
+                '\n',
+                '\n',
+                'print(1)\n']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual([doc.top_padding, doc.bottom_padding],
+                             [2, 2])
+
+    def test_DocBaseClass_instantiate_padding_inline_PYTHON3_8(self):
+        # To test that bottom_padding sets to nothing if docstring is
+        # followed by inline docstring.
+        data = ['def some_function:\n',
+                '\n',
+                '   """\n',
+                '   documentation in single line\n',
+                '   """ # This is inline docstring',
+                '\n',
+                '\n',
+                'print(1)']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual([doc.top_padding, doc.bottom_padding],
+                             [1, 0])
+
+    def test_DocBaseClass_instantiate_padding_inline_PYTHON3_9(self):
+        # Paddings will not be instantiated for docstring_type=others
+        data = ['\n',
+                '   """\n',
+                '   documentation in single line\n',
+                '   """ # This is inline docstring',
+                '\n',
+                '\n',
+                'print(1)']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual([doc.top_padding, doc.bottom_padding],
+                             [0, 0])
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_10(self):
+        data = ['class xyz:\n',
+                '   """\n',
+                '   This docstring is of docstring_type class\n',
+                '   """\n']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual(doc.docstring_type, 'class')
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_11(self):
+        data = ['def xyz:\n',
+                '   """\n',
+                '   This docstring is of docstring_type function\n',
+                '   """\n']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual(doc.docstring_type, 'function')
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_12(self):
+        data = ['\n',
+                '   """\n',
+                '   This docstring is of docstring_type others\n',
+                '   """\n',
+                '\n',
+                'print(1)']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual(doc.docstring_type, 'others')
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_13(self):
+        data = ['## Documentation for a function.\n',
+                '#\n',
+                '#  More details.\n',
+                'def func():\n',
+                '    pass\n']
+
+        docstyle_PYTHON3_doxygen = DocstyleDefinition.load('PYTHON3',
+                                                           'doxygen')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'doxygen'):
+            self.assertEqual(doc.docstring_type, 'function')
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_14(self):
+        data = ['## Documentation for a class.\n',
+                '#\n',
+                '#  More details.\n',
+                'class PyClass:\n',
+                '\n']
+
+        docstyle_PYTHON3_doxygen = DocstyleDefinition.load('PYTHON3',
+                                                           'doxygen')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'doxygen'):
+            self.assertEqual(doc.docstring_type, 'class')
+
+    def test_DocBaseClass_instantiate_docstring_type_PYTHON3_15(self):
+        data = ['def some_function():\n',
+                '"""\n',
+                'documentation\n',
+                '"""\n',
+                'class myPrivateClass:\n',
+                '    pass']
+
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual(doc.docstring_type, 'function')
+
     def test_generate_diff(self):
         data_old = ['\n', '""" documentation in single line  """\n']
         for doc_comment in DocBaseClass.extract(
--- a/tests/bearlib/languages/documentation/DocumentationCommentTest.py
+++ b/tests/bearlib/languages/documentation/DocumentationCommentTest.py
@@ -19,6 +19,9 @@ class DocumentationCommentTest(unittest.TestCase):
     ReturnValue = DocumentationComment.ReturnValue
 
     Metadata = DocstyleDefinition.Metadata
+    ClassPadding = DocstyleDefinition.ClassPadding
+    FunctionPadding = DocstyleDefinition.FunctionPadding
+    DocstringTypeRegex = DocstyleDefinition.DocstringTypeRegex
 
 
 class GeneralDocumentationCommentTest(DocumentationCommentTest):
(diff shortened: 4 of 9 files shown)


coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -58,6 +60,8 @@ def __init__(self, language: str, docstyle: str, markers: (Iterable, str),
                                  'actually {}).'.format(length))
 
         self._metadata = metadata
+        self._top_padding = None if padding == () else int(padding[0])
+        self._bottom_padding = None if padding == () else int(padding[1])
```
damngamerz: @jayvdb This answers your question :) . `.coalang` will have official standards defined.
By default `top_padding` and `bottom_padding` of DocComment are initialized with zero. If we don't have style defined. It will have default values. Also `padding`'s major motive was if a developer doesn't wants to follow official standard and decides suppose I want to have 2 blank lines at the bottom of docstring. He will set those padding variables and it will be amended by the bear part.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -21,7 +21,7 @@ class DocstyleDefinition:
 
     @enforce_signature
     def __init__(self, language: str, docstyle: str, markers: (Iterable, str),
-                 metadata: Metadata):
+                 metadata: Metadata, padding: (Iterable, str)):
```
jayvdb: Is it possible to provide a default value for this new arg, thereby avoiding breaking the API?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
jayvdb: Why split this into to variables rather than store a tuple?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: let me see though. But I don't think that would cause any problem cause `load` function provides an empty tuple if it finds nothing. Which I later break into the properties of `top_padding` and `bottom_padding`. which is later taken care by https://github.com/coala/coala/pull/4557#discussion_r129518147

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: using separate variables it becomes easily available at bear side. `DocComment.docstyle_definition.top_padding/bottom_padding` Also they are read as strings from the file(need int). I need to get separate values in the bear part anyhow. So why not split them in the first place.

coalib/bearlib/languages/documentation/DocBaseClass.py
```diff
@@ -335,3 +339,44 @@ def process_documentation(self, *args, **kwargs):
...
+            end_index = doc_comment.range.end.line
+            while start_index >= 0 and content[start_index] == '\n':
+                top_padding += 1
+                start_index -= 1
+            while end_index < len(content) and content[end_index] == '\n':
+                # This condition will take place if theres an inline docstring
+                # following documentation.
+                if ((doc_comment.marker[2]+'\n') != content[end_index-1][-4:]
+                        and bottom_padding == 0):
+                    break
```
damngamerz: This is important I couldn't find any other way to do this :)

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: yes this breaks the API. I tried to provide a default value. But the thing is `load` returns an empty tuple if it finds nothing which overwrites that default value. The best solution to providing default values through https://github.com/coala/coala/pull/4557/files#diff-bb503d77ea6d0498c9865cab08c904f4R64

damngamerz:
@jayvdb We can continue our discussion here.... As `default.coalang` will take care of the intended blank lines.
So for python I believe is that PEP-257 is old. Even we don't follow it's official multi line docstring. And I feel what we use here at coala is much better(which is an enhanced inherited version of PEP - 257 standards).
atm we don't have any mechanism to describe the docstring type. Is it normal function docstring or class function docstring?
so any ideas what can be `top_padding` and `bottom_padding` which can be applied to all docstrings in coala repo, coala-bears? my bets are on (0,1). As generally we need to have a blank line after the docstyle(such as in class function).
Or if we don't implement this now(leave it as (0,0)) cause this is a trivial task. And find some way to get docstring origin(function docstring or class function docstring).

tests/bearlib/languages/documentation/DocBaseClassTest.py
```diff
@@ -14,18 +14,19 @@
 
 
 class DocBaseClassTest(unittest.TestCase):
+    BaseClass = DocBaseClass()
```
jayvdb: this should be in method `setUpClass` or `setUp` .

tests/bearlib/languages/documentation/DocumentationCommentTest.py
```diff
@@ -19,6 +19,7 @@ class DocumentationCommentTest(unittest.TestCase):
     ReturnValue = DocumentationComment.ReturnValue
 
     Metadata = DocstyleDefinition.Metadata
+    BaseClass = DocBaseClass()
```
jayvdb: this (and the lines above) should be in method `setUpClass` or `setUp`

tests/bearlib/languages/documentation/documentation_extraction_testdata/doxygen.py
```diff
@@ -2,35 +2,42 @@
...
 #
 #  More details.
+
 def func():
     pass
 
 ## Documentation for a class.
 #
 # More details.
+
```
jayvdb: nope; wrong per https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#pythonblocks

(and also this fixup *must* undo the bad changes in the previous patch to default.py : https://github.com/coala/coala/pull/4549#pullrequestreview-52493280 )

jayvdb:
> So for python I believe is that PEP-257 is old.

No, you are wrong.  On https://www.python.org/dev/peps/pep-0257/ , you can see it's status is "Active".  It has not been replaced.  It is maintained at https://github.com/python/peps/commits/master/pep-0257.txt
And https://github.com/pycqa/pydocstyle implements it.  And there are vibrant discussions about the finer details of it, and pydocstyle allowing ignoring some of its rules to allow for alternative styles (and recently has a few extra rules to enforce other styles), but nobody dares say pep 257 is not the official style guide of Python.

> And I feel what we use here at coala is much better(which is an enhanced inherited version of PEP - 257 standards).

Again, wrong.
What coala uses is not particularly good.  The reason it is bad is because we have not undertaken a project to standardise the docstrings.  No bears are enabled.  The docstrings can contain anything.  Only manual code review can catch problems.  Your project is intended to solve that.  Or we can work towards enabling PyDocStyleBear.
There is also a bug in QuotesBear which imposes a silly PEP 257 violation.
In fact Stefan and I were complaining about the bears bad behaviour recently on another patch very recently.  We were asking for changes without realising that QuotesBear prohibited it.

> atm we don't have any mechanism to describe the docstring type.

A design problem, for sure.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
jayvdb: You can used NamedTuple to make it easier for them to be used.
The reason to keep them as a tuple is so that everywhere they are a pair, from style config to bear.  consistency.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -9,12 +9,9 @@ def foobar_explosion(radius):
     """
     A nice and neat way of documenting code.
     :param radius: The explosion radius. """
-
-
     def get_55():
         """A function that returns 55."""
```
jayvdb: this blank line is also not acceptable.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -73,7 +68,6 @@ def foobar_triangle(side_A, side_B, side_C):
     :return: returns perimeter
     """
```
jayvdb: this blank line is also not acceptable.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -43,7 +39,6 @@ def best_docstring(param1, param2):
              Cut to the Next Line.
     """
```
jayvdb: this blank line is also not acceptable.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -29,7 +25,6 @@ def get_55():
 """ Docstring inline with triple quotes.
```
jayvdb: Note that this, and the one above, are *not* a docstring.  They are incorrectly described as such (https://github.com/coala/coala/issues/4566), and shouldnt be parsed as such (is this being done?).

Also my understanding of https://www.python.org/dev/peps/pep-0008/#blank-lines is that two blank lines are expected after the prior def, and before the next def.

```bash
$ pycodestyle default.py
...
default.py.1:28:1: E302 expected 2 blank lines, found 1
default.py.1:42:1: E302 expected 2 blank lines, found 1
...
```

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
```diff
@@ -9,12 +9,8 @@ def foobar_explosion(radius):
...
     A nice and neat way of documenting code.
     :param radius: The explosion radius. """
-
-
     def get_55():
         """A function that returns 55."""
-
-
         return 55
     return get_55() * radius
```
jayvdb: now the fun part ..

```bash
$ pycodestyle default.py
default.py.1:17:1: E305 expected 2 blank lines after class or function definition, found 1
...
```

that is part of this PR.  this is undisputed territory that all module level def & classes have two whitespace before and after them.

Also, there should be a blank line after `def get_55` is finished, before `return get_55() * radius` .  This is a greyish area; I have seen people argue for and against this.  But it becomes clearer (and I am pretty sure Guido comment on this specific case) when there are two inner def's inside a def ; they need to be separated by a blank line.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
damngamerz: (is this being done?). Yes its being parsed atm.

tests/bearlib/languages/documentation/documentation_extraction_testdata/default.py
damngamerz: yes I will amend those changes in the test cases :+1:

damngamerz:
@jayvdb I and my mentor has decided that defining type of docstring(module, def or class) can be a separate issue. It's totally a new design issue. And we have some awesome ideas about it's implementation. Let's get this PR merged I will amend those changes tonight :wink:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -186,13 +227,38 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
         metadata = cls.Metadata(*(str(docstyle_settings.get(req_setting, ''))
                                   for req_setting in metadata_settings))
 
+        class_padding = (cls.ClassPadding('', '') if
+                         docstyle_settings.get('class_padding', '').value == ''
+                         else cls.ClassPadding(
+                         *(int(padding) for padding in tuple(
+                             docstyle_settings.get('class_padding', ''))))
+                         )
```
NiklasMM: These constructs are super hard to read. Would look much nicer as a try/except block

coalib/bearlib/languages/documentation/DocBaseClass.py
```diff
@@ -43,9 +44,17 @@ def extract(content, language, docstyle):
...
         docstyle_definition = DocstyleDefinition.load(language, docstyle)
-        return extract_documentation_with_markers(
+        doc_comment_iterator = extract_documentation_with_markers(
                         content, docstyle_definition)
 
+        doc_comment_iterator = self.instantiate_padding(
+            content, doc_comment_iterator)
+
+        doc_comment_iterator = self.instantiate_kind(
+            content, doc_comment_iterator)
```
damngamerz: > This can be done much more efficiently by turning this method into a generator and changing instantiate_padding and instantiate_kind to process only one DocComment at a time.
This way we only need to iterate once over the docstrings.

Yes even I'm a bit worried about this... Although Im not good with generators and it's always been a pain in my head...But if it solves it why not we do it...Just we have to note that finally we need an updated DocComment iterator/generator which will have instantiated type and paddings.

>Actually, why do we need those three separate functions to parse the Docstring? Can't we do all the work in one go in extract_documentation_with_markers?

No, we can't as we need `range` , an instance of DocComment and whole file (`content`), to instantiate type and padding.

coalib/bearlib/languages/documentation/DocBaseClass.py
```diff
@@ -84,3 +93,95 @@ def process_documentation(self, *args, **kwargs):
...
+            top_padding = 0
+            bottom_padding = 0
+            start_index = doc_comment.range.start.line - 2
+            end_index = doc_comment.range.end.line
+            while start_index >= 0 and content[start_index] == '\n':
+                top_padding += 1
+                start_index -= 1
+            while end_index < len(content) and content[end_index] == '\n':
+                # This condition will take place if theres an inline docstring
+                # following documentation.
```
damngamerz: `content[end_index] `represents the line after end marker.
yup in this case it wont be true as
`end_index > len(content)`
`bottom_padding` will be instantiated to 0 in this case :)

coalib/bearlib/languages/documentation/DocBaseClass.py
```diff
@@ -84,3 +93,95 @@ def process_documentation(self, *args, **kwargs):
...
+    def instantiate_kind(self, content, doc_comment_iterator):
+        """
+        Instantiates kind of DocumentationComment instances, which defines
+        the kind of docstring: function, class or others.
+
+        :param content:
+            Contains the file content.
+        :param doc_comment_iterator:
+            An updated iterator returning each DocumentationComment found
+            in the content.
```
Nosferatul: it was added :+1:

coalib/bearlib/languages/documentation/DocBaseClass.py
NiklasMM: You have all you need at the end of [extract_documentation_with_markers](https://github.com/damngamerz/coala/blob/blankline/coalib/bearlib/languages/documentation/DocumentationExtraction.py#L253). I see no reason, why we should return an incomplete DocComment there and then use two more methods to finalize it.

coalib/bearlib/languages/documentation/DocBaseClass.py
NiklasMM: uhhh, I still don't understand. The padding *is* 0

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -127,6 +148,37 @@ def metadata(self):
...
+        A namedtuple consisting of values about blank lines before and after
+        the documentation of ``docstring_type`` function.
+
+        These values are official standard of following blank lines before and
+        the documentation of ``docstring_type`` function.
+        """
+        return self._function_padding
+
+    @property
+    def docstring_type_regex(self):
```
NiklasMM: Why do we need getters for all those members?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -186,13 +238,44 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+                            )
+
+        docstring_type_regex = (
+            cls.DocstringTypeRegex('', '')
+            if docstyle_settings.get('docstring_type_regex', '').value == ''
+            else cls.DocstringTypeRegex(
+                *(str(sign) for sign in tuple(
+                    docstyle_settings.get('docstring_type_regex', '')))
+                  )
+                                )
```
NiklasMM: [Comment from previous review unadressed](https://github.com/coala/coala/pull/4557#discussion_r130921669)

coalib/bearlib/languages/documentation/DocBaseClass.py
damngamerz: I will add inline comments :+1:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: As everything we are fetching from `.coalang` is a standard. And we don't want anyone to set those up manually. Hence setting them as getters :)
PS- Even the previous things like markers, comment's signature....etc we are fetching from `.coalang` are introduced in DocstyleDefinition as getters.

NiklasMM: hm, ok... since Python doesn't have private members, the whole concept is a bit useless. But if you want to do it like that it's fine.

coalib/bearlib/languages/documentation/DocBaseClass.py
```diff
@@ -43,9 +45,17 @@ def extract(content, language, docstyle):
...
         docstyle_definition = DocstyleDefinition.load(language, docstyle)
-        return extract_documentation_with_markers(
+        doc_comment_iterator = extract_documentation_with_markers(
                         content, docstyle_definition)
 
+        doc_comment_iterator = DocBaseClass.instantiate_padding(
+            content, doc_comment_iterator)
+
+        doc_comment_iterator = DocBaseClass.instantiate_docstring_type(
+            content, doc_comment_iterator)
```
NiklasMM: This is still unchanged and very inefficient. This method should be generator.

You have two options: 

1) Change `instantiate_padding` and `instantiate_docstring_type` to only process one docstring at a time, so this can become a generator.
2) move the functionality from `instantiate_padding` and `instantiate_docstring_type` into `extract_documentation_with_markers` and use `yield from`

I prefer the second option.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -186,13 +238,41 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+                *(int(padding) for padding in tuple(
+                    docstyle_settings.get('class_padding', ''))))
+        except IndexError:
+            class_padding = cls.ClassPadding('', '')
+
+        try:
+            docstyle_settings['function_padding']
+            function_padding = cls.FunctionPadding(
+                *(int(padding) for padding in tuple(
+                    docstyle_settings.get('function_padding', ''))))
```
NiklasMM: you don't have to use `get` with a default value, that's why you're in a try/except block. And why are you casting this to a tuple? isn't it already an iterable?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -186,13 +238,41 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+        try:
+            docstyle_settings['class_padding']
+            class_padding = cls.ClassPadding(
+                *(int(padding) for padding in tuple(
+                    docstyle_settings.get('class_padding', ''))))
+        except IndexError:
+            class_padding = cls.ClassPadding('', '')
+
+        try:
+            docstyle_settings['function_padding']
```
NiklasMM: what is this for?

coalib/bearlib/languages/documentation/DocBaseClass.py
damngamerz: Yes I missed your previous comment we can do that...I see :+1:

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: yes no need to use get now :tada: 
yup this isn't an iterable. we need to cast into a tuple.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: yup we can remove this I guess...
as now I will use `docstyle_settings['function_padding']` instead of `docstyle_settings.get()`
which will throw a `KeyError` :+1:

SanketDG: this is what should throw the `KeyError` right?

This shouldn't be it, I was just showing an example

coalib/bearlib/languages/documentation/DocstyleDefinition.py
NiklasMM: what is it then?

damngamerz: when this reads `function_padding` from `.coalang` the values are just read as a strings.
suppose `function_padding = 1, 0`
This will be read as 
`docstyle_settings = {..., function_padding : '1, 0',...}`
we cast then into a tuple ('1', '0') so they get as iterables and then we typecast them into integers.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
             marker_dict,
             docstyle_definition)
         if doc:
+            # Instantiate padding
+            top_padding = 0
+            bottom_padding = 0
+            start_index = doc.range.start.line - 2
```
NiklasMM: please explain the `-2` in a comment

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
+            end_index = doc.range.end.line
+            while start_index >= 0 and content[start_index] == '\n':
```
NiklasMM: What about lines containing spaces?

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
+                top_padding += 1
+                start_index -= 1
+            # If the end_index is instantiated above the len(content) i.e.
+            # In case where ending marker of docstring is at the last line.
+            # Then the bottom_padding will be default to 0. This will also
+            # prevent IndexError raised by content[end_index].
+            while end_index < len(content) and content[end_index] == '\n':
+                # This condition will take place if theres an inline docstring
+                # following documentation.
+                if ((doc.marker[2]+'\n') != content[end_index-1][-4:]
```
NiklasMM: I still don't understand why this is needed. Please give an example.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
+                        and bottom_padding == 0):
+                    break
+                bottom_padding += 1
+                end_index += 1
+            doc.top_padding = top_padding
+            doc.bottom_padding = bottom_padding
```
NiklasMM: You could remove these two lines by replacing all occurrences of `top_padding` with `doc.top_padding` and dito for `bottom_padding`. There's no real need for a temporary variable.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
+            doc.range = TextRange.from_values(
+                start_index + 2,
+                1 if top_padding > 0 else doc.range.start.column,
+                end_index,
+                1 if bottom_padding > 0 else doc.range.end.column)
+
+            # Instantiate docstring_type
```
NiklasMM: This comment is inaccurate, you're just compiling regexes here

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -250,4 +251,58 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
+            start_index = doc.range.start.line - 2
+
+            # End line differs when mid marker and end marker is different
+            if doc.marker[1] == doc.marker[2]:
+                end_index = doc.range.end.line - 1
+            else:
+                end_index = doc.range.end.line
+
+            # We check for the class regex and function regex
+            # before and after the documentation.
```
NiklasMM:
```Python
def some_function():
  """
    documentation
  """
  class myPrivateClass:
     pass

This docstring would falsely be identified as a type class.

I suppose to mitigate this, you'll need to specify the position of the docstring in the coalang files, too.

File: coalib/bearlib/languages/documentation/DocstyleDefinition.py [hunk 27]
```diff
@@ -186,13 +238,38 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
         metadata = cls.Metadata(*(str(docstyle_settings.get(req_setting, ''))
                                   for req_setting in metadata_settings))
 
+        try:
+            class_padding = cls.ClassPadding(
+                *(int(padding) for padding in tuple(
```
NiklasMM: In your [previous comment](https://github.com/coala/coala/pull/4557#discussion_r131541556) you said that `docstyle_settings['class_padding']` is a string like "1, 0", if this is true, `tuple("1, 0")` will be `("1", ",", " ", "0")`, which `int` will choke on. Rather use `docstyle_settings['class_padding'].split(",")`

Also I'm wondering why this isn't caught by tests.

tests/bearlib/languages/documentation/DocBaseClassTest.py
```diff
@@ -259,6 +259,102 @@ def test_DocBaseClass_extraction_PYTHON3_4(self):
...
+        docstyle_PYTHON3_default = DocstyleDefinition.load('PYTHON3',
+                                                           'default')
+
+        for doc in DocBaseClass.extract(data, 'PYTHON3', 'default'):
+            self.assertEqual([doc.top_padding, doc.bottom_padding],
+                             [2, 2])
+
+    def test_DocBaseClass_instantiate_padding_inline_PYTHON3_7(self):
+        # To test that bottom_padding sets to nothing if docstring is
+        # followed by inline docstring.
```
NiklasMM: Why is this the desired behavior?

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -186,13 +238,38 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
+                    docstyle_settings['class_padding'])))
+        except IndexError:
+            class_padding = cls.ClassPadding('', '')
```
damngamerz: because `0, 0` will mean that official defined padding is `0, 0`. So better leave the defaults to be `'', ''`

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: we don't/shouldn't care about the blank lines having whitespaces. As we have many bears which will fix them. And then DocumentationAPI will calibrate correct paddings in the end.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: Its already there.
https://github.com/coala/coala/pull/4557/files#diff-7b910a6c281c2179ed513e9ee826e64eR289

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: I don't think so....we will require such a thing. Although I will add this in the test cases if it doesn't work we will have to think about the position.

tests/bearlib/languages/documentation/DocBaseClassTest.py
damngamerz: we have such a case in DocStyleBear.
I don't wanna break the existing tests :wink: 
https://github.com/coala/coala-bears/blob/master/tests/documentation/test_files/DocumentationStyleBear/bad_file2.py.test#L21

coalib/bearlib/languages/documentation/DocumentationExtraction.py
NiklasMM: You can't assume that users are also using other bears. The bear needs to work independently

tests/bearlib/languages/documentation/DocBaseClassTest.py
NiklasMM: how is such a case handled by documentation extraction?

NiklasMM: What I mean is, if you parse and re-assemble this docstring, what will happen?

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: Yes this fails. We have two ways to solve this
First one is a tricky but will work, I think we can fix this by little prioritizing the algorithm
something like:
1. we search the line closest to the docstring first. (whichever has least padding will be scanned by both the regexes first.)
2. If paddings are found to be equal scanning the line before starting marker with regexes first(i.e. with both `function` and `class`). Than last line will be scanned.
This will support the generic nature.

Second introducing `docstring_type_position`.
If we found its written `top` then we only scan top lines of the docstring with regexes or vice-versa.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: cool then I will use `.strip(' ')` :tada:

coalib/bearlib/languages/documentation/DocumentationExtraction.py
damngamerz: Im going with the second option looks much more future proof :+1:

tests/bearlib/languages/documentation/DocBaseClassTest.py
damngamerz: oh yes! thats simple
The affected `range` is always taken from starting marker to the end marker(except in case of padding). DocumentationAPI doesn't care about whats after the end marker. Hence if its followed by inline comment...It's still is at the same position after the assembling.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: Hmm now even I m wondering the same. You are correct about `tuple('1,0')` will be `("1", ",", " ", "0")` but what I figured out is `docstyle_settings` is a `Setting` object. Now when I do `tuple(docstyle_settings['class_padding'].value)` gives me `("1", ",", " ", "0")` but when I only do `tuple(docstyle_settings['class_padding'])` gives me `('1', '0')`(some magic :stuck_out_tongue: ). So I think let it be that way.

coalib/bearlib/languages/documentation/DocumentationExtraction.py
NiklasMM: introducing `docstring_type_position` is the correct solution. The other will not work, you'll only make it work for Python. Maybe `docstring_position` should suffice, I doubt that styles will position the docstring on different sides for different types.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
NiklasMM: uh no, please investigate

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -255,5 +256,63 @@ def extract_documentation_with_markers(content, docstyle_definition):
...
             ignore_string_match = ignore_regex.search(content[start_line])
 
+            # Instantiate padding
+            doc.top_padding = 0
+            doc.bottom_padding = 0
+            # minus 2 because we want to check the line before the marker.
+            start_index = doc.range.start.line - 2
+            end_index = doc.range.end.line
+            while (start_index >= 0 and
+                    content[start_index].strip(' ') == '\n'):
```
NiklasMM: You can remove the `' '`

coalib/bearlib/languages/documentation/DocstyleDefinition.py
damngamerz: I see why it's like that https://github.com/coala/coala/blob/41e4752a40530ec915fa01582bc73708e00c26b0/coalib/settings/Setting.py#L121

`Setting` offers common data type conversions. It has `list_delimeters` for list/tuple conversion. That's why we are getting `('1', '0')` while doing `tuple(docstyle_settings['class_padding'])` :tada:

NiklasMM: Wow, talk about overengineering.... anyway... it's ok then

NiklasMM:
@damngamerz one last thing: The commit message needs an update: `kind` doesn't exist anymore

damngamerz:
@NiklasMM Updated commit message :+1:

NiklasMM:
ack d53307c

SanketDG:
ack 070a19e

SanketDG:
@rultor merge

Issue #4580, opened 2017-07-29
Ask for ``language`` if running coala with aspect


adhikasp:
`language` is mandatory setting if section contain `aspects`. Ask input if doesn't exist.

adhikasp:
@jayvdb I am thinking of closing this because it's make UI not uniform with others case in coala. For example, if users don't specify other important setting like `files` or `bears`, coala just skip the section instead of asking user to fill it.

Issue #4463, opened 2017-07-14
Validate language setting


jayvdb:
The `language` setting is used for accessing the `Language` definitions.

Currently two bears use this, `AnnotationBear` (and three dependent bears) and `MypyBear`.  They have their own tests for this value.

Those bears currently define the setting as `language: str`.  They should be able to define a setting `language: Language` (or similar), and have automatic validation occur, with the tests for that validation being done in the coala repository.

Also as part of the aspects project, `language` will be used behind the scenes in coala even when bears do not explicitly use the `language` setting.  The validation of the setting needs to occur in a consistent location, with consistent error messages, whenever the `language` setting is *used* (but there should be no error if `language` has an invalid value but it isnt being used at all).

note there is more complicated work occurring elsewhere wrt aspects & languages (e.g. https://github.com/coala/coala/issues/4147).  This task is strictly about ensuring all `language` setting validation results in the same error message , and that validation only occurs once per `Section`.

adhikasp:
Correction, as part of aspects project, `language` will always need to be validated (and shown error) IF coala is run under "aspect-mode" because it will become the basis of bear selection, even if none of the selected bear will use `language`.

adhikasp:
Correction, in aspect-mode `language` should still thrown error if invalid because it will became the basis of bear selection (based by aspect capability and language supported by bear).

jayvdb:
This was raised due to https://github.com/coala/coala/pull/4397#discussion_r127374759

jayvdb:
See also https://github.com/coala/coala-bears/issues/1635

adhikasp:
@jayvdb what is the preferred behavior if validation fail? Does we just throw error log or drop to interactive input for users to correct it?

Makman2:
You don't have to deal with the error behaviour specifically, as you will write a type-converter function that will just check whether the passed language string is valid or not. If it's not valid, you throw some error (I don't know the exact error type now, you can look that up in other type-converter funcs), and coala will propagate the properly up to the user :+1:

Makman2:
Ah it should be a `ValueError` I believe :+1:

adhikasp:
Oh yeah right, I was thinking of doing it in `ConfigurationGathering` instead of type-converter

adhikasp:
[...]

PR #4582, opened 2017-07-29
Aspects: Add more robust setting checks

adhikasp:
Closes https://github.com/coala/coala/issues/4580 https://github.com/coala/coala/issues/4463

### DIFF
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -48,22 +48,14 @@ def validate_aspect_config(section):
     :return:        The validity of section.
     """
     aspects = section.get('aspects')
-    section_language = section.get('language')
 
     if not len(aspects):
         return False
 
-    if not len(section_language):
+    if not section.language:
         logging.warning('Setting `language` is not found in section `{}`. '
                         'Usage of aspect-based setting must include '
                         'language information.'.format(section.name))
-        return False
-
-    try:
-        Language[section_language]
-    except UnknownLanguageError as exc:
-        logging.warning('Section `{}` contain invalid language setting. '
-                        '{}'.format(section.name, exc))
         return False
 
     if len(section.get('bears')):
@@ -72,6 +64,24 @@ def validate_aspect_config(section):
                         'takes priority and will overwrite any '
                         'explicitly listed bears.'.format(section.name))
     return True
+
+
+def _set_section_language(sections):
+    """
+    Validate ``language`` setting and inject them to section if valid.
+
+    :param sections: List of sections that potentially contain ``language``.
+    """
+    for section_name, section in sections.items():
+        section_language = section.get('language')
+        if not len(section_language):
+            continue
+
+        try:
+            section.language = Language[section_language]
+        except UnknownLanguageError as exc:
+            logging.warning('Section `{}` contain invalid language setting: '
+                            '{}'.format(section_name, exc))
 
 
 def merge_section_dicts(lower, higher):
@@ -474,6 +484,7 @@ def gather_configuration(acquire_settings,
         arg_list = sys.argv[1:] if arg_list is None else arg_list
     sections, targets = load_configuration(arg_list, arg_parser=arg_parser,
                                            args=args)
+    _set_section_language(sections)
     aspectize_sections(sections)
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings)
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -5,6 +5,7 @@ import logging
 
 from coalib.collecting.Collectors import (
     collect_all_bears_from_sections, filter_section_bears_by_languages)
+from coalib.bearlib.languages.Language import Language, UnknownLanguageError
 from coalib.misc import Constants
 from coalib.output.ConfWriter import ConfWriter
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
@@ -30,14 +31,47 @@ def aspectize_sections(sections):
     :param sections:  List of section that potentially contain aspects setting.
     :return:          The new sections.
     """
-    for section_name, section in sections.items():
-        section.aspects = extract_aspects_from_section(section)
-        if section.aspects is not None and len(section.get('bears')):
-            logging.warning("'aspects' and 'bears' configuration is detected "
-                            "in section '{}'. Aspect-based configuration will "
-                            'takes priority and will overwrite any '
-                            'explicitly listed bears'.format(section_name))
+    for _, section in sections.items():
+        if validate_aspect_config(section):
+            section.aspects = extract_aspects_from_section(section)
+        else:
+            section.aspects = None
     return sections
+
+
+def validate_aspect_config(section):
+    """
+    Validate if a section contain required setting to run in aspects mode.
+
+    :param section: The section that potentially contain aspect
+                    setting.
+    :return:        The validity of section.
+    """
+    aspects = section.get('aspects')
+    section_language = section.get('language')
+
+    if not len(aspects):
+        return False
+
+    if not len(section_language):
+        logging.warning('Setting `language` is not found in section `{}`. '
+                        'Usage of aspect-based setting must include '
+                        'language information.'.format(section.name))
+        return False
+
+    try:
+        Language[section_language]
+    except UnknownLanguageError as exc:
+        logging.warning('Section `{}` contain invalid language setting. '
+                        '{}'.format(section.name, exc))
+        return False
+
+    if len(section.get('bears')):
+        logging.warning('`aspects` and `bears` setting is detected '
+                        'in section `{}`. Aspect-based configuration will '
+                        'takes priority and will overwrite any '
+                        'explicitly listed bears.'.format(section.name))
+    return True
 
 
 def merge_section_dicts(lower, higher):
--- a/coalib/settings/Section.py
+++ b/coalib/settings/Section.py
@@ -46,8 +46,11 @@ def append_to_sections(sections,
 
 def extract_aspects_from_section(section):
     """
-    Extracts aspects and their related settings from a section and create an
-    AspectList from it.
+    Extract aspects settings from a section into an AspectList.
+
+    Note that the section is assumed to already have valid and complete aspects
+    related setting. This checking could be done by
+    :meth:`coalib.settings.ConfigurationGathering.validate_aspect_config`.
 
     :param section: Section object.
     :return:        AspectList containing aspectclass instance with
@@ -55,15 +58,6 @@ def extract_aspects_from_section(section):
     """
     aspects = section.get('aspects')
     language = section.get('language')
-
-    # Skip aspects initialization if not configured in section
-    if not len(aspects):
-        return None
-
-    if not len(language):
-        raise AttributeError('Language was not found in configuration file. '
-                             'Usage of aspect-based configuration must '
-                             'include language information.')
 
     aspect_instances = AspectList(exclude=section.get('excludes'))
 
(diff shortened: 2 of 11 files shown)


coalib/settings/Setting.py
```diff
@@ -44,6 +45,21 @@ def glob_list(obj, *args, **kwargs):
...
+    """
+    Convert a string into ``Language`` object.
+
+    :param name:        String containing language name.
+    :return:            ``Language`` object.
+    :raises ValueError: If the ``name`` contain invalid language name.
+    """
+    try:
+        return Language[name]
+    except AttributeError:
```
jayvdb: This line of your patch is correct, but will need to change if https://github.com/coala/coala/pull/4643 is merged, as that fixes a fairly fundamental exception problem.

coalib/settings/Setting.py
```diff
@@ -44,6 +45,21 @@ def glob_list(obj, *args, **kwargs):
     return obj.__glob_list__(*args, **kwargs)
 
 
+def language_obj(name):
```
jayvdb: I'm not keen on this name.  This name needs to be used in `Bear.run()` override arg lists, and becomes part of our documentation.  Can it simply be `language` ?  And be used as `def run(.... , language: language):` ; while that looks like it might be wrong, it should work and do the right thing.

The other alternative is that the `run` argument validator should work correctly with `def run(.... , language: Language):` , which would require improvements to the argument validator so that dict subclass, like `Language`, are checked using the value to key-index the dict, and thus `KeyError` would be the expected exception of incorrect values instead of `ValueError`.

Also, in order to properly solve https://github.com/coala/coala/issues/4463 , this PR should include a test which is a Bear using `language: language`.

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,66 @@
...
+    :return:                The validity of section.
+    """
+    aspects = section.get('aspects')
+    aspects_language = section.get('language')
+
+    if not len(aspects):
+        return False
+
+    if not len(aspects_language):
+        logging.warning('Language setting is not found in section `{}`. '
```
jayvdb: `Language setting ..` -> `Setting 'language' ..`

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -346,12 +354,47 @@ def test_default_section_deprecation_warning(self):
...
+        self.section['aspects'] = ''
+        self.assertFalse(check_aspect_config(self.section, acquire_settings))
+
+    def test_check_aspect_config_no_language(self):
+        self.section['language'] = ''
+        with simulate_console_inputs('python'), self.assertLogs(
+                logging.getLogger()) as cm:
+            self.assertTrue(check_aspect_config(
+                            self.section, acquire_settings))
+            self.assertRegex(cm.output[0], 'Language setting is not found in '
```
jayvdb: start the string on a new line. (also below..)

tests/settings/SettingTest.py
```diff
@@ -3,9 +3,10 @@
 import unittest
 from collections import OrderedDict
 
+from coalib.bearlib.languages import Language
 from coalib.settings.Setting import (
     Setting, path, path_list, url, typed_dict, typed_list, typed_ordered_dict,
-    glob, glob_list)
+    glob, glob_list, language)
```
jayvdb: on a new line, as it isnt a glob.

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,66 @@
...
-                            'explicitly listed bears'.format(section_name))
+    for _, section in sections.items():
+        if check_aspect_config(section, acquire_settings):
+            section.aspects = extract_aspects_from_section(section)
+        else:
+            section.aspects = None
     return sections
 
 
+def check_aspect_config(section, acquire_settings):
```
jayvdb: this isnt a `check_..` , as it does UI and it modifies the `section`

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,68 @@
...
+            LogPrinter(),
+            {'language': ['The file\'s language in section `{}`'
+                          .format(section.name),
+                          'all aspect bears']},
+            section)
+        aspects_language = language_dict['language']
+        section.append(Setting('language', aspects_language))
+
+    try:
+        language(aspects_language)
```
jayvdb: Is the returned Language object cached?
It should be. The objective of the issue is that the lookup is done only once.

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,68 @@
...
+    except ValueError as exc:
+        logging.warning('{} Coala will fallback to using `bears` setting.'
+                        .format(exc))
+        return False
+
+    if len(section.get('bears')):
+        logging.warning('`aspects` and `bears` configuration is detected '
```
jayvdb: Why?
It could filter the aspects-determined bear list , using only explicitly mentioned bears.

adhikasp: Currently `collect_bears_by_aspects` could only search from all-bears list. This would need to be done in separate issue.

coalib/settings/ConfigurationGathering.py
adhikasp: I think the original issue doesn't specify *caching* the target Language object, only providing consistent location (which is using `language()`) and error message.

The missing things is 

> that validation only occurs once per Section

Which is not true in this PR because each bears will run validation independently. Its more complicated to read list of bear->check if any of them need language and I think doesn't provide any additional benefit.

jayvdb: validation running once per Section requires caching.

jayvdb: This does not require knowing if there was any bears that read `language`.  If there is a `language` setting, validate it, and cache the Language object in the Section.

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,68 @@
...
+                          .format(section.name),
+                          'all aspect bears']},
+            section)
+        aspects_language = language_dict['language']
+        section.append(Setting('language', aspects_language))
+
+    try:
+        language(aspects_language)
+    except ValueError as exc:
+        logging.warning('{} Coala will fallback to using `bears` setting.'
```
jayvdb: What if there is no `bears` setting?

coalib/settings/ConfigurationGathering.py
jayvdb: I expect the bear filtering to occur after the bears have been collected.
There has been some recent improvements in this area in `coalib/parsing/FilterHelper.py`
Could you investigate and make it a stretch goal please.

coalib/settings/ConfigurationGathering.py
```diff
@@ -22,24 +23,68 @@
...
-                            'explicitly listed bears'.format(section_name))
+    for _, section in sections.items():
+        if verify_aspect_config(section, acquire_settings):
+            section.aspects = extract_aspects_from_section(section)
+        else:
+            section.aspects = None
     return sections
 
 
+def verify_aspect_config(section, acquire_settings):
```
jayvdb: 'verify_' is the same as 'check_' mentioned previously https://github.com/coala/coala/pull/4582#discussion_r135069786 
They both imply state is being checked, and no modifications are occurring.

coalib/settings/ConfigurationGathering.py
adhikasp: It will be catched by `warn_nonexistent_targets` (if entire .coafile doesn't have bears/aspects). If only 1 section doesn't have it, coala will skip it.

coalib/settings/ConfigurationGathering.py
adhikasp: TBH I can't find any concise name/verb without it being too long (`verify_and_modify_aspect_config` doesnt seems good)...

Could you point a good name here?

coalib/settings/ConfigurationGathering.py
adhikasp: Well it will be a bit misleading if `bears is not present, but it still skip the whole section.

coalib/settings/ConfigurationGathering.py
adhikasp: This will need some design first. Filtering by name **after** bears have been collected could make some aspect unfulfilled (collect_all -> filter_aspects -> filter_bears) or if you mean collect_all -> filter_bears -> filter_aspects that make more sense.

jayvdb: I had not thought about aspects going unfulfilled.  collect_all -> filter_bears -> filter_aspects would be better.  Anyway ... create an issue about it ... it can be worked on after gsoc.

coalib/settings/ConfigurationGathering.py
jayvdb: are you sure?  A section has `aspects`, no `bears` and an invalid value for `language`.  Is this error message here going to be shown?  If so, the error message text is wrong.

coalib/settings/ConfigurationGathering.py
jayvdb: aspectize_section?

adhikasp:
I haven't organize the commits yet, but all the code should have been there

coalib/settings/ConfigurationGathering.py
```diff
@@ -30,16 +31,60 @@ def aspectize_sections(sections):
...
+    """
+    for section_name, section in sections.items():
+        section_language = section.get('language')
+        if not len(section_language):
+            continue
+
+        try:
+            section.language = Language[section_language]
+        except UnknownLanguageError as exc:
+            logging.warning('Section `{}` contain invalid language setting. '
```
jayvdb: `setting. ` should be `setting: `

coalib/settings/ConfigurationGathering.py
```diff
@@ -30,16 +31,60 @@ def aspectize_sections(sections):
...
     return sections
 
 
+def validate_aspect_config(section):
+    """
+    Validate if a section contain required setting to run in aspects mode.
+
+    :param section:         The section that potentially contain aspect
+                            setting.
+    :param acquire_setting: Method used to acquire missing setting.
```
jayvdb: not needed now, if https://github.com/coala/coala/issues/4580 is being rejected.

adhikasp: Forget to remove this...

coalib/settings/ConfigurationGathering.py
adhikasp: And this too...

adhikasp:
Add new commit to fix https://github.com/coala/coala/issues/4681

coalib/settings/ConfigurationGathering.py
```diff
@@ -30,16 +31,59 @@ def aspectize_sections(sections):
...
+
+    if len(section.get('bears')):
+        logging.warning('`aspects` and `bears` setting is detected '
+                        'in section `{}`. Aspect-based configuration will '
+                        'takes priority and will overwrite any '
+                        'explicitly listed bears.'.format(section.name))
+    return True
+
+
+def validate_and_inject_language(sections):
```
jayvdb: This can be just `_set_section_language` .  validation is assumed.  doesnt need to be part of public API

coalib/settings/Section.py
```diff
@@ -46,24 +46,18 @@ def append_to_sections(sections,
 
 def extract_aspects_from_section(section):
     """
-    Extracts aspects and their related settings from a section and create an
-    AspectList from it.
+    Extracts aspects settings from a section into an AspectList.
```
jayvdb: Singular 'Extract', not plural

jayvdb:
ack 67ef72b 5f90a8b 25893e2 c8f1405 fc75951 82ec615 9979e16

jayvdb:
reack 256854c c7135e9 1bd2e87 5476acb 11f564d 141f44f a697b09

jayvdb:
@rultor merge

PR #4589, opened 2017-07-29
WIP: DO NOT MERGE

Makman2:
PR for showing the problem of https://github.com/pytest-dev/pytest-cov/issues/165


Makman2:
Fixed upstream on pytest-cov. However an upgrade is not necessary, because coverage was fixed by using `ThreadPoolExecutor` as a workaround (and it will be kept as it's also useful to test the core on normal Python threads).

Issue #4495, opened 2017-07-20
InvalidFilterException: Add convenience methods generating messages


Makman2:
The `__init__` method could take parameters that construct a message conveniently:
```
def __init__(self, filter_name=None, available_filters=None):
    pass
```
And use it accordingly where filters are present.

Makman2:
CC @shreyans800755

ayush1999:
I'd love to work on this.
@Makman2 @shreyans800755  Can you tell me how to proceed?

Makman2:
The exception itself shall build the messages that are currently used inside the filter procedures. For that build an appropriate `__init__` function with appropriate parameters.

newbazz:
I would like to work on this issue.

Makman2:
@ayush1999 you want to get it going again? Or can I reassign? ;)

ayush1999:
I'm on it! Sorry for the delay.

newbazz:
So are you woking on it? @ayush1999

ayush1999:
Yeah! I'll amend the PR shortly. Sorry for the delay. (was caught up in some other work)

aashutoshrathi:
@newbazz Should I create these files (tests/parsing/FilterTest.py, coalib/parsing/InvalidFilterException.py and coalib/parsing/FilterHelper.py) again?

newbazz:
It sud have been created!!

aashutoshrathi:
Can you explain what exactly needs to be changed, in which file?

Makman2:
Please use `git grep` to look for the keywords mentioned in the issue description, then you'll find everything. If you don't understand an issue completely on your own, rather pick a new one :)

jayvdb:
@shreyans800755 @Makman2 , we have two PRs here:
https://github.com/coala/coala/pull/4821
https://github.com/coala/coala/pull/4610
I'll let you guys sort this one out.

PR #4610, opened 2017-08-02
InvalidFilterException.py: Constructor added

ayush1999:
InvalidFilterException.py : Constructor added.
Added the `__init__` method inside `InvalidFilterException.py`.
Now, We can directly pass arguments to `InvalidFilterException`, rather than passing the exception error message as a separate  argument.
Closes: #4495

### DIFF
--- a/coalib/parsing/FilterHelper.py
+++ b/coalib/parsing/FilterHelper.py
@@ -27,10 +27,8 @@ class FilterHelper:
                 get_all_bears)
             all_bears = get_all_bears(LogPrinter())
         if not cls.is_valid_filter(filter_name):
-            raise InvalidFilterException('{!r} is an invalid filter. '
-                                         'Available filters: {}'.format(
-                                             filter_name,
-                                             cls.get_all_filters_str()))
+            raise InvalidFilterException(
+                   filter_name, cls.get_all_filters_str())
         if not filter_args or len(filter_args) == 0:
             return all_bears
         return cls.available_filters[filter_name](all_bears, filter_args)
--- a/coalib/parsing/InvalidFilterException.py
+++ b/coalib/parsing/InvalidFilterException.py
@@ -1,2 +1,17 @@
 class InvalidFilterException(Exception):
-    pass
+
+    def __init__(self, filter_name, available_filters):
+        self.filter_name = filter_name
+        self.available_filters = available_filters
+        if self.available_filters is None:
+            message = 'Invalid Filter'
+
+        if self.available_filters is not None:
+            message = 'Invalid filter. Available filters are : ' + \
+                     ' ,'.join(self.available_filters)
+
+        else:
+            message = self.filter_name + \
+                      ' is an invalid filter. Available filters: ' + \
+                      ' ,'.join(self.available_filters)
+        super().__init__(message)
--- a/tests/parsing/FilterTest.py
+++ b/tests/parsing/FilterTest.py
@@ -125,3 +125,11 @@ class FilterTest(unittest.TestCase):
             # All bear plus 1 line holding the closing colour escape sequence.
             self.assertEqual(len(stdout.strip().splitlines()),
                              TEST_BEARS_COUNT + 1)
+
+    def check_by_style(self):
+        retval, stdout, stderr = execute_coala(
+            coala.main, 'coala', '-B' '--filter-by', 'can_detect')
+        self.assertEqual(retval, 0)
+        self.assertRaises(InvalidFilterException,
+                          InvalidFilterException,
+                          str(filter), FilterHelper.get_all_filters_str())


ayush1999:
Also, I have closed the initial PR.

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,9 @@
 class InvalidFilterException(Exception):
-    pass
+    def __init__(self, filter, available_filters):
+        self.filter = filter
+        self.available_filters = available_filters
```
Makman2: can't you just call `super().__init__(message)`? Then you don't need to override `__str__`.

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,9 @@
 class InvalidFilterException(Exception):
-    pass
+    def __init__(self, filter, available_filters):
```
Makman2: Please support `None` values for those. And if one of those parameters is `None`, it shall not be printed.

Makman2: (the corresponding information shall not be printed*)

ayush1999: So, if any one of the parameters is zero, what should the output be?

Makman2: Would do following:
```
# both None
message = 'Invalid filter'
# only available_filters=None
message = '{!r} is an invalid filter'.format(filter)  # or using your message: '{} is an invalid filter'
# only filter=None
message = 'Invalid filter. Available filters: {}'.format(available_filters)
# both not None
message = '{!r} is an invalid filter. Available filter: {}'.format(available_filters)
```

shreyans800755: `filter` -> `filter_name` to avoid confusion in future.

shreyans800755: I think we should pass the list of all the filters instead of passing it in string format. Makes more sense to format it over here.

ayush1999:
@shreyans800755 Would passing them as list or a string make a difference?

ayush1999:
@Makman2 If I use `super().__init__ ` , I will have to change the message as:
` filter_name + message + available filters`.
How can I do this using `super()` ? And does using `__str__` present any difficulties?

Makman2:
Overriding `__str__` bypasses the exceptions default behaviour which we can use in this case.
Using `super()` works super easy: You take the code of `__str__` and put it inside `__init__`. But instead of returning the message, you just feed it into `super().__init__(message)`.

shreyans800755:
> Would passing them as list or a string make a difference?

Yes, it allows the exception class to format it differently if needed.

One more thing, please follow guides given at coala.io/commit and update the commit message accordingly.

coalib/parsing/FilterHelper.py
```diff
@@ -27,10 +27,8 @@ def apply_filter(cls, filter_name, filter_args, all_bears=None):
                 get_all_bears)
             all_bears = get_all_bears(LogPrinter())
         if not cls.is_valid_filter(filter_name):
-            raise InvalidFilterException('{!r} is an invalid filter. '
-                                         'Available filters: {}'.format(
-                                             filter_name,
-                                             cls.get_all_filters_str()))
+            raise InvalidFilterException(
+                str(filter_name), cls.get_all_filters_str())
```
shreyans800755: No need to convert `filter_name` to `str` explicitly, its already a string

shreyans800755:
@Makman2 IMO we don't need the condition `available_filters == None` as we are passing it ourselves and not relying on user input for that.

shreyans800755:
@Makman2 IMO we don't need the condition `available_filters == None` as we are passing it ourselves and not relying on user input for that.

shreyans800755:
IMO, we don't even need the condition `filter_name == None` as arg parser takes care of that by enforcing at least one param which in this case will be stored in `filter_name` variable. No need to make extra check for that.

shreyans800755:
@ayush1999 I'm meant to change the commit message and not the PR title.
It should be of the format:
```
filename: result achieved
description
Closes/Fixes <issue link>
```

shreyans800755:
@ayush1999 I'm meant to change the commit message and not the PR title.
It should be of the format:
```
filename: result achieved
description
Closes/Fixes <issue link>
```

Makman2:
I just thought of making the exception a bit more mature by supporting different combinations of `filter_name` and `available_filters`, so you can easily use it yourself if you don't need/have any of those parameters. But how you like, we can also remove those extra conditions :+1: @shreyans800755

shreyans800755:
@ayush1999 Please add the reference to linked issue to make it easy to track. Have a look at coala.io/commit

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
...
-    pass
+class InvalidFilterException(BaseException):
+
+    def __init__(self, filter_name, available_filters):
+        self.filter = filter_name
+        self.available_filters = list(available_filters.split(', '))
+        if self.filter is None and self.available_filters is None:
+            message = 'Invalid Filter'
+
+        if self.filter is None and self.available_filters is not None:
```
shreyans800755: `self.filter` is never going to be `None`. Because argparser will make sure that there will be at least one filter available.

coalib/parsing/FilterHelper.py
```diff
@@ -27,10 +27,8 @@ def apply_filter(cls, filter_name, filter_args, all_bears=None):
                 get_all_bears)
             all_bears = get_all_bears(LogPrinter())
         if not cls.is_valid_filter(filter_name):
-            raise InvalidFilterException('{!r} is an invalid filter. '
-                                         'Available filters: {}'.format(
-                                             filter_name,
-                                             cls.get_all_filters_str()))
+            raise InvalidFilterException(
+                   filter_name, cls.get_all_filters_str())
```
Makman2: indentation: 4 spaces relative to `raises` :)

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
-class InvalidFilterException(Exception):
-    pass
+class InvalidFilterException(BaseException):
```
Makman2: No please keep it as an `Exception`, `BaseException` is used for "special" exceptions like `KeyboardInterrupt`.

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
...
+
+    def __init__(self, filter_name, available_filters):
+        self.filter = filter_name
```
Makman2: why not `self.filter_name`? :)

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
...
+        self.available_filters = list(available_filters.split(', '))
```
Makman2: I think you should already accept a list, instead of a comma-separated string^^ And below you can construct the string with `', '.join(available_filters) `

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
...
+            message = 'Invalid Filter'
+
+        if self.available_filters is not None:
+            message = 'Invalid filter. Available filters are : ' + \
+              str(self.available_filters)
+
+        else:
+            message = self.filter + \
+                      ' is an invalid filter. Available filters: ' + \
+                      str(self.available_filters)
```
Makman2: @shreyans800755 thought not to use different conditions, so we have always the same message and require the user of this exception to pass always the filter-name and available-filters. Both is fine with me, although @shreyans800755's variant is easier to implement.

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,19 @@
...
+
+        super().__init__(message)
+        raise self
```
Makman2: never do this in an exception. The user shall raise exceptions himself, now the construction of this exception object always raises an exception.

Makman2:
You have two empty files committed: `encrypted` and `ff`. Please remove them.

tests/parsing/FilterTest.py
```diff
@@ -125,3 +125,11 @@ def test_filter_by_can_detect_null(self):
...
             self.assertEqual(len(stdout.strip().splitlines()),
                              TEST_BEARS_COUNT + 1)
+
+    def check_by_style(self):
+        retval, stdout, stderr = execute_coala(
+            coala.main, 'coala', '-B' '--filter-by', 'can_detect')
+        self.assertEqual(retval, 0)
+        self.assertRaises(InvalidFilterException,
+                          InvalidFilterException,
+                          str(filter), FilterHelper.get_all_filters_str())
```
Makman2: could you use a custom string instead of invoking `str` on the python built-in function `filter`?
```
>>> str(filter)
"<class 'filter'>"
```

coalib/parsing/InvalidFilterException.py
```diff
@@ -1,2 +1,17 @@
...
+
+    def __init__(self, filter_name, available_filters):
+        self.filter_name = filter_name
+        self.available_filters = available_filters
+        if self.available_filters is None:
+            message = 'Invalid Filter'
+
+        if self.available_filters is not None:
+            message = 'Invalid filter. Available filters are : ' + \
+                     ' ,'.join(self.available_filters)
```
Makman2: indentation + please use braces instead of `\` for line continuation (same below)

Makman2: Have you read my previous comment regarding those if-conditions?

ayush1999:
@palash25 Sure! Thanks.

_Sent from my Lenovo P1a42 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

tests/parsing/FilterTest.py
palash25: @ayush1999  your commit message `changes made` doesn't follow the coala commit guidelines. Please refer to [this](http://api.coala.io/en/latest/Developers/Writing_Good_Commits.html) and edit your commit message accordingly.

jayvdb:
Removing from review queue as we have two PRs for the same issue.  See the issue.

Makman2:
Because of not being assigned to the issue and it has be taken by someone else, closing.

Issue #4613, opened 2017-08-04
Use RuntimeError for unexpected problems in Core


jayvdb:
https://github.com/coala/coala/pull/4192 used warnings for unexpected errors.
They should use `RuntimeError` so the problems are seen and reported.

Makman2:
I guess this is blocked by https://github.com/coala/coala/issues/4529. Once fast-fail is supported, we can reliably throw anything we like, and replace the `logging.warning`s with `RuntimeError`s 👍

Issue #4529, opened 2017-07-22
Fail fast in Session._schedule_bears


sils:
there's a warning about stuff that could go wrong if some other stuff isn't properly coded; those kinds of warnings are usually never used and result in unneeded complexity

sils:
CC @Makman2 who doesn't like this idea

sils:
https://github.com/coala/coala/pull/4526/commits/45a49faa269a66d62866cd6164a9278027c4406b#diff-94f6470cc4dcfa1e33ff5eb050fd5191R209

Makman2:
I would argue to keep it: It's also not testable when throwing an exception instead. Also it doesn't guarantee a fast fail because the code runs inside `Future` callbacks.

Imo it's a reasonable safety measure^^

Makman2:
I changed my mind, fast failing is always better. A bit complicated though due to multiprocessing ;)

palash25:
@Makman2 so instead of giving a warning about the dependencies not being resolved in ``_schedule_bears`` we want the session to fail/terminate?

Makman2:
Yes.

jayvdb:
Note that https://github.com/coala/coala/pull/4614 is likely the same problem as this issue.  Already fixed.

jayvdb:
@palash25 , you wanted this one ? :-)

palash25:
I was thinking of giving this a shot after GSoC.

jayvdb:
I'm going to need it for https://github.com/coala/coala/pull/4169 , or https://github.com/coala/coala/pull/4614 needs to get unblocked ( or a decent reason given why it is still blocked ).

jayvdb:
In order to unblock myself, I have an alternative at https://github.com/coala/coala/pull/4169

PR #4212, opened 2017-05-10
Makman2/nextgen3-10

Makman2:
The new bears ;)

PR #4614, opened 2017-08-04
Core: Use RuntimeError for unexpected errors

jayvdb:
Also add a specific coverage rule to skip these RuntimeError.

Fixes https://github.com/coala/coala/issues/4613
Related to https://github.com/coala/coala/issues/4166


coalib/core/Core.py
```diff
@@ -266,8 +266,8 @@ def _cleanup_bear(self, bear):
             # Check the DependencyTracker additionally for remaining
             # dependencies.
             resolved = self.dependency_tracker.are_dependencies_resolved
-            if not resolved:  # pragma: no cover
-                logging.warning(
+            if not resolved:
+                raise RuntimeError(  # Unreachable code
                     'Core finished with run, but it seems some dependencies '
                     'were unresolved: {}. Ignoring them, but this is a bug, '
```
SanketDG: not sure if "ignoring them" now makes sense because we abort anyway

setup.cfg
```diff
@@ -61,3 +61,4 @@ exclude_lines =
     pragma ${PLATFORM_SYSTEM}: no cover
     pragma ${OS_NAME}: no cover
     pragma Python [0-9.,]*${PYTHON_VERSION}[0-9.,]*: no cover
+    raise RuntimeError.*Unreachable code
```
Makman2: that's bad, not all RuntimeError shall not be covered. Rather do a `pragma` imo

coalib/core/Core.py
```diff
@@ -212,8 +212,8 @@ def _schedule_bears(self, bears):
         """
         for bear in bears:
             if self.dependency_tracker.get_dependencies(
-                    bear):  # pragma: no cover
-                logging.warning(
+                    bear):
+                raise RuntimeError(  # Unreachable code
```
Makman2: The problem is currently that the core doesn't exit properly with this exception, as this function gets also run on the processes and following cleanups aren't executed.

Makman2:
unack d04cf2a

Makman2:
Using exceptions is unfortunately not so easy and does not solve the problem completely. Everything is executed in subprocesses/threads, and we need to take care of proper cleanup.

I already started to work on this once, but https://github.com/coala/coala/pull/4212 wasn't merged yet and I didn't want to let the PRs stack for the core^^

setup.cfg
jayvdb: They are not all covered.  They must have a comment which says "Unreachable code" otherwise they wont be included by this rule.
"pragma: no cover" says nothing.

coalib/core/Core.py
jayvdb: What "following cleanups".

The current logic will keep iterating on the for loop.  Possibly cleaning up other bears?

So I should defer the raise until after the loop?

setup.cfg
Makman2: Ah sorry misread :+1:

coalib/core/Core.py
Makman2: If we raise we need to break out immediately of the event loop. As the event loop will run on external processes, there may be other `_schedule_bears` functions running concurrently and trying to run the remaining bears.

I would suggest doing this:
```python
def _throw_exception(self, ex):
    # do cleanups, like unscheduling all bears by cleaning up self.running_tasks etc.

    # either raise here now:
    raise ex
    # or defer to self.run
    self._exception = ex

# inside self.run:
def run(self):
    try:
        self.event_loop.run_forever()
        if self._exception:
            raise self._exception
    finally:
        self.event_loop.close
```

jayvdb: A lot of that already exists.  We certainly should not rely on no exceptions occurring inside any function; never know when/where one will occur (i.e. `KeyboardInterrupt`).
How about this commit on top
https://github.com/jayvdb/coala/commit/76afe4855dd0e4ccc8aa3f5f5e9ca4a7f47e1299

jayvdb: hmm, that commit causes a problem :/

jayvdb: Ya, using `close()` was the problem; changing it to `stop()` fixes it. https://github.com/jayvdb/coala/commit/4fa4fc94ef0be5abd8aac5c3d93bedf5ca0dc471

nemaniarjun:
@Makman2 whats the call on this one? I think we should close this. As the original issue is blocked by "fast failing" issue.

Makman2:
This can be merged once fail-fast-support is implemented, we still want to have this PR in fact :)

Marking this blocked.

Makman2:
Blocked by https://github.com/coala/coala/issues/4529

jayvdb:
I have an alternative approach for the pragma problem at https://github.com/coala/coala/pull/4169

Issue #3065, opened 2016-11-20
no need for # +ELLIPSIS


userzimmermann:
many doctests use `# +ELLIPSIS` to support `...` in outputs. This seems to be default and therefore obsolete :)

jayvdb:
how are you planning to remove it?
As far as I am aware, doctest runner will fail unless it exists.

Makman2:
Does not work, `# +ELLIPSIS` is not a default at least for our test setup. Is it possible to set some flag in `setup.cfg`?

adtac:
There's https://github.com/scikit-learn/scikit-learn/blob/master/setup.cfg#L23

sils:
IMO remove it, it works with our test setup and if someone uses another test setup he can still configure it to be default.

PR #4641, opened 2017-08-16
doctest: Remove unnecessary +ELLIPSIS

jayvdb:
doctest option +ELLIPSIS is enabled by default in pytest

Fixes https://github.com/coala/coala/issues/3065


jayvdb:
Initially acked on #4643

Issue #4661, opened 2017-08-24
Create decorator to map bear's setting with aspect/taste


adhikasp:
```python
@map_setting_to_aspect(
        remove_all_unused_imports=(get_aspect('UnusedImport')
                                   .remove_non_standard_import),
        remove_unused_variables=get_aspect('UnusedLocalVariable'))
def run(self, filename, file,
        remove_all_unused_imports: bool=True,
        remove_unused_variables: bool=True):
    pass
```

The `run` argument will be overrided by aspect's setting (only if it present). This make bear's aspectization much more easy.

PR #4662, opened 2017-08-24
Aspect: Create ``map_setting_to_aspect`` decorator

adhikasp:
```python
@map_setting_to_aspect(
        remove_all_unused_imports=(get_aspect('UnusedImport')
                                   .remove_non_standard_import),
        remove_unused_variables=get_aspect('UnusedLocalVariable'))
def run(self, filename, file,
        remove_all_unused_imports: bool=True,
        remove_unused_variables: bool=True):
    pass
```

The `run` argument will be overrided by aspect's setting 
(only if it present). This make bear's aspectization much more easy.

Closes https://github.com/coala/coala/issues/4661

### DIFF
--- a/coalib/bearlib/aspects/__init__.py
+++ b/coalib/bearlib/aspects/__init__.py
@@ -6,6 +6,7 @@ from pkgutil import iter_modules
 from types import ModuleType
 
 from .base import aspectbase
+from .decorators import map_setting_to_aspect
 from .exceptions import (AspectTypeError, AspectNotFoundError,
                          MultipleAspectFoundError)
 from .meta import aspectclass
@@ -21,7 +22,8 @@ from .root import Root
 __all__ = ['Root', 'Taste', 'TasteError',
            'aspectclass', 'aspectbase', 'AspectTypeError',
            'AspectNotFoundError', 'MultipleAspectFoundError',
-           'AspectList']
+           'AspectList', 'map_setting_to_aspect',
+           ]
 
 
 class aspectsModule(ModuleType):
--- a/coalib/bearlib/aspects/decorators.py
+++ b/coalib/bearlib/aspects/decorators.py
@@ -1,0 +1,51 @@
+
+from functools import wraps
+
+from .taste import Taste
+from .meta import aspectclass
+from coalib.settings.FunctionMetadata import FunctionMetadata
+
+
+def map_setting_to_aspect(**aspectable_setting):
+    """
+    Map function arguments with aspect and override it if appropriate.
+
+    This decorator can be used by ``Bear.run()`` to automatically map and
+    override bear's setting value with their equivalent aspect or taste.
+
+    The order of setting override from the lowest to highest is:
+    - Setting default (in bear's run argument)
+    - Aspect/taste default (if aspect is activated in Section)
+    - Explicit aspect/taste default (if aspect is activated in Section)
+    - Explicit setting
+
+    :param aspectable_setting:
+        A dictionary of settings as keys and their equivalent aspect or taste
+        as value.
+    """
+    def _func_decorator(func):
+        @wraps(func)
+        def _new_func(self, *args, **kwargs):
+            if self.section.aspects:
+                aspects = self.section.aspects
+                for arg, aspect_value in aspectable_setting.items():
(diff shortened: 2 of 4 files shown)


jayvdb:
Please dont put code fragments in the commit message.
Just describe the enhancements.
(and your code fragment in the commit body has incorrect indentation anyway.)

coalib/bearlib/aspects/decorators.py
```diff
@@ -0,0 +1,33 @@
+from .taste import Taste
+from .meta import aspectclass
+
+
+def map_setting_to_aspect(**aspectable_setting):
+    """
+    Override function arguments value with coala's aspect and taste.
+
+    This decorators could be used by ``Bear.run()`` to automatically map and
```
hemangsk: decorator*  ?

coalib/bearlib/aspects/__init__.py
```diff
@@ -21,7 +22,7 @@
 __all__ = ['Root', 'Taste', 'TasteError',
            'aspectclass', 'aspectbase', 'AspectTypeError',
            'AspectNotFoundError', 'MultipleAspectFoundError',
-           'AspectList']
+           'AspectList', 'map_setting_to_aspect']
```
jayvdb: add trailing comma's and put the final `]` on a new line so it doesnt need updating all the time.

coalib/bearlib/aspects/decorators.py
```diff
@@ -0,0 +1,37 @@
+
+from functools import wraps
+
+from .taste import Taste
+from .meta import aspectclass
+
+
+def map_setting_to_aspect(**aspectable_setting):
+    """
+    Override function arguments value with coala's aspect and taste.
```
jayvdb: 'Override' ?

What happens if a setting is explicitly in .coafile, and is in an aspect? The aspect overrides the explicit setting?  I hope not.

coalib/bearlib/aspects/decorators.py
```diff
@@ -0,0 +1,37 @@
...
+    override bear's setting value with their equivalent aspect or taste.
+
+    :param aspectable_setting:
+        A dictionary of settings as keys and their equivalent aspect or taste
+        as value.
+    """
+    def _func_decorator(func):
+        @wraps(func)
+        def _new_func(self, *args, **kwargs):
+            if hasattr(self, 'section') and self.section.aspects:
```
jayvdb: Why `hasattr(self, 'section')` ; i think it should raise an exception if there is no `.section`?

coalib/bearlib/aspects/decorators.py
adhikasp: Um yes it is. It will **always** override explicit or default setting with aspect/taste setting if coala is run under aspect, except for those setting that doesn't mapped here (could be used by a very specific linter setting like `pep_ignore`) . Do we want different behavior here?

coalib/bearlib/aspects/decorators.py
adhikasp: Yeah right `hasattr(self, 'section')` is not necessary here.

coalib/bearlib/aspects/decorators.py
adhikasp: Priority: old-setting default -> explicit old-setting -> aspect taste default -> explicit aspect taste

jayvdb: explicit user choices should never be overriden automatically.
I think it should be:

explicit setting -> explicit aspect taste (if aspects  activated in Section) -> aspect taste default (if aspects activated in Section) -> setting default

(please stop calling them 'old-setting' - `Setting` is still a core component, and people will still keep using them at least until aspects are perfect for *everyones* needs, which wont be soon, and even then a Setting may still be used internally by the `run` method, until we also deprecate the `run` method. )

adhikasp: Okay I agree, will change it.

jayvdb: Then this function docstring needs to change.

jayvdb:
ack 76b1e99

jayvdb:
@rultor merge

Issue #4552, opened 2017-07-24
Remove all use of line continuation


jayvdb:
`\` in Python is an abomination.

From a `ast` perspective it makes it very hard to separate logical lines from physical lines, requiring any ast reader to have lots of extra crazy code to handle this one piece of unnecessary syntax.

Add it to the banned word list in `.coafile`

```
$ git grep '\\$' coalib
coalib/bearlib/aspects/Smell.py:            System.out.printf("Welcome %s %s, a %d-year-old %s "\
coalib/bearlib/aspects/Smell.py:"from %s who works as a%s\n",firstName, lastName, age, gender, city, \
coalib/bearlib/languages/Language.py:            assert isclass(arg), \
coalib/bearlib/languages/Language.py:                'This decorator is made for classes. Did you mean to use ' \
coalib/bearlib/languages/Language.py:    OrderedDict([(<Version('2.7')>, '#'), (<Version('3.3')>, '#'), \
coalib/bearlib/languages/documentation/DocumentationComment.py:        >>> from coalib.bearlib.languages.documentation.DocumentationComment \
coalib/bearlib/languages/documentation/DocumentationComment.py:        >>> from coalib.bearlib.languages.documentation.DocstyleDefinition \
coalib/coala_main.py:from coalib.results.result_actions.PrintMoreInfoAction import  \
coalib/coala_main.py:from coalib.results.result_actions.PrintDebugMessageAction import \
coalib/misc/BuildManPage.py:        $ python setup.py build_manpage --output=coala.1 \
coalib/results/Result.py:    >>> r = Result('origin','{arg1} and {arg2}', \
coalib/settings/Annotations.py:    ValueError: Couldn't convert value 'str' to any specified type or find it \
coalib/testing/LocalBearTestHelper.py:        assert bear_output_generator is not None, \
coalib/testing/LocalBearTestHelper.py:                      tempfile_kwargs=tempfile_kwargs) as (file, fname), \
```

mikanaji:
Can I get this one?

newbazz:
I would like to work on this if no ones working.

IamRaviTejaG:
@newbazz I'm working on it, please feel free to work on any other issue. :)

shreyateeza:
I would like to work on this if @IamRaviTejaG is not working

Makman2:
Unassigning due to inactivity.

raduklb:
I would like to work on this issue.

jayvdb:
[...]

PR #4670, opened 2017-08-25
Remove use of line continuation

IamRaviTejaG:
Remove use of line continuation (/'s) completely.

Closes https://github.com/coala/coala/issues/4552

### DIFF
--- a/coalib/bearlib/languages/documentation/DocumentationExtraction.py
+++ b/coalib/bearlib/languages/documentation/DocumentationExtraction.py
@@ -203,7 +203,7 @@ def _extract_doc_comment_from_line(content, line, column, regex,
         if doc_comment:
             return end_line, end_column, doc
         else:
-            malformed_comment = MalformedComment(dedent("""\
+            malformed_comment = MalformedComment(dedent("""
                 Please check the docstring for faulty markers. A starting
                 marker has been found, but no instance of DocComment is
                 returned."""), line)
--- a/coalib/results/Result.py
+++ b/coalib/results/Result.py
@@ -1,8 +1,7 @@
 import uuid
 from os.path import relpath
 
-from coala_utils.decorators import (
-    enforce_signature, generate_ordering, generate_repr, get_public_members)
+from coala_utils.decorators import (enforce_signature, generate_ordering, generate_repr, get_public_members)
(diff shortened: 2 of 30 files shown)


coalib/results/Result.py
```diff
@@ -1,8 +1,7 @@
 import uuid
 from os.path import relpath
 
-from coala_utils.decorators import (
-    enforce_signature, generate_ordering, generate_repr, get_public_members)
+from coala_utils.decorators import (enforce_signature, generate_ordering, generate_repr, get_public_members)
```
adhikasp: Using `(` to break line is okay (and in fact the preferred way).

coalib/bearlib/languages/documentation/DocumentationExtraction.py
```diff
@@ -203,7 +203,7 @@ def _extract_doc_comment_from_line(content, line, column, regex,
         if doc_comment:
             return end_line, end_column, doc
         else:
-            malformed_comment = MalformedComment(dedent("""\
+            malformed_comment = MalformedComment(dedent("""
```
adhikasp: I think this will make whatever intended to do here became changed, maybe they dont want the newline character in this docs (not sure though)

new.txt
```diff
@@ -0,0 +1,5 @@
+Remove use of line continuation
```
adhikasp: This file is not needed, you can write what changes in commit message.

IamRaviTejaG:
@adhikasp Yeah, thanks for the suggestion. Will work on it accordingly.

nealkarpe:
Hi @IamRaviTejaG.
1) As @adhikasp pointed out, you're exceeding the line char limit. There's a reason that the commit checklist asks you to run coala locally before making a PR (the PycodestyleBear would've reported this error :smiley:).It's always a good idea to run coala and then commit!
2) I think in the issue, you were supposed to change the .coafile itself to add `\` to the banned word list. Please check the issue again.
3) Also, change `/` to `\` in the commit message :grinning:

prnvdixit:
@nealkarpe This PR is closed 😉 Please review the actual PR corresponding to the issue at https://github.com/coala/coala/pull/4720 😄 👍

Issue #4724, opened 2017-09-21
Unable to collect Bear from X - Bear package names clash with bear's dependency module name


refeed:
Related to https://github.com/coala/coala/issues/2053

https://travis-ci.org/coala/coala-bears/jobs/276993965#L3832
```
The command "bash .ci/tests.sh" exited with 0.
1.95s$ .ci/generate_bear_requirements.py
WARNING:root:Unable to collect bears from /home/travis/build/coala/coala-bears/bears/general/LinkBackupBear.py. Probably the file is malformed or the module code raises an exception.
INFO:root:Exception was:
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coalib/collecting/Collectors.py", line 134, in icollect_bears
    for bear in _import_bears(matching_file, kinds):
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coalib/collecting/Collectors.py", line 26, in _import_bears
    types=list):
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coalib/collecting/Importers.py", line 160, in iimport_objects
    attributes, local)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coala_utils/decorators.py", line 26, in yield_once_generator
    for item in iterator(*args, **kwargs):
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coalib/collecting/Importers.py", line 119, in _iimport_objects
    module = _import_module(file_path)
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/coalib/collecting/Importers.py", line 31, in _import_module
    return __import__(module_name)
  File "/home/travis/build/coala/coala-bears/bears/general/LinkBackupBear.py", line 9, in <module>
    import tubeup.__main__
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/tubeup/__main__.py", line 30, in <module>
    import internetarchive.cli
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/internetarchive/cli/__init__.py", line 27, in <module>
    from internetarchive.cli import ia, ia_configure, ia_delete, ia_download, ia_list, \
  File "/home/travis/virtualenv/python3.5.3/lib/python3.5/site-packages/internetarchive/cli/ia_metadata.py", line 50, in <module>
    from backports import csv
[...]

PR #4725, opened 2017-09-21
Importers: Clean sys.path after importing in _import_module

refeed:
Clean sys.path after importing in Importers._import_module so
coala-bears package names don't clash with Bear dependencies
name.

Fixes https://github.com/coala/coala/issues/4724

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


jayvdb:
@gitmate-bot rebase

coalib/collecting/Importers.py
```diff
@@ -28,7 +28,14 @@ def _import_module(file_path):
                 module_name = cased_module_name
                 break
 
-    return __import__(module_name)
+    module = __import__(module_name)
+
+    # Change `module_dir` index in `sys.path` to the last index, so bear's
+    # dependencies don't clash with coala-bears packages that have the same
+    # name with it.
+    sys.path.append(sys.path.pop(sys.path.index(module_dir)))
```
jayvdb: the `if` on line 18 needs to be removed, so that `module_dir` is always forced to the front of the list, so that another item in the path cant be confused with the `module_dir`.

Maybe also try : first "assert" that `module_dir` is not in the sys.path, and issue a really strong warning , force insertion at the front of the list, and then remove it here afterwards.

refeed: By "the front of the list" does it mean after the last index of the list?

jayvdb: front of the list = first path that Python uses to find the package name.

refeed: oh ok, I see

coalib/collecting/Importers.py
```diff
@@ -15,8 +15,13 @@ def _import_module(file_path):
     module_name = os.path.splitext(os.path.basename(file_path))[0]
     module_dir = os.path.dirname(file_path)
 
+    # Force insert `module_dir` at front of the `sys.path`
     if module_dir not in sys.path:
         sys.path.insert(0, module_dir)
+    elif sys.path.index(module_dir) == 0:
+        pass
+    else:
+        sys.path.insert(0, sys.path.pop(sys.path.index(module_dir)))
```
jayvdb: Need a warning if `module_dir` was in `sys.path`, because it is going to be removed on line 40, and the user probably either 1) was manipulating sys.path to workaround this bug (so they can now drop their workaround code), or 2) has some other reason and they are going to be surprised that we're deleting it from `sys.path`

refeed: okay

coalib/collecting/Importers.py
```diff
@@ -28,6 +25,18 @@ def _import_module(file_path):
                 module_name = cased_module_name
                 break
 
+    if module_dir not in sys.path:
```
jayvdb: `module_dir` should never be in `sys.path`.  If it is, that is a bug.

refeed: ok, so I think we need to remove this `if` block and assuming that `module_dir` would never be in `sys.path`?

jayvdb: `module_dir` should *never* be in `sys.path`.

Add an assert that `module_dir` is not in `sys.path`.

Destroy this `if` block, and rewrite this patch based on the above, fixing any test which violates this rule.

nemaniarjun:
@refeed the tests are failing, are you still on this? or should we close it? :)

jayvdb:
ping @refeed

PR #4726, opened 2017-09-25
coala: Add filter-by option for json and format

shreyans800755:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/coala_modes.py
+++ b/coalib/coala_modes.py
@@ -46,8 +46,6 @@ def mode_json(args, debug=False):
     import json
 
     from coalib.coala_main import run_coala
-    from coalib.misc.DictUtilities import inverse_dicts
-    from coalib.misc.Exceptions import get_exitcode
     from coalib.output.Logging import configure_json_logging
     from coalib.output.JSONEncoder import create_json_encoder
 
@@ -55,25 +53,10 @@ def mode_json(args, debug=False):
         log_stream = configure_json_logging()
 
     JSONEncoder = create_json_encoder(use_relpath=args.relpath)
-    results = []
 
-    if args.show_bears:
-        try:
-            from coalib.parsing.FilterHelper import apply_filter
+    results, exitcode, _ = run_coala(args=args, debug=debug)
 
-            local_bears, global_bears = apply_filter(
-                'language', args.filter_by_language)
-            bears = inverse_dicts(local_bears, global_bears)
-            for bear, _ in sorted(bears.items(),
-                                  key=lambda bear_tuple:
-                                  bear_tuple[0].name):
-                results.append(bear)
-        except BaseException as exception:  # pylint: disable=broad-except
-            return get_exitcode(exception)
-    else:
-        results, exitcode, _ = run_coala(args=args, debug=debug)
-
-    retval = {'bears': results} if args.show_bears else {'results': results}
+    retval = {'results': results}
 
     if args.log_json:
         retval['logs'] = [json.loads(line) for line in
--- a/coalib/coala_modes.py
+++ b/coalib/coala_modes.py
@@ -64,7 +64,7 @@ def mode_json(args, debug=False):
 
     if args.output:
         filename = str(args.output[0])
-        with open(filename, 'w+') as fp:
+        with open(filename, 'w') as fp:
             json.dump(retval, fp,
                       cls=JSONEncoder,
                       sort_keys=True,
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -407,6 +407,16 @@ def print_results_formatted(log_printer,
                 exception)
 
 
+def print_bears_formatted(bears, format=None):
+    format_str = format or ('name:{name}:can_detect:{can_detect}:'
+                            'can_fix:{can_fix}:description:{description}')
+    print('\n\n'.join(format_str.format(name=bear.name,
+                                        can_detect=bear.CAN_DETECT,
+                                        can_fix=bear.CAN_FIX,
+                                        description=bear.get_metadata().desc)
+                      for bear in bears))
+
+
 def print_affected_files(console_printer,
                          log_printer,
                          result,
@@ -911,7 +921,8 @@ def show_bear(bear,
 def print_bears(bears,
                 show_description,
                 show_params,
-                console_printer):
+                console_printer,
+                args=None):
     """
     Presents all bears being used in a stylized manner.
 
@@ -922,6 +933,7 @@ def print_bears(bears,
     :param show_params:      True if the parameters and their description
                              should be shown.
     :param console_printer:  Object to print messages on the console.
+    :param args:             Args passed to coala command.
     """
     if not bears:
         console_printer.print('No bears to show. Did you forget to install '
@@ -929,20 +941,41 @@ def print_bears(bears,
                               'coala-bears`.')
         return
 
-    for bear, sections in sorted(bears.items(),
-                                 key=lambda bear_tuple:
-                                 bear_tuple[0].name.lower()):
-        show_bear(bear,
-                  show_description,
-                  show_params,
-                  console_printer)
+    results = [bear for bear, _ in sorted(bears.items(),
+                                          key=lambda bear_tuple:
+                                          bear_tuple[0].name.lower())]
+    if args and args.json:
+        from coalib.output.JSONEncoder import create_json_encoder
+        JSONEncoder = create_json_encoder(use_relpath=args.relpath)
+        json_output = {'bears': results}
+        import json
+        json_formatted_output = json.dumps(json_output,
+                                           cls=JSONEncoder,
+                                           sort_keys=True,
+                                           indent=2,
+                                           separators=(',', ': '))
+        if args.output:
+            filename = args.output[0]
+            with open(filename, 'w+') as fp:
+                fp.write(json_formatted_output)
+        else:
+            print(json_formatted_output)
+    elif args and args.format:
+        print_bears_formatted(results)
+    else:
+        for bear in results:
+            show_bear(bear,
+                      show_description,
+                      show_params,
+                      console_printer)
 
 
 def show_bears(local_bears,
                global_bears,
                show_description,
                show_params,
-               console_printer):
+               console_printer,
+               args=None):
     """
     Extracts all the bears from each enabled section or the sections in the
     targets and passes a dictionary to the show_bears_callback method.
@@ -956,10 +989,11 @@ def show_bears(local_bears,
     :param show_params:      True if the parameters and their description
                              should be shown.
     :param console_printer:  Object to print messages on the console.
+    :param args:             Args passed to coala command.
     """
     bears = inverse_dicts(local_bears, global_bears)
 
-    print_bears(bears, show_description, show_params, console_printer)
+    print_bears(bears, show_description, show_params, console_printer, args)
 
 
 def show_language_bears_capabilities(language_bears_capabilities,
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -956,7 +956,7 @@ def print_bears(bears,
                                            separators=(',', ': '))
         if args.output:
             filename = args.output[0]
-            with open(filename, 'w+') as fp:
+            with open(filename, 'w') as fp:
                 fp.write(json_formatted_output)
         else:
             print(json_formatted_output)
--- a/tests/coalaJSONTest.py
+++ b/tests/coalaJSONTest.py
@@ -81,7 +81,8 @@ class coalaJSONTest(unittest.TestCase):
     def test_show_language_bears(self):
         with bear_test_module():
             retval, stdout, stderr = execute_coala(
-                coala.main, 'coala', '--json', '-B', '-l', 'java', '-I')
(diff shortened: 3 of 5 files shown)


shreyans800755:
@gitmate-bot rebase

shreyans800755:
@Makman2 please review this PR.

coalib/coala_modes.py
```diff
@@ -46,34 +46,17 @@ def mode_json(args, debug=False):
...
-                results.append(bear)
-        except BaseException as exception:  # pylint: disable=broad-except
-            return get_exitcode(exception)
-    else:
-        results, exitcode, _ = run_coala(args=args, debug=debug)
 
-    retval = {'bears': results} if args.show_bears else {'results': results}
+    results, exitcode, _ = run_coala(args=args, debug=debug)
+
+    retval = {'results': results}
```
Makman2: do we change json output with that? If we have `show-bears` enabled, it's put into the `results` field in JSON, not into `bears` any more.

shreyans800755: No. If `show-bears` is specified along with json mode, execution flow won't come to this file. Have a look at https://github.com/coala/coala/pull/4726/files/fa62b7ba0d032b0d65a8c25c57f121b738a04202#diff-aed82cdd504caf613603b250c092f460L56. I've shifted this condition below `show-bears` condition in `coala.py` file.

coalib/output/ConsoleInteraction.py
```diff
@@ -407,6 +407,16 @@ def print_results_formatted(log_printer,
...
 
 
+def print_bears_formatted(bears, format=None):
+    format_str = format or ('name:{name}:can_detect:{can_detect}:'
+                            'can_fix:{can_fix}:description:{description}')
+    print('\n\n'.join(format_str.format(name=bear.name,
+                                        can_detect=bear.CAN_DETECT,
+                                        can_fix=bear.CAN_FIX,
+                                        description=bear.get_metadata().desc)
+                      for bear in bears))
```
Makman2: is it possible to have 2 commits, one for implementing `filter-by` for `format` and one for `json`?

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,58 @@ def print_bears(bears,
...
+                              cls=JSONEncoder,
+                              sort_keys=True,
+                              indent=2,
+                              separators=(',', ': '))
+            else:
+                print(json.dumps(json_output,
+                                 cls=JSONEncoder,
+                                 sort_keys=True,
+                                 indent=2,
+                                 separators=(',', ': ')))
```
Makman2: can't you use `json.dumps` once and just write it to a file? This deduplicates the call ^^

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,58 @@ def print_bears(bears,
...
     if not bears:
         console_printer.print('No bears to show. Did you forget to install '
                               'the `coala-bears` package? Try `pip3 install '
                               'coala-bears`.')
         return
 
+    results = []
+    if args and args.json:
+        from coalib.output.JSONEncoder import create_json_encoder
+        JSONEncoder = create_json_encoder(use_relpath=args.relpath)
```
Makman2: can't you create the JSONEncoder below? Then you need one if less

shreyans800755:
@Makman2 Pushed the changes you requested. I can split the commit if its really required.

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,58 @@ def print_bears(bears,
...
+            show_bear(bear,
+                      show_description,
+                      show_params,
+                      console_printer)
+
+    if args:
+        if args.json:
+            json_output = {'bears': results}
+            import json
+            if args.output:
```
Makman2: what if output is provided for `format`?

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,58 @@ def print_bears(bears,
...
+                              indent=2,
+                              separators=(',', ': '))
+            else:
+                print(json.dumps(json_output,
+                                 cls=JSONEncoder,
+                                 sort_keys=True,
+                                 indent=2,
+                                 separators=(',', ': ')))
+        elif args.format:
+            print_bears_formatted(results)
```
Makman2: maybe put here
```
for bear in results:
    show_bear(...)
```
from above, to make this more readable in code. So you have the bear-gathering phase and the printing phase. However you need return statements then for each case from above, e.g.:
```
print_bears_formatted(...)
return
```

shreyans800755: Yeah makes sense. I'm kinda checking `args and (args.json or args.format)` condition for all the bears individually, which is unnecessary.

coalib/output/ConsoleInteraction.py
shreyans800755: As per help section,
```
-o, --output          write results to the given file (must be called with
                        --json)
```

IMO, if we want this option to be used with other entrypoints in the similar context, we should file different issue.

coalib/output/ConsoleInteraction.py
shreyans800755: I can't see scenario of error over here. So, by default coala.py returns 0 after calling show_bears function from ConsoleInteraction module. I haven't changed that part.

Makman2: > I can't see scenario of error over here. So, by default coala.py returns 0 after calling show_bears function from ConsoleInteraction module. I haven't changed that part.

What do you mean?

shreyans800755: https://github.com/coala/coala/pull/4726/commits/f039489890654bc7396f7a482c1c62fcc96bf571#diff-aed82cdd504caf613603b250c092f460L88 returns success - zero after calling `show_bears` function with appropriate entry point through `args` param. So, we don't need to return any error code from `show_bears` function.

shreyans800755:
@gitmate-bot rebase

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,49 @@ def print_bears(bears,
...
+        JSONEncoder = create_json_encoder(use_relpath=args.relpath)
+        json_output = {'bears': results}
+        import json
+        json_formatted_output = json.dumps(json_output,
+                                           cls=JSONEncoder,
+                                           sort_keys=True,
+                                           indent=2,
+                                           separators=(',', ': '))
+        if args.output:
+            filename = str(args.output[0])
```
Makman2: `args.output[0]`? Why not `args.output`?

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,49 @@ def print_bears(bears,
...
+            with open(filename, 'w+') as fp:
```
Makman2: `w+`, so write and append? That would corrupt existing json files, the parent element has to be either an item or an array in json syntax.

coalib/output/ConsoleInteraction.py
```diff
@@ -922,27 +933,49 @@ def print_bears(bears,
...
+                fp.write(json_formatted_output)
+        else:
+            print(json_formatted_output)
+    elif args and args.format:
+        print_bears_formatted(results)
```
Makman2: hmm so we don't support `args.output`... I won't change this commit that much it's still very good, however we have to streamline the functionality of different output formats and arguments like `args.output`.

coalib/output/ConsoleInteraction.py
shreyans800755: `args.output` is list of file names given as args. It is used in similar manner in other places (https://github.com/coala/coala/pull/4726/commits/a581d099b1d8b8dab3a910f3ef2e2a974e67338f#diff-a297897b536e74044ca36e47697b169dL83) as well.

coalib/output/ConsoleInteraction.py
shreyans800755: It is same as https://github.com/coala/coala/pull/4726/commits/a581d099b1d8b8dab3a910f3ef2e2a974e67338f#diff-a297897b536e74044ca36e47697b169dL84. @tulphoon has made last changes there. Should I add new commit to correct it at both the locations ?

coalib/output/ConsoleInteraction.py
shreyans800755: @Makman2

Makman2: Shouldn't we then iterate over each file and write to each of them??

coalib/output/ConsoleInteraction.py
Makman2: yeah sounds good :+1:

coalib/output/ConsoleInteraction.py
shreyans800755: True. But do we really need an option like `output` ?
If user is using cli, there's always an option to use Linux i/o redirectives. It's actually easy to remember and shorter to write.

Makman2: though we should still support it for now (with multiple files) and discussing later the usefulness of `--output`

Makman2: Ah and you should adapt tests to use multiple output files to test that properly ;)

shreyans800755: Ah! Only one argument is supported after `--output` using the parameter `nargs=1` while building parser. But it is stored in a list by default irrespective of number of args.
Do we still want to implement this for multiple files ?

shreyans800755:
@Makman2 
Any idea why appveyor is failing ?

coalib/output/ConsoleInteraction.py
Makman2: Ah because of `nargs`... I think if someone omits `nargs=1` in the parameter definition altogether, it requires by default a single value and doesn't use an array to store it. However, this is not job of this PR to change that, so keep it ;)

Makman2: Maybe add a comment then? Not sure...

Makman2:
Ugh damn upstream break... Got this error on another PR too...

Makman2:
1. see suggestion of gitmate ;)
2. you should maybe again use `coala` as a commit tag
3. Add the reason why you are now truncating the files before instead of appending, I think this is important (especially without an issue)

coalib/output/ConsoleInteraction.py
Makman2: Isn't `args.output[0]` already a string? Then you don't need to invoke `str`

tests/coalaJSONTest.py
```diff
@@ -150,3 +151,17 @@ def test_output_file(self):
...
+                coala.main, 'coala', '--json', '-B', '--filter-by', 'language',
+                'java', '-I', '--output', 'bears.json')
+
+        with open('bears.json') as fp:
+            data = json.load(fp)
+
+        self.assertEqual(retval, 0)
+        self.assertEqual(len(data['bears']), 2)
+        self.assertFalse(stderr)
+        os.remove('bears.json')
```
Makman2: remove the file immediately once you've loaded contents, then the file doesn't remain on test fails (which would cause then test fails, because fo the append mode!)

tests/coalaJSONTest.py
```diff
@@ -150,7 +151,6 @@ def test_output_file(self):
         self.assertFalse(retval2)
         self.assertFalse(stdout2)
         self.assertFalse(stderr2)
-        os.remove('file.json')
```
Makman2: this change should go into a separate commit

tests/coalaJSONTest.py
```diff
@@ -160,8 +160,8 @@ def test_show_language_bears_output_file(self):
 
         with open('bears.json') as fp:
             data = json.load(fp)
+        os.remove('bears.json')
 
         self.assertEqual(retval, 0)
         self.assertEqual(len(data['bears']), 2)
         self.assertFalse(stderr)
-        os.remove('bears.json')
```
Makman2: and this change should be in the previous commit introducing the test

coalib/coala_modes.py
```diff
@@ -64,7 +64,7 @@ def mode_json(args, debug=False):
 
     if args.output:
         filename = str(args.output[0])
-        with open(filename, 'w+') as fp:
+        with open(filename, 'w') as fp:
```
Makman2: actually, do we wanna test that? Simply write to the same file twice, and check if content was replaced and not appended^^

shreyans800755: IMO, it doesn't make sense to add tests for this as its python internals. Just for safety, I've tested it locally and seems to be working fine.

Makman2: >  it doesn't make sense to add tests for this as its python internals.

It definitely reflects on the user. That's why we should test it.
However, let's open an issue for that afterwards, imo not so important right now as it's currently not officially standardized I believe whether we append or overwrite output files.

Makman2:
For the commit `Delete output file immediately after loading data` I definitely want to see the tag `coalaJSONTest`. So `coalaJSONTest: Delete output file immediately` (imo the context of tests is important here)

Makman2:
ack 8492fe9 50d9e32 3458c29

Makman2:
@rultor merge

PR #4780, opened 2017-10-23
core: Explicitly shutdown executor

Makman2:
The event-loop can't know that is has to close the executor, as it's
always passed explicitly to `run_in_executor()`.

### DIFF
--- a/tests/core/CoreTest.py
+++ b/tests/core/CoreTest.py
@@ -619,3 +619,23 @@ class CoreOnThreadPoolExecutorTest(CoreTest):
     def setUp(self):
         super().setUp()
         self.executor = ThreadPoolExecutor, tuple(), dict(max_workers=8)
+
+
+# This test class only runs test cases once, as the tests here rely on specific
+# executors / having the control over executors.
+class CoreOnSpecificExecutorTest(CoreTestBase):
+    def test_custom_executor_closed_after_run(self):
+        bear = MultiTaskBear(Section('test-section'),
+                             {'some-file': []},
+                             tasks_count=1)
+
+        # The executor should be closed regardless how many bears are passed.
+        for bears in [set(), {bear}]:
+            executor = ThreadPoolExecutor(max_workers=1)
+
+            self.execute_run(bears, executor)
+
+            # Submitting new tasks should raise an exception now.
+            with self.assertRaisesRegex(
+                    RuntimeError, 'cannot schedule new futures after shutdown'):
+                executor.submit(lambda: None)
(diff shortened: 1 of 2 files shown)


tests/core/CoreTest.py
```diff
@@ -619,3 +619,23 @@ class CoreOnThreadPoolExecutorTest(CoreTest):
...
+
+        # The executor should be closed regardless how many bears are passed.
+        for bears in [set(), {bear}]:
+            executor = ThreadPoolExecutor(max_workers=1)
+
+            self.execute_run(bears, executor)
+
+            # Submitting new tasks should raise an exception now.
+            with self.assertRaisesRegex(
+                    RuntimeError, 'cannot schedule new futures after shutdown'):
```
shreyans800755: Shouldn't we throw (catch here as well) more specific exception if a job is submitted to shut down executor ?

Makman2: Can't be changed, the error is coming from the `ThreadPoolExecutor` itself^^

shreyans800755: Oh :disappointed:

shreyans800755:
@Makman2 
Is there a possibility that given custom executor might be shut down twice? Once over here and once by the code passing it. Because code passing it might also have ownership to the same object.
If so, can it give RuntimeError or any other exception ? Can we have a test case covering that part ?

Makman2:
> Is there a possibility that given custom executor might be shut down twice? Once over here and once by the code passing it.

Yes, you can do
```
my_executor = SomeExecutor(...)
core.run(..., my_executor)
my_executor.shutdown()
```
But that won't raise an exception. Only if you submit tasks. Anyway, it's documented (in this commit) that the executor will be shut down once `core.run` is finished, to prevent stale executors consuming resources.

> Because code passing it might also have ownership to the same object.

There's no concept in Python regarding "ownership". Everybody with an object reference can use it as it likes.
But conceptually, if somebody creates the executor, the executor belongs to him and he has to do additional tasks to handle it properly if necessary (except shutting it down, which is done by the core).

But I think I know where your confusion comes from:
> ... if a job is submitted to shut down executor ?

The executor shutdown happens in a done-callback of a job. If a job finished (like an analysis on a file), it's called and the execution state is checked (how many bears left to process, resolve dependencies, run new jobs, etc etc). If the state-checking sees that there's nothing more to do, it closes the event loop first. This causes the main thread to continue after `event_loop.run_forever()`. However, the executor is still running as we feed it into the event loop externally. Inside `core.run`, we are shutting it down too directly after that. No exceptions are generated from the core. Only if the user tries to reuse his custom executor after `core.run` - using `submit` or similar to create new jobs on it (unrelated from the core) - he will get an error. But it will be always the user's fault (assuming the core has no weird bugs ;D).

shreyans800755:
ack 3df2610

Makman2:
@gitmate-bot rebase

shreyans800755:
@gitmate-bot rebase

Makman2:
@rultor merge

Issue #4505, opened 2017-07-21
NextGen-Core: Do not close event loop if not all bears were scheduled


Makman2:
There is a corner case when multiple bear dependencies are passed and one of them schedules no tasks.
If the bear scheduling no tasks is processed first, `cleanup_bear` is called that will close the event loop as no other running tasks are registered (yet), although the other bear has still to be scheduled.

This needs to be handled correctly and tests need to be written for this case.

PR #4819, opened 2017-10-31
core: Do not close event loop too early

Makman2:
... when scheduling all dependency bears. Bears that offload no tasks
may cause the event-loop to close too early while other bears had no
chance to submit their tasks.

Fixes https://github.com/coala/coala/issues/4505

### DIFF
--- a/coalib/core/Core.py
+++ b/coalib/core/Core.py
@@ -230,19 +230,21 @@ class Session:
 
                 self.running_tasks[bear] = tasks
 
+                # Cleanup bears without tasks after all bears had the chance to
+                # schedule their tasks. Not doing so might stop the run too
+                # early, as the cleanup is also responsible for stopping the
+                # event-loop when no more tasks do exist.
+                if not tasks:
+                    logging.debug('{!r} scheduled no tasks.'.format(bear))
+                    bears_without_tasks.append(bear)
+                    continue
+
                 for task in tasks:
                     task.add_done_callback(functools.partial(
                         self._finish_task, bear))
 
                 logging.debug('Scheduled {!r} (tasks: {})'.format(bear,
                                                                   len(tasks)))
-
-                # Cleanup bears without tasks after all bears had the chance to
-                # schedule their tasks. Not doing so might stop the run too
-                # early, as the cleanup is also responsible for stopping the
-                # event-loop when no more tasks do exist.
-                if not tasks:
-                    bears_without_tasks.append(bear)
 
         for bear in bears_without_tasks:
             self._cleanup_bear(bear)
--- a/coalib/core/Core.py
+++ b/coalib/core/Core.py
@@ -211,6 +211,8 @@ class Session:
         :param bears:
             A list of bear instances to be scheduled onto the process pool.
         """
+        bears_without_tasks = []
+
         for bear in bears:
             if self.dependency_tracker.get_dependencies(
                     bear):  # pragma: no cover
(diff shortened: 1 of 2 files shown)


coalib/core/Core.py
```diff
@@ -235,11 +237,18 @@ def _schedule_bears(self, bears):
...
+                # Cleanup bears without tasks after all bears had the chance to
+                # schedule their tasks. Not doing so might stop the run too
+                # early, as the cleanup is also responsible for stopping the
+                # event-loop when no more tasks do exist.
                 if not tasks:
-                    # We need to recheck our runtime if something is left to
-                    # process, as when no tasks were offloaded the event-loop
-                    # could hang up otherwise.
-                    self._cleanup_bear(bear)
+                    bears_without_tasks_to_cleanup.append(bear)
```
shreyans800755: Can we keep this condition before looping over the tasks ?

Something like:
```
if not tasks:
    bears_without_tasks_to_cleanup.append(bear)
    continue
```

Might not save too many clock cycles, but it will make code cleaner.

Makman2: Yeah why not, if you like I would add then an additional debug-message `BearXYZ scheduled no tasks.`, so we still keep the information that a bear has been processed :+1:

Makman2: Hm just thinking if this should be done in a separate commit... This commit diff is smaller when not shifting the logic away^^

shreyans800755: IMO this can be done in the same commit as we are adding this condition in this commit. It doesn't make sense to add something in first commit and change the same in subsequent one especially in same PR.

shreyans800755: And there won't be a case where we only want to rollback one of the commit and not the other one. So, its better to keep single commit.

Makman2: > as we are adding this condition in this commit

which condition? The `if` is already there (`if not tasks`)

shreyans800755: Oh my bad. Can be done in different commit.

coalib/core/Core.py
```diff
@@ -211,6 +211,8 @@ def _schedule_bears(self, bears):
         :param bears:
             A list of bear instances to be scheduled onto the process pool.
         """
+        bears_without_tasks_to_cleanup = []
```
shreyans800755: `bears_without_tasks_to_cleanup` -> `bears_without_tasks`

coalib/core/Core.py
```diff
@@ -235,11 +237,18 @@ def _schedule_bears(self, bears):
...
+                # event-loop when no more tasks do exist.
                 if not tasks:
-                    # We need to recheck our runtime if something is left to
-                    # process, as when no tasks were offloaded the event-loop
-                    # could hang up otherwise.
-                    self._cleanup_bear(bear)
+                    bears_without_tasks_to_cleanup.append(bear)
+
+        for bear in bears_without_tasks_to_cleanup:
+            # We need to recheck our runtime if something is left to process, as
```
shreyans800755: Do we still need this comment ?

tests/core/CoreTest.py
```diff
@@ -595,6 +602,33 @@ def test_run_generate_tasks_dynamically_with_dependency_results(self):
         self.assertEqual(
             len(bear.dependency_results[BearA]), 1)
 
+    def test_run_many_dependency_bears_with_zero_tasks(self):
```
shreyans800755: `test_run_many_dependency_bears_with_zero_tasks` -> `test_run_multiple_dependency_bears_with_zero_tasks`

tests/core/CoreTest.py
```diff
@@ -120,6 +120,13 @@ def generate_tasks(self):
         return (((i,), {}) for i in range(tasks_count))
 
 
+class DependentOnManyZeroTaskBearsTestBear(TestBearBase):
```
shreyans800755: `DependentOnManyZeroTaskBearsTestBear` -> `DependentOnMultipleZeroTaskBearsTestBear`

coalib/core/Core.py
Makman2: hm yeah... I think this is more or less self explaining :+1:

tests/core/CoreTest.py
Makman2: Does it matter? :D

shreyans800755: Just trying to be more accurate with words. :smiley:

Makman2: Kk ;)

shreyans800755:
ack aa720e2

shreyans800755:
ack a7b8e0b

Makman2:
@gitmate-bot rebase

Makman2:
@rultor merge

Issue #4651, opened 2017-08-19
Add CLI to print only selected bears given in args


co-robo:
Opened by @shreyans800755 at [gitter/coala/coala](https://gitter.im/coala/coala?at=5998c94ca7b406262da83962)

shreyans800755:
CC @Makman2 @meetmangukiya

mikanaji:
Can i get this issue?

Makman2:
The issue is currently assigned, please pick another one :)
It's also not labeled right now.

shreyans800755:
@mikanaji I'm waiting for https://github.com/coala/coala/pull/4726 to be completed first. Technically, this issue doesn't require that PR to be merged. But, its good to have. Otherwise, I'll have to resolve a lot of conflicts unnecessarily. Once that PR is merged, which should be done pretty soon, I'll push my changes to this PR.

mikanaji:
@shreyans800755, Thanks!

shreyans800755:
@mykromo Please go ahead with some issue you may find interesting.

mykromo:
Alright. Thanks!

shreyans800755:
@mykromo I've created a PR. Feel free to review.

PR #4832, opened 2017-11-04
coala: Option to print only selected bears

shreyans800755:
It adds option to print only selected bears using `--show-bears -bears`.

Closes https://github.com/coala/coala/issues/4651

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -54,8 +54,7 @@ def main(debug=False):
         configure_logging(not args.no_color)
 
         if args.show_bears:
-            from coalib.settings.ConfigurationGathering import (
-                get_all_bears)
+            from coalib.settings.ConfigurationGathering import get_all_bears
             kwargs = {}
             if args.bears:
                 kwargs['bear_globs'] = args.bears
--- a/coalib/coala.py
+++ b/coalib/coala.py
@@ -56,7 +56,10 @@ def main(debug=False):
         if args.show_bears:
             from coalib.settings.ConfigurationGathering import (
                 get_all_bears)
-            filtered_bears = get_all_bears()
+            kwargs = {}
+            if args.bears:
+                kwargs['bear_globs'] = args.bears
+            filtered_bears = get_all_bears(**kwargs)
             if args.filter_by_language:
                 logging.warning(
                     "'--filter-by-language ...' is deprecated. "
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -289,12 +289,15 @@ def get_all_bears_names():
     return [bear.name for bear in get_all_bears()]
 
 
-def collect_all_bears_from_sections(sections, log_printer=None):
+def collect_all_bears_from_sections(sections,
+                                    log_printer=None,
+                                    bear_globs=('**',)):
     """
     Collect all kinds of bears from bear directories given in the sections.
 
     :param sections:    List of sections so bear_dirs are taken into account
     :param log_printer: Log_printer to handle logging
+    :param bear_globs:  List of glob patterns.
     :return:            Tuple of dictionaries of local and global bears.
                         The dictionary key is section class and
                         dictionary value is a list of Bear classes
@@ -305,7 +308,7 @@ def collect_all_bears_from_sections(sections, log_printer=None):
         bear_dirs = sections[section].bear_dirs()
         local_bears[section], global_bears[section] = collect_bears(
             bear_dirs,
-            ['**'],
+            bear_globs,
             [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
             warn_if_unused_glob=False)
     return local_bears, global_bears
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -397,11 +397,15 @@ def get_config_directory(section):
     return config if os.path.isdir(config) else os.path.dirname(config)
 
 
-def get_all_bears(log_printer=None, arg_parser=None, silent=True):
+def get_all_bears(log_printer=None,
+                  arg_parser=None,
+                  silent=True,
+                  bear_globs=('**',)):
     """
     :param log_printer: The log_printer to handle logging.
     :param arg_parser:  An ``ArgParser`` object.
     :param silent:      Whether or not to display warnings.
+    :param bear_globs:  List of glob patterns.
     :return:            Tuple containing dictionaries of local bears
                         and global bears.
     """
@@ -409,7 +413,7 @@ def get_all_bears(log_printer=None, arg_parser=None, silent=True):
                                      arg_parser=arg_parser,
                                      silent=silent)
     local_bears, global_bears = collect_all_bears_from_sections(
-        sections)
+        sections, bear_globs=bear_globs)
     return local_bears, global_bears
 
 
--- a/tests/output/ConsoleInteractionTest.py
+++ b/tests/output/ConsoleInteractionTest.py
@@ -11,6 +11,7 @@ from testfixtures import LogCapture, StringComparison
 
 from coalib.bearlib.spacing.SpacingHelper import SpacingHelper
 from coalib.bears.Bear import Bear
+from coalib import coala
 from coala_utils.ContextManagers import (
     make_temp, retrieve_stdout, simulate_console_inputs)
 from coalib.output.ConsoleInteraction import (
@@ -41,6 +42,8 @@ from pygments.filters import VisibleWhitespaceFilter
 from pygments.lexers import TextLexer
 from pygments.style import Style
 from pygments.token import Token
+
+from tests.TestUtilities import bear_test_module, execute_coala
 
 
 STR_GET_VAL_FOR_SETTING = ('Please enter a value for the setting \"{}\" ({}) '
@@ -1082,3 +1085,20 @@ class PrintFormattedResultsTest(unittest.TestCase):
                 {abspath('file'): ('def fun():\n', '    pass  \n')})
             self.assertEqual(stdout.getvalue(),
                              "('    pass  \\n',)\n")
+
+    def test_show_bears_specified_in_args(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--bears',
+                'JavaTestBear', '--no-color')
+            self.assertEqual(retval, 0)
+            self.assertEqual(stdout.strip(), 'JavaTestBear')
+
+    def test_show_bears_specified_in_args_regex(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--bears',
+                '*Java*', '*Space*', '--no-color')
+            self.assertEqual(retval, 0)
+            self.assertEqual(['JavaTestBear', 'SpaceConsistencyTestBear'],
+                             [bear.strip() for bear in stdout.splitlines()])


shreyans800755:
I'll add the tests soon.

coalib/coala.py
```diff
@@ -55,8 +56,22 @@ def main(debug=False):
...
+                sections, _ = load_configuration(arg_list=None)
+                local_bears = {}
+                global_bears = {}
+                for (key, value) in sections:
+                    bear_dirs = value.bear_dirs()
+                    local_bears[key], global_bears[key] = collect_bears(
+                        bear_dirs,
+                        args.bears,
+                        [BEAR_KIND.LOCAL, BEAR_KIND.GLOBAL],
+                        warn_if_unused_glob=False)
```
mrtes: Take a look at [collect_all_bears_from_sections](https://github.com/coala/coala/blob/master/coalib/collecting/Collectors.py#L292-L311)

Basically you are just duplicating this code.
I think it would be more beneficial to refactor that function (or some other, related function) to take a bear-glob parameter. This would also help to keep code related to bear collecting in the Collectors.py module.

mrtes:
Two things:

1.  Maybe rename `collect_all_bears_from_sections` to `collect_bears_from_sections` or something. That would clarify the change.

2. You could potentially move the remaining code in the `args.bears` branch to a new function in `ConfigurationGathering.py`.  Or maybe refactor the `get_all_bears` function, to take the glob ;)

I dont know. What do you think about that?

shreyans800755:
@mrtes Point 2 sounds good.

mrtes:
Alright :)
Looks good to me
ack a91bb02

Makman2:
Coala is always written with a lowercase c ;D

Makman2:
Also some verb in the commit message would be good, like "Add option to ..."
You could in fact leave the tag, as every commit is affecting coala itself ;)

Makman2:
unack a91bb02

coalib/coala.py
```diff
@@ -54,9 +54,11 @@ def main(debug=False):
         configure_logging(not args.no_color)
 
         if args.show_bears:
-            from coalib.settings.ConfigurationGathering import (
-                get_all_bears)
-            filtered_bears = get_all_bears()
+            from coalib.settings.ConfigurationGathering import get_all_bears
```
Makman2: unnecessary change. If it fits into a single line, please do a separate commit for it :+1:
(which is defnitely reasonable, having this in a single line reads way better)

Makman2: (I mean just the import line, not the `filtered_bears` thing :D)

coalib/coala.py
```diff
@@ -54,9 +54,11 @@ def main(debug=False):
...
+            if args.bears:
+                filtered_bears = get_all_bears(bear_globs=args.bears)
+            else:
+                filtered_bears = get_all_bears()
```
Makman2: -->
```
filtered_bears = (get_all_bears(bear_globs=args.bears)
                  if args.bears else
                  get_all_bears())
```

Makman2: or even
```
kwargs = {}
if args.bears:
    kwargs['bear_globs'] = args.bears
filtered_bears = get_all_bears(**kwargs)
```

coalib/collecting/Collectors.py
```diff
@@ -289,12 +289,15 @@ def get_all_bears_names():
     return [bear.name for bear in get_all_bears()]
 
 
-def collect_all_bears_from_sections(sections, log_printer=None):
+def collect_all_bears_from_sections(sections,
+                                    log_printer=None,
+                                    bear_globs=['**']):
```
Makman2: if possible, try to use a tuple so the default value is immutable :+1:

coalib/settings/ConfigurationGathering.py
```diff
@@ -397,19 +397,23 @@ def get_config_directory(section):
     return config if os.path.isdir(config) else os.path.dirname(config)
 
 
-def get_all_bears(log_printer=None, arg_parser=None, silent=True):
+def get_all_bears(log_printer=None,
+                  arg_parser=None,
+                  silent=True,
+                  bear_globs=['**']):
```
Makman2: also here, if possible use a tuple :)

shreyans800755:
> Also some verb in the commit message would be good, like "Add option to ..."

This commit doesn't add the option. Its just using existing option.

> You could in fact leave the tag, as every commit is affecting coala itself ;)

Tag is for the file coala.py and not in the context of whole project.

neelanshsahai:
unack https://github.com/coala/coala/pull/4832/commits/86de9cc9dd3c644d880d40e399cf7049156bd521

shreyans800755:
@neelanshsahai 
Please look at https://github.com/coala/coala/pull/4832#discussion_r149208761.

I've deliberately kept 2 commits as 2nd one of not part of this issue.
And it's already rebased.

neelanshsahai:
:worried: 
Yupp!!
missed that point. Very Well.

Makman2:
@neelanshsahai btw: manual rebase is usually not necessary any more, as we have @gitmate-bot to do that. It's only necessary if there are conflicts which have to be resolved by hand.

Makman2:
Nice I've found a bug in gitmate-bot...

shreyans800755:
Why are manual reviews shown as failed even after pushing new changes or rebasing ?
Ideally it should be based on commit SHA. And its status must be reset after replacing old commit with the new one.

Makman2:
@shreyans800755 gitmate-bot is smart enough to detect code changes. If nothing has changed, it sees it's the same commit even though the SHA is different. Because it was unacked before, it stays unacked. Feel free to discuss that behaviour at [code.gitmate.io](code.gitmate.io), not sure what's better myself :+1:

shreyans800755:
@Makman2 
Not just automated rebased. I've tried to push changes by amending the commit as well before couple of days. But, still it failed both the manual reviews as older equivalent commits were unacked earlier.
I'll discuss this on gitmate.

coalib/settings/ConfigurationGathering.py
```diff
@@ -397,19 +397,23 @@ def get_config_directory(section):
     return config if os.path.isdir(config) else os.path.dirname(config)
 
 
-def get_all_bears(log_printer=None, arg_parser=None, silent=True):
+def get_all_bears(log_printer=None,
+                  arg_parser=None,
+                  silent=True,
+                  bear_globs=tuple('**')):
```
Makman2: this is not what's intended, I believe you thought this creates a single-element tuple with `**` in it, right? ;)
However, `tuple()` collects any iterabel and puts it into a tuple. `string` is an iterable, it allows you to iterate over each char. So in this case you'll get:
`('*', '*')`

It has to be `('**',)` (note the comma)

Makman2: (and without `tuple`, so just `bear_globs=('**',)`)

coalib/collecting/Collectors.py
```diff
@@ -289,12 +289,15 @@ def get_all_bears_names():
...
-def collect_all_bears_from_sections(sections, log_printer=None):
+def collect_all_bears_from_sections(sections,
+                                    log_printer=None,
+                                    bear_globs=tuple('**')):
     """
     Collect all kinds of bears from bear directories given in the sections.
 
     :param sections:    List of sections so bear_dirs are taken into account
     :param log_printer: Log_printer to handle logging
+    :param bear_globs:  List of patterns/bear names to match bears
```
Makman2: missing period at end.
\+
What kind of patterns? Mention that these are globs :+1: So just `List of glob patterns ...`

coalib/collecting/Collectors.py
```diff
@@ -289,12 +289,15 @@ def get_all_bears_names():
     return [bear.name for bear in get_all_bears()]
 
 
-def collect_all_bears_from_sections(sections, log_printer=None):
+def collect_all_bears_from_sections(sections,
+                                    log_printer=None,
+                                    bear_globs=tuple('**')):
```
Makman2: same here, see above

coalib/settings/ConfigurationGathering.py
```diff
@@ -397,19 +397,23 @@ def get_config_directory(section):
...
     """
     :param log_printer: The log_printer to handle logging.
     :param arg_parser:  An ``ArgParser`` object.
     :param silent:      Whether or not to display warnings.
+    :param bear_globs:  List of patterns/bear names to match bears
```
Makman2: see above

Makman2:
ack 21d285c

Makman2:
unack 4e8221e

Makman2:
Could you maybe write a test that fails when using a `*` as a glob pattern for the default behaviour? So we catch your erroneous default case :+1:

shreyans800755:
> Could you maybe write a test that fails when using a * as a glob pattern for the default behaviour?

Not failing. I replaced default value of `bear_globs` and still tests are passing. Not sure whether to call it bug in glob matching or a feature :smile:

shreyans800755:
One more problem:

Following commands are not giving any bears:
`coala -B --bears *`
`coala -B --bears **`

But,
`coala -B --bears *Clang*`
is giving list of bears having word `Clang` in it.

shreyans800755:
@gitmate-bot rebase

shreyans800755:
@gitmate-bot rebase

Makman2:
> Got the problem. Actually, in first 2 cases, value of args.bears is expanded to list of files in current dir. So, it is not matching in any bears.

Yeah wildcard expansion happens in shell.

> Not failing. I replaced default value of bear_globs and still tests are passing. Not sure whether to call it bug in glob matching or a feature 😄

Wheee that's bad... that's kind of unexpected I believe 😅 However in that case let's file another issue for it to clarify up on that :+1:

tests/output/ConsoleInteractionTest.py
```diff
@@ -1082,3 +1082,25 @@ def test_source_lines(self):
                 {abspath('file'): ('def fun():\n', '    pass  \n')})
             self.assertEqual(stdout.getvalue(),
                              "('    pass  \\n',)\n")
+
+    def test_show_bears_specified_in_args(self):
+        from coalib import coala
+        from tests.TestUtilities import bear_test_module, execute_coala
```
Makman2: why not importing those at top? :O

tests/output/ConsoleInteractionTest.py
```diff
@@ -1082,3 +1082,25 @@ def test_source_lines(self):
...
+    def test_show_bears_specified_in_args_regex(self):
+        from coalib import coala
+        from tests.TestUtilities import bear_test_module, execute_coala
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--bears',
+                '*Java*', '*Space*')
+            self.assertEqual(retval, 0)
+            # 2 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 3)
```
Makman2: I would like to see if we test for the output strings, to see whether the correct bears were matched. If you worry about color, enable `--no-color` in arguments (I hope that's the right flag)

tests/output/ConsoleInteractionTest.py
```diff
@@ -1082,3 +1082,25 @@ def test_source_lines(self):
...
+            # 1 bear plus 1 line holding the closing colour escape sequence.
+            self.assertEqual(len(stdout.strip().splitlines()), 2)
+
+    def test_show_bears_specified_in_args_regex(self):
+        from coalib import coala
+        from tests.TestUtilities import bear_test_module, execute_coala
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--bears',
+                '*Java*', '*Space*')
```
Makman2: If possible (though not necessary now):
Can we make this more reliable? As when we add some bear with that name the test will fail.
Maybe it also helps to add a comment explaining what is likely failing so people not involved here immediately understand what has changed.

shreyans800755:
@Makman2 Filed new issue for glob pattern investigation at https://github.com/coala/coala/issues/4878.

tests/output/ConsoleInteractionTest.py
shreyans800755: I've modified tests added in this PR not to fail in such cases. I'm testing the name of the bear instead of number of bears now.

tests/output/ConsoleInteractionTest.py
```diff
@@ -1082,3 +1086,20 @@ def test_source_lines(self):
...
+            self.assertEqual(retval, 0)
+            self.assertEqual(stdout.strip(), 'JavaTestBear')
+
+    def test_show_bears_specified_in_args_regex(self):
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '-B', '--bears',
+                '*Java*', '*Space*', '--no-color')
+            self.assertEqual(retval, 0)
+            for bear in stdout.strip().splitlines():
```
jayvdb: This should be changed into an exact match, so the order is locked down and cant randomly change.

Probably want to compare an array of names against
`[l.strip() for l in stdout.strip().splitlines()]`

shreyans800755: I don't think its a good idea.
If someone wants to add a bear matching this criteria, then their changes will break this test if we do an exact match. And it can be difficult to understand the context if the person is newcomer. And this is testing what we are actually doing by running coala.

shreyans800755:
@gitmate-bot rebase

shreyans800755:
@gitmate-bot rebase

tests/output/ConsoleInteractionTest.py
jayvdb: Correct.  That is intentional.  Then we can see that new test bears have an impact on the test results.  It should require that the PR adding a test bear adds only one extra line to the output here.  Or they can choose a different name for their test bear.  Not hard.

shreyans800755:
@jayvdb 
I've made requested changes.

> I expect there would be some cli doc changes as part of this PR.

I think its better to put docs changes in coala/documentation. We can have different issue and PR for that.

tests/output/ConsoleInteractionTest.py
```diff
@@ -9,8 +9,10 @@
 
 from testfixtures import LogCapture, StringComparison
 
+
```
Makman2: unnecessary change (and also bad here imo :D)

Makman2:
@shreyans800755 looks good to me (only one style change)
lgtm after that ;)

Makman2:
(Sorry for the late review, was a bit unavailable :3)

shreyans800755:
@Makman2 Made the change and rebased.

Makman2:
ack fa77a8a 8d15735

Issue #4673, opened 2017-08-25
support ignoring whole directories


RJ722:
Currently, `ignore` in `.coafile` takes in glob patterns as input, but it does not ignore whole directory.

For example:
`.coafile` looks like this:

```
[Default]
ignore = tests/foo/test_files/*
```

and there is no need of test_files being a python package because it is not being loaded as a package, but `PythonPackageInitBear` would complain:
```
Project wide:
**** PythonPackageInitBear [Section: PythonPackageInit | Severity: NORMAL] ****
!    ! Directory "tests/foo/test_files" does not contain __init__.py file
```

Even passing `ignore = ignore = tests/foo/test_files/` or `ignore = ignore = tests/foo/test_files` does not works.

RJ722:
CC @Makman2

shreyans800755:
@RJ722 @Makman2 
We can ignore whole directory like this:
```
[Default]
ignore = tests/foo/test_files/**
```

It is used in coala project itself at https://github.com/coala/coala/blob/master/.coafile#L45.

But I think its nice to have it with single * (asterisk) at the end.

Makman2:
Does this also work for the case described above @RJ722 ?

Makman2:
I would say if someone wants to ignore the whole directory, I would also support doing
```
[Default]
ignore = tests/foo/test_files/
# or without slash
ignore = tests/foo/test_files
```

RJ722:
> Does this also work for the case described above @RJ722 ?

Hmm No, not working. :-(

himankpathak:
@RJ722 have you also tested the following changes as stated by Makman2 ?

```
[Default]
ignore = tests/foo/test_files/
# or without slash
ignore = tests/foo/test_files
```
Or should I test them?

RJ722:
Yeah, tried it. Not working.

Makman2:
It was a suggestion for what we should support, it's not working yet @himankpathak

himankpathak:
@Makman2 oh okay, I will keep looking for a workaround.

mrtes:
Well, there is a function `collect-dirs`, which seems to be intended for that purpose.
It is just unused currently. 
I guess we could first collect directories with this function (and ignore directory globs), then collect the files for the remaining directories and ignore the file patterns.

mrtes:
Slightly more complicated than I imagined, but still doable.
Is there a reason we don't use the native "glob" and "fnmatch" modules?

[...]

PR #4842, opened 2017-11-09
Collectors.py: Add support to ignore directories

mrtes:
This change allows users to specify directory globs in the form:
.../test/
.../test
.../test/**
.../test/*

These patterns now ignore the content of the whole directory and all
subdirectories.

Closes https://github.com/coala/coala/issues/4673

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -3,11 +3,13 @@ import logging
 import os
 import pkg_resources
 import itertools
+import re
 
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.collecting.Importers import iimport_objects
 from coala_utils.decorators import yield_once
 from coalib.misc.Exceptions import log_exception
+from coalib.misc.IterUtilities import partition
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
 from coalib.parsing.Globbing import fnmatch, iglob, glob_escape
 
@@ -36,15 +38,17 @@ def _import_bears(file_path, kinds):
 
 
 @yield_once
-def icollect(file_paths, ignored_globs=None, match_cache={}):
+def icollect(file_paths, ignored_globs=None, match_cache={},
+             match_function=fnmatch):
     """
     Evaluate globs in file paths and return all matching files.
 
-    :param file_paths:    File path or list of such that can include globs
-    :param ignored_globs: List of globs to ignore when matching files
-    :param match_cache:   Dictionary to use for caching results
-    :return:              Iterator that yields tuple of path of a matching
-                          file, the glob where it was found
+    :param file_paths:      File path or list of such that can include globs
+    :param ignored_globs:   List of globs to ignore when matching files
+    :param match_cache:     Dictionary to use for caching results
+    :param match_function:  The function to use for glob matching
+    :return:                Iterator that yields tuple of path of a matching
+                            file, the glob where it was found
     """
     if isinstance(file_paths, str):
         file_paths = [file_paths]
@@ -54,8 +58,45 @@ def icollect(file_paths, ignored_globs=None, match_cache={}):
             match_cache[file_path] = list(iglob(file_path))
 
         for match in match_cache[file_path]:
-            if not ignored_globs or not fnmatch(match, ignored_globs):
+            if not ignored_globs or not match_function(match, ignored_globs):
                 yield match, file_path
+
+
+def match_dir_or_file_pattern(path, ignore_patterns=None):
+    """
+    Tries to match the given path with the directory (substring match) or file
+    (enforced full match) patterns.
+
+    :param path:                Valid file path
+    :param ignore_patterns:     List of regex patterns that match a file or a
+                                directory
+    :return:                    True if any of the given pattern match
+    """
+    def escape(pattern):
+        return pattern.replace('\\', '\\\\')
+
+    expanded_ignores = list_glob_results(ignore_patterns)
+
+    file_patterns, dir_patterns = partition(
+        expanded_ignores,
+        os.path.isfile)
+
+    return (
+        any((re.match(escape(pattern), path) for pattern in dir_patterns)) or
+        any((re.fullmatch(escape(pattern), path) for pattern in file_patterns)))
+
+
+def list_glob_results(values=None):
+    """
+    Expands the globs of all given values and concatenates the results.
+
+    :param values:  List of file-globs or files.
+    :return:        List of matched files.
+    """
+    return functools.reduce(
+        lambda seed, value: seed + list(iglob(value)),
+        values if values else (),
+        [])
 
 
 def collect_files(file_paths, log_printer=None, ignored_file_paths=None,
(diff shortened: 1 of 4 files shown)


Asalle:
ack ae9a04cf8907cbc6ef2068bae6885bff728741bc

coalib/parsing/Globbing.py
```diff
@@ -216,8 +217,8 @@ def fnmatch(name, globs):
     """
     globs = (globs,) if isinstance(globs, str) else tuple(globs)
 
-    if len(globs) == 0:
-        return True
+    if not globs:
```
jayvdb: Is this necessary?  It changes the return semantics of this function for edge cases.
best avoided, as it might break coala-quickstart or some other user of these functions which are not being tested by CI.

coalib/collecting/Collectors.py
```diff
@@ -54,11 +57,57 @@ def icollect(file_paths, ignored_globs=None, match_cache={}):
...
+    """
+    Partitions the iterable into two iterables based on the given predicate
+
+    :param predicate:   A function that takes an item of the iterable and
+                        returns a boolean
+    :return:            Two iterators pointing to the original iterable
+    """
+    a, b = itertools.tee(((predicate(item), item) for item in iterable))
+
+    return ([item for pred, item in a if pred],
```
jayvdb: This runs the predicate twice for each item; when done for 1000s of files, that is significant.

coalib/collecting/Collectors.py
```diff
@@ -54,11 +57,57 @@ def icollect(file_paths, ignored_globs=None, match_cache={}):
...
+    :param values:  List of file-globs or files.
+    :return:        List of matched files.
+    """
+    return functools.reduce(
+        lambda seed, value: seed + list(iglob(value)),
+        values,
+        [])
+
+
+def collect_files(file_paths, log_printer=None, ignored_file_paths=(),
```
jayvdb: `ignored_file_paths=()` changes the semantics of the function, which is why you had to change some of the existing tests.

It is generally a bad idea to rely on args default being a empty collection.  The default should be None, and inside the function it does the right thing for any False-ish value.

i.e., inside, add `if not foo: foo = (, )` , as this is an internal matter.

As a general rule, a function should never distinguish between different types of False-ish provided in data given in arguments unless it was part of a very clear contract.

tests/collecting/CollectorsTest.py
```diff
@@ -35,7 +36,6 @@ def test_file_invalid(self):
         with LogCapture() as capture:
             self.assertEqual(collect_files(file_paths=['invalid_path'],
                                            log_printer=self.log_printer,
-                                           ignored_file_paths=None,
```
jayvdb: undo this please.

coalib/collecting/Collectors.py
```diff
@@ -90,7 +147,7 @@ def collect_files(file_paths, log_printer=None, ignored_file_paths=None,
     return limited_files
 
 
-def collect_dirs(dir_paths, ignored_dir_paths=None):
+def collect_dirs(dir_paths, ignored_dir_paths=()):
```
jayvdb: as above; empty list defaults are bad.

jayvdb:
unack https://github.com/coala/coala/commit/ae9a04cf8907cbc6ef2068bae6885bff728741bc

coalib/collecting/Collectors.py
mrtes: I didn't test this, but the predicate should only be evaluated in the generator expression supplied to `itertools.tee` -> once per item (I guess I could remove one set of brackets there?).

In the return statement, only the _result_ of the predicate (i.e. a boolean) is compared, which should be fast.

There could be an  issue with large input iterables, that contain mostly _truthy_ or _falsy_ values (in the context of our predicate), see [the documentation](https://docs.python.org/2/library/itertools.html#itertools.tee).

coalib/collecting/Collectors.py
mrtes: Can you explain why this is generally bad?
I understand that argument in case of empty list defaults, because the list is only allocated once and all changes values will carry over to the next function call.
However, this is not the case with tuples.

The advantage of `None` is that we are forced to check for Falsy-Values.
However, we also accept all other Falsy values, which can lead to weird behaviour, like accepting `0` and throwing when another number is provided.

In my opinion, using a tuple default value if a collection is expected makes the contract a lot clearer for people using the function.

coalib/parsing/Globbing.py
mrtes: No its not necessary -> I will remove it.
Is there a list, which modules are used by other applications?
That would make it clearer, which functions can be modified safely.

mrtes:
@jayvdb  I changed my PR according to your comments (with the exception of the partition function) and restructured the code a bit.

coalib/parsing/Globbing.py
```diff
@@ -196,7 +196,8 @@ def translate(pattern):
 
 def fnmatch(name, globs):
     """
-    Tests whether name matches one of the given globs.
+    Tests whether name matches one of the given globs. An empty glob will
```
jayvdb: would be nice to add your addition starting on a new line, so the old line isnt modified.

tests/collecting/CollectorsTest.py
```diff
@@ -94,10 +120,10 @@ def test_limited(self):
                                         '*py')],
                           self.log_printer,
                           limit_file_paths=[os.path.join(
-                                                self.collectors_test_dir,
-                                                'others',
-                                                '*',
-                                                '*2.py')]),
+                              self.collectors_test_dir,
```
jayvdb: this change isnt necessary.
please undo

tests/collecting/CollectorsTest.py
```diff
@@ -339,7 +365,8 @@ def test_filter_section_bears_by_languages(self):
         self.assertEqual(str(local_bears['test_section'][0]),
                          "<class 'bears2.Test2LocalBear'>")
 
-        global_bears = filter_section_bears_by_languages(global_bears, ['Java'])
+        global_bears = filter_section_bears_by_languages(
```
jayvdb: this change wasnt necessary; please undo

coalib/processes/Processing.py
```diff
@@ -382,7 +382,6 @@ def instantiate_processes(section,
     """
     filename_list = collect_files(
         glob_list(section.get('files', '')),
-        None,
```
jayvdb: you can drop this.  A newcomer will do it.
https://github.com/coala/coala/issues/5341

mrtes:
Implemented your suggestions.
Btw. the partition function I added, could be moved to some utils module, so that it can be used in other contexts, without requiring a dependency to the collectors module.
I'm just not sure which module would be suited for this.
Any suggestions?

jayvdb:
`coalib.misc.IterUtilities`?

or you can simply leave it and prefix it with `_`, and we worry about its final resting home another day.

mrtes:
Alright.
Going with your first suggestion :)

mrtes:
Not sure what the problems of circle ci and gitmate are.
My local tests succeed, and I haven't really changed much, except for moving the `partition` function.
Anyway the circle ci failures do not seem to be related to failing tests?

jayvdb:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/coala/pull/4842/commits/3d8c8189e09dcd228752200dfdfefd9212e4dc7d

jayvdb:
@gitmate-bot ff

Issue #4668, opened 2017-08-24
Output message in json if args.json is true and no bears to show


shreyans800755:
Following ~~https://github.com/coala/coala/pull/4650~~ https://github.com/coala/coala/pull/4726.

With above PR, `--filter-by` option is enabled for json and format entrypoints of coala. But when there are no bears to show, following message is shown even if `json` flag given in the command:
```
No bears to show. Did you forget to install the `coala-bears` package? Try `pip3 install coala-bears`.
```

Update the message to be in proper format according to args given in the command - json or format or plaintext

CC @Makman2 

EDIT by @Makman2:
Point directly to blocking issue.

Makman2:
Maybe we just put no message at all into the empty Json list?

shreyans800755:
Should be good. What about `--format` ?

Makman2:
hm maybe also no output? As it's some "systematic" output usually for other executables

Asalle:
Why the status is `blocked`?

Makman2:
Good point, I probably forgot to remove it.

Makman2:
Ah wait no the relevant PR is still not merged.

shreyans800755:
@Asalle This issue is waiting for https://github.com/coala/coala/issues/4499 to be resolved.

nvzard:
Is this issue still blocked?

shreyans800755:
@nvzard not any more. Please assign yourself and go ahead with the work. Ping me on PR or gitter for review once you have some progress to be shown

shreyans800755:
Also feel free to ask doubts if you're not clear with something.

nvzard:
Thanks for supporting :)

nvzard:
Can you please tell me what message do you want it to show?

nvzard:
Sorry I was busy with my exams. I've started working on this issue. Can you please tell me what message do you want it to show?

Makman2:
☝️ https://github.com/coala/coala/issues/4668#issue-252743994
> With above PR, --filter-by option is enabled for json and format entrypoints of coala. But when there are no bears to show, following message is shown even if json flag given in the command:
> `No bears to show. Did you forget to install the `coala-bears` package? Try `pip3 install coala-bears`.`

> Maybe we just put no message at all into the empty Json list?

shreyans800755:
@Makman2 Please remove the blocked label.

PR #4877, opened 2017-11-16
ConsoleInteraction: Remove no bears message

nvzard:
Disables "No bears to show." message when
--json flag is used.

Fixes https://github.com/coala/coala/issues/4668

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -932,7 +932,7 @@ def print_bears(bears,
     :param console_printer:  Object to print messages on the console.
     :param args:             Args passed to coala command.
     """
-    if not bears:
+    if not bears and not (args and args.json):
         console_printer.print('No bears to show. Did you forget to install '
                               'the `coala-bears` package? Try `pip3 install '
                               'coala-bears`.')
--- a/tests/output/ConsoleInteractionTest.py
+++ b/tests/output/ConsoleInteractionTest.py
@@ -24,6 +24,7 @@ from coalib.output.ConsoleInteraction import (BackgroundSourceRangeStyle,
                                               BackgroundMessageStyle,
                                               highlight_text)
 from coalib.output.printers.ListLogPrinter import ListLogPrinter
+from coalib.parsing.DefaultArgParser import default_arg_parser
 from coalib.results.Diff import Diff
 from coalib.results.Result import Result
 from coalib.results.result_actions.ApplyPatchAction import ApplyPatchAction
@@ -953,6 +954,12 @@ class ShowBearsTest(unittest.TestCase):
         with retrieve_stdout() as stdout:
             show_bears({}, {}, True, True, self.console_printer)
             self.assertIn('No bears to show.', stdout.getvalue())
+
+    def test_show_bears_with_json(self):
+        args = default_arg_parser().parse_args(['--json'])
+        with retrieve_stdout() as stdout:
+            show_bears({}, {}, True, True, self.console_printer, args)
+            self.assertEqual('{\n  "bears": []\n}\n', stdout.getvalue())
 
     @patch('coalib.output.ConsoleInteraction.show_bear')
     def test_show_bears(self, show_bear):


Makman2:
1. It's `--json` ;)
   If commit message is too long then, you can remove the `.py` ending in the tag ;)
2. ~~It~~ ~~d~~Disable**s** ...
3. Use `Closes` instead of `Fixes`, the addressed issue is not a bug.

Makman2:
unack e8d3286

nvzard:
Made the changes. Thank for fast response.

SanketDG:
Look at the CI tests, there is still something wrong.

@Makman2 has already pointed it out!

shreyans800755:
@nvzard 
In commit shortlog:
`ConsoleInteractions`->`ConsoleInteraction`

RJ722:
Also, is there an alternate message we need to show when we have `--json` flag?

RJ722:
Also, you need to add tests! 😉

nvzard:
At the moment we just put no message at all into the empty Json.
>Also, is there an alternate message we need to show when we have --json flag?

Thanks a lot @RJ722 , this is my first code related issue and I really needed something like this. Thanks a lot again :)

Makman2:
> Also, is there an alternate message we need to show when we have --json flag?

No that's what we want to get rid off with this PR ;)
An empty json list will emit something into the terminal, probably `[]`, which is quite understandable meaning "no bears" ;)

jayvdb:
Please run the tests locally before pushing to be reviewed.

nvzard:
Sorry, I'll do that from now on.

RJ722:
I think a better way to represent this would be to write it this way:
```py
    json_flag = args and args.json
    if not bears and not json_flag:
        console_printer.print('No bears to show. Did you forget to install '
                              'the `coala-bears` package? Try `pip3 install '
                              'coala-bears`.')
        return
```
It is more clear and does not raise the aforementioned error.
Also, you can change the test here: https://github.com/coala/coala/blob/master/tests/output/ConsoleInteractionTest.py#L936

May be, add a case where you pass on an arg flag and assert that nothing is printed on `stdout`.

nvzard:
@RJ722 Thanks. I thought of this :P Your solution is more elegant.
```
if not bears:
    if args and args.json:
        pass
    else:
        console_printer.print('No bears to show. Did you forget to install '
                                          'the `coala-bears` package? Try `pip3 install '
                                          'coala-bears`.')
    return
```
It passed test_show_bears_empty but failed some other

nvzard:
If I create a test case 'test_show_bears_with_args' and pass args as True, it will still show the error message. How can I pass args.json in the function instead of True?
>May be, add a case where you pass on an arg flag and assert that nothing is printed on stdout.

nvzard:
If I create a test case 'test_show_bears_with_args' or 'test_show_bears_with_json_flag' and pass args as True, it will still show the error message. How can I pass args.json in the function instead of True?
>May be, add a case where you pass on an arg flag and assert that nothing is printed on stdout.

nvzard:
@shreyans800755 
While writing tests, If I want to pass 'args' in a method I can simply use 'True'. But what if I want to pass 'args.json' specifically in the method. What should I use?
```
def test_show_bears_empty(self):
        with retrieve_stdout() as stdout:
            show_bears({}, {}, True, True, self.console_printer, True)
            self.assertIn('No bears to show.', stdout.getvalue())
```
`def show_bears(local_bears, global_bears, show_description, show_params, console_printer, args=None):`

shreyans800755:
@nvzard 
Try this:
args = parse_args(['--json'])

Source: https://stackoverflow.com/questions/18160078/how-do-you-write-tests-for-the-argparse-portion-of-a-python-module

Solution I mentioned looked clean and good to me. You can look for more answers if there is any better one.

Vedarth:
It looks good to me ack https://github.com/coala/coala/pull/4877/commits/563055e812be5c164f316ffc84676e7d7378c1db

RJ722:
Tests are yet to be added @Vedarth  😉 

unack https://github.com/coala/coala/commit/563055e812be5c164f316ffc84676e7d7378c1db

shreyans800755:
@nvzard 
Look at https://github.com/coala/coala/pull/4877#issuecomment-345277467.
We need to show something like `{}` or `[]` to indicate there are no bears matching the filter criteria.
You are not showing anything here.

unack bfcf40e

tests/output/ConsoleInteractionTest.py
```diff
@@ -938,6 +939,12 @@ def test_show_bears_empty(self):
             show_bears({}, {}, True, True, self.console_printer)
             self.assertIn('No bears to show.', stdout.getvalue())
 
+    def test_show_bears_with_json(self):
+        args = default_arg_parser().parse_args(['--json'])
+        with retrieve_stdout() as stdout:
+            show_bears({}, {}, True, True, self.console_printer, args)
+            self.assertIn('', stdout.getvalue())
```
shreyans800755: Output should be `{}` or `[]` and not empty.

nvzard: Done.

tests/output/ConsoleInteractionTest.py
```diff
@@ -938,6 +939,12 @@ def test_show_bears_empty(self):
             show_bears({}, {}, True, True, self.console_printer)
             self.assertIn('No bears to show.', stdout.getvalue())
 
+    def test_show_bears_with_json(self):
+        args = default_arg_parser().parse_args(['--json'])
+        with retrieve_stdout() as stdout:
+            show_bears({}, {}, True, True, self.console_printer, args)
+            self.assertIn('[]', stdout.getvalue())
```
shreyans800755: How come your test is passing with this change I suggested in earlier comment without any code changes ?
And actual output I'm getting on my local is:
```
{
  "bears": []
}
```

shreyans800755:
unack 3c23b78

shreyans800755:
@nvzard 
Oh got it. You are using `self.assertIn`.
Instead, use `self.assertEqual` and match the exact string.

tests/output/ConsoleInteractionTest.py
shreyans800755: https://github.com/coala/coala/pull/4877#issuecomment-347036274

nvzard: Oh sorry, my bad. I'll try to fix this :)

nvzard:
Oh sorry, my bad. Thanks for pointing it out. I'll try to fix this :)

nvzard:
@shreyans800755  I've made the requested changes, can you please review this PR. Its been quite some time :P

coalib/output/ConsoleInteraction.py
```diff
@@ -952,6 +953,8 @@ def print_bears(bears,
             filename = args.output[0]
             with open(filename, 'w') as fp:
                 fp.write(json_formatted_output)
+        elif not results:
+            print(json_output['bears'])
```
shreyans800755: I don't think we need to take care of this separately. This way, we'll two different output if we are printing it on console or in file.
Something like this should be fine:
```
{
  bears: []
}
```
And update the tests accordingly.

shreyans800755:
unack b4c713a

nvzard:
@shreyans800755 made the requested changes :)

coalib/output/ConsoleInteraction.py
```diff
@@ -929,7 +929,8 @@ def print_bears(bears,
     :param console_printer:  Object to print messages on the console.
     :param args:             Args passed to coala command.
     """
-    if not bears:
+    json_flag = args and args.json
+    if not bears and not json_flag:
```
Makman2: up to you: maybe inlining `args and args.json`

nvzard: Done :+1:

nvzard: Thanks for your suggestion :) It made the code more clear and removed the unnecessary use of a variable.

coalib/output/ConsoleInteraction.py
```diff
@@ -929,7 +929,8 @@ def print_bears(bears,
     :param console_printer:  Object to print messages on the console.
     :param args:             Args passed to coala command.
     """
-    if not bears:
+
```
Makman2: I would remove this newline ;)

nvzard: Done :+1:

Makman2:
ack c86a3b9

Makman2:
@gitmate-bot rebase

Issue #4356, opened 2017-06-12
coala --files does\not\support\backslash\paths\on\windows


userzimmermann:
The backslashes are just discarded:

```
> coala -I --bears PEP8Bear --files docs\conf.py
Executing section cli...
[WARNING][17:14:44] No files matching 'C:\Users\Zimmermann\Projects\coala\lib\docsconf.py' were found. If this rule is not required, you can remove it from section [cli] in your .coafile to deactivate this warning.
```

@sils Funny that no one complained about that so far :D

jayvdb:
I think you'll find this is the same rabbit hole as https://github.com/coala/coala/pull/4171 , and IMO the only sane solution is to use pathlib (https://github.com/coala/coala/issues/4358)

Makman2:
Just escape the path. coala interpretes backslashes, so this is actually intended currently. I'm not sure if #4171 is good as we are starting to be inconsistent because we parse different settings differently.

userzimmermann:
@Makman2 What's so bad about different value types for different settings, like normal string type (w/ escaping) or path type (w/o escaping)?

cc @jayvdb

Makman2:
this makes stuff super inconsistent, and I can really feel that this is a bad idea and we will hit issues in the future^^

Makman2:
Actually I'm for keeping our current escape sequence support (although not really mature right now) everywhere inside a coafile, regardless which setting.

Makman2:
Changed priority, as there's a workaround (by using double backslashes).

userzimmermann:
@Makman2 This is nothing but a torture for Windows cmdline users!

userzimmermann:
@Makman2 
> this makes stuff super inconsistent, and I can really feel that this is a bad idea and we will hit issues in the future^^

We will hit even more issues in the future if we don't correct the mistakes of the past ;)

userzimmermann:
@Makman2 Oh! BTW... Maybe we have some little misunderstanding ;)

> Actually I'm for keeping our current escape sequence support (although not really mature right now) everywhere inside a coafile, regardless which setting.

That's fine! I would never use backslashes as path separators in a `.coafile`

It's only about the `--files` cmdline flag

Makman2:
> This is nothing but a torture for Windows cmdline users!

Your fault choosing windows 😅

> We will hit even more issues in the future if we don't correct the mistakes of the past ;)

[...]

PR #4883, opened 2017-11-19
DefaultArgParser: Uni(xi)fy directory separators

userzimmermann:
Use a new `PathArg` type for all cmdline arguments that represent file system paths. It replaces OS-native directory separators with Unix-style forward slashes. Windows-style backslashes would otherwise be interpreted as escaping syntax and therefore removed when turning cmdline arguments into coala settings, what makes interactively working with coala in a Windows cmdline a pain ;)

Closes https://github.com/coala/coala/issues/4356

cc @Makman2 @jayvdb

### DIFF
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 
 from coalib.misc import Constants
 from coalib.collecting.Collectors import get_all_bears_names
@@ -20,6 +21,23 @@ class CustomFormatter(argparse.RawDescriptionHelpFormatter):
             # Option string arguments (like "-f, --files")
             parts = action.option_strings
             return ', '.join(parts)
+
+
+class PathArg(str):
+    """
+    Uni(xi)fying OS-native directory separators in path arguments.
+
+    Removing the pain from interactively using coala in a Windows cmdline,
+    because backslashes are interpreted as escaping syntax and therefore
+    removed when arguments are turned into coala settings
+
+    >>> import os
+    >>> PathArg(os.path.join('path', 'with', 'separators'))
+    'path/with/separators'
+    """
+
+    def __new__(cls, path):
+        return str.__new__(cls, path.replace(os.path.sep, '/'))
 
 
 def default_arg_parser(formatter_class=None):
@@ -142,19 +160,19 @@ To run coala without user interaction, run the `coala --non-interactive`,
             lambda *args, **kwargs: get_all_bears_names())  # pragma: no cover
 
     inputs_group.add_argument(
-        '-f', '--files', nargs='+', metavar='FILE',
+        '-f', '--files', type=PathArg, nargs='+', metavar='FILE',
         help='files that should be checked')
 
     inputs_group.add_argument(
-        '-i', '--ignore', nargs='+', metavar='FILE',
+        '-i', '--ignore', type=PathArg, nargs='+', metavar='FILE',
         help='files that should be ignored')
 
     inputs_group.add_argument(
-        '--limit-files', nargs='+', metavar='FILE',
+        '--limit-files', type=PathArg, nargs='+', metavar='FILE',
         help="filter the `--files` argument's matches further")
 
     inputs_group.add_argument(
-        '-d', '--bear-dirs', nargs='+', metavar='DIR',
+        '-d', '--bear-dirs', type=PathArg, nargs='+', metavar='DIR',
         help='additional directories which may contain bears')
 
     outputs_group = arg_parser.add_argument_group('Outputs')
--- a/tests/coalaCITest.py
+++ b/tests/coalaCITest.py
@@ -61,7 +61,7 @@ class coalaCITest(unittest.TestCase):
             retval, stdout, stderr = execute_coala(coala.main, 'coala',
                                                    '--non-interactive',
                                                    '-c', os.devnull,
-                                                   '-f', re.escape(filename),
+                                                   '-f', filename,
                                                    '-b',
                                                    'SpaceConsistencyTestBear',
                                                    '--settings',
@@ -87,7 +87,7 @@ class coalaCITest(unittest.TestCase):
                                                    '--non-interactive',
                                                    '-c', os.devnull,
                                                    '-b', 'LineCountTestBear',
-                                                   '-f', re.escape(filename),
+                                                   '-f', filename,
                                                    debug=debug)
             self.assertIn('This file has 1 lines.',
                           stdout,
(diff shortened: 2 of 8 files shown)


userzimmermann:
This makes also all the `re.escape(filename)` hacks in tests obsolete (whereby using `re.escape` is anyway not a good solution for that purpose, since it escapes every non-alnum char). And since those hacks are actually also not working anymore together with `PathArg`, those hacks are removed as part of the one single commit of this PR. Splitting the commit would in any case lead to failing tests for the first one

cc @Makman2 @jayvdb

Makman2:
`re.escape` should rather be exchanged with `coala_utils.string_processing.escape(..., '\\')` (avoiding to escape other stuff that is just special for regex), but this is a different issue ;)

userzimmermann:
@Makman2 You mean https://github.com/coala/coala/issues/4890 ? ;)

userzimmermann:
@Makman2 BTW: I hope you are meanwhile convinced enough that this PR is more than efficient and harmless enough to get merged ;P Have a ❤️ for all the brave Windows coalaians out there 😄

userzimmermann:
@Makman2 Now have that ❤️ finally! ;P

coalib/parsing/DefaultArgParser.py
```diff
@@ -22,6 +23,23 @@ def _format_action_invocation(self, action):
...
+    because backslashes are interpreted as escaping syntax and therefore
+    removed when arguments are turned into coala settings
+
+    >>> import os
+    >>> PathArg(os.path.join('path', 'with', 'separators'))
+    'path/with/separators'
+    """
+
+    def __new__(cls, path):
+        return str.__new__(cls, path.replace(os.path.sep, '/'))
```
Makman2: why not using `__init__`?

Makman2: or generally a conversion function instead of a type

coalib/parsing/DefaultArgParser.py
```diff
@@ -22,6 +23,23 @@ def _format_action_invocation(self, action):
             return ', '.join(parts)
 
 
+class PathArg(str):
```
Makman2: not sure if we should call it `path_arg` instead (thinking of `str`, `int`, etc because these are also all lowercase)

Makman2: and maybe we should just call it path? We could maybe use it then also as a settings type for bears (I hope this works....)

coalib/parsing/DefaultArgParser.py
userzimmermann: @Makman2 _\*whisper\* ;) Let's silently forget your `__init__` question ;) `str` is immutable. So `__init__` can't change it's value anymore. So the only way to override it's construction is, well, by overriding the actual constructor method, which is actually `__new__`_

Function or class? Well, you won't believe me how long I though about that... But you have to make a decision :D In the end I just thought a class looks more like a `type=` than a function ;)

coalib/parsing/DefaultArgParser.py
userzimmermann: @Makman2 I wouldn't make it a general setting type. I wouldn't allow `paths\with\backslashes` in config files. All your actual worries will come true then. And mine are the same :)

That's why the `PathArg` name, because it should really only be for the `ArgParser`, for cmdline args

userzimmermann: @Makman2 According to https://docs.python.org/3.6/distutils/setupscript.html#writing-the-setup-script ...

_"Note that any pathnames (files or directories) supplied in the setup script should be written using the Unix convention, i.e. slash-separated. The Distutils will take care of converting this platform-neutral representation into whatever is appropriate on your current platform before actually using the pathname. This makes your setup script portable across operating systems, which of course is one of the major goals of the Distutils. In this spirit, all pathnames in this document are slash-separated."_

**No** to backslashes in config file. **Yes** to backslashes on the Windows cmdline :)

userzimmermann: @Makman2 So an according bear setting type would again only make sense for the cmdline. When there will ever be a way to provide bear settings on the cmdline ;)

userzimmermann: @Makman2 Hmm... Actually I'm working on supporting bear settings as cmdline-like flags in `%%...Bear` cell magics as part of `bearsh`... So we should really think about some well-balanced typing mechanism there :)

It's like:

```python
In [*]: %%...Bear --setting value
   ...: code-to-anylize...
   ...: ...
```

userzimmermann: @Makman2 On the other hand, IPython always uses forward slashes in it's path auto-completion...

userzimmermann: But those should be maybe converted back to backslashes under Windows... Maybe not every external linter tool works with forward slashed paths on Windows... Hmm... The other way around :D We should later really think about such stuff...

userzimmermann:
BTW: Everybody reading through this PR and thinking now (or even already before) _OMG! Microsoft, what have you done?_ should read https://blogs.msdn.microsoft.com/larryosterman/2005/06/24/why-is-the-dos-path-character/ :) Those backslashes have a very interesting background

coalib/parsing/DefaultArgParser.py
Makman2: > *whisper* ;) Let's silently forget your __init__ question ;) str is immutable. So __init__ can't change it's value anymore. So the only way to override it's construction is, well, by overriding the actual constructor method, which is actually __new__

Already thought so it has to do with str's immutability, but just wanted to clarify, I actually forgot nearly everything about python's metaclassing magic 😅 

I would just say for the sake of simplicity use a function, however we might want to change str representations to be nice and cool, and a function can't do that ;) so let's keep it then :)

coalib/parsing/DefaultArgParser.py
Makman2: phew man you put many comments here ... 😅 

> I wouldn't make it a bear setting type. It would allow paths\with\backslashes in config files, right? All your actual worries will come true then. And mine are the same :)

Yeah that's a good thought, let's teach people the only way: The Linux way :P :P 
No but really for consistency this is a good point :+1:

Though we can consider reusing the type in a later issue :+1:

Makman2:
>  And are not really Microsoft's fault

No comment on why they kept it up then 😅

Makman2:
ack f385e00

userzimmermann:
reack https://github.com/coala/coala/pull/4883/commits/d26f7a7cab868e51c72d1a644d92ca927dd81acd

userzimmermann:
reack https://github.com/coala/coala/pull/4883/commits/b5a3bc7f4eaa3251730fc36a3d66b2345c8108d6

PR #4897, opened 2017-11-21
Build setup.py and requirements.txt using moban

jayvdb:
coala-mobans provides a set of metadata common to
all repositories, with local overrides.

requirements.txt is generated using the default
moban provided requirements.txt.jj2.

### DIFF
--- a/.moban.dt/core-setup.py.jj2
+++ b/.moban.dt/core-setup.py.jj2
@@ -1,0 +1,41 @@
+{% extends 'coala-setup.py.jj2' %}
+
+{% block local_exec_block_1 %}
+assert_supported_version()
+
+
+class BuildPyCommand(setuptools.command.build_py.build_py):
+
+    def run(self):
+        if platform.system() != 'Windows':
+            self.run_command('build_manpage')
+        setuptools.command.build_py.build_py.run(self)
+
+{% endblock %}
+
+{% block local_exec_block_2 %}
+
+# Generate API documentation only if we are running on readthedocs.io
+on_rtd = getenv('READTHEDOCS', None) is not None
+if on_rtd:
+    call(BuildDocsCommand.apidoc_command)
+    if 'dev' in '{{ build_version }}':
+        current_version = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        call(['python3', '.misc/adjust_version_number.py', 'coalib/VERSION',
+              '-b {}'.format(current_version)])
+        VERSION = get_version()
+
+{% endblock %}
+
+{% block custom_requirements %}
+if __name__ == '__main__':
+    if platform.system() != 'Windows':
+        data_files = [('man/man1', ['coala.1'])]
+    else:
+        data_files = None
+{% endblock %}
+
+{% block custom_setup_commands %}
+                    'build_manpage': BuildManPage,
+                    'build_py': BuildPyCommand,
+{% endblock %}
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,10 @@ before_script:
   # Restore the original requirements.txt
   - mv requirements.orig requirements.txt
   - python .misc/check_setuptools.py
+  - git clone https://github.com/moremoban/setupmobans ../setupmobans
+  - git clone https://gitlab.com/coala/mobans ../coala-mobans
+  - moban
+  - git diff --exit-code
 
 script:
   - py.test
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ before_install:
   - if [[ $TRAVIS_OS_NAME == osx ]]; then pip install -r requirements.txt; fi
 
   # For bears in .coafile requiring npm dependencies
-  - npm install
+  - npm install --no-save
 
 before_script:
   # Restore the original requirements.txt
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ import setuptools.command.build_py
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 
-from coalib import VERSION, assert_supported_version, get_version
+from coalib import assert_supported_version, get_version
 from coalib.misc.BuildManPage import BuildManPage
 
 try:
@@ -22,6 +22,7 @@ try:
 except (ValueError, UnicodeError):
     locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')
 
+VERSION = '0.12.0.dev99999999999999'
 
 assert_supported_version()
 
@@ -45,25 +46,29 @@ class PyTestCommand(TestCommand):
 
 class BuildDocsCommand(setuptools.command.build_py.build_py):
     apidoc_command = (
-        'sphinx-apidoc', '-f', '-o', 'docs', '--no-toc', 'coalib'
+        'sphinx-apidoc', '-f', '-o', 'docs',
+        '--no-toc',
+        'coalib'
     )
-    doc_command = ('make', '-C', 'docs', 'html', 'SPHINXOPTS=-W')
+    make_command = ('make', '-C', 'docs', 'html', 'SPHINXOPTS=-W')
 
     def run(self):
-        errOne = call(self.apidoc_command)
-        errTwo = call(self.doc_command)
-        sys.exit(errOne or errTwo)
+        err_no = call(self.apidoc_command)
+        if not err_no:
+            err_no = call(self.make_command)
+        sys.exit(err_no)
 
 
 # Generate API documentation only if we are running on readthedocs.io
 on_rtd = getenv('READTHEDOCS', None) is not None
 if on_rtd:
     call(BuildDocsCommand.apidoc_command)
-    if 'dev' in VERSION:
+    if 'dev' in '0.12.0.dev99999999999999':
         current_version = datetime.datetime.now().strftime('%Y%m%d%H%M%S')
         call(['python3', '.misc/adjust_version_number.py', 'coalib/VERSION',
               '-b {}'.format(current_version)])
         VERSION = get_version()
+
 
 with open('requirements.txt') as requirements:
     required = requirements.read().splitlines()
@@ -74,13 +79,15 @@ with open('test-requirements.txt') as requirements:
 with open('README.rst') as readme:
     long_description = readme.read()
 
-
+extras_require = None
+data_files = None
 if __name__ == '__main__':
     if platform.system() != 'Windows':
         data_files = [('man/man1', ['coala.1'])]
     else:
         data_files = None
 
+if __name__ == '__main__':
     setup(name='coala',
           version=VERSION,
           description='Linting and Fixing Code for All Languages',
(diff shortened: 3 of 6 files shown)


jayvdb:
c.f. https://github.com/coala/coala-bears/pull/2134 , which does the same for coala-bears, using the same templates

.moban.dt/core-setup.py.jj2
```diff
@@ -0,0 +1,48 @@
+{% extends 'coala-setup.py.jj2'%}
```
yukiisbored: Where's the `coala-setup.py.jj2`?

package.json
```diff
@@ -2,5 +2,5 @@
   "name": "coala",
   "dependencies": {
     "csslint": "~1"
-    }
+  }
```
jayvdb: Im not sure why this has changed.

.travis.yml
```diff
@@ -68,6 +68,10 @@ before_script:
   # Restore the original requirements.txt
   - mv requirements.orig requirements.txt
   - python .misc/check_setuptools.py
+  - git clone https://github.com/moremoban/setupmobans ../setupmobans
+  - git clone https://github.com/jayvdb/coala-mobans ../coala-mobans
```
jayvdb: before this can be merged; that repo needs to be moved to coala org

.moban.dt/core-setup.py.jj2
```diff
@@ -0,0 +1,49 @@
...
+{%endblock%}
+
+{%block local_exec_block_1%}
+assert_supported_version()
+
+
+class BuildPyCommand(setuptools.command.build_py.build_py):
+
+    def run(self):
+        if platform.system() != 'Windows':
```
sims1253: what do we do if it is windows?

.moban.yml
```diff
@@ -0,0 +1,48 @@
+overrides: coala.yaml
+
+name: coala
+contact: coala.analyzer@gmail.com
```
sims1253: no `@coala.io` address?

setup.py
```diff
@@ -45,26 +44,30 @@ def run_tests(self):
 
 class BuildDocsCommand(setuptools.command.build_py.build_py):
     apidoc_command = (
-        'sphinx-apidoc', '-f', '-o', 'docs', '--no-toc', 'coalib'
+        'sphinx-apidoc', '-f', '-o', 'docs',
+        '--no-toc',
```
sims1253: I don't really get this reformatting. It is not consistent in itself. Lines neither hold the same number of elements nor are they the same length.
  Or is it grouping by type of parameter?

setup.py
```diff
@@ -45,26 +44,30 @@ def run_tests(self):
...
-    doc_command = ('make', '-C', 'docs', 'html', 'SPHINXOPTS=-W')
+    make_command = ('make', '-C', 'docs', 'html', 'SPHINXOPTS=-W')
 
     def run(self):
-        errOne = call(self.apidoc_command)
-        errTwo = call(self.doc_command)
-        sys.exit(errOne or errTwo)
+        err_no = call(self.apidoc_command)
+        if not err_no:
+            err_no = call(self.make_command)
```
sims1253: Even though the new version is better performance wise (no idea how much) I find the fist one a lot easier to read.

.moban.yml
jayvdb: see https://github.com/coala/coala/blob/master/setup.py#L88

Changing that is a separate issue (and probably an issue already exists, and possibly already a stale PR, or 5..).

.moban.dt/core-setup.py.jj2
jayvdb: Nothing.  Note that this template re-creates [the *existing* setup.py](https://github.com/coala/coala/blob/308e1e3/setup.py), and those lines were last modified in https://github.com/coala/coala/commit/e6bd2d3f6395ccc094e31a01402962fb60d40c8e.

setup.py
jayvdb: This is being formatted by moban, and the logic is re-used between core and -bears.
See https://github.com/coala/coala-bears/blob/5ca3eb7/setup.py#L28 where bears has a different set of parameters for this.

setup.py
jayvdb: This is the one used by coala-bears: https://github.com/coala/coala-bears/blob/5ca3eb7/setup.py#L34

Here is us fixing the same bug twice .. ;-)

https://github.com/coala/coala/commit/069e468c3a8f21d1e7882501340d1e52d16bd910 - Dec 24, 2016
https://github.com/coala/coala-bears/commit/0a99c06853f57474cd4be44fb3164bbf70cca922 - Mar 22, 2016

Invoking a command unnecessarily makes for bad reading, and the second command could also cause side effects which are undesirable and hard to understand because the exit code is from the first command.

.moban.dt/core-setup.py.jj2
```diff
@@ -0,0 +1,49 @@
+{% extends 'coala-setup.py.jj2'%}
+
+{%block local_import_block%}
```
Monal5031: I don't see this block on the template which you are extending? i.e. https://github.com/jayvdb/coala-mobans/blob/master/templates/coala-setup.py.jj2

setup.py
```diff
@@ -45,26 +44,30 @@ def run_tests(self):
 
 class BuildDocsCommand(setuptools.command.build_py.build_py):
     apidoc_command = (
-        'sphinx-apidoc', '-f', '-o', 'docs', '--no-toc', 'coalib'
+        'sphinx-apidoc', '-f', '-o', 'docs',
```
sangamcse: Shouldn't it be placed with line 46?
> See https://github.com/coala/coala-bears/blob/5ca3eb7/setup.py#L28

setup.py
```diff
@@ -74,15 +77,17 @@ def run(self):
 with open('README.rst') as readme:
     long_description = readme.read()
 
-
+extras_require = None
```
sangamcse: One blank line needed.

.moban.dt/core-setup.py.jj2
```diff
@@ -0,0 +1,41 @@
+{% extends 'coala-setup.py.jj2'%}
```
sangamcse: I don't have much idea about this format but there is a space before `extends`, is it required?

setup.py
Monal5031: I'm guessing earlier it was because of the line length, now it can be on the same line.

setup.py
```diff
@@ -74,15 +77,17 @@ def run(self):
...
 if __name__ == '__main__':
     if platform.system() != 'Windows':
         data_files = [('man/man1', ['coala.1'])]
     else:
         data_files = None
 
+if __name__ == '__main__':
     setup(name='coala',
-          version=VERSION,
+          version='0.12.0.dev99999999999999',
```
Monal5031: I'm confused over why to hardcode the version when [Line 68](https://github.com/coala/coala/pull/4897/files#diff-2eeaed663bd0d25b7e608891384b7298R68) will get us the same version.

.moban.dt/core-setup.py.jj2
jayvdb: Is there are a style guide for jinja2? :)

We have a bear, and I have enabled it.

Does it need enhancing? :P

setup.py
jayvdb: Where?

setup.py
jayvdb: hmm.  actually no, that only happens when 'dev' is in the version.
but, line 68 is definitely problematic; will need to think through how that should work in a moban controlled system.

setup.py
```diff
@@ -134,7 +142,7 @@ def run(self):
               'Topic :: Scientific/Engineering :: Information Analysis',
               'Topic :: Software Development :: Quality Assurance',
               'Topic :: Text Processing :: Linguistic'],
-          cmdclass={'build_manpage': BuildManPage,
+          cmdclass={'docs': BuildDocsCommand,
```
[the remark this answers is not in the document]
jayvdb: bears has docs and test, and now coala has docs and other stuff and test.  That was the easy way to avoid too much differences in the output.  But I'll look into alpha sorting.  I suspect it means moving all these commands into metadata instead of template blocks, and I dont know it is worth the complexity.  But worth looking at again.

setup.py
sangamcse: Oh, I thought earlier line [76 and 77](https://github.com/coala/coala/pull/4897/files#diff-2eeaed663bd0d25b7e608891384b7298L77) was blank and was good. But after changes, it is not detected as required blank line, then it might be good. Not sure. 😛

.moban.dt/core-setup.py.jj2
sangamcse: Don't know. Found this https://github.com/pallets/jinja#nutshell 
> Is there a style guide for jinja2? :)

setup.py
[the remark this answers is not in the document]
Monal5031: Maybe we can keep the docs and test in meta, and let the other stuff stay in the template, since those two are same for both?

setup.py
Monal5031: I have a suggestion don't know if it would be good to do so.

can we always adjust the version (using `.misc/adjust_version_number.py`) rather than adjusting only when there is a dev, that way maybe we won't need to hardcode it.

yukiisbored: @Monal5031 This is controlled by the template. It doesn't matter since it'll be generated. If the file isn't updated, CI will fail anyway.

yukiisbored:
ack https://github.com/coala/coala/pull/4897/commits/121b6028f31842148c4147ec4674fced6f280745

yukiisbored:
ack https://github.com/coala/coala/pull/4897/commits/1254fb93dfef76ec2e8ae3f2934575b74772e188

setup.py
jayvdb: See earler https://github.com/coala/coala/pull/4897#discussion_r159525549

setup.py
```diff
@@ -93,8 +98,9 @@ def run(self):
                             'makman@alice.de'),
           url='http://coala.io/',
           platforms='any',
-          packages=find_packages(exclude=['build.*', 'tests', 'tests.*']),
+          packages=find_packages(exclude=('build.*', 'tests', 'tests.*')),
           install_requires=required,
+          extras_require=extras_require,
```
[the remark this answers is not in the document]
jayvdb: Ya, bears needs this.  Detectable, but not without a re-write, and we have another rewrite planned for the dependencies anyway, so not worth it atm

.moban.dt/core-setup.py.jj2
jayvdb: I've added spaces near the `{%` and `%}`.

setup.py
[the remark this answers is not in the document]
jayvdb: Turns out this isnt easy.

jayvdb:
reack https://github.com/coala/coala/pull/4897/commits/0eff18479d263cb258fdf93653091656c9a456d2

jayvdb:
reack https://github.com/coala/coala/pull/4897/commits/7a09875f2f9d4c69dec0d1685103665a35e4c4c0

setup.py
```diff
@@ -134,7 +144,7 @@ def run(self):
               'Topic :: Scientific/Engineering :: Information Analysis',
               'Topic :: Software Development :: Quality Assurance',
               'Topic :: Text Processing :: Linguistic'],
-          cmdclass={'build_manpage': BuildManPage,
+          cmdclass={'docs': BuildDocsCommand,
```
Makman2: I'm wondering why you actually move the `docs` item here. Looks redundant.

Makman2:
ack 0eff18479d263cb258fdf93653091656c9a456d2

Makman2:
@gitmate-bot ff

Issue #3847, opened 2017-03-04
Detect unsupported clang and report a useful error message


jayvdb:
If the wrong clang version is installed, the error message is like "Please ensure that your python bindings are compatible with your libclang.so version" , which doesnt help the average user.

In addition to the exception message provided by the libclang python wrapper, we need a message which says something like "coala requires clang 3.4"

yashLadha:
I can work upon this @jayvdb

yashLadha:
Is the version checking should be os specific, or we are targeting only linux users.

yashLadha:
How about showing user that the machine doesn't have necessary clang version instead of showing the whole stack trace which is not meant for an average user.

proishan11:
Hello! @jayvdb . Does the code here <https://github.com/coala/coala-bears/blob/master/bears/c_languages/ClangBear.py#L11-L21> need to be changed?

RaiVaibhav:
@proishan11 ping :smile:

Makman2:
Unassigning due to inactivity.

nvzard:
https://github.com/coala/coala/blob/master/coalib/results/Diff.py#L159
Is this the part where I should be working on?

jayvdb:
@nvzard , you've hit a slightly different problem (https://github.com/coala/coala/issues/3182) in the same location , but the error message is equally as bad and needs improvement (which is what this bug is about)  ;-)

the easiest approach is in `Diff.from_clang_fixit`, but there is also `SourceRange.from_clang_range`.  Catch the exception, and then emit a useful error message, and then re-raise the exception.

Another approach is to provide a handler in the bears.  I've been quite keen on moving this clang stuff into a new global bear which acts as a dependency (see BEAR_DEPS) for all the clang bears, so the core coala doesnt need to rely on clang.

```
$ git grep from_clang
bears/c_languages/ClangBear.py:            affected_code = tuple(SourceRange.from_clang_range(range)
bears/c_languages/ClangBear.py:                diffs = {filename: Diff.from_clang_fixit(fixits[0], file)}
bears/c_languages/ClangComplexityBear.py:                affected_code = (SourceRange.from_clang_range(cursor.extent),)
```

PR #4915, opened 2017-11-28
from_clang: Report useful error

nvzard:
Additional simple message

Closes https://github.com/coala/coala/issues/3847

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/results/Diff.py
+++ b/coalib/results/Diff.py
@@ -2,12 +2,18 @@ import copy
 import difflib
 import logging
 
+from pkg_resources import get_distribution
 from unidiff import PatchSet
 
+from clang.cindex import LibclangError
 from coalib.results.LineDiff import LineDiff, ConflictError
 from coalib.results.SourceRange import SourceRange
 from coalib.results.TextRange import TextRange
 from coala_utils.decorators import enforce_signature, generate_eq
+
+
+if get_distribution('libclang-py3').version != '3.4.0':
+    raise LibclangError('coala requires clang 3.4.0')
 
 
 @generate_eq('_file', 'modified', 'rename', 'delete')
--- a/coalib/results/SourceRange.py
+++ b/coalib/results/SourceRange.py
@@ -1,9 +1,15 @@
 from os.path import relpath
+from pkg_resources import get_distribution
 
+from clang.cindex import LibclangError
 from coala_utils.decorators import enforce_signature, get_public_members
 from coalib.results.SourcePosition import SourcePosition
 from coalib.results.TextRange import TextRange
 from coalib.results.AbsolutePosition import AbsolutePosition
+
+
+if get_distribution('libclang-py3').version != '3.4.0':
+    raise LibclangError('coala requires clang 3.4.0')
 
 
 class SourceRange(TextRange):
--- a/tests/results/DiffTest.py
+++ b/tests/results/DiffTest.py
@@ -3,8 +3,12 @@ import logging
 import unittest
 import unittest.case
 
+from clang.cindex import LibclangError
+from importlib import reload
 from unidiff.errors import UnidiffParseError
+from unittest.mock import patch
 
+import coalib.results.Diff
 from coalib.output.JSONEncoder import create_json_encoder
 from coalib.results.Diff import ConflictError, Diff, SourceRange
 
@@ -567,6 +571,13 @@ class DiffTest(unittest.TestCase):
         fixit = tu.diagnostics[0].fixits[0]
         clang_fixed_file = Diff.from_clang_fixit(fixit, file).modified
         self.assertEqual(fixed_file, clang_fixed_file)
+
+    @patch('pkg_resources.get_distribution')
+    def test_clang_version_check(self, get_distribution):
+        get_distribution.configure_mode(version='1.2')
+        with self.assertRaisesRegexp(LibclangError,
+                                     'coala requires clang 3.4.0'):
+            reload(coalib.results.Diff)
 
     def test_equality(self):
         a = ['first', 'second', 'third']
--- a/tests/results/SourceRangeTest.py
+++ b/tests/results/SourceRangeTest.py
@@ -2,10 +2,14 @@ import unittest
 from collections import namedtuple
 from os.path import abspath
 
+from clang.cindex import LibclangError
+from importlib import reload
+from unittest.mock import patch
 from coalib.results.SourcePosition import SourcePosition
 from coalib.results.SourceRange import SourceRange
 from coalib.results.AbsolutePosition import AbsolutePosition
 from coalib.results.Diff import Diff
+import coalib.results.SourceRange
 
 
 class SourceRangeTest(unittest.TestCase):
@@ -41,6 +45,13 @@ class SourceRangeTest(unittest.TestCase):
         uut = SourceRange.from_clang_range(ClangRange(start, end))
         compare = SourceRange.from_values('t.c', 1, 2, 3, 4)
         self.assertEqual(uut, compare)
+
+    @patch('pkg_resources.get_distribution')
+    def test_clang_version_check(self, get_distribution):
+        get_distribution.configure_mode(version='1.2')
+        with self.assertRaisesRegexp(LibclangError,
+                                     'coala requires clang 3.4.0'):
+            reload(coalib.results.SourceRange)
 
     def test_from_absolute_position(self):
         text = ('a\n', 'b\n')


coalib/results/Diff.py
```diff
@@ -166,16 +166,20 @@ def from_clang_fixit(cls, fixit, file):
         """
         assert isinstance(file, (list, tuple))
```
yashasingh: The assert statement can throw `AssertionError`. Seems we can have it in the `try` branch.

coalib/results/Diff.py
```diff
@@ -166,16 +166,20 @@ def from_clang_fixit(cls, fixit, file):
...
+                                      fixit.range.end.line])
+            endindex = fixit.range.end.column - len(file[fixit.range.end.line-1])-1
+
+            newvalue = (oldvalue[:fixit.range.start.column-1] +
+                        fixit.value +
+                        oldvalue[endindex:])
+            new_file = (file[:fixit.range.start.line-1] +
+                        type(file)(newvalue.splitlines(True)) +
+                        file[fixit.range.end.line:])
+        except LibclangError as err:
```
yashasingh: Accordingly, after having assert statement in try branch. It could be modified as `except Error as err:` and inside you may handle `LibclangError` separately. This will make the code more robust IMO :grinning:

coalib/results/Diff.py
```diff
@@ -166,16 +166,20 @@ def from_clang_fixit(cls, fixit, file):
...
+            print("coala requires clang 3.4")
```
newbazz: We can use logging instead of print statements.

coalib/results/SourceRange.py
```diff
@@ -54,11 +54,17 @@ def from_clang_range(cls, range):
...
-                               range.end.line,
-                               range.end.column)
+        try:
+            from_values = cls.from_values(range.start.file.name,
+                                          range.start.line,
+                                          range.start.column,
+                                          range.end.line,
+                                          range.end.column)
+        except LibclangError as err:
+            print("coala requires clang 3.4")
```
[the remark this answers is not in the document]
newbazz: logging can be used here as well

coalib/results/Diff.py
nvzard: @yashasingh  I totally agree with you, but I think handling errors due to Assertion would be a different issue and this PR's goal is to detect unsupported clang and report a useful error message. Maybe we can have a different issue for catching AssertionError?

coalib/results/Diff.py
```diff
@@ -166,16 +166,21 @@ def from_clang_fixit(cls, fixit, file):
...
+                len(file[fixit.range.end.line-1])-1
+
+            newvalue = (oldvalue[:fixit.range.start.column-1] +
+                        fixit.value +
+                        oldvalue[endindex:])
+            new_file = (file[:fixit.range.start.line-1] +
+                        type(file)(newvalue.splitlines(True)) +
+                        file[fixit.range.end.line:])
+        except LibclangError as err:
+            print('coala requires clang 3.4')
```
Makman2: Use logging please instead of `print`

nvzard: Done :+1:

coalib/results/Diff.py
Makman2: Actually this is intended without a try block. The idea is that we want to see the error and making coala stop, otherwise we wouldn't use an `assert` here ;) Otherwise it's the same like doing:
```
try:
    raise Exception
except Exception:
    pass
```

coalib/results/Diff.py
Makman2: no please keep it. This is exactly the intention of the `except` syntax when you provide a specific exception. Catching broad exceptions is usually a bad thing as other errors that you don't know yet are silently catched, and then the program might go into an unpredicted state (which can cause security issues or whatever...).
Robust code doesn't always mean catching any exception ;)

coalib/results/Diff.py
```diff
@@ -166,16 +166,21 @@ def from_clang_fixit(cls, fixit, file):
...
+                len(file[fixit.range.end.line-1])-1
+
+            newvalue = (oldvalue[:fixit.range.start.column-1] +
+                        fixit.value +
+                        oldvalue[endindex:])
+            new_file = (file[:fixit.range.start.line-1] +
+                        type(file)(newvalue.splitlines(True)) +
+                        file[fixit.range.end.line:])
+        except LibclangError as err:
+            logging.warning('coala requires clang 3.4')
```
Makman2: that's definitely an error, no warning ;)

coalib/results/Diff.py
```diff
@@ -166,16 +166,21 @@ def from_clang_fixit(cls, fixit, file):
...
-                    fixit.value +
-                    oldvalue[endindex:])
-        new_file = (file[:fixit.range.start.line-1] +
-                    type(file)(newvalue.splitlines(True)) +
-                    file[fixit.range.end.line:])
+        try:
+            oldvalue = '\n'.join(file[fixit.range.start.line-1:
+                                      fixit.range.end.line])
+            endindex = fixit.range.end.column - \
+                len(file[fixit.range.end.line-1])-1
```
Makman2: please use braces instead of backslashes to break statements over multiple lines.

newbazz:
check the checks in the checklist as well ;)

coalib/results/Diff.py
```diff
@@ -166,16 +166,21 @@ def from_clang_fixit(cls, fixit, file):
...
+
+            newvalue = (oldvalue[:fixit.range.start.column-1] +
+                        fixit.value +
+                        oldvalue[endindex:])
+            new_file = (file[:fixit.range.start.line-1] +
+                        type(file)(newvalue.splitlines(True)) +
+                        file[fixit.range.end.line:])
+        except LibclangError as err:
+            logging.warning('coala requires clang 3.4')
+            raise err
```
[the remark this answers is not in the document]
nvzard: I totally agree with you suggestion. To check the version  we can use
```
except LibclangError as err:
    if 'libclang.so version' in str(err):
        logging.error('coala requires clang 3.4')
        raise err 
```
to be more precise but we first need to catch the [LibclangError](https://github.com/llvm-mirror/clang/blob/a6d3ef724655a64ac137945615d3ff002c4f7993/bindings/python/clang/cindex.py#L4019) somehow so that we can know the error is due to the incompatible version.

We could have used the `__version__` attribute but it is not present in [clang.cindex](https://github.com/llvm-mirror/clang/blob/a6d3ef724655a64ac137945615d3ff002c4f7993/bindings/python/clang/cindex.py) module.

coalib/results/Diff.py
```diff
@@ -166,16 +166,22 @@ def from_clang_fixit(cls, fixit, file):
...
+            newvalue = (oldvalue[:fixit.range.start.column-1] +
+                        fixit.value +
+                        oldvalue[endindex:])
+            new_file = (file[:fixit.range.start.line-1] +
+                        type(file)(newvalue.splitlines(True)) +
+                        file[fixit.range.end.line:])
+        except LibclangError as err:
+            if 'libclang.so version' in str(err):
+                logging.error('coala requires clang 3.4')
+                raise err
```
Makman2: no that's not what I meant 😅 I meant doing something like this:
```
if libclang.version < 3.4:  # <-- this field probably is called differently
    raise LibclangError('libclang for clang 3.4')

# no try block
oldvalue = ...
...
```

coalib/results/SourceRange.py
```diff
@@ -54,11 +54,18 @@ def from_clang_range(cls, range):
...
+        try:
+            from_values = cls.from_values(range.start.file.name,
+                                          range.start.line,
+                                          range.start.column,
+                                          range.end.line,
+                                          range.end.column)
+        except LibclangError as err:
+            if 'libclang.so version' in str(err):
+                logging.error('coala requires clang 3.4')
+                raise err
```
Makman2: this double catching is redundant.

coalib/results/Diff.py
nvzard: We could have used the __version__ attribute but it is not present in clang.cindex module.
I'll try to find a solution to this problem :) Thanks for your suggestion.

nvzard: Current approach is based on this: https://github.com/coala/coala/issues/3847#issuecomment-345529019

Makman2: Btw we may want to put the check into module scope, so we directly raise an exception, but not sure. Up to you to decide it. But yeah John is right, if we move the dependency into the bear that would be awesome, but not task of this PR ;)

nvzard: @Makman2  I tried finding a way so that we can do version check using `if libclang.version < 3.4` but was unable to find one, should we stick with the try block?
Or do you want me to change `libclang-py3~=3.4.0` to `libclang-py3=3.4` in requirements.txt ?

nvzard:
Yeah, I am waiting for confirmation on the approach to this problem, will then write tests based on the approach :)

nvzard:
@Makman2  just adding the code in a try block is raising an assertion error? :/ 
https://ci.appveyor.com/project/coala/coala/build/1.0.10885/job/qvtdji2s1k4iakrf#L285

nvzard:
@shreyateeza Tests for existing code https://github.com/coala/coala/blob/master/tests/results/DiffTest.py are failing. I tried locally with a fresh git clone repo.

nvzard:
@shreab373 just adding the code in a try block is raising an assertion error in tests? :/
https://ci.appveyor.com/project/coala/coala/build/1.0.10885/job/qvtdji2s1k4iakrf#L285

Can you please help with this situation? I'm stuck here.
I'll write tests for this once I get approval from @Makman2  that this approach is correct.

Makman2:
> @Makman2 I tried finding a way so that we can do version check using something like
if libclang.version < 3.4:
but was unable to find one, should we stick with the try block?

GitHub doesn't let me answer on that comment directly, however:
If you can't find a way now it's okay, keep it in a `try` block. But if possible I would like to check it somehow at module-import-stage so people immediately know that things aren't going to work out in all cases. But if that's also too complicated, then nvm ;)

> Or do you want me to change libclang-py3~=3.4.0 to libclang-py3=3.4 in requirements.txt ?

No the current spec is already good, the problem is that if users might have a different version installed, that we don't crash completely but give him a hint how to fix it, which is what you're doing right now :)

nvzard:
@Makman2  How do I test a piece of code which executes only when different version of a particular module is installed on the build?
Lasse suggested to use `# pragma: no cover` as the test would be way more work than necessary for such a small thing. He even suggested to mock `get_distribution` but that would 
very tedious and complicated. 
What's your opinion?

sils:
please note that the smallness doesn't matter but the impact of the code multiplied by the likelyhood of it containing bugs. This piece is very simple, easily reviewed and even if it breaks almost nobody will care so the overall risk isn't high.

ishanSrt:
If your `if` clause is never false, you can add this workaround: add another parameter to the function and set its default value to be `get_distribution('libclang-py3').version`, test this against `'3.4.0'`
and then in the test case give any other random value to that param. 
cc @Makman2 your thoughts on this?

Makman2:
Use `unittest.patch`, then you can modify any function you like to throw exceptions / doing whatever on certain conditions :+1:

ishanSrt:
@nvzard did you run `pytest --cov --cov-report html`? and look which lines of code have not been executed or partially executed?

tests/results/SourceRangeTest.py
```diff
@@ -2,6 +2,9 @@
 from collections import namedtuple
 from os.path import abspath
 
+from clang.cindex import LibclangError
+from pkg_resources import get_distribution
```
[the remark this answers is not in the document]
ishanSrt: listen to the PyUnusedCodeBear 😀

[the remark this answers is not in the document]
nvzard: Trying to use the unused import :p

coalib/results/Diff.py
```diff
@@ -166,6 +168,9 @@ def from_clang_fixit(cls, fixit, file):
         """
         assert isinstance(file, (list, tuple))
 
+        if get_distribution('libclang-py3').version != '3.4.0':
+            raise LibclangError('coala requires clang 3.4')
```
Makman2: I would like to see this at module scope, this may be a significant performance hit for LocalBears using this^^

coalib/results/SourceRange.py
```diff
@@ -54,6 +56,9 @@ def from_clang_range(cls, range):
 
         :param range: A cindex.SourceRange object.
         """
+        if get_distribution('libclang-py3').version != '3.4.0':
+            raise LibclangError('coala requires clang 3.4')
```
Makman2: See above

Makman2:
unack c1b74b89bf34987f19ca277a6d7d5d5f9abc88eb

coalib/results/Diff.py
```diff
@@ -2,14 +2,24 @@
...
 from coalib.results.TextRange import TextRange
 from coala_utils.decorators import enforce_signature, generate_eq
 
 
+def ClangVersionCheck():
+    if get_distribution('libclang-py3').version != '3.4.0':
+        raise LibclangError('coala requires clang 3.4')
+
+
+ClangVersionCheck()
```
Makman2: you can inline it, so just put here what's inside the `ClangVersionCheck()` function (which doesn't meet btw our naming scheme ;) --> `def check_clang_version`)

coalib/results/Diff.py
```diff
@@ -2,14 +2,24 @@
...
 
+from clang.cindex import LibclangError
 from coalib.results.LineDiff import LineDiff, ConflictError
 from coalib.results.SourceRange import SourceRange
 from coalib.results.TextRange import TextRange
 from coala_utils.decorators import enforce_signature, generate_eq
 
 
+def ClangVersionCheck():
+    if get_distribution('libclang-py3').version != '3.4.0':
```
Makman2: and above 3.4?

coalib/results/SourceRange.py
```diff
@@ -1,11 +1,21 @@
...
 from coalib.results.TextRange import TextRange
 from coalib.results.AbsolutePosition import AbsolutePosition
 
 
+def ClangVersionCheck():
+    if get_distribution('libclang-py3').version != '3.4.0':
+        raise LibclangError('coala requires clang 3.4')
+
+
+ClangVersionCheck()
```
Makman2: same here, please inline

tests/results/DiffTest.py
```diff
@@ -3,10 +3,13 @@
...
 import unittest.case
 
+from clang.cindex import LibclangError
 from unidiff.errors import UnidiffParseError
+from unittest.mock import patch
 
 from coalib.output.JSONEncoder import create_json_encoder
-from coalib.results.Diff import ConflictError, Diff, SourceRange
+from coalib.results.Diff import ClangVersionCheck, ConflictError, Diff, \
+        SourceRange
```
Makman2: could you use braces instead of `\`?

coalib/results/Diff.py
Makman2: Ah you did it to test things easier... but should be possible I think to test it anyway even without a functino, there should be infos out there how to do it for such cases

Makman2:
unack eac5dbe9fd124770e643a317ed7c651f39f4a02b

tests/results/SourceRangeTest.py
```diff
@@ -42,6 +44,15 @@ def test_from_clang_range(self):
         compare = SourceRange.from_values('t.c', 1, 2, 3, 4)
         self.assertEqual(uut, compare)
 
+    @patch('coalib.results.SourceRange.get_distribution')
+    def test_clang_version_check(self, get_distribution):
+        import importlib
+        import coalib.results.SourceRange
+        with self.assertRaisesRegexp(LibclangError,
+                                     'coala requires clang 3.4.0'):
+            get_distribution.configure_mode(version='1.2')
```
ishanSrt: if this line is setting the libclang version to 1.2 shouldn't the assert statement be after this to actually raise the error?

ishanSrt: can you explain why are you trying to reload the module `coalib.results.Diff`

tests/results/DiffTest.py
```diff
@@ -568,6 +570,15 @@ def test_from_clang_fixit(self):
         clang_fixed_file = Diff.from_clang_fixit(fixit, file).modified
         self.assertEqual(fixed_file, clang_fixed_file)
 
+    @patch('pkg_resources.get_distribution')
+    def test_clang_version_check(self, get_distribution):
+        import importlib
+        import coalib.results.Diff
```
Makman2: 1. please from the imports to top
2. `from importlib import reload`

tests/results/SourceRangeTest.py
```diff
@@ -42,6 +44,15 @@ def test_from_clang_range(self):
         compare = SourceRange.from_values('t.c', 1, 2, 3, 4)
         self.assertEqual(uut, compare)
 
+    @patch('pkg_resources.get_distribution')
+    def test_clang_version_check(self, get_distribution):
+        import importlib
+        import coalib.results.SourceRange
```
Makman2: See above

Makman2:
unack 9c5ba6be5ed9f944a4b4603d63d4201b38202c06

Makman2:
ack 86575f308b478de4ed41bf36105629fb9c2bd616

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

Issue #4886, opened 2017-11-19
No tests for coala --ignore


userzimmermann:
Recognized while creating https://github.com/coala/coala/pull/4883

There should be added some tests for this cmdline option. This is its definition:

https://github.com/coala/coala/blob/4e0c88e5d5e575e6fd20b5abb437d0ef54cb6077/coalib/parsing/DefaultArgParser.py#L148-L150

Thekiddiejsandpython:
Hi! I'm a coala newbie 😄 Do I make a fork, then make a pull request? Thanks 🍦 🕶  😀

Thekiddiejsandpython:
I'm doing a task 4 GCI - google code-in

Vedarth:
I would like to work on this issue. Kindly assign it to me. 😄

userzimmermann:
@Vedarth Thanks for your interested! But please take a look @ https://github.com/coala/coala/pull/4930#pullrequestreview-90337947 before starting work on this

cc @Makman2

anctartica:
@Vedarth are you still working on this?

Vedarth:
@anctartica Yes I am.

Vedarth:
@userzimmermann Thanks for your advice. It was really helpful. In https://github.com/coala/coala/pull/4930#pullrequestreview-90337947 you were against using `LineCountTestBear`. Apparently all we need to check is only
```
Executing section cli...
```
should get printed. So should I use some other testbear or just by checking the `stdout`?

PR #4930, opened 2017-11-30
tests/coalaTest.py: Add test case for ignore flag

AVturbine:
Adds a temporary file, then immediately ignores it.
If everything goes right, coala should do nothing.

Fixes https://github.com/coala/coala/issues/4886

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


tests/coalaTest.py
```diff
@@ -323,3 +323,16 @@ def test_coala_without_color(self):
...
+
+    def test_coala_ignore_file(self):
+        with bear_test_module(), \
+                prepare_file(['#fixme  '], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                    coala.main, 'coala',
+                    '--non-interactive',
+                    '-f', re.escape(filename),
+                    '-i', re.escape(filename),
+                    '-b', 'ErrorTestBear')
```
Mixih: Don't use `ErrorTestBear`. This bear is designed to test coala handling of errors due to missing runtime deps. Instead please use `LineCountTestBear` and check stdout to make sure no output is found for that file.

tests/coalaTest.py
```diff
@@ -323,3 +323,16 @@ def test_coala_without_color(self):
...
+            self.assertIn(
+                    "No files matching '{}' were found.".format(filename),
+                    stderr)
```
Mixih: also assert that the expected output is not found in stdout. if you make the previous change, that would be `'This file has 1 lines.'`

Mixih: Assert that the retval is 0.

tests/coalaTest.py
```diff
@@ -326,13 +326,14 @@ def test_coala_without_color(self):
...
+                    '-c', os.devnull,
                     '--non-interactive',
                     '-f', re.escape(filename),
                     '-i', re.escape(filename),
-                    '-b', 'ErrorTestBear')
-            self.assertIn(
-                    "No files matching '{}' were found.".format(filename),
-                    stderr)
+                    '-b', 'LineCountTestBear')
+            self.assertNotIn(stdout, 'This file has''lines')
```
adtac: Is the second argument right? It just amounts to `'This file haslines'`

Mixih: should be `this file has 1 lines`

newbazz: @Mixih then howcome the tests passed?

AVturbine: Idk how it passes, I'll have to take a look.

AVturbine: Oh nevermind, if the ignore flag doesn't work, the bear prints `this file has 2 lines`, since the assert is looking for `this file haslines`, it still passes. Sorry about that.

AVturbine:
Hope this works. Like 15 mins left on the clock. Squashed my commits down into one.

Makman2:
ack 94098a5

Makman2:
Sorry it took so long @AVturbine , we are lacking manpower :/

Makman2:
@gitmate-bot rebase

tests/coalaTest.py
```diff
@@ -322,3 +322,17 @@ def test_coala_without_color(self):
...
+
+    def test_coala_ignore_file(self):
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                    coala.main, 'coala',
+                    '-c', os.devnull,
+                    '--non-interactive',
+                    '-f', re.escape(filename),
+                    '-i', re.escape(filename),
```
Makman2: Ah there were recent changes in our parsing on Windows: You can leave the `re.escape` calls, just `filename` is enough :)

Makman2:
unack 845ffb1

tests/coalaTest.py
```diff
@@ -322,3 +322,17 @@ def test_coala_without_color(self):
...
+                    '-b', 'LineCountTestBear')
+            self.assertNotIn(stdout, 'This file has 1 lines')
```
userzimmermann: You can't rely on this. The output of `LineCountTestBear` might change in the future. Then it won't get recognized in this test if for some reason the ignore flag doesn't work correctly anymore

Without ignore, `stdout` is currently like:

```
Executing section cli...

..\..\..\..\zimmer~1\appdata\local\temp\tmp0v9kz559
**** LineCountTestBear [Section: cli | Severity: INFO] ****
!    ! This file has 1 lines.
```

Which can also change in the future. So it's generally not a good idea to only explicitly check for some specific text to not appear in `stdout`

With ignore `stdout` is just:

```
Executing section cli...
```

So better explicitly check that nothing more than the latter appears in `stdout`

@Makman2

tests/coalaTest.py
userzimmermann: You need a test for the long `--ignore` flag as well

And as @Makman2 told: No `re.escape` of paths anymore. https://github.com/coala/coala/pull/4883 made that obsolete, and using `re.escape` for that purpose was anyway not a good idea ;)

Issue #4887, opened 2017-11-19
No tests for coala --limit-files


userzimmermann:
Recognized while creating https://github.com/coala/coala/pull/4883

There should be added some tests for this cmdline option. This is its definition:

https://github.com/coala/coala/blob/4e0c88e5d5e575e6fd20b5abb437d0ef54cb6077/coalib/parsing/DefaultArgParser.py#L152-L154

PR #4942, opened 2017-12-05
coalaCITest.py: Add --limit-files test

LifetimeAlone:
coalaCITest.py: Add --limit-files test

Add --limit-files setting
unit tests

Closes https://github.com/coala/coala/issues/4887

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/tests/coalaCITest.py
+++ b/tests/coalaCITest.py
@@ -137,3 +137,41 @@ class coalaCITest(unittest.TestCase):
                 r'were not provided.'
         ):
             self.test_fail_acquire_settings(debug=True)
+
+    def test_limit_files_affirmative(self):
+        sample_text = '\t#include <a>'
+        with open('match.cpp', 'w') as match, \
+                open('noMatch.cpp', 'w') as no_match:
+            match.write(sample_text)
+            no_match.write(sample_text)
+        with bear_test_module():
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '--non-interactive',
+                '-c', os.devnull,
+                '--limit-files', 'match*',
+                '-f', 'match.cpp', 'noMatch.cpp',
+                '-b', 'SpaceConsistencyTestBear',
+                '--settings', 'use_spaces=True')
+            os.remove('match.cpp')
+            os.remove('noMatch.cpp')
+            self.assertIn('match.cpp', stdout)
+            self.assertNotIn('noMatch.cpp', stdout)
+            self.assertIn("Applied 'ShowPatchAction'", stderr)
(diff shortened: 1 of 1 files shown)


tests/coalaCITest.py
```diff
@@ -137,3 +137,17 @@ def test_fail_acquire_settings_debug(self):
...
         ):
             self.test_fail_acquire_settings(debug=True)
+
+    def test_limit_files(self):
+        with bear_test_module(), \
+             prepare_file(['\t#include <a>'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '--non-interactive',
+                '-c', os.devnull,
+                '--limit-files', 'some_file_name',
```
adtac: shouldn't this be `filename` so that it actually tests whether or not `--limit-files` is considered?

also, there should be an assert checking if the `\t` was reported; it should be converted to spaces because of `use_spaces=True`, so there will be some logging to console

LifetimeAlone: No, it shouldn't be filename AFAIK. The test shows, that if `-f` args do not match `--limit-files` pattern, files won't be put into coala

LifetimeAlone: This test is only about `--limit-files`, `\t` report is being checked in different tests

adtac: Ah, my mistake.

One tiny review: it's `--limit-files`, not `--limit--files` in your commit shortlog :)

jayvdb: You are not testing that `--limit-files` does what people use it for.
The only conclusion that is certain from this test, alone, is that using `--limit-files` results in zero results.

tests/coalaCITest.py
```diff
@@ -137,3 +137,41 @@ def test_fail_acquire_settings_debug(self):
                 r'were not provided.'
         ):
             self.test_fail_acquire_settings(debug=True)
+
+    def test_limit_files(self):
+        sample_text = '\t#include <a>'
+        match = open('match.cpp', 'w')
```
jayvdb: use `with open(...):` syntax

jayvdb: actually I think we also have a temporary file context manager, which is used by other tests.

LifetimeAlone: temporary file context manager doesn't support named files, AFAIK, it creates a temp file only if `name=None`

jayvdb:
ack https://github.com/coala/coala/pull/4942/commits/c959af3b8af76a154ae983adb7bfc204a3662d31

Issue #4560, opened 2017-07-26
Make `bear.BEAR_DEPS` instance bound


Makman2:
Originated from https://github.com/coala/coala/issues/4436

PR #4970, opened 2017-12-14
core: Make bear dependencies instance bound

Makman2:
Bears can now define dependencies at instantiation for `BEAR_DEPS`,
depending on different section settings.

The Bear class makes now a copy of the dependencies specified in
`BEAR_DEPS`, so users can directly modify the set inside `__init__`
(which prevents modifying the class wide dependencies that are always
needed).

For that, the core test bears had to be adapted, because before
picklibility of those test bears was not necessary to run the
core.

Closes https://github.com/coala/coala/issues/4560

### DIFF
--- a/tests/core/BearTest.py
+++ b/tests/core/BearTest.py
@@ -26,6 +26,10 @@ class Bear2(Bear):
     pass
 
 
+class BearWithDependencies(Bear):
+    BEAR_DEPS = {Bear1}
+
+
 class BearWithAnalysis(Bear):
 
     def analyze(self, x: int, y: int, z: int=33):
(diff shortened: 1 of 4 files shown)


tests/core/BearTest.py
```diff
@@ -254,3 +258,18 @@ def test_json(self):
             'REQUIREMENTS': set()}
 
         self.assertEqual(result, expected)
+
+    def test_class_bear_deps_immutability(self):
+        section = Section('test-section')
+        uut = BearWithDependencies(section, {})
+
+        self.assertNotEqual(id(uut.BEAR_DEPS),
```
Udayan12167: Are you trying to check if belong to possibly the same memory address with id?

Makman2: yes

Makman2: people shall be able to modify `self.BEAR_DEPS` without modifying `cls.BEAR_DEPS`, so a copy is made. To distinct the lists, I use `id` here.

Makman2: Oh my bad, why didn't I thought of using `assertIs`/`assertIsNot`...

Makman2:
ack 438156512456bd3eb50a9f3b635a59c33392a3e1

Makman2:
@gitmate-bot ff

Issue #4928, opened 2017-11-30
Don't require a `--files` flag if we're running only global bears


Makman2:

newbazz:
running global bears means not specifing the `--bears` param?

Makman2:
No. Some bears are so called `GlobalBear`s. They operate on the whole repo instead of individual files. However, some `GlobalBear`s don't process files at all (like the `GitCommitBear`), and those wouldn't need files to run.

ishanSrt:
@Makman2 is this bear specific or absolutely all global bears don't require a `--files` flag

ishanSrt:
@Makman2 I am assuming my previous comment, and I am opening a PR. Also assign this issue to me. If only global bears are being run and only the files argument is not provided, coala will run!

Makman2:
> @Makman2 is this bear specific or absolutely all global bears don't require a --files flag

It's bear specific.

Makman2:
Maybe we shouldn't require `--files` in general and assume "no files" by default? Not sure though...

PR #4997, opened 2017-12-21
No need of --files flag with only global bears

ishanSrt:
Fixes https://github.com/coala/coala/issues/4928


ishanSrt:
@pradeepgangwar be more precise.

pradeepgangwar:
quoting it from the link i mentioned in previous comment.
> A commit message consists of 3 parts:
>  - shortlog
>  - commit body
>  - issue reference

your commit message should follow this style.

ishanSrt:
@pradeepgangwar commit body is optional only to describe the things that are not self explanatory. Read that entire doc, carefully.

coalib/settings/ConfigurationGathering.py
```diff
@@ -492,8 +500,12 @@ def gather_configuration(acquire_settings,
     aspectize_sections(sections)
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings)
+    only_global_bears = True;
+    for sec, value in local_bears.items():
+        if value:
+            only_global_bears = False;
     save_sections(sections)
-    warn_nonexistent_targets(targets, sections)
+    warn_nonexistent_targets(targets, sections, only_global_bears)
```
pradeepgangwar: I think you should add global bears option to `gather_configuration` function too, and then see build log, they fail because of this.

ishanSrt: what is meant by global bears option?

ishanSrt: the checks are failing because of the test cases. all bears till now have to have a `--files` tag. now I am changing that so the test yell at me why didn't coala quit when the `--files` was not specified

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,13 +182,19 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
 
     # Can't be summarized as python will evaluate conditions lazily, those
     # functions have intended side effects though.
-    files_config_absent = warn_config_absent(sections, 'files')
```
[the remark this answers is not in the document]
ishanSrt: already made those changes

ishanSrt:
Gitmate is still showing outdated suggested changes. Ignore them

ishanSrt:
@Makman2 review needed

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,13 +182,19 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
...
-    if files_config_absent or bears_config_absent:
+    files_config_absent, arg_files = warn_config_absent(sections,
+                                                        'files',
+                                                        only_global_bears)
+    bears_config_absent, arg_bears = warn_config_absent(sections,
+                                                        ['bears', 'aspects'],
+                                                        only_global_bears)
+    if (files_config_absent or bears_config_absent) and (
+        not (only_global_bears and files_config_absent
+             and arg_files == ['files'])):
```
khanchi97: @ishanSrt if `files_config_absent` is `True` then this condition will be always `false`

ishanSrt: I don't think thats gonna happen. if other conditions like `only_global_bears` and `arg_files == ['files']` turn false

coalib/settings/ConfigurationGathering.py
```diff
@@ -201,12 +210,13 @@ def warn_config_absent(sections, argument, log_printer=None):
...
 
-    formatted_args = ' or '.join('`--{}`'.format(arg) for arg in argument)
-    logging.warning('coala will not run any analysis. Did you forget '
-                    'to give the {} argument?'.format(formatted_args))
-    return True
+    if not (only_global_bears and argument == ['files']):
+        formatted_args = ' or '.join('`--{}`'.format(arg) for arg in argument)
+        logging.warning('coala will not run any analysis. Did you forget '
+                        'to give the {} argument?'.format(formatted_args))
+    return True, argument
```
khanchi97: why you are returning `argument` because you are not using it except in tests?

ishanSrt: `argument` tells me what is actually missing. if that is `== ['files']` only then (of course with only global bears) there should not be a system exit

coalib/settings/ConfigurationGathering.py
khanchi97: oh yeah right :)

coalib/settings/ConfigurationGathering.py
khanchi97: I meant to ask only why you are returning `argument`.What is the use of returning this?

ishanSrt: @khanchi97 I just told you 😄 that. If any argument is not present, I want the value of that and if that is equal to `['files']` and I am using only global bears (so that I can check its value collectively later on, on line 191)I will not raise system error

ishanSrt: if I am unable to explain please ask more specifically or explain how can I achieve the same results without returning that argument 🤔

khanchi97: Oh, you are using it in line 191.

ishanSrt:
after this merge a successive issue needs to be opened to provide a warning that coala ran on nothing 👍

khanchi97:
>after this merge a successive issue needs to be opened to provide a warning that coala ran on nothing 

We can open this issue now and add status `blocked` until this PR get merged.

coalib/settings/ConfigurationGathering.py
```diff
@@ -162,14 +162,17 @@ def save_sections(sections):
...
     not existent within the given sections.
 
-    :param targets:     The targets to check.
-    :param sections:    The sections to search. (Dict.)
-    :param log_printer: The log printer to warn to.
+    :param targets:           The targets to check.
+    :param sections:          The sections to search. (Dict.)
+    :param log_printer:       The log printer to warn to.
+    :param only_global_bears: True if only global bears are
+                              are being run
```
Makman2: I would order the parameters like defined in the signature ;)

coalib/settings/ConfigurationGathering.py
```diff
@@ -201,12 +210,13 @@ def warn_config_absent(sections, argument, log_printer=None):
         argument = [argument]
     for section in sections.values():
         if any(arg in section for arg in argument):
-            return False
+            return (False, argument)
```
Makman2: 1. Redundant braces
2. Return value changed, please update the docstring.

But this kinda looks wrong to change the return value here, I mean `argument` was given before to the function, right? No need to return it again.

coalib/settings/ConfigurationGathering.py
```diff
@@ -201,12 +210,13 @@ def warn_config_absent(sections, argument, log_printer=None):
...
 
-    formatted_args = ' or '.join('`--{}`'.format(arg) for arg in argument)
-    logging.warning('coala will not run any analysis. Did you forget '
-                    'to give the {} argument?'.format(formatted_args))
-    return True
+    if not (only_global_bears and argument == ['files']):
+        formatted_args = ' or '.join('`--{}`'.format(arg) for arg in argument)
+        logging.warning('coala will not run any analysis. Did you forget '
+                        'to give the {} argument?'.format(formatted_args))
```
Makman2: Will coala still run analysis, even though this message doesn't appear if only global bears given?

Makman2:
unack 6065b06041944b9206e27518fd0ddbc07fec896e

coalib/settings/ConfigurationGathering.py
ishanSrt: does it matter? as no files are provided. is there a difference between coala running analysis on no files and coala running no analysis at all? @Makman2

ishanSrt:
@Makman2 made the changes 👍 
As for your concern whether coala is still running analysis I am pretty sure it is still doing that. 
Confirmed by adding random print statements in `PythonPackageInitBear`. and running that without the `--files` option although couldn't test it for `GitCommitBear` due to https://gitter.im/coala/coala?at=5a6426026117191e61912dbd

coalib/settings/ConfigurationGathering.py
```diff
@@ -162,14 +162,17 @@ def save_sections(sections):
...
     Prints out a warning on the given log printer for all targets that are
     not existent within the given sections.
 
-    :param targets:     The targets to check.
-    :param sections:    The sections to search. (Dict.)
-    :param log_printer: The log printer to warn to.
+    :param targets:           The targets to check.
+    :param sections:          The sections to search. (Dict.)
+    :param only_global_bears: True if only global bears are
+                              are being run
```
Makman2: Missing period

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,13 +182,16 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
...
     # functions have intended side effects though.
-    files_config_absent = warn_config_absent(sections, 'files')
-    bears_config_absent = warn_config_absent(sections, ['bears', 'aspects'])
-    if files_config_absent or bears_config_absent:
+    files_config_absent = warn_config_absent(sections, 'files',
+                                             only_global_bears)
+    bears_config_absent = warn_config_absent(sections, ['bears', 'aspects'],
+                                             only_global_bears)
+    if (files_config_absent or bears_config_absent) and (
+            not (only_global_bears and files_config_absent)):
```
Makman2: Could you explain me this addition? Just curious what it does^^ Can't get behind it^^

coalib/settings/ConfigurationGathering.py
```diff
@@ -203,9 +209,10 @@ def warn_config_absent(sections, argument, log_printer=None):
         if any(arg in section for arg in argument):
             return False
 
-    formatted_args = ' or '.join('`--{}`'.format(arg) for arg in argument)
-    logging.warning('coala will not run any analysis. Did you forget '
-                    'to give the {} argument?'.format(formatted_args))
+    if not (only_global_bears and argument == ['files']):
```
Makman2: maybe `... and 'files' in argument`, as `argument` may be a list of multiple args?

coalib/settings/ConfigurationGathering.py
```diff
@@ -252,7 +259,8 @@ def load_configuration(arg_list,
         user_sections = load_config_file(
             Constants.user_coafile, silent=True)
 
-        default_config = str(base_sections['default'].get('config', '.coafile'))
+        default_config = str(
+            base_sections['default'].get('config', '.coafile'))
```
Makman2: unnecessary change

coalib/settings/ConfigurationGathering.py
ishanSrt: Yes the first part of the condition i.e. what was already there before deals with, if either the files parameter or bears parameter is missing, it will raise system exit whereas what i have added is `and (not (only_global_bears and files_config_absent)):` not to raise system exit in case of of only global bears are absent and files parameter is absent

coalib/settings/ConfigurationGathering.py
ishanSrt: only `'files'` is passed as an argument on line 182/185 and its being converted to a list in line 201/207

coalib/settings/ConfigurationGathering.py
Makman2: I think here's a mistake. I tried to reduce the expression, and I got:
```python
(file_config_absent and not only_global_bears) or
(bears_config_absent and not files_config_absent)
```
And the 2nd term is a bit wondering me: If both are True, it depends on `only_global_bears`. Shouldn't we always log something if we don't give any bears at all?

coalib/settings/ConfigurationGathering.py
```diff
@@ -492,8 +499,16 @@ def gather_configuration(acquire_settings,
...
     local_bears, global_bears = fill_settings(sections,
                                               acquire_settings)
+
+    only_global_bears = False
+    for sec, value in global_bears.items():
+        if value:
+            only_global_bears = True
+    for sec, value in local_bears.items():
+        if value:
+            only_global_bears = False
```
Makman2: 1. You don't need `sec`, so instead of `items()`, you can iterate over `values()`.
2. You should break once `value == True`
3. You can use `any`:

```python
only_global_bears = any(global_bears.values()) and not any(local_bears.values())
```

Makman2:
unack e27b5a3c143ceb13d258257a10546ddf6923bb5b

coalib/settings/ConfigurationGathering.py
ishanSrt: @Makman2 if both are true, `only_global_bears` would be False, so an error would be logged

ishanSrt: and if files is given, and bears are not given, again, `only_global_bears` would be False and hence error would be logged

ishanSrt: all this due to added lines 502-509

ishanSrt: which are now on line 503 & 504

coalib/settings/ConfigurationGathering.py
ishanSrt: ^^^ the line numbers may be changed due to the changes I made

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,13 +182,16 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
 
     # Can't be summarized as python will evaluate conditions lazily, those
     # functions have intended side effects though.
-    files_config_absent = warn_config_absent(sections, 'files')
-    bears_config_absent = warn_config_absent(sections, ['bears', 'aspects'])
-    if files_config_absent or bears_config_absent:
+    files_config_absent = warn_config_absent(sections, 'files',
+                                             only_global_bears)
```
Makman2: maybe do
```
if not only_global_bears:
    warn_config_absent(sections, 'files)
```
Then we save the `only_global_bears` param for that function.

ishanSrt: @Makman2 but how will `warn_config_absent` know that only global bears are being run?? we need to suppress the warning message in `warn_config_absent` if `only_global_bears` is true!

ishanSrt: In case you don't want me to pass around `only_global_bears`

Makman2: but we don't have to call `warn_config_absent` in case `only_global_bears` is `True`, or not? I guess I'm missing something ;)

ishanSrt: now I get what you mean, in that case I can remove the condition inside the function `warn_config_absent`

ishanSrt: btw I don't think its making any difference 😶

ishanSrt: sorry, that way I won't have to pass around `only_global_bears` 👍

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,7 +182,8 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
 
     # Can't be summarized as python will evaluate conditions lazily, those
     # functions have intended side effects though.
-    files_config_absent = warn_config_absent(sections, 'files')
+    files_config_absent = warn_config_absent(
+        sections, 'files') if not only_global_bears else false
```
Makman2: it's `False`, not `false` :)
Actually I'm wondering why tests don't fail

coalib/settings/ConfigurationGathering.py
```diff
@@ -162,14 +162,17 @@ def save_sections(sections):
     conf_writer.close()
 
 
-def warn_nonexistent_targets(targets, sections, log_printer=None):
+def warn_nonexistent_targets(targets, sections,
+                             only_global_bears, log_printer=None):
```
jayvdb: never add parameters in the middle of the list; that breaks the API.  Add them to the end.

ishanSrt: @jayvdb but positional parameter should be before named parameters. Am I missing something over here?

Makman2: Maybe make it a keyword parameter and shift it to the end?
However, I think we have done deprecation efforts way too long. We are still having `0.x` versions, and as a maintainer I would claim our right to break API (as long as we properly mention it in the release notes) especially because we are officially still "alpha" ^^ Those internal things are likely to be not used outside, and people can still use the older version if they are too lazy to update. If we can pin package versions, others can do the same ;)
But @jayvdb that's a different discussion I'm kicking off here, your decision :)

ishanSrt: I will do it but can you explain me in what context are you all talking about? why does this cause API breakage?

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -25,6 +25,7 @@
     gather_configuration,
     get_filtered_bears,
     load_configuration,
+    warn_nonexistent_targets
```
jayvdb: add trailing comma, so the next person doesnt need to modify your line.

coalib/settings/ConfigurationGathering.py
```diff
@@ -162,14 +162,17 @@ def save_sections(sections):
...
     """
     Prints out a warning on the given log printer for all targets that are
     not existent within the given sections.
 
-    :param targets:     The targets to check.
-    :param sections:    The sections to search. (Dict.)
-    :param log_printer: The log printer to warn to.
+    :param targets:           The targets to check.
+    :param sections:          The sections to search. (Dict.)
+    :param only_global_bears: True if only global bears are
```
jayvdb: wrong position in params order

coalib/settings/ConfigurationGathering.py
```diff
@@ -179,7 +182,8 @@ def warn_nonexistent_targets(targets, sections, log_printer=None):
 
     # Can't be summarized as python will evaluate conditions lazily, those
     # functions have intended side effects though.
-    files_config_absent = warn_config_absent(sections, 'files')
+    files_config_absent = warn_config_absent(
```
jayvdb: all of the rest of this function is useless if `only_global_bears`, because that implicitly says there is at least one bear, so `bears_config_absent` can never be `True`.

Bail out early if `bears_config_absent` is True.

ishanSrt: i see, all of this function is useless, cause we dont require files and we already know that there is atleast one bear provided by the user if `only_global_bears` is true

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -480,3 +481,51 @@ def test_get_filtered_bears(self):
...
+        except SystemExit:
+            raised = True
+        self.assertFalse(raised)
+
+    def test_gather_configuration_with_only_global_bears(self):
+        args = (lambda *args: True, self.log_printer)
+        sections, local_bears, global_bears, targets = (
+            gather_configuration(
+                *args,
+                arg_list=['-b PyromaBear']))
```
jayvdb: use a dummy bear, not a real one

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -480,3 +481,51 @@ def test_get_filtered_bears(self):
...
+        print(global_bears)
```
jayvdb: test locally, not as a pull request :P

Issue #4957, opened 2017-12-09
--apply-patches and ApplyPatchAction does not work as expected


blazeu:
# Steps to Reproduce

1. Using https://github.com/coala/gci-leaders codebase
1. With the following `.coafile` (minimal version of what's in the repo)
```
[all]
files = lib/*.js, static/js/*.js
eslint_config = .eslintrc
bears = ESLintBear
```
2. Run `coala --apply-patches`

# Actual Results
The problem was not fixed automatically, instead it's asking what to do, like the usual `coala` run.

# Expected Results
The problem is fixed automatically without user's intervention.

# Additional Information:
```
❯ coala --version         
0.12.0.dev20171202193440
```

- `default_actions = *: ApplyPatchAction` also does not work.
- Patching with `coala` using interactive CLI **does work as expected**.

Makman2:
When running over the codebase of `gci-leaders` I don't get any errors. Could you post a file where `ESLintBear` fails?

blazeu:
Oh, I forgot to explain.
I create my own error to test.

You can add semicolon in any of the js files for example.

Makman2:
Confirmed.

shreyateeza:
@blazeu or @Makman2 Can anyone please explain me more in detail?

Makman2:
☝️ https://github.com/coala/coala/issues/4957#issue-280720415
I think the report is quite understandable :+1:

PR #4998, opened 2017-12-22
Fix --apply-patches not working

ishanSrt:
Fixes https://github.com/coala/coala/issues/4957

### DIFF
--- a/tests/processes/ProcessingTest.py
+++ b/tests/processes/ProcessingTest.py
@@ -11,7 +11,7 @@ import unittest
 
 from pyprint.ConsolePrinter import ConsolePrinter
 
-from testfixtures import LogCapture, StringComparison
+from testfixtures import LogCapture, StringComparison, TempDirectory
 
 from coalib.bears.Bear import Bear
 from coalib.output.printers.LogPrinter import LogPrinter
@@ -22,6 +22,7 @@ from coalib.processes.Processing import (
     execute_section, filter_raising_callables, get_default_actions,
     get_file_dict, print_result, process_queues, simplify_section_result,
     yield_ignore_ranges)
+from coalib.results.Diff import Diff
 from coalib.results.HiddenResult import HiddenResult
 from coalib.results.Result import RESULT_SEVERITY, Result
 from coalib.results.result_actions.ApplyPatchAction import ApplyPatchAction
@@ -29,6 +30,7 @@ from coalib.results.result_actions.PrintDebugMessageAction import (
     PrintDebugMessageAction)
 from coalib.results.result_actions.ResultAction import ResultAction
 from coalib.results.SourceRange import SourceRange
+from coalib.results.SourcePosition import SourcePosition
 from coalib.settings.ConfigurationGathering import gather_configuration
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
@@ -437,7 +439,8 @@ class ProcessingTest(unittest.TestCase):
                    {'file1': [Result('a', 'b')], 'file2': None},
                    {'file3': [Result('a', 'c')]},
                    None)
-        yielded, yielded_unfixed, all_results = simplify_section_result(results)
+        yielded, yielded_unfixed, all_results = simplify_section_result(
+            results)
         self.assertEqual(yielded, True)
         self.assertEqual(yielded_unfixed, True)
         self.assertEqual(len(all_results), 2)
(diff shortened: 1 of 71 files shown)


ishanSrt:
the changes are made for obvious reason, the glob `*` will not match anything that has`/` or `\` whereas ESLintBear has `(prettier/prettier)` in it

Makman2:
Please add tests, so this doesn't sneak in again ;)

Makman2:
unack dcb0b55

ishanSrt:
@Makman2 I am having trouble making a test. Is there any test already present so I get a hang of most of the things
What I need to do is test out what the function: https://github.com/coala/coala/blob/2531a3d09f4239576f42803e03944244781b3181/coalib/processes/Processing.py#L106 return. I don't understand what should I provide for the parameter, `file_dict` as I see many tests giving file paths that don't exist and tests running fine for eg. https://github.com/coala/coala-quickstart/blob/50b358a60454932c0c33fe9443f305abd1103b11/tests/generation/SettingsTest.py#L47
Also I need to get the parameter `results` which I am not sure how am I gonna obtain that.
If I am using a dummy file then how do I obtain a dummy result object for that?
I see a dummy Result object being made here https://github.com/coala/coala/blob/2531a3d09f4239576f42803e03944244781b3181/tests/processes/ProcessingTest.py#L679 but I am unsure how do I make one so as to replicate the conditions in the https://github.com/coala/coala/issues/4957

ishanSrt:
@Makman2 would really appreciate a little bit of help here, although i know you're a bit busy.

Makman2:
Alright :) Some explanations on how things work in detail inside coala:
1. coala loads .coafile's (there's not only the one the user provides, but also system-wide coafiles and user-wide coafiles, but this is not so important)
2. coala reads all sections, and assigns the values to `Section` objects.
3. Some settings are special, like `files`, `bears` or `default_actions`, which control also coala-internal logic.
4. During coala's code flow, it will retrieve the value of `files` which is still a glob yet, and passes is to globbing functions to retrieve all paths of all files the user wants to analyze that exist on the current user's file tree.
5. For all file paths retrieved, coala now loads all files into memory, and places them into the "file-dict":
    ```python
    {'filepathA': ['content:\n', 'line2\n', 'line3\n'], ...}
    ```
    Note that at this stage, files have gone virtual, and coala won't read those files any more (at least they have no relevance for the bears after this).
6. coala now instantiates all requested bears, fills them up with all settings from the user (via the `Section` objects) and with the file-dict.
7. Those bear instances get scheduled into the core, so they are actually run.
8. The bears now emit `Result` objects, defining what codestyle-violations, hints, etc etc have been found.
9. coala collects these results and takes care of presenting them to the user or doing other result-handling (e.g. applying patches immediately)

So this is a little overview, trying to address your questions specifically:

> I don't understand what should I provide for the parameter, file_dict

It has to be at least a dict, mapping file-names/paths to the respective file contents. Like mentioned, files here are completely virtual, you can actually place whatever you like there. E.g.

```python
{'filepathA': ['content:\n', 'line2\n', 'line3\n']}
```
is referring to a set containing a single file, with the name `filepathA` containing
```
content:
line2
line3
```
This file doesn't have to exist for real in the OS filesystem. At the `apply_patches` function coala has already loaded real files into this dict.

> Also I need to get the parameter results which I am not sure how am I gonna obtain that.

This is a list of `Result` objects. Those are only holding information about an issue, like `message`, `severity`, etc. You can create a `Result` object to your likings, like said, it's just holding information. Those are returned by the bear, once they have analyzed files that were passed via the `file-dict`. In fact bears can be mathematically seen as a function performing this mapping:
one `file-dict` --> many `Result`s. So you can imagine an example run of coala with some unknown bear that has generated results for your `file-dict`. Having that said: You can provide any results you like, they can contain anything.

> I only need to know if the given function returns an empty list if i replicate the conditions in the issue

That's what I don't know, I haven't written the function :3

ishanSrt:
@Makman2 test added succesfully!

ishanSrt:
@nvzard I don't have wany idea what you are trying to tell me. Elaborate a little bit.

tests/processes/ProcessingTest.py
```diff
@@ -11,9 +11,11 @@
 
 from pyprint.ConsolePrinter import ConsolePrinter
 
-from testfixtures import LogCapture, StringComparison
+from testfixtures import LogCapture, StringComparison, TempDirectory
 
 from coalib.bears.Bear import Bear
+from coalib.output.ConsoleInteraction import (acquire_settings,
```
khanchi97: Hey, @ishanSrt I think this one is unused import :)

tests/processes/ProcessingTest.py
```diff
@@ -22,13 +24,17 @@
...
     get_file_dict, print_result, process_queues, simplify_section_result,
     yield_ignore_ranges)
+from coalib.results.Diff import Diff
 from coalib.results.HiddenResult import HiddenResult
 from coalib.results.Result import RESULT_SEVERITY, Result
 from coalib.results.result_actions.ApplyPatchAction import ApplyPatchAction
 from coalib.results.result_actions.PrintDebugMessageAction import (
     PrintDebugMessageAction)
 from coalib.results.result_actions.ResultAction import ResultAction
+from coalib.results.result_actions.GeneratePatchesAction import (
```
khanchi97: this one too ;)

tests/processes/ProcessingTest.py
ishanSrt: hmmm, travis pointed out one unused import to me, I wonder why it didn't point out these ones

ishanSrt:
@khanchi97 why did travis not point out those unused imports? is it not functioning properly? anyways made the requested changes

ishanSrt:
Sorry I messed something up, apparently I synced my local remote repository but Pull Request doesn't keep track of changes to the target branch, I am gonna have to open another PR for this 🤪

Makman2:
Please mention the PR link here so we can keep track :+1:

ishanSrt:
@Makman2 https://github.com/coala/coala/pull/5017

PR #5017, opened 2017-12-29
Fix --apply-patches not working

ishanSrt:
Change ApplyPatchAction glob from * to **.
Add a test case which ensures that ApplyPatchAction
works for ESLintBear.

Fixes https://github.com/coala/coala/issues/4957

### DIFF
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -249,7 +249,7 @@ To run coala without user interaction, run the `coala --non-interactive`,
 
     misc_group.add_argument(
         '-a', '--apply-patches', action='store_const',
-        dest='default_actions', const='*: ApplyPatchAction',
+        dest='default_actions', const='**: ApplyPatchAction',
         help='apply all patches automatically if possible')
 
     misc_group.add_argument(
--- a/coalib/results/result_actions/GeneratePatchesAction.py
+++ b/coalib/results/result_actions/GeneratePatchesAction.py
@@ -29,7 +29,7 @@ def create_arg_parser(files, bears):
     args = default_arg_parser().parse_args()
     args.files = files
     args.bears = bears
-    args.default_actions = '*: ApplyPatchAction'
+    args.default_actions = '**: ApplyPatchAction'
 
     return args
 


ishanSrt:
ping @khanchi97

tests/processes/ProcessingTest.py
```diff
@@ -816,3 +818,50 @@ def test_autoapply_override(self):
...
         self.assertEqual(newres, [])
+
+
+class ESLintBear_apply_patches(unittest.TestCase):
+
+    def setUp(self):
+        self.section = Section('name')
+        self.section.append(Setting('files', '*.js, **/*.js, **/**/*.js'))
+        self.section.append(Setting('eslint_config', '.eslintrc'))
+        self.section.append(Setting('bears', 'ESLintBear'))
```
Makman2: This requires `ESLintBear` to be installed, right? I would like to ask you to write a custom bear instead so coala keeps independent from `coala-bears`, even for testing.

tests/processes/ProcessingTest.py
```diff
@@ -816,3 +818,50 @@ def test_autoapply_override(self):
                                       self.section, self.log_printer, {}, [],
                                       console_printer=self.console_printer)
         self.assertEqual(newres, [])
+
+
+class ESLintBear_apply_patches(unittest.TestCase):
```
Makman2: 1. --> `ESLintBearApplyPatches`, don't mix casing conventions ;)
2. Also please append `Test` to the class name, I think it's not picked up otherwise^^

Makman2: EDIT:
Seems the test is picked up, anyway, please adapt to our naming scheme ;)

ishanSrt:
@Makman2 I added a custom bear. Tell me if its alright although I feel that I have done many things in the test case that were not required to test out the function of `autoapply_actions`. Most of it because earlier I did it to test it out for ESLintBear and now I added the custom bear which kind of mimics ESLintBear

ishanSrt:
Nevermind @Makman2 . I solved that. Review this and ignore gitmate as setting the value of origin with a `/` is necessary for this test

ishanSrt:
Nevermind @Makman2 I fixed that but now I am questioning myself whether a test like this is even needed? if I pass the glob `**`, `DummyBear (dummy/dummy)` will match it whereas if I pass `*`, it won't. It depends entirely on the user if he gives the correct file name. Same over here, it depends entirely on the user to give the correct glob. What I am testing over here is like `=`, i.e. a property of python. I guess instead of this I should rather provide a warning if a user provides the glob `*` saying some properties won't work unless you provide `**`

tests/processes/ProcessingTest.py
```diff
@@ -816,3 +818,104 @@ def test_autoapply_override(self):
                                       self.section, self.log_printer, {}, [],
                                       console_printer=self.console_printer)
         self.assertEqual(newres, [])
+
```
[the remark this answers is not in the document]
aashraybhandar1: @ishanSrt  The code does not comply with PEP8 standards. Kindly make the necessary change.

tests/processes/ProcessingTest.py
```diff
@@ -816,3 +817,37 @@ def test_autoapply_override(self):
...
+
+class apply_patchesTest(unittest.TestCase, Bear):
+
```
aashraybhandar1: Maybe a better way to write the class name might be ApplyPatchesTest

tests/processes/ProcessingTest.py
```diff
@@ -816,3 +817,37 @@ def test_autoapply_override(self):
...
+                                            lines.splitlines(True),
+                                            output.splitlines(True)))])
+            self.result = [self.result]
+            ret = autoapply_actions(self.result,
+                                    self.file_dict,
+                                    {},
+                                    self.section,
+                                    self.log_printer)
+
+        self.assertEqual(ret, [])
```
aashraybhandar1: Okay I have a couple of doubts can you help me clarify. The basic aim is to check if no output is printed with a js file having error when you are telling coala to fix the changes automatically. Like where is the test.js file containing the error :sweat_smile:  . I know this might be a very stupid doubt but if you could help me clarify it

Makman2:
@ishanSrt In fact that's right, however our default should be like this, so if the user relies on it no surprises happen. If the user enters it manually, it's his responsibility (although it's our fault that the syntax is a bit unexpecting here).

> I guess instead of this I should open another PR to provide a warning if a user provides the glob * in the .coafile saying some properties won't work unless you provide **

This PR is still good and we definitely want it. Indeed we can (and should) open an issue to discuss the behaviour with the user's intentions for this problem :+1:

ishanSrt:
@Makman2 but we still needed the changes in `DefaultArgParser.py` and `GeneratePatchesAction.py`
>So if a user provides a command line argument --apply-patches this would work while if someone writes in the .coafile default_actions = *: ApplyPatchAction
it simply won't work

Makman2:
My bad closing this PR, that was totally an accident.

ishanSrt:
@Makman2 This PR is complete unless of course you let me test it for ESLintBear 👍

Makman2:
ack bdc17102d246e9c87ba39e5f77fd36bce0b37008

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

Issue #5034, opened 2018-01-03
Ignore comments for more than one bear don't work properly


Makman2:
From https://github.com/coala/coala-bears/issues/2209:

> Even though the ignore statement has been added, the same bear is prompting.
> ```
> [   1] print(·'Hey'·)··//·Ignore·CPDBear··//·Ignore·PycodestyleBear·(E201)
> 
> Downloads/new.py
> [   2] print('Hey')
> **** CPDBear [Section: cli | Severity: NORMAL] ****
> !    ! Duplicate code found.
> [    ] *0. Do (N)othing
> [    ]  1. (O)pen file
> [    ]  2. Print (M)ore info
> [    ]  3. Add (I)gnore comment
> [    ] Enter number (Ctrl-D to exit): 3
> !    ! Please enter a value for the parameter 'language' (No description given.): CPP
> [    ] An ignore comment was added to your source code.
> ```
> And look what it did to the file
> ```
> print( 'Hey' )  // Ignore CPDBear  // Ignore PycodestyleBear (E201)  // Ignore CPDBear
> print('Hey')
> ```
> Two statements ignoring the same bear have been added.
> I think that coala is reading only the second ignore statement.

Makman2:
CC @sils 
If I recall correctly, you were involved into writing the ignore feature. Do you maybe have an idea?

sils:
no idea, sorry

khanchi97:
I wanna give it a try ;)

ishanSrt:
can't get PMD to work correctly right now, can you duplicate this with any other bear? also what exactly was your file before. @pareksha
cause
```
[   1] print(·'Hey'·)··//·Ignore·CPDBear··//·Ignore·PycodestyleBear·(E201)
```
this looks like a coala prompt

pareksha:
@ishanSrt take a look at this
```
tests/js/PrettierLintBearTest.py
|  14| foo(reallyLongArg(),·omgSoManyParameters(),·IShouldRefactorThis(),·isThereSeriouslyAnotherOne());··//·Ignore·PycodestyleBear··//·Ignore·LineLengthBear
|    | [NORMAL] PycodestyleBear (E501):
|    | E501 line too long (150 > 79 characters)
|    | *0: Do nothing
|    |  1: Open file(s)
|    |  2: Add ignore comment
|    | Enter number (Ctrl-D to exit):
```

pareksha:
@ishanSrt this was my file before running coala again
```
[...]

PR #5069, opened 2018-01-11
IgnoreResultAction.py: Fix ignore to use 'and'

pareksha:
This updates the 'Add Ignore Comment' setting
to insert 'and' conjunction for ignoring the second bear
if an ignore comment was already present for
another bear in the same line.

Fixes  https://github.com/coala/coala/issues/5034

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +57,27 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        if 'Ignore' in str(original_file_dict[filename][
+                    source_range.start.line-1].rstrip()) and 'Bear' in str(
+                    original_file_dict[filename][
+                        source_range.start.line-1].rstrip()):
+            ignore_comment = ignore_comment[
+                ignore_comment.index('Ignore'):].replace('Ignore', ' and')
```
ishanSrt: don't use `and`, use `comma` for now as it seems like `Processing.py` is only inferring comma separated values, line 476
```
return list(StringConverter(toignore, list_delimiters=', '))
```

pareksha: @ishanSrt both 'and' as well as ``comma`` working fine, let me ask @Makman2 regarding this

pareksha: I used 'and' because it's mentioned in the user's tutorial also.

ishanSrt: Sorry just checked https://docs.python.org/3/library/stdtypes.html & https://gitlab.com/coala/coala-utils/blob/master/coala_utils/string_processing/StringConverter.py . The strings will be treated as an iterable here.

pareksha: So both 'and' and ``comma`` will work right?

ishanSrt: Yes, they were not working for me earlier because of some changes I made on my branch 😀

ishanSrt: Sorry, later. I was solving an issue and saw that `and` wasn't working for me which reminded me that I had suggested the usage of `and` on this issue. So I asked to go for commas.

pareksha: @ishanSrt are you absolutely sure that 'and' wasn't working, please give the content in the file as well as the result.

ishanSrt: it wasn't working because of the changes I had made. its working on the master branch

pareksha: Okay, so 'and' will do right?

ishanSrt: yes

pareksha: Great, please review again and acknowledge, if satisfied. Thanks for the review. :smiley:

newbazz: dont you need to replace `Ignore` eith `and` and not with ` and`?

pareksha: @newbazz , either I have to put space before ``and`` ( as I did ) or I have to put space in ``ignore_diff.change_line`` ( after ``new_line`` and before ``ignore_comment`` ), both will work as ``Ignore abcBear and xyzBear``.There must be a space between ``abcBear`` and ``and``. :smile:

newbazz: just add comment of what this code does ;P

pareksha: @newbazz Do I have to add the comment in the code? I have already added comment in the test case though.

newbazz: In the code ;)

pareksha: @newbazz, changes made :smile: . Thanks for the review.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +57,31 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
         ignore_diff = Diff(original_file_dict[filename])
-        ignore_diff.change_line(
-            source_range.start.line,
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+        """
+        Apply the ignore comment using 'and' with the previous ignore,
+        if the same line already has a previous ignore comment.
+        """
```
Makman2: If this is a normal comment, rather use `#` instead of `"""`, as those are actually strings in Python.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +57,31 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+
+        if 'Ignore' in str(original_file_dict[filename][
+                    source_range.start.line-1].rstrip()) and 'Bear' in str(
+                    original_file_dict[filename][
+                        source_range.start.line-1].rstrip()):
```
Makman2: maybe rather do a regular expression match? Yeah I think that would be easier + better

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +57,31 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+            ignore_comment = ignore_comment[
+                ignore_comment.index('Ignore'):].replace('Ignore', ' and')
+            initial_line = original_file_dict[filename][
+                source_range.start.line-1].rstrip() + ' '
+            new_line = initial_line[:-(len(
+                initial_line) - initial_line.rfind('Bear') - len('Bear'))]
+            ignore_diff.change_line(
+                source_range.start.line,
+                original_file_dict[filename][source_range.start.line-1],
+                new_line + ignore_comment)
```
Makman2: And can't you use `re.sub` to make a replacement? Then you don't have to strip and strap and whatever :) You can reuse the regex mentioned above ;)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +57,31 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+        else:
+            ignore_diff.change_line(
+                source_range.start.line,
+                original_file_dict[filename][source_range.start.line-1],
+                original_file_dict[filename][
+                    source_range.start.line-1].rstrip() + '  ' + ignore_comment)
```
Makman2: In any way you are going to change a line here, right?
I would move the `change_line` part out of the if, and inside the if you only compute the patched line / ignore comment.

Makman2: So you would do (roughly):
```
if ignore_comment_existing:
    # assemble more complex ignore comment using 'and'
else:
    # assemble fist ignore comment

ignore_diff.change_line(..., line_with_ignore_comment)
```

Makman2:
unack a7baaef1c68afc4097c266e89a7e2d7f7b9543b9

coalib/results/result_actions/IgnoreResultAction.py
pareksha: @Makman2 this is a docstring specifying what the next lines of code will do.

pareksha:
Why has the gitmate bot labelled my PR as ``area/CLI`` ``difficulty/medium`` ``type/bug``. Aren't these labels for Issues?

Makman2:
We synchronize them from the respective issues :+1: Somebody might want to review only certain things in coala ;)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,23 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
-            source_range.start.line,
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        original_line = str(original_file_dict[filename][
+                        source_range.start.line-1].rstrip())
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if re.compile(r'(Ignore\s)(.+Bear)').search(original_line):
```
Makman2: you can use `re.search` directly instead of compiling it before ;) That's only useful if you keep the returned expression-object around for performance reasons ;)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,23 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+            ignore_comment = re.compile(r'(.+Ignore)\s(.+Bear).*').sub(
+                r'\2 and', ignore_comment).rstrip()
+            changed_line = re.compile(r'(Ignore\s)(.+Bear)').sub(
```
Makman2: It seems you're reusing nearly the same regex everywhere. Can you try to find a "master-regex" that you can unse everywhere?

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,23 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        original_line = str(original_file_dict[filename][
+                        source_range.start.line-1].rstrip())
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if re.compile(r'(Ignore\s)(.+Bear)').search(original_line):
+            ignore_comment = re.compile(r'(.+Ignore)\s(.+Bear).*').sub(
```
Makman2: Do we require that ignores need to end with `Bear`? I think it depends on the origin^^

Makman2:
unack a87ade1e9424b0e14787ef9b34f00c67f2e3f78f

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,21 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
         filename = source_range.file
 
         ignore_diff = Diff(original_file_dict[filename])
-        ignore_diff.change_line(
-            source_range.start.line,
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        original_line = str(original_file_dict[filename][
```
newbazz: It is better to use `original_line=re.compile()`

pareksha: Yes, it is but only when you need to create an object with the required pattern which you have to use further in your code. As here I am using the pattern only once, so it's better to use it without creating object. Bit of memory will be saved!

pareksha: @newbazz but I have to get the entire string. Don't need a regex to do that.

Makman2: Why are you calling `str` on a `str`-object? :O

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,21 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+                        source_range.start.line-1].rstrip())
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if re.search(r'Ignore.+Bear', original_line):
```
Makman2: From previous comment:
> Do we require that ignores need to end with `Bear`? I think it depends on the origin^^

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,21 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
+            changed_line = re.sub(r'(.+Ignore\s)(.+)', r'\1' +
+                                  result.origin + r' and \2', original_line)
+
+        # Add a new ignore comment if no previous ignore exists.
+        else:
+            changed_line = original_line + '  ' + ignore_comment
+
+        ignore_diff.change_line(source_range.start.line, original_line,
+                                changed_line)
```
Makman2: Shouldn't you maybe modify `self.get_ignore_comment` instead?

coalib/results/result_actions/IgnoreResultAction.py
pareksha: @Makman2 doesn't every bear has ``Bear`` at the end? If not, then I have to search only for ``Ignore`` in the line. Please convey.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -57,11 +58,21 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        original_line = str(original_file_dict[filename][
+                        source_range.start.line-1].rstrip())
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if re.search(r'Ignore.+Bear', original_line):
+            changed_line = re.sub(r'(.+Ignore\s)(.+)', r'\1' +
+                                  result.origin + r' and \2', original_line)
```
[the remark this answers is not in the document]
pareksha: Alright.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: Sorry. :face_with_head_bandage: Will correct it.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: I have searched https://coala.io/#/languages , every bear has a ``Bear`` with it. If I am missing something, do let me know :wink:

coalib/results/result_actions/IgnoreResultAction.py
pareksha: @Makman2 I have to just modify the previous comment. There is no need of getting a comment from the ``self.get_ignore_comment``. I only need the ``origin``. So, felt it best to modify ``self.apply``.

coalib/results/result_actions/IgnoreResultAction.py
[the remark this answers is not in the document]
pareksha: @Makman2 there is no need to use ``re.escape`` on ``result.origin`` because ``result.origin`` is a string, not a pattern, if I do 
```
In [2]: re.escape('Ignore PylintBear (W505)')
Out[2]: 'Ignore\\ PylintBear\\ \\(W505\\)'
```

coalib/results/result_actions/IgnoreResultAction.py
Makman2: I think modifying `get_ignore_comment` is the right thing to do. This function will return the ignore comment itself, and dependent on whether there were existing ignore comments, you assemble the line differently and generate the ignore comment differently (using `and` as a conjunction).

Makman2: Especially you use `get_ignore_comment` only in the case that there were no previous ignore comments, so it's currently a corner-case function :) And this shouldn't be the case ;)

Makman2:
Commit message: You have two spaces between `Fixes` and the issue link. Please use just one ;)

Makman2:
Sorry for taking so long, and that the review is so sparse, but I want to see the modifications on `get_ignore_comment` first before I further review, I expect that a few things change with that ;)

pareksha:
@Makman2 I have made the modifications on ``get_ignore_comment``. Please review and no need for sorry. ;)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -78,19 +89,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
 
         return file_diff_dict
 
-    def get_ignore_comment(self, origin, language):
+    def get_ignore_comment(self, origin, language, original_line):
```
Makman2: you can give `original_line` a default value of `''` I think, would make sense.

Makman2: Then you don't have to use `xyz` below, which is actually a bit confusing :)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -78,19 +89,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
 
-        >>> IgnoreResultAction().get_ignore_comment("Bear", "c")
+        >>> IgnoreResultAction().get_ignore_comment("Bear", "c", "xyz\n")
         '// Ignore Bear\n'
+
+        And Appending Comments
+
+        >>>
+        IgnoreResultAction().get_ignore_comment("New", "c", "// Ignore Bear\n")
+        'New and'
```
Makman2: Tbh, I don't understand this result 😅

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -78,19 +89,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
         """
+        # Searches for previous ignore comment
+        ignore_regex = re.compile(r'Ignore\s(\S+)')
+        ignore_search = ignore_regex.search(original_line)
+        if ignore_search and 'Bear' in ignore_search.group(1):
+            ignore_comment = origin + ' and'
+            return ignore_comment
```
Makman2: hm what about separating functionality a bit more:
`get_ignore_comment` (which imo should be renamed to `create_ignore_comment`, but that's a different issue :P), is responsible for creating the ignore comment. It also supports doing this for multiple bears, so `origin` --> `origins` and accepts a list (/any iterable) of origins to create the ignore comment for.

Then there's another function `read_ignore_comments` (which should actually be called `get_ignore_comments` :P), which extracts all existing ignore comments from a line. So when the action wants to add ignores, following happens:
1. `read_ignore_comments` is called, which returns all origins currently ignored (together with a `SourceRange`/`TextRange` to indicate where the comment is actually located)
2. Those origins are fed to `get_ignore_comment`, and returns the assembled new comment itself.
3. If `read_ignore_comments` did not return existing comments, the new comment is appended, otherwise the old one is replaced with the help of the positional information from `read_ignore_comments` with the new combined comment.

This is a bit more flexible and easier to maintain I believe, this would actually lay the foundation for better processing of ignore comments :)

Makman2: What do you think? :)

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -46,22 +47,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
         ignore_diff = Diff(original_file_dict[filename])
-        ignore_diff.change_line(
-            source_range.start.line,
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if ignore_comment == result.origin + ' and':
+            changed_line = re.sub(r'(Ignore\s)(\S+)', r'\1{} \2'.format(
```
nityeshaga: If 2 or more ignore comments are already present in the line, then it would make the comment look like - `Ignore ThisBear and ThatBear and ThatBear`. Don't you think it be better if the comment looked like -- `Ignore ThisBear, ThatBear and ThatBear`?

palash25: This was discussed here before: https://github.com/coala/coala/pull/5069#discussion_r161372051

I wonder why it wasn't added :thinking:   cc @pareksha

pareksha: I had discussed this with @Makman2 on coala/coala thread, he said 'and' would be alright as either we have to put ``,`` or ``and``. It will look like either ``Ignore ThisBear and ThatBear and ThatBear`` or ``Ignore ThisBear, ThatBear, ThatBear``. It can't be ``Ignore ThisBear, ThatBear and ThatBear`` as noone knows whether it's the last bear or there are yet more to come.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: Yes, you are right @Makman2, your suggestion is certainly better, but don't you think your suggested design restructures the design on which ``ignore_comments`` work whereas the functionality it would offer would be the same as I have done here. This issue is for modifying ``IgnoreResultAction`` to use conjuctions. I think a separate issue must be created which involves totally restructuring the ``IgnoreResultAction`` as per your suggestion. What do you say?

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -46,22 +47,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
...
-
         ignore_diff = Diff(original_file_dict[filename])
-        ignore_diff.change_line(
-            source_range.start.line,
-            original_file_dict[filename][source_range.start.line-1],
-            original_file_dict[filename][source_range.start.line-1].rstrip() +
-            '  ' + ignore_comment)
+
+        # Add the new ignore using 'and' with the previous ignore.
+        if ignore_comment == result.origin + ' and':
```
palash25: Since both `and` and `comma` work fine I think we should prefer `comma` over `and`.

coalib/results/result_actions/IgnoreResultAction.py
nityeshaga: Yes we don't know if its the last bear, so how about making the method behave a bit more intelligently -- initially it might have `Ignore ThisBear and ThatBear` and then when another bear comes, the message might be changed to `Ignore ThisBear, ThatBear and YetAnotherBear`. What do you think @pareksha ?

palash25: This is a naive approach that I'm suggesting but maybe we can count the number of ignores and according to that add combinations of `,` and `and`(s). For e.g. if just one ignore is present then we don't need a comma and just add an `and`. Otherwise replace `and` with `,` and add an `and` before the last ignore.

This might not be the optimal solution, but it might lead us to a better solution.

nityeshaga: I am not sure if counting the number of *ignore*s would work as we want the new ignore comment to have just one `Ignore` word in the comment. That being said, I think we may do something on similar lines and check for the absence/presence of an `and` in the comment. 
If present, we can replace it with a `,` and add an `and` to the end. 
Else, we can just add an `and` to the end.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: You are right @Makman2. I have made the changes accordingly, except that ``read_ignore_comments`` is being called by ``get_ignore_comment`` as I need the comment delimiter for the regex, otherwise it may give some unexpected behaviour.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -46,22 +47,32 @@ def apply(self, result, original_file_dict, file_diff_dict, language: str,
         """
         Add (I)gnore comment
         """
+        source_range = next(filter(lambda sr: exists(sr.file),
+                                   result.affected_code))
+        filename = source_range.file
 
-        ignore_comment = self.get_ignore_comment(result.origin, language)
+        original_line = original_file_dict[filename][
+                        source_range.start.line-1].rstrip()
```
jayvdb: this is incorrect indentation.  `source_range` is not at the same scope level as `original_file_dict`

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
                         language=language))
                 ignore_comment = None
         return ignore_comment
+
+    def read_ignore_comments(self, original_line, comment_delimiter):
+        r"""
```
jayvdb: why use `r` here?

tests/results/result_actions/IgnoreResultActionTest.py
```diff
@@ -101,6 +101,40 @@ def test_ignore(self):
...
+            }
+
+            file_diff_dict = {}
+
+            # Modify already existing patch
+            uut.apply(Result.from_values('origin', 'msg', f_a, 2),
+                      file_dict, file_diff_dict, 'python')
+            self.assertEqual(
+                file_diff_dict[f_a].modified,
+                ['1\n', '2  # Ignore KeywordBear,PEP8Bear and origin\n', '3\n'])
```
jayvdb: there should be a space before `PEP8Bear`

pareksha:
@jayvdb changes made. Please have a look.
Btw really happy to see those green ticks again. Thanks for the quick fix :smiley:

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
                         language=language))
                 ignore_comment = None
         return ignore_comment
+
+    def read_ignore_comments(self, original_line, comment_delimiter):
+        """
+        Reads Ignore Comment(s) already present in the line and
+        accordingly returns list of bears previously ignored
+        """
+
```
jayvdb: remove blank line here. It violates PEP8/PEP257

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
...
+        if comment_delimiter not in original_line:
+            return
+        else:
```
jayvdb: as you do a `return` above, this `else` is useless, except it causes unnecessary indentation.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
...
+            ignore_regex = re.compile(re.escape(comment_delimiter)
+                                      + r'\sIgnore(\s\S+)+')
+            ignore_comments = [ignore_comment.group() for ignore_comment in
+                               ignore_regex.finditer(original_line.rstrip())]
+            ignored_bears = []
+            bear_regex = re.compile(r'\S+Bear')
+            for ignore_comment in ignore_comments:
+                for bear in bear_regex.finditer(ignore_comment):
+                    ignored_bears.append(bear.group())
+            return ignored_bears
```
jayvdb: add a blank line before the return.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
                         language=language))
                 ignore_comment = None
         return ignore_comment
+
+    def read_ignore_comments(self, original_line, comment_delimiter):
+        """
+        Reads Ignore Comment(s) already present in the line and
```
jayvdb: docstrings should use imperative tone per PEP257; i.e. no plural `Reads` & `returns`.

End the imperative command with a full stop.

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
...
+    def read_ignore_comments(self, original_line, comment_delimiter):
+        """
+        Read Ignore Comment(s) already present in the line and
+        accordingly return a list of bears previously ignored.
+        """
+        if comment_delimiter not in original_line:
+            return
+
+        ignore_regex = re.compile(re.escape(comment_delimiter)
+                                  + r'\sIgnore(\s\S+)+')
```
jayvdb: this literal `Ignore` is problematic, as coala supports a variety of 'ignore' strings.

e.g. `start Ignoring foo`.

See `tests/processes/ProcessingTest.py` `test_yield_ignore_ranges`

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
...
+        ignore_comments = [ignore_comment.group() for ignore_comment in
+                           ignore_regex.finditer(original_line.rstrip())]
+        ignored_bears = []
+        bear_regex = re.compile(r'\S+Bear')
```
jayvdb: A bear name does not need to end in `Bear`.  See `tests/test_bears/TestBearDep.py`

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +140,23 @@ def get_ignore_comment(self, origin, language):
                         language=language))
                 ignore_comment = None
         return ignore_comment
+
+    def read_ignore_comments(self, original_line, comment_delimiter):
```
jayvdb: this method is not about the action.  It is a generic algorithm which should be part of the relevant module which does ignore handling.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: This functionality is for the ``Add Ignore Comment`` that coala adds itself, not for the Ignore statements we write. If we choose ``Add Ignore Comment`` option, coala adds ``somedelimiter + Ignore thisBear``

coalib/results/result_actions/IgnoreResultAction.py
pareksha: I placed it here as it is explicitly for ignores added by coala only. I felt this is the most relevant file. Else please direct me to the file.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: So every bear has a ``Bear`` in it, right? I will make the regex accordingly

coalib/results/result_actions/IgnoreResultAction.py
jayvdb: This PR is to fix it so it can *amend* existing Ignore comments correctly.
The existing Ignore comments could have been written by a human, or by coala's Ignore feature.  It doesnt matter.

coalib/results/result_actions/IgnoreResultAction.py
jayvdb: No, every bear does not have `Bear` in it.  We have functions which get the list of bears which have been loaded.  That list should be used.

Or better, assume all words are bear names here.  I dont think your algorithm here needs to restrict which words are valid bear names.

coalib/results/result_actions/IgnoreResultAction.py
pareksha: Yes, the existing ignore comments can be written by humans or by coala but the amending ignore statement is written by coala only. And coala writes ignore comments for each line separately whereas ``start ignoring foo`` is for multiline ignores. Right?

coalib/results/result_actions/IgnoreResultAction.py
jayvdb: No it isnt.  This is *action* code.  Any function which doesnt relate specifically to doing the action should be located elsewhere.  Find the modules which have code related to detecting, loading and processing ignore comments.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +141,33 @@ def get_ignore_comment(self, origin, language):
                         language=language))
                 ignore_comment = None
         return ignore_comment
+
+    def read_ignore_comments(self, original_line, start_comment,
```
jayvdb: https://github.com/coala/coala/pull/5069#discussion_r185209788 not done yet.

As discussed on private chat, maybe create a new module for new ignore-related code to live in.

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +141,33 @@ def get_ignore_comment(self, origin, language):
...
+
+        ignore_regex = re.compile(re.escape(start_comment)
+                                  + r'\s(\wgnore|\woqa)(\s\S+)+'
+                                  + re.escape(end_comment))
+        ignore_comments = [ignore_comment.group() for ignore_comment in
+                           ignore_regex.finditer(original_line.rstrip())]
+        ignored_bear_list = []
+
+        for ignore_comment in ignore_comments:
+            ignored_bears = ignore_comment.replace(',',
```
jayvdb: split this into several statements for easier reading

coalib/results/result_actions/IgnoreResultAction.py
```diff
@@ -115,3 +141,33 @@ def get_ignore_comment(self, origin, language):
...
+        accordingly return a list of bears previously ignored.
+        """
+        if start_comment not in original_line:
+            return
+
+        if end_comment is not '':
+            end_comment = ' ' + end_comment
+
+        ignore_regex = re.compile(re.escape(start_comment)
+                                  + r'\s(\wgnore|\woqa)(\s\S+)+'
```
jayvdb: this string should be in a module level constant.

coalib/parsing/inline_comments.py
```diff
@@ -0,0 +1,30 @@
...
+    accordingly return a list of bears previously ignored.
+    """
+    if start_comment not in original_line:
+        return
+
+    if end_comment is not '':
+        end_comment = ' ' + end_comment
+
+    ignore_regex = re.compile(re.escape(start_comment)
+                              + r'\s(\wgnore|\woqa)(\s\S+)+'
```
jayvdb: as previously noted at https://github.com/coala/coala/pull/5069#discussion_r186591327
this should be a module level constant

coalib/parsing/inline_comments.py
```diff
@@ -0,0 +1,30 @@
+import re
+
+
+def read_ignore_comments(original_line, start_comment, end_comment=''):
+    """
+    Read Ignore Comment(s) already present in the line and
```
jayvdb: Comment(s) -> comment .

This algorithm has no support for multiple comments on the same line.

coalib/parsing/inline_comments.py
```diff
@@ -0,0 +1,30 @@
...
+    accordingly return a list of bears previously ignored.
```
jayvdb: use `:return:` and also add metadata for the parameters that are used by sphinx to generate documentation.

coalib/parsing/inline_comments.py
```diff
@@ -0,0 +1,30 @@
+import re
```
jayvdb: and Makman2 [decided](https://gitter.im/coala/coala?at=5af2e45540f24c4304647b52) on `comments` for the module name

coalib/parsing/inline_comments.py
pareksha: ``ignore_comments`` holds all the comments (regex matches) on the same line.

pareksha:
@jayvdb done the changes you told. Please review.

jayvdb:
As mentioned on gitter, please beg for more reviews from the other people who have pending patches related to ignore comments. i.e. bond with your peers in the community ;-)

pareksha:
You had suggested the changes that is why I was asking you for a review. But anyways, I think I need to do some reviews myself.

Issue #5044, opened 2018-01-05
.coafile not showing the correct error


ishanSrt:
An earlier typo in .coafile which is being fixed by https://github.com/coala/coala/issues/5043 doesn't show the correct error message. Instead it shows an error based on the previous line
eg. if your .coafile has this 
```
use_spaces = true
default_action: **: ApplyPatchAction
```
it will show
```
 Unable to convert parameter 'use_spaces' into type <class 'bool'>.
```
 if your .coafile is 
```
files = *.txt
default_action: **: ApplyPatchAction
```
It will show **Unable to find that file/ No matching file**

or if your .coafile is 
```
bears = SpaceConsistencyBear
default_action: **: ApplyPatchAction
```
it will show 
```
No bears matching 'SpaceConsistencyBear
default_action: **: ApplyPatchAction' were found. Make sure you have coala-bears installed or you have typed the name correctly.
```
Instead it should point out the error

ishanSrt:
I would like to self assign myself to this issue 😌

ishanSrt:
@Makman2 mark blocked by https://github.com/coala/coala/issues/5043

Makman2:
Oh I'm just thinking: The parser allows to continue values on next lines without specifying escape characters. This is probably why it points out that the error lies in the previous lines, because this is where the setting-assignment effectively begins, and it thinks those new malformed setting-syntaxes belong to the value.

ishanSrt:
Yes exactly! That's what I could make out from https://github.com/coala/coala/issues/5044#issue-286363165
Btw how does the parser know that a line is complete?

ishanSrt:
@Makman2 even in case of valid lines

Makman2:
I guess it tries to match a pattern that indicates a new value (roughly `.*=.*`). If it does not match, it assumes the next line to be a value for the last setting parsed.

Makman2:
I'm really not sure what to do now. An escape character would be probably nicest for multi-line values, but breaks coafiles quite badly...

Oh but just thinking: What about noting in case of errors, where the relevant lines are? For our case, it would display `lines 1-2`. Then people probably grasp the cause earlier.

ishanSrt:
@Makman2 didn't get what you meant by this:
>What about noting in case of errors, where the relevant lines are? For our case, it would display lines 1-2. Then people probably grasp the cause earlier.

ishanSrt:
[...]

Issue #5043, opened 2018-01-05
.coafile Correct command


ishanSrt:
`default_action: *: ShowPatchAction` this is not the correct format and should be changed to 
`default_actions = *: ShowPatchAction`

yugeshk:
I would like to take this up.

Makman2:
@yugeshk we are a bit low on newcomer issues and you've already opened a PR for another one, is it okay if you try to find a `difficulty/low` issue? :)

yugeshk:
@Makman2 yes sure!

Hynix07:
I would like to take this up!!

ishanSrt:
@Hynix07 still working on it?

Makman2:
Unassigning due to inactivity.

PR #5070, opened 2018-01-11
ConfParser.py: Display correct error

ishanSrt:
Fixes https://github.com/coala/coala/issues/5044

### Basic idea behind my changes:
if I find a `:` before I find a `#` or a `=` I print out the error

### DIFF
--- a/coalib/parsing/ConfParser.py
+++ b/coalib/parsing/ConfParser.py
@@ -32,6 +32,23 @@ class ConfParser:
         self.__rand_helper = None
         self.__init_sections()
 
+    def check_valid_lines(self, lines):
+        """
+        This function checks for some invalid config file commands like
+        ``default_action: *: ShowPatchAction``
+        which show error in the previous line. This function gives the
+        correct error in such cases.
+        """
+        for line in lines:
+            occ = line.find(':')
+            line_substr = line[:occ]
+            occ_equals = line_substr.find('=')
+            occ_comment = line_substr.find('#')
+            if occ != -1 and not ((occ_equals != -1 and occ_equals < occ) or
+                                  (occ_comment != -1 and occ_comment < occ)):
+                logging.error('Invalid command "{}" in config file'.format
+                              (line.strip('\n')))
+
     def parse(self, input_data, overwrite=False):
         """
         Parses the input and adds the new data to the existing.
@@ -50,6 +67,8 @@ class ConfParser:
 
         with open(input_data, 'r', encoding='utf-8') as _file:
             lines = _file.readlines()
+
+        self.check_valid_lines(lines)
 
         if overwrite:
             self.__init_sections()
--- a/tests/parsing/ConfParserTest.py
+++ b/tests/parsing/ConfParserTest.py
@@ -191,3 +191,11 @@ class ConfParserTest(unittest.TestCase):
                                             'already been defined in section '
                                             'name. The previous setting will '
                                             'be overridden.')
+
+    def test_check_valid_lines(self):
+        logger = logging.getLogger()
+        with self.assertLogs(logger, 'ERROR') as self.cm:
+            self.uut.check_valid_lines(['default_action: *: ShowPatchAction'])
+        self.assertEqual(self.cm.output[0], 'ERROR:root:Invalid command '
+                                            '"default_action: *: ShowPatc'
+                                            'hAction" in config file')


ishanSrt:
@Makman2 I tried to solve it the hard way first. I started making regex for valid config file lines but that was getting way to complex. In the end I ditched all of that 🤯😔 and made this quick fix

ishanSrt:
@Makman2 
First I tried to do this the hard way but that involved writing regex for all kinds of valid coala lines. It was getting way too difficult, so I ditched it and wrote this fix.

The line for which I have written the code is itself is the cause for failing the test. It just says `Warning: treated as error`. Maybe I am unsure how to solve this or should this issue be blocked by https://github.com/coala/coala/issues/5043 ? If it is blocked by that issue then I am unsure how is my code-coverage 100% for the changes I made which leads me to believe that when travis and circle run coala on the coala directory that too affects the code coverage! 😲

Makman2:
Nah it's the documentation generation, something is maybe badly formatted there.

coalib/parsing/ConfParser.py
```diff
@@ -32,6 +32,23 @@ def __init__(self,
         self.__rand_helper = None
         self.__init_sections()
 
+    def check_valid_lines(self, lines):
+        """
+        This function checks for some invalid config file commands like
+        default_action: *: ShowPatchAction
+        which show error in the previous line. This function gives the
```
Makman2: --> **an** error

coalib/parsing/ConfParser.py
```diff
@@ -32,6 +32,23 @@ def __init__(self,
...
+        correct error and raises System Exit in this case.
```
Makman2: 1.
```
System Exit
```
-->
```
``SystemExit``
```

ishanSrt: I don't raise a system exit now. so I will remove it

coalib/parsing/ConfParser.py
```diff
@@ -32,6 +32,23 @@ def __init__(self,
...
+        """
+        for line in lines:
+            occ = line.find(':')
+            line_substr = line[:occ]
+            occ_equals = line_substr.find('=')
+            occ_comment = line_substr.find('#')
+            if occ != -1 and not ((occ_equals != -1 and occ_equals < occ) or
+                                  (occ_comment != -1 and occ_comment < occ)):
+                logging.error('Invalid command "{}" in config file'.format
+                              (line.strip('\n')))
```
Makman2: whee... this looks so hacky... Although our CLI code is in a crappy state, this patch is sailing near the wind... this is not how this should work optimally :/ Isn't there any way directly modifying the parsing routines? Without using a ton of regexes? 😅 

PS: @Makman2 has learned new English: "sailing near the wind"

Makman2: But at least the patch isn't complicated...

tests/coalaTest.py
```diff
@@ -304,10 +305,13 @@ def test_coala_with_color(self):
...
+            only_warns = True
             for err in errors:
-                self.assertNotRegex(err, '^\[WARNING\]')
-            self.assertEqual(
-                retval, 0, 'coala must return zero when there are no errors')
+                if not re.match('^\[WARNING\]', err):
+                    only_warns = False
+            if only_warns:
+                self.assertEqual(retval, 0, 'coala must return zero '
+                                 'when there are no errors')
```
Makman2: Why is there an error introduced now in coala?

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -362,7 +362,8 @@ def test_default_section_deprecation_warning(self):
                                  self.log_printer,
                                  arg_list=[])
 
-        self.assertIn('WARNING', cm.output[0])
+        output = '\n'.join(cm.output)
+        self.assertIn('WARNING', output)
```
Makman2: why this change?

Makman2:
unack 15d177ae6feb3bc812fb172b4b46316c3b0b0708

tests/settings/ConfigurationGatheringTest.py
ishanSrt: @Makman2 coala is giving both warnings and errors through stderr. so I needed a condition such as there is atleast one warning created. I am giving out an error now so `cm.output[0]` is no longer a warning

ishanSrt: I couldn't find something to match atleast one from a list in assert

tests/coalaTest.py
ishanSrt: for  the reason that coala gives both errors and warnings through stderr

Makman2: I think I'm missing something, but I can't see a change that causes this test to fail^^

ishanSrt:
@Makman2 made the requested changes. Also tell me if I am wrong, you asked me to remove the changes I made in coalaTest.py because if an invalid config line tries to creep into the .coafile in the future, the tests are going to fail instead of just giving out an error. Am I right?

coalib/parsing/ConfParser.py
```diff
@@ -32,6 +32,23 @@ def __init__(self,
...
+        correct error in such cases.
+        """
+        for line in lines:
+            occ = line.find(':')
+            line_substr = line[:occ]
+            occ_equals = line_substr.find('=')
+            occ_comment = line_substr.find('#')
+            if occ != -1 and not ((occ_equals != -1 and occ_equals < occ) or
+                                  (occ_comment != -1 and occ_comment < occ)):
+                logging.error('Invalid command "{}" in config file'.format
```
jayvdb: include line number

coalib/parsing/ConfParser.py
```diff
@@ -32,6 +32,23 @@ def __init__(self,
...
+        ``default_action: *: ShowPatchAction``
+        which show error in the previous line. This function gives the
+        correct error in such cases.
+        """
+        for line in lines:
+            occ = line.find(':')
+            line_substr = line[:occ]
+            occ_equals = line_substr.find('=')
+            occ_comment = line_substr.find('#')
+            if occ != -1 and not ((occ_equals != -1 and occ_equals < occ) or
```
jayvdb: I expect this approach is going to fail on

```
default_action =
  *: ShowPatchAction
```
(which I assume is currently valid)

and why cant I have?

```
a = b
# 1:1 relationship
```
(again, I assume the the parser currently understands that correctly)

ishanSrt:
@jayvdb as suggested by Makman I have to completely change all this https://github.com/coala/coala/issues/5044#issuecomment-361647237

ishanSrt:
Anyways this approach was wrong 😅

Issue #5041, opened 2018-01-05
The glob `*` does not match all bears


ishanSrt:
Many bears' `result.origin` is not just the name of the bear. For e.g.. ESLintBear has result.origin "_ESLintBear (prettier/prettier)_ and the glob `*` doesn't match anything with file seperators in it i.e. `/` or `\`, we have to use glob `**`. The PR https://github.com/coala/coala/pull/5017 fixes the issue with `--apply-patches` command line argument but if the `.coafile` has something like this 
```
default_actions = *: ApplyPatchAction
```
This may not match all bears. This needs to be solved. Many things can be done to solve this eg:

1. The `result.origin` of those bears can be changed

2. I don't think that any user will think this through and use glob `*` specifically to ignore such bears so we can internally change the glob `*` to `**

3. We can prompt the user about the above problem if they ever use the glob `*`

status/needs discussion

ishanSrt:
@Makman2 assign this issue to me and tell me which of the above 3 solutions should I apply?

Makman2:
Okay that is not so easy. Usually if we say "glob" we want to be compliant to globbing-standards. So this is dangerous and might confuse:

> I don't think that any user will think this through and use glob * specifically to ignore such bears so we can internally change all the glob * to **

Regarding this one:

> The result.origin of those bears can be changed

How do you want to change it?

Makman2:
So we implement our own simple glob-like matching that only has `*` as a feature, and everything else is full-text-matching.

ishanSrt:
@Makman2 wouldn't it be sufficient to replace all `*` with `**` whether only `*` or combination of `pattern` and `*` ?

Makman2:
:+1:

ishanSrt:
Reassignment needed on this issue

akshatkarani:
Can i work on this issue?

akshatkarani:
@ishanSrt  I would like to work on it. Are you still on it?

PR #5114, opened 2018-01-26
Globbing.py: Convert glob '*' to '**'

ishanSrt:
Some bears' ``result.origin`` have os.sep in them which makes
the glob '*' not match against them. This commit removes the
usage of this glob altogether and converts it into `**`.

Closes https://github.com/coala/coala/issues/5041


coalib/parsing/Globbing.py
```diff
@@ -212,9 +212,12 @@ def fnmatch(name, globs):
                         More than two or just one sequence can be given.
     -  '?':             Matches any single character.
     -  '*':             Matches everything but os.sep.
+                        [NOTE: Not implementing this anymore,
+                        converting it internally into '**']
```
pareksha: Could you change `it` to ``'*'`` to be more precise?

tests/parsing/GlobbingTest.py
```diff
@@ -195,8 +195,8 @@ def test_questionmark(self):
 
     def test_asterisk(self):
         pattern = 'a*b'
-        matches = ['axb', 'ayb']
-        non_matches = ['aXbX', os.path.join('a', 'b')]
+        matches = ['axb', 'ayb', os.path.join('a', 'b')]
+        non_matches = ['aXbX']
```
pareksha: Why do you have to change the position of ``os.path.join('a', 'b')`` from ``non_matches`` to ``matches``?

ishanSrt: the commit body explains that

newbazz: Then how will pattern `a**b` and `a*b` ?

ishanSrt:
@newbazz that won't make any difference while comparing globs

tests/parsing/GlobbingTest.py
ishanSrt: I don't understand what you mean by that question

newbazz:
Why any loop hole?

ishanSrt:
@newbazz 😆😂 this isn't a "loophole" at all since the behaviour won't change at all.

ishanSrt:
The glob ** means 0 to any number of any characters so two of them also mean the same. Anyways I made the changes.

ishanSrt:
@newbazz a lot of tests are failing if I try to implement that and I want to check whether I can get away without having to fix 😛 them plus it i't making any logical difference

ishanSrt:
cc @Makman2

coalib/processes/Processing.py
```diff
@@ -138,7 +138,8 @@ def autoapply_actions(results,
             action = default_actions[result.origin]
         except KeyError:
             for bear_glob in default_actions:
-                if fnmatch(result.origin, bear_glob):
+                if fnmatch(result.origin, bear_glob.replace('*', '**')
+                           if bear_glob == '*' else bear_glob):
```
newbazz: Sudnt this be `if not "**" in bear_globs then replace '*' with '**' ` ?

ishanSrt: @newbazz wip. Please review only after I ping you😉

ishanSrt: actually the solution is not simple as that, I have to check for each '*' I find whether the next is also no '*' and only then I need to convert that. I am trying to find an elegant one liner for that. Also need to write a test over here.

newbazz: That is what i am telling bro, `if `**` not in x and '*' in x then replace` but then there can be something like `a*b**c` right?

ishanSrt: yes and I don't want to write an ugly loop

newbazz: probably you can split by '**' and then write a for loop ;P

newbazz: Then join ;)

ishanSrt: thats a great option!
so something like `''.join([glob.replace('*','**') for glob in bear_glob.split('**')])`

newbazz: yeah exactly!! :)

Makman2: You could also use `re.subs(r'\*[^*]', '**', text)`

Makman2: Ah maybe the regex has to be expanded to
```
\*([^*]|$)
```

ishanSrt: @Makman2 that is still gonna match the last `*` in any number of consecutive `*`s

Issue #3704, opened 2017-02-05
Improve emphasis of material in Newcomer's Guide


arvind-iyer:
There are too many "Notes" and "See Also" blocks all using the same pattern that subliminally make one(or atleast me) to just scroll past all of them. 

Couple of ideas:
* Multipage guide (navbar, index/contents links on side so people can jump to page
* Different color schemes for different types of blocks used currently
* at least changing notes to important or something that indicates urgency/importance

arvind-iyer:
@sims1253

sims1253:
thank you :)

sils:
CC @Adrianzatreanu , docs lead

yash-nisar:
This has been solved IMO, check http://api.coala.io/en/latest/Developers/Newcomers_Guide.html. Regarding the links part on the side of the page, a separate issue has been opened. I think this can be closed.

kriti21:
Is this issue available? I would like to work on it.

Makman2:
@kriti21 please check-up @yash-nisar's comment :+1:

kriti21:
Let me break the task needed here and my queries also. Please provide your guidance @arvind-iyer @yash-nisar @Makman2 
1) The newcomer guide is still contained in a single page. Should I make it a multi page document even though the links are present to jump to the required section?
2) Please suggest the color scheme that should be followed to differentiate Notes and See Also blocks.
3) For notes, IMO tags like important and general-help would be sufficient.

Makman2:
> The newcomer guide is still contained in a single page. Should I make it a multi page document even though the links are present to jump to the required section?

In that case imo we can keep it in a single document.

> Please suggest the color scheme that should be followed to differentiate Notes and See Also blocks.

What about something like:
- Very important -> red
- Normally important / caution / warning -> yellow
- Note and See Also -> Blue (like it is now I believe)

> For notes, IMO tags like important and general-help would be sufficient. Any other that you suggest?

"general-help" should be called "note" ;) But not that I can at least think of, please introduce as many as necessary / making sense :)

PR #5119, opened 2018-01-30
Newcomers_Guide: Improves emphasis of material

kriti21:
This adds color schemes for different types of blocks
and catgorizes them into important, caution and notes.

Closes https://github.com/coala/coala/issues/3704

### DIFF
--- a/docs/Developers/Newcomers_Guide.rst
+++ b/docs/Developers/Newcomers_Guide.rst
@@ -51,7 +51,7 @@ to coala, while getting you familiar with the workflow!
 
 For more information about Pull Requests, keep reading!
 
-.. note::
+.. tip::
 
     **You do not need to read the coala codebase to get started** - this guide
     is intended to help you do that without reading tons of meaningless code.
@@ -75,7 +75,7 @@ to learn how to set up an environment to work on coala.
 Most importantly, keep notes on what could be changed to make coala usage
 easier!  What documentation was missing? What was hard to understand?
 
-.. note::
+.. seealso::
 
     *Struggling with this?* We have a very verbose guide on this topic in
     `our Google Code In resources <https://github.com/coala/coala/wiki/Google-Code-In-Task-Use-coala>`_
@@ -175,7 +175,7 @@ Now it is time to pick an issue.
 It is the best way to familiarise yourself with the codebase.
 You can view `all Newcomer issues on GitHub <https://coala.io/new>`_.
 
-.. note::
+.. important::
 
     You need to be logged in before you follow the Newcomer issues link.
 
@@ -192,7 +192,7 @@ Leave a comment that you would like to be assigned to the issue. This way
 we don't have multiple people working on the same issue at the same time.
 Now you can start working on it!
 
-.. note::
+.. caution::
 
     As stated above, you should never work on an issue without being
     assigned. Fortunately, corobo is here to help you! If you are
@@ -228,7 +228,7 @@ Now you can start working on it!
     Before starting your first commit, check out this
     link: `Writing good commits <http://coala.io/commit>`_.
 
-.. seealso::
+.. important::
 
     An important part of working on issues is documenting your work
     in such a way that it is easy for others to read and understand.
@@ -256,7 +256,7 @@ repository, go to the official repository of coala/coala-bears and click on the
 where ``myfork`` is the name of your fork, and ``fork_link`` is a link to your
 fork repository.
 
-.. note::
+.. important::
    It is important that you **DO NOT** make your changes on the master branch
    of your forked repository to avoid the following cases:
 
@@ -310,7 +310,7 @@ into your bash prompt. This will analyze your code and help you fix it.
 Step 4. Sending Your Changes
 ----------------------------
 
-.. note::
+.. caution::
 
    Before committing your changes, please check that you are indeed in a
    development branch created in step 3. To check if you are in a branch, type:
@@ -387,7 +387,7 @@ see that's the case from the ``process/pending review`` label. Now all you have
 to do is wait, or let the other developers know on Gitter that you have
 published your changes.
 
-.. note::
+.. important::
 
     Do not tag the reviewers every time you push a change. They review PRs
     consistently whenever they have time!
@@ -399,7 +399,7 @@ Now there are two possibilities:
 - your ``Pull Request`` doesn't get accepted, and therefore you will
   need to modify it as per the review comments
 
-.. note::
+.. caution::
 
     Wait until the reviewer has reviewed your whole Pull Request
     and has labeled it ``process/wip``. If you push again before that happens,
@@ -449,7 +449,7 @@ Now you should `tell us about your experience <https://coala.io/newform>`_ and
 go for `a low issue <https://coala.io/low>`__ - they are really rewarding!
 
 
-.. note::
+.. attention::
 
     Do not delete the fork subsequent to Pull Request for review or
     after it is merged!
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -69,7 +69,7 @@ h1,h2,h3,h4,h5,h6 {
     margin-bottom: 2em;
 }
 
-div.admonition, div.note {  /* Ignore CSSLintBear */
+div.admonition {  /* Ignore CSSLintBear */
     background-color: #263238;
     border: none;
     border-radius: 0;
@@ -81,14 +81,34 @@ div.admonition, div.note {  /* Ignore CSSLintBear */
     transition: box-shadow .25s;
 }
 
+div.attention {  /* csslint allow: overqualified-elements */
+    background-color: #d9534f;
+}
+
+div.caution {  /* csslint allow: overqualified-elements */
+    background-color: #f0ad4e;
+}
+
+div.important {  /* csslint allow: overqualified-elements */
+    background-color: #d9534f;
+}
+
+div.tip {  /* csslint allow: overqualified-elements */
+    background-color: #428bca;
+}
+
+div.note {  /* csslint allow: overqualified-elements */
+    background-color: #5bc0de;
+}
+
 div.admonition p.admonition-title {  /* Ignore CSSLintBear */
     font-family: 'Roboto', serif;
-    font-weight: 100;
+    font-weight: 200;
     margin: 0 0 5px 0;
 }
 
 div.admonition p {
-    font-weight: 200;
+    font-weight: 300;
     margin-top: 1em;
 }
 


docs/_static/custom.css
```diff
@@ -81,6 +81,30 @@ div.admonition, div.note {  /* Ignore CSSLintBear */
...
 
+div.caution {  /* Ignore CSSLintBear */
+    background-color: #b44c0e;
+    border: none;
+    border-radius: 0;
+    box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12), 0 3px 1px -2px rgba(0, 0, 0, 0.2);
+    color: white;
+    font-family: Roboto;
+    padding: 2em;
+    position: relative;
```
newbazz: Relative positioning isnt default one?

docs/_static/custom.css
```diff
@@ -81,6 +81,30 @@ div.admonition, div.note {  /* Ignore CSSLintBear */
...
+
+div.important {  /* Ignore CSSLintBear */
+    background-color: #783238;
+    border: none;
+    border-radius: 0;
+    box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.14), 0 1px 5px 0 rgba(0, 0, 0, 0.12), 0 3px 1px -2px rgba(0, 0, 0, 0.2);
+    color: white;
+    font-family: Roboto;
+    padding: 2em;
+    position: relative;
```
newbazz: here as well

docs/Developers/Newcomers_Guide.rst
```diff
@@ -208,27 +208,27 @@ Now you can start working on it!
...
-            corobo file issue <repo> <title>
-            <description>
+            **corobo file issue <repo> <title>**
+            **<description>**
 
-        * You shouldn't close any PR, instead mark them as work in progress::
+        * You shouldn't close any PR, instead mark them as work in progress
 
-            corobo mark wip <full url>
+            **corobo mark wip <full url>**
```
newbazz: I ud suggest boldning in a different commit <if required>, what you say @Makman2 ?

adhikasp: I think that is unnecessary. This could be considered as part of emphasizing and relevant to the commit.

newbazz: @adhikasp (Y)

ishanSrt:
does this change the content of the link: http://api.coala.io/en/latest/Developers/Newcomers_Guide.html or just for https://github.com/coala/coala/blob/master/docs/Developers/Newcomers_Guide.rst or are the two same?

kriti21:
@ishanSrt both of the above links you mentioned refer to the same thing but the changes I have made are for http://api.coala.io/en/latest/Developers/Newcomers_Guide.html Because this is sort of a design issue, major change was on css rendering part.

docs/Developers/Newcomers_Guide.rst
```diff
@@ -310,13 +308,11 @@ into your bash prompt. This will analyze your code and help you fix it.
 Step 4. Sending Your Changes
 ----------------------------
 
-.. note::
+.. caution::
 
    Before committing your changes, please check that you are indeed in a
    development branch created in step 3. To check if you are in a branch, type:
 
-   ::
```
Makman2: different issue :)

docs/_static/custom.css
```diff
@@ -81,6 +81,14 @@ div.admonition, div.note {  /* Ignore CSSLintBear */
     transition: box-shadow .25s;
 }
 
+div.caution {  /* Ignore CSSLintBear */
+    background-color: #b44c0e;
+}
+
+div.important {  /* Ignore CSSLintBear */
+    background-color: #783238;
+}
```
Makman2: Isn't this maybe a bit too dark?

Btw I would like to see a screenshot how it looks like when it's rendered :)

docs/Developers/Newcomers_Guide.rst
```diff
@@ -175,7 +175,7 @@ Now it is time to pick an issue.
 It is the best way to familiarise yourself with the codebase.
 You can view `all Newcomer issues on GitHub <https://coala.io/new>`_.
 
-.. note::
+.. caution::
```
Makman2: Maybe this is "important"?

We should definitely clear what messages are "caution" and what "important"

docs/Developers/Newcomers_Guide.rst
```diff
@@ -208,27 +208,27 @@ Now you can start working on it!
     You can do amazing stuff using corobo.
 
         * Issue assigning as stated earlier.
-        * File issues::
```
Makman2: Different issue removing the colons :)

docs/Developers/Newcomers_Guide.rst
Makman2: Hm this should be rendered as code, so it should be kept. Imo a different issue.

docs/Developers/Newcomers_Guide.rst
```diff
@@ -208,27 +208,27 @@ Now you can start working on it!
...
+            **<description>**
 
-        * You shouldn't close any PR, instead mark them as work in progress::
+        * You shouldn't close any PR, instead mark them as work in progress
 
-            corobo mark wip <full url>
+            **corobo mark wip <full url>**
 
-        * To label a PR as pending review::
+        * To label a PR as pending review
```
Makman2: Different issue

docs/Developers/Newcomers_Guide.rst
```diff
@@ -208,27 +208,27 @@ Now you can start working on it!
...
 
-            corobo mark pending <full url>
+            **corobo mark pending <full url>**
 
-        * To see all of the corobo commands::
+        * To see all of the corobo commands
 
-            corobo help
+            **corobo help**
```
Makman2: Different issue

docs/Developers/Newcomers_Guide.rst
```diff
@@ -208,27 +208,27 @@ Now you can start working on it!
...
 
     Before starting your first commit, check out this
     link: `Writing good commits <http://coala.io/commit>`_.
 
-.. seealso::
+.. caution::
```
Makman2: Rather "note" or "important"

docs/Developers/Newcomers_Guide.rst
```diff
@@ -449,7 +445,7 @@ Now you should `tell us about your experience <https://coala.io/newform>`_ and
 go for `a low issue <https://coala.io/low>`__ - they are really rewarding!
 
 
-.. note::
+.. caution::
```
Makman2: Maybe we want a new tag "critical" or so, because this is so damn important imo 😅

Makman2:
unack 80ead210ddcea628751cb015bcc2b678ee908ee3

docs/Developers/Newcomers_Guide.rst
kriti21: I will add critical ;) Is that okay ?

kriti21: @Makman2 I am using tag "critical" for this one. Can you suggest me the background color to be used with this? Red is already used for important and yellow for caution.

docs/Developers/Newcomers_Guide.rst
kriti21: @Makman2 These colons are not for text but to use a different block defined in css. If i keep these colons, the background will be as defined here https://github.com/coala/coala/blob/1db00d706353bb0d8a175c6d561df21e51cad6e7/docs/_static/custom.css#L108 or else the left padding is disturbed.
And that is why, the rendered code also needs to be bold. Additionally, bold text appears better over yellow background. But this i can keep normal for this PR and open a separate issue to bold this text. What do you suggest?

kriti21:
colorbrightness

kriti21:
Also "critical" is not supported in the Specific Admonitions. I have used "Attention" instead. Here are all the available options http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions , if you could suggest anything better. Or I could make a generic admonition if that would be fine because we do not really need critical tag that often.

docs/_static/custom.css
jayvdb: We should netlify the docs in this repo.
I had a WIP for that, but it was blocked by https://github.com/coala/devops/issues/90

docs/Developers/Newcomers_Guide.rst
jayvdb: @kriti21 has an issue been created about this yet?
If not, please create one.  if you describe clearly what needs to be done, and show them this PR, it is a newcomer issue, and we'll find a newcomer to do it, and you need to remove this stuff from your PR.

docs/_static/custom.css
kriti21: I have added a screenshot in the comments anyway. Is that okay or any changes required ? @jayvdb

docs/Developers/Newcomers_Guide.rst
kriti21: But adding the extra colon makes the background of all the following corobo commands grey colored. Is is okay to push a PR like that (not looking good) ? @jayvdb

jayvdb: Create a separate issue.  The extra colon is a syntax *bug*, and is unrelated to your issue.

kriti21:
@HizkiFW I have added `/* Ignore CSSLintBear */ ` to avoid https://github.com/coala/coala/pull/5119#pullrequestreview-110256850 error. Using only `.important` or `.caution` will not work as these are used in other developer guides. I have changed the colors though. Please review. :)

HizkiFW:
Regarding the CSS, I found out that using the class names alone doesn't work because the theme (`alabaster.css`) and `basic.css` both have the element name on their style definitions, thus overriding the class-only definitions.

This could be fixed by removing the `div`s from those files, but they're part of another project (https://github.com/bitprophet/alabaster). Maybe we could make a fork of the theme and use that instead.

Also, @kriti21:
> Using only `.important` or `.caution` will not work as these are used in other developer guides.

what do you mean by "other developer guides"? Which ones?

kriti21:
>what do you mean by "other developer guides"? Which ones?
I just meant that is will not work that way i.e. using `.important` only. Said something wrong there. Please neglect that @HizkiFW 
`custom.css` is a common file used for all the guides here https://github.com/coala/coala/tree/master/docs/Developers . Making any changes might have more impacts than required. What do you suggest I should do?

HizkiFW:
You're right, changing that is definitely going to break more than just a few things. Just leave it like that but also add a comment with a link to this issue.

docs/Developers/Newcomers_Guide.rst
```diff
@@ -281,8 +281,6 @@ fork repository.
    In order to avoid the above mentioned cases you can create a new branch
    where you will work on the issue. To do that run:
 
-   ::
```
Phantom-42: This fix is unrelated to improving emphasis of material. I think it is better to have this in a separate change.

docs/_static/custom.css
```diff
@@ -81,6 +81,18 @@ div.admonition, div.note {  /* Ignore CSSLintBear */
     transition: box-shadow .25s;
 }
 
+div.attention {  /* Ignore CSSLintBear */
```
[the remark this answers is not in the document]
jayvdb: I think we can use `!important` to override other stylesheets.
@blazeu ?

Ignoring the bear is not a good option, as it means all of its rules are ignored.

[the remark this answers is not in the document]
blazeu: Yeah sure, and `div.attention` is a bit excessive, no need to put the element name unless necessary

kriti21:
@Phantom-42 I have implemented everything that you suggested. Please have a look.
@jayvdb Could you please help me with `!important` ? I am not getting what went wrong here.

blazeu:
What happens if you remove important and add `div.` back, does it makes CSSLintBear angry again? is that the reason why there's ignore line before?

Using important is better than having excessive selector IMO. Try adding `/* csslint allow: important */*`

Also, IIRC, StyleLintBear has nicer default than CSSLintBear and CSSLint repo is kind of dead.

blazeu:
Okay, I just looked at the default css for this theme, and I found out why we needed `div` or we must use `! important`. The default css has `div` in front of everything. Sorry for catching up late.

In this case using `div.note` is okay, just need to put that csslint ignore line. `/* csslint allow: overqualified-elements */`

kriti21:
@jayvdb @blazeu Please review :)

jayvdb:
ack https://github.com/coala/coala/pull/5119/commits/1b94767a1319799d192752565a8e226cef893f00

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

Issue #2535, opened 2016-07-23
Make `--find-config` default


hypothesist:

Makman2:
you sure? this will backpropagate all directories, you could grab some unintended coafile^^

hypothesist:
But IMO being even one level below the root directory will complain about no coafiles.

hypothesist:
@Makman2 so should we go forward with this?

Makman2:
yes, @sils1297 is also voting for this :3

Daeke:
I'll take a look at this, if you want to assign it to me!

amacneil:
I came across this issue looking for this feature, but I think the current `--find-config` behavior is not what most people would expect.

Right now there does not seem to be a way to run coala on a subdirectory that is not the root of your project. There is a `--files` option, but then I need to specify specific files (rather than relying on the patterns in `.coafile`).

What currently happens:

```sh
cd project/subdir
coala # fails, the default coafile '.coafile' was not found
coala --find-config # runs coala for the entire project
```

What I would expect:

```sh
cd project/subdir
coala # ideally this would work automatically (as this issue requests)
coala --find-config # I would expect this to only run bears against the current folder
```

So, there should be a way to run coala against a subdirectory of my project, and use the `.coafile` from the project root. Ideally both of these methods would have the same effect:

```sh
cd project/subdir && coala
coala project/subdir
```

Any thoughts on this? Am I missing something obvious?

sils:
Hey @amacneil , you can use `coala --limit-files "subdir/**"`, then coala will use the config (which you can also give with `-c`) and limit the set of files for every section by the limit-files glob. This is especially used in editor plugins to only check that file but do it properly ;)

Can you tell us where you searched and where we could have put the info for you to find? Or how can we make this more visible? I think this is what you wanted.

amacneil:
@sils thanks, that seems to do what I wanted.

I'm not sure, that flag seems obvious now that I know it exists 😄 I guess I would expect to see it on this page, which explains clearly how to override the files, but not how to run the default configured file list against a specific directory:

https://docs.coala.io/en/latest/Users/Glob_Patterns.html

RohanVB:
Still being worked on? 
CC @vreethesh

vreethesh:
Actually looked into it more than once. Couldn't come to a proper solution
to solve it. With help, I can solve it.

With regards,
Reethesh Venkataraman
[...]

PR #5120, opened 2018-01-30
Make `--find-config` default

khanchi97:
Set `--find-config` default and disable `--find-config` where
`--no-config` is encountered and change some values where tests were
failing just because of making `--find-config` default.

Closes https://github.com/coala/coala/issues/2535

### DIFF
--- a/tests/settings/ConfigurationGatheringTest.py
+++ b/tests/settings/ConfigurationGatheringTest.py
@@ -74,7 +74,8 @@ class ConfigurationGatheringTest(unittest.TestCase):
             str(sections['cli']),
             "cli {bears : 'JavaTestBear', config : " +
             repr(PathArg(temporary)) +
-            ", files : '*.java', save : 'True', test : '5'}")
+            ", files : '*.java', find_config : 'True', "
+            "save : 'True', test : '5'}")
 
         with make_temp() as temporary:
             sections, local_bears, global_bears, targets = (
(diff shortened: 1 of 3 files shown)


tests/settings/ConfigurationGatheringTest.py
```diff
@@ -114,7 +115,7 @@ def test_system_coafile_parsing(self):
             arg_list=[])
 
         self.assertEqual(str(sections['test']),
-                         "test {value : '1', testval : '5'}")
```
jayvdb: something is wrong if you need to modify this.

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -228,7 +229,8 @@ def test_back_saving(self):
         with open(filename, 'r') as f:
             lines = f.readlines()
         self.assertEqual(['[cli]\n',
-                          'config = some_bad_filename\n'], lines)
+                          'config = some_bad_filename\n',
+                          'find_config = True\n'], lines)
```
jayvdb: as `find_config` defaults to True, there is no need to include it in every section, not even in `cli` section.

tests/settings/ConfigurationGatheringTest.py
jayvdb: sorry for the delay; we've all been under strain.  Catching up though!

Issue #2290, opened 2016-06-04
Allow bear names to be used without the "Bear" suffix


sils:
and remove `Bear` from the names, useless

sils1297:
Maybe let the user type `--bears=Clang` and coala will search for a ClangBear so he doesn't have to type so much?

CC @underyx @sims1253 for his discussion

sims1253:
I think giving the bears the tool they use as an attribute is enough, as one can list them as needed then.
Then maybe just name the bears after what they are doing. Although i don't have an idea what to do with multible bears doing the same for different/same languages.

Makman2:
so every bear expression shall be expanded to `bear-wanted(Bear|bear)?` (regex style)

Makman2:
and I'm not sure whether removing `Bear` from the class names is useless, just calling stuff `Clang` or `Alex` is not cool :O

Makman2:
at least it helps us organizing our stuff better^^

fneu:
Now we're getting to where the bodies are burried... [Because of the koala's resemblance to a bear, it is often called the koala bear, although it is a marsupial, and not a true bear.](https://en.wikipedia.org/wiki/Koala#Etymology)

jayvdb:
I've put this at medium difficulty as I dont know how difficult it might be.
But if this issue is only about the Bear suffix in CLI/config, a low difficulty approach would be to add the missing Bear suffix in the CLI/config loading phase, so that the internal data structures all continue to have the same Bear suffix and thus should not break.

anctartica:
please assign this to me

PR #5134, opened 2018-02-05
Collectors.py: Use bear names without suffix

anctartica:
Allow bear names to be used without the "Bear"
suffix at the end.

Closes https://github.com/coala/coala/issues/2290

### DIFF
--- a/coalib/collecting/Collectors.py
+++ b/coalib/collecting/Collectors.py
@@ -177,6 +177,23 @@ def collect_bears(bear_dirs, bear_globs, kinds, log_printer=None,
         bears_found[index].append(bear)
         bear_globs_with_bears.add(glob)
 
+    unused_globs = set(bear_globs) - set(bear_globs_with_bears)
+    suffix_globs = {}
+
+    for glob in unused_globs:
+        if glob is not '**' and glob is not '*':
+            if glob.endswith('bear'):  # pragma nt: no cover
+                new_glob = glob[:-4] + 'B' + glob[-3:]
+                suffix_globs[new_glob] = glob
+            elif not glob.endswith('Bear'):
+                suffix_globs[glob + 'Bear'] = glob
+
+    for bear, glob in icollect_bears(bear_dirs,
+                                     set(suffix_globs.keys()), kinds):
+        index = kinds.index(_get_kind(bear))
+        bears_found[index].append(bear)
+        bear_globs_with_bears.add(suffix_globs[glob])
+
     if warn_if_unused_glob:
         _warn_if_unused_glob(bear_globs, bear_globs_with_bears,
                              'No bears matching \'{}\' were found. Make sure '
--- a/tests/collecting/CollectorsTest.py
+++ b/tests/collecting/CollectorsTest.py
@@ -271,6 +271,16 @@ class CollectBearsTest(unittest.TestCase):
             ['other_kind'],
             self.log_printer)[0]), 0)
 
+    def test_bear_suffix(self):
+        self.assertEqual(
+            len(collect_bears(os.path.join(self.collectors_test_dir, 'bears'),
+                              ['namebear'], ['kind'],
+                              self.log_printer)[0]), 1)
+        self.assertEqual(
+            len(collect_bears(os.path.join(self.collectors_test_dir, 'bears'),
+                              ['name'], ['kind'],
+                              self.log_printer)[0]), 1)
+
     def test_all_bears_from_sections(self):
         test_section = Section('test_section')
         test_section.bear_dirs = lambda: os.path.join(self.collectors_test_dir,
--- a/tests/collecting/collectors_test_dir/bears/nameBear.py
+++ b/tests/collecting/collectors_test_dir/bears/nameBear.py
@@ -1,0 +1,3 @@
+from bear1 import TestBear as ImportedTestBear
+
+__additional_bears__ = [ImportedTestBear]


anctartica:
still working on this...

coalib/parsing/CliParsing.py
```diff
@@ -65,16 +65,19 @@ def parse_cli(arg_list=None,
...
                 arg_value = ','.join([str(val) for val in arg_value])
-
             append_to_sections(sections,
                                arg_key,
                                arg_value,
                                origin,
                                section_name='cli',
                                from_cli=True)
-
     return sections
```
pareksha: Don't remove blank spaces. They enhance code readability.

coalib/parsing/CliParsing.py
```diff
@@ -65,16 +65,19 @@ def parse_cli(arg_list=None,
                                   origin,
                                   line_parser)
         else:
+            if arg_key == 'bears':
+                if isinstance(arg_value, list):
+                    arg_value = [
```
pareksha: You can instead do ``if arg_key == 'bears' and isinstance(arg_value, list):``. 
Nested if is not required here.

coalib/parsing/ConfParser.py
```diff
@@ -114,6 +114,13 @@ def __parse_lines(self, lines, origin):
                 current_keys = keys
 
             for section_override, key in current_keys:
+                if(key == 'bears') and value:
+                    bears = value.split(',')
+                    bears = [Bear.strip() for Bear in bears]
+                    bears = [Bear + 'Bear' if Bear and not Bear.endswith('Bear')
+                             else Bear for Bear in bears]
```
pareksha: You can do ``bears = [Bear.strip() + 'Bear' if not Bear.endswith('Bear') else Bear for Bear in value.split(',')]``
This way you avoid three times assignment to ``bear``.

coalib/parsing/ConfParser.py
```diff
@@ -114,6 +114,13 @@ def __parse_lines(self, lines, origin):
...
+                    value = ','.join(Bear for Bear in bears)
+
```
pareksha: ``.join`` accepts iterables as argument. So, do ``'-'.join(bears)``

anctartica: we need a comma there

anctartica: thanks for the review!

coalib/parsing/CliParsing.py
```diff
@@ -65,9 +65,12 @@ def parse_cli(arg_list=None,
                                   origin,
                                   line_parser)
         else:
+            if arg_key == 'bears' and isinstance(arg_value, list):
+                arg_value = [
+                    Bear + 'Bear' if Bear and not Bear.endswith('Bear')
+                    else Bear for Bear in arg_value]
             if isinstance(arg_value, list):
                 arg_value = ','.join([str(val) for val in arg_value])
-
```
pareksha: No need to remove this blank line too.

coalib/parsing/ConfParser.py
pareksha: No problem :smile:

coalib/parsing/CliParsing.py
```diff
@@ -65,6 +65,10 @@ def parse_cli(arg_list=None,
                                   origin,
                                   line_parser)
         else:
+            if arg_key == 'bears' and isinstance(arg_value, list):
+                arg_value = [
+                    Bear + 'Bear' if Bear and not Bear.endswith('Bear')
+                    else Bear for Bear in arg_value]
```
newbazz: Leave a line after this.

coalib/parsing/CliParsing.py
```diff
@@ -94,6 +98,12 @@ def parse_custom_settings(sections,
         (_, key_tuples, value, append, _) = line_parser._parse(
             setting_definition)
         for key_tuple in key_tuples:
+            if(key_tuple[1] == 'bears') and value:
```
newbazz: write a comment above this for what are you doing here.

coalib/parsing/ConfParser.py
```diff
@@ -114,6 +114,12 @@ def __parse_lines(self, lines, origin):
                 current_keys = keys
 
             for section_override, key in current_keys:
+                if(key == 'bears') and value:
```
newbazz: Comment here as well.

coalib/parsing/CliParsing.py
```diff
@@ -65,6 +65,12 @@ def parse_cli(arg_list=None,
                                   origin,
                                   line_parser)
         else:
+            # if 'Bear' suffix is not already present then add it here
+            if arg_key == 'bears' and isinstance(arg_value, list):
+                arg_value = [
+                    Bear + 'Bear' if Bear and not Bear.endswith('Bear')
+                    else Bear for Bear in arg_value]
```
Makman2: What I'm just thinking, but I guess isn't really easy / possible, is to use our automatic settings parsing. We would introduce a new type `BearList`, that would automatically handle this. But like said, just an idea ;)

coalib/parsing/CliParsing.py
```diff
@@ -65,6 +65,12 @@ def parse_cli(arg_list=None,
                                   origin,
                                   line_parser)
         else:
+            # if 'Bear' suffix is not already present then add it here
```
Makman2: what about `bear`? (lower-case)

coalib/parsing/CliParsing.py
```diff
@@ -65,6 +65,12 @@ def parse_cli(arg_list=None,
...
+            if arg_key == 'bears' and isinstance(arg_value, list):
```
Makman2: what if we provide just a single bear name? Then it isn't a list, is it?

coalib/parsing/CliParsing.py
Makman2: variable names should be lower case, so `Bear` --> `bear`

Makman2: Optional: You could break the stuff a bit to make it more readable:
```
bear + 'Bear'
if bear and not bear.endswith('Bear') else
bear
for bear in arg_value
```

Makman2: Also: Why
`if bear and not bear.endswith(...)`, so why the `if bear` part? Can `bear` be `None` or an empty string?

coalib/parsing/ConfParser.py
```diff
@@ -114,6 +114,13 @@ def __parse_lines(self, lines, origin):
                 current_keys = keys
 
             for section_override, key in current_keys:
+                # if 'Bear' suffix is not already present then add it here
+                if(key == 'bears') and value:
+                    bears = [Bear.strip() + 'Bear'
```
Makman2: same heere, looks duplicated

coalib/parsing/ConfParser.py
```diff
@@ -114,6 +114,13 @@ def __parse_lines(self, lines, origin):
                 current_keys = keys
 
             for section_override, key in current_keys:
+                # if 'Bear' suffix is not already present then add it here
+                if(key == 'bears') and value:
```
Makman2: you should add a space after `if`

tests/parsing/ConfParserTest.py
```diff
@@ -39,6 +39,9 @@ class ConfParserTest(unittest.TestCase):
     key1 = value1
     key2 = value1
     key1 = value2
+
+    [bearstest]
+    bears = A , BBear, C ,,
```
Makman2: actually, I would consider an empty string to be an invalid bear.

Makman2:
unack cd3b7fce1e4661af7eede5e3386c11cf6983dde3

coalib/parsing/CliParsing.py
anctartica: where does this automatic settings parsing happen?

tests/collecting/CollectorsTest.py
```diff
@@ -239,21 +239,21 @@ def test_bear_invalid(self):
     def test_simple_single(self):
         self.assertEqual(len(collect_bears(
             [os.path.join(self.collectors_test_dir, 'bears')],
-            ['bear1'],
+            ['ABear'],
```
jayvdb: why is this changing?

That implies your algorithm is operating differently than before.

i.e. a regression.

tests/collecting/CollectorsTest.py
```diff
@@ -264,6 +264,13 @@ def test_no_duplications(self):
             ['kind'],
             self.log_printer)[0]), 2)
 
+    def test_bear_suffix(self):
+        self.assertEqual(len(collect_bears(
```
jayvdb: this is shocking indentation.
(yes I know that it is used above, but no need to repeat it)

tests/collecting/collectors_test_dir/bears/BBear.py
```diff
@@ -1,4 +1,4 @@
-from bear1 import TestBear as ImportedTestBear
+from ABear import TestBear as ImportedTestBear
```
jayvdb: why must this change?

old bear names are no longer supported?

tests/collecting/collectors_test_dir/bears/metabear.py
```diff
@@ -1,3 +0,0 @@
-from bear1 import TestBear as ImportedTestBear
```
jayvdb: old bear names are not supported?

tests/collecting/CollectorsTest.py
anctartica: For this test case the bear file was named as `bear1`. Now when collecting the bear it would change this to `bear1Bear` which is not what we want. So, I renamed the test case bear file to `ABear`. 
The algorithm should work for all previous cases as before:
- Input `JSStandardBear` this has the `Bear` suffix at the end so does not get changed.
- Input `JSStandard` this does not have the `Bear suffix so changes to `JSStandardBear` when collecting.
-Input `JSStandardbear` changes to `JSStandardBear` when collecting

jayvdb: no that is covering cases which are in our bears repos, but isnt covering cases that were previously allowed by custom bears created and used by other people, possibly confidential and not published anywhere.

tests/collecting/CollectorsTest.py
anctartica: since there are 3 opening round brackets you want me to have 3 indents after this line? @jayvdb 
`self.assertEqual(len(collect_bears(
                    os.path.join(
                        self.collectors_test_dir,
                        'bears'),
                    ['namebear'], ['kind'],
                    self.log_printer)[0]), 1)`

jayvdb:
so much better!

coalib/collecting/Collectors.py
```diff
@@ -177,6 +177,23 @@ def collect_bears(bear_dirs, bear_globs, kinds, log_printer=None,
         bears_found[index].append(bear)
         bear_globs_with_bears.add(glob)
 
+    unused_globs = set(bear_globs) - set(bear_globs_with_bears)
+    suffix_globs = {}
+
+    for glob in unused_globs:
+        if glob is not '**' and glob is not '*':
+            if glob.endswith('bear'):
```
jayvdb: I think you will need to use `#pragma nt: no cover` here

I assume this is because of filename casing on windows, but it might specifically be caused by other coala code adding case-insensitivity to work around windows.

In any case, this is worth a pragma no cover.

jayvdb:
ack https://github.com/coala/coala/pull/5134/commits/6e35cf223e1c035081bf517b5f8cae8a66d4348a

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

PR #5135, opened 2018-02-07
Allow to supply a different "coalang_dir"

### LINKED ISSUE #2617: DocstyleDefinition.get_available_definitions: Allow to supply a different "coalang_dir"
Like in `DocstyleDefinition.load` to search in other directories for coalang files.

jackthepirate6:
Added a new parameter coalang_dir to get_available_definitions.

Closes #2617

### DIFF
--- a/coalib/bearlib/languages/documentation/DocstyleDefinition.py
+++ b/coalib/bearlib/languages/documentation/DocstyleDefinition.py
@@ -301,14 +301,18 @@ class DocstyleDefinition:
                    function_padding, docstring_type_regex, docstring_position)
 
     @staticmethod
-    def get_available_definitions():
+    def get_available_definitions(coalang_dir=None):
         """
         Returns a sequence of pairs with ``(docstyle, language)`` which are
         available when using ``load()``.
 
+        :param coalang_dir:        Path to directory with coalang docstyle
+                                   definition files. This replaces the default
+                                   path if given.
         :return: A sequence of pairs with ``(docstyle, language)``.
         """
-        pattern = os.path.join(os.path.dirname(__file__), '*.coalang')
+        pattern = os.path.join(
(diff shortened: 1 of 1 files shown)


jackthepirate6:
Tried it, but still failing. It may be because I haven't written tests .

newbazz:
Nope i dont think so coz then the travis sud have failed as well

jackthepirate6:
It is giving the following error
imageedit_2_6047057669

coalib/bearlib/languages/documentation/DocstyleDefinition.py
```diff
@@ -301,14 +301,18 @@ def load(cls, language: str, docstyle: str, coalang_dir=None):
...
     @staticmethod
-    def get_available_definitions():
+    def get_available_definitions(coalang_dir=None):
         """
         Returns a sequence of pairs with ``(docstyle, language)`` which are
         available when using ``load()``.
 
+        :param coalang_dir:        Path to directory with coalang docstyle
+                                   definition files. This replaces the default
+                                   path if given.
```
newbazz: `if not given` if i am not wrong.

newbazz:
@jackthepirate6 yeah exactly so theres a container error and not tests failing error.

jackthepirate6:
When do we need 2 container

newbazz:
for two independent and parallel computations.

jackthepirate6:
The default value is None if not given, so it will do the old work

newbazz:
yeah i dont find any problem in that.

coalib/bearlib/languages/documentation/DocstyleDefinition.py
newbazz: make this change and push. Lets see what happens this time.

Issue #5162, opened 2018-02-14
Add `tags` for VIM's tagfiles in .gitignore


nityeshaga:
vim users need to create tag files to jump around the codebase. This file must not be committed. Hence, must be included in `.gitignore` file.

nityeshaga:
@Makman2 I know I asked for it on the chat but I have already completed one newcomer issue. Would it be okay if I do another one?

Makman2:
In this case you quite need it to enhance your development workflow, so it's okay :)

nityeshaga:
Yaay! Thanks a lot @Makman2 ! :smile:

PR #5164, opened 2018-02-14
.gitignore: Add names to ignore tag files in vim

nityeshaga:
Added tagfile names that need to be ignored by git in case the
developer is using VIM editor and needs to generate tags using ctags,
etags or gtags.

Closes https://github.com/coala/coala/issues/5162

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,16 @@ Goopfile.lock
 venv/
 rultor_secrets.sh
 .vscode
+
+#Ignore tags created by ctags, etags and gtags
+TAGS
+.TAGS
+!TAGS/
+tags
+.tags
+!tags/
+gtags.files
+GTAGS
+GRTAGS
+GPATH
+GSYMS


.gitignore
```diff
@@ -38,3 +38,16 @@ Goopfile.lock
...
+.TAGS
+!TAGS/
+tags
+.tags
+!tags/
+gtags.files
+GTAGS
+GRTAGS
+GPATH
+GSYMS
```
sks444: Can't we have a single pattern to ignore all the temporary files created by vim instead of listing them one by one?
see https://stackoverflow.com/questions/4824188/git-ignore-vim-temporary-files

nityeshaga: Yes you can but not in this case. I don't think so. Unlike the `.swp` files and the like, tag files are not temporary files in vim. They are real editable permanent files that contain tags to certain function and variable definitions present in the project which are then used during navigation.

Makman2: But those are quite many... are these really all temporary files by vim? I've seen people provide only `tags` in some gitignore-files.

nityeshaga: Yes only tags might work too. It will for my case. I just included others in case someone else is using tags configured in a different way. I searched a little bit and took help of [this suggested .gitignore](https://github.com/github/gitignore/blob/master/Global/Tags.gitignore) file.

nityeshaga: @Makman2 you think I should change it?

sks444: @nityeshaga , maybe. Take a look at this Vim.gitignore file where they have only used `tags` for auto generated tag files https://github.com/github/gitignore/blob/master/Global/Vim.gitignore

nityeshaga: If someone is using the `gtags` program for creating such tags, he/she may have these 3 files- 
* `GPATH`, 
* `GRTAGS` and 
* `GTAGS`

[Further reading](http://www.gnu.org/software/global/globaldoc_toc.html)

Similarly other files may also be generated based on some specific settings that may be used by a user. Actually there are 3 programs that are generally used for creating such tags files- `ctags` (probably the most popular), `etags` and `gtags`. AFAIK `ctags`, in its most basic setting, will generate a file named `tags`.

So, the question comes down to whether we want our gitignore to work with all these different settings or just the most common one?

Makman2: Ah okay those are different tagging systems, sounds okay :+1:

Makman2:
ack ecc9094605b08d53f2cf8f5a681c5a7385c66f94

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

Issue #4881, opened 2017-11-18
master fails on Fedora 26


SanketDG:
Fresh install of `coala` and `coala-bears` master on a new venv:

```
[WARNING][11:33:59] Implicit 'Default' section inheritance is deprecated. It will be removed soon. To silence this warning remove settings in the 'Default' section from your coafile. You can use dots to specify inheritance: the section 'all.python' will inherit all settings from 'all'.
[ERROR][11:34:02] An unknown error occurred. This is a bug. We are sorry for the inconvenience. Please contact the developers for assistance. During execution of coala an exception was raised. This should never happen. When asked for, the following information may help investigating:
[INFO][11:34:02] Exception was:
Traceback (most recent call last):

  File "/home/sanketdg/projects/coala/coalib/coala.py", line 84, in main
    args)

  File "/home/sanketdg/projects/coala/coalib/output/ConsoleInteraction.py", line 990, in show_bears
    print_bears(bears, show_description, show_params, console_printer, args)

  File "/home/sanketdg/projects/coala/coalib/output/ConsoleInteraction.py", line 950, in print_bears
    separators=(',', ': '))

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 201, in encode
    chunks = list(chunks)

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 325, in _iterencode_list
    yield from chunks

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 438, in _iterencode
    yield from _iterencode(o, _current_indent_level)

  File "/home/sanketdg/.pyenv/versions/3.6.1/lib/python3.6/json/encoder.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)

[...]

PR #5170, opened 2018-02-15
JSONEncoder.py: Prevent error from package_managers

anctartica:
coala -B --json command fails while accessing package_managers attribute. This should be handled in JSONEncoder.

Closes https://github.com/coala/coala/issues/4881

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


anctartica:
@Makman2 i have added a separate case for handling objects with the `package_managers` property in `JSONEncoder.py`. Do you think this is fine? The `coala -B --json` command is able to show the results although I still have to add the tests which is causing the build to fail.

jayvdb:
@gitmate-bot rebase

jayvdb:
Needs an extra test to have 100% test coverage.  See build logs.

coalib/output/JSONEncoder.py
```diff
@@ -21,6 +21,12 @@ def default(self, obj):
                 params = self._filter_params(
                     fdata.optional_params, fdata.non_optional_params)
                 return obj.__json__(**params)
+            elif 'package_managers' in dir(obj):
+                return {member: getattr(obj, member)
+                        for member in dir(obj)
+                        if not member == 'package_managers'
```
jayvdb: `not foo == bar` ?? try `!=` ?

However, this is the wrong level to be putting a special case for 'package_managers'.  That should be specified by the caller.  Can the `__json__` attribute (see above) be used to skip the package_managers and other undesirable bits?

If not, you can add explicit keywords to `create_json_encoder` for names to be skips, or some other mechanism of bypassing the items we dont want.

Another approach is modify the container of `package_managers` to to indicate that the attribute `package_managers` does not need to be serialised, because it will automatically be recreated again.  This is the `__setstate__` and `__getstate__` methods described at https://docs.python.org/2/library/pickle.html#object.__getstate__

anctartica: > If not, you can add explicit keywords to create_json_encoder for names to be skips, or some other mechanism of bypassing the items we dont want.
I looked into this and the only way seem to override the `iterencode` method which is not very clean approach. Also, I am not sure if this would be very reusable or required at all in the future. 

> Another approach is modify the container of package_managers to to indicate that the attribute package_managers does not need to be serialised, because it will automatically be recreated again. This is the __setstate__ and __getstate__
I think this would be a better solution to solve this problem. Also, then we'd have to use the pickle module rather than json here.

anctartica: One solution would be to add the `__json__` attribute to the `DistributionRequirement` class and there avoid adding the `package_managers`.
Another solution could be that we pass a list of elements that are to be avoided to `get_public_members()`
Other solutions I could find were to override `iterencode` in `JSONEncoder` which is not very clean or reusable.
> Another approach is modify the container of package_managers to to indicate that the attribute package_managers does not need to be serialised, because it will automatically be recreated again. This is the __setstate__ and __getstate__ methods described at https://docs.python.org/2/library/pickle.html#object.__getstate__

Not sure about this. Requires using pickle we get binary serialisation. So, do we output that?

jayvdb:
Wrt to pickle, we would not be using pickle, but enhancing our encoder to use `__getstate__` like pickle does. I expect the stdlib json encoder does this already

Issue #5205, opened 2018-02-25
Redefinition of the same function `format_lines()` in 3 files


nityeshaga:
The function `format_lines()` is defined in the exact same way in [ShowAppliedPatchesAction.py](https://github.com/coala/coala/blob/8db41c06a825ac7791005d5cb91291371bea3c5f/coalib/results/result_actions/ShowAppliedPatchesAction.py#L8), [ConsoleInteraction.py](https://github.com/coala/coala/blob/8db41c06a825ac7791005d5cb91291371bea3c5f/coalib/output/ConsoleInteraction.py#L117) and [coala_main.py](https://github.com/coala/coala/blob/8db41c06a825ac7791005d5cb91291371bea3c5f/coalib/coala_main.py#L46). (*in ShowAppliedPatchesAction.py version, there is a very minor difference- the amount of padding is 4 in it whereas others contain 5, but IMHO it is inconsequential*). 

We should keep only one copy, I suggest the `coala_main()` one because:
1. it is "coala_main"
2. it contains a some comment- `# type: (object, object, object) -> object`

We can then import the function from `coala_main()` in the other 2 files. This reduces redundancy of code.

nityeshaga:
cc @Makman2 

If this is a valid issue, can I work on it?

nityeshaga:
On second thought, `format_lines()` might really belong in [ConsoleInteraction.py](https://github.com/coala/coala/blob/8db41c06a825ac7791005d5cb91291371bea3c5f/coalib/output/ConsoleInteraction.py) file because it contains all the other functions relevant for console interactions.

Naveenaidu:
@nityeshaga The reason the format_lines is present in the coala_main.py is to print the options when the `apply_single` is true. Well the `format_lines()` does belong to ConsoleInteraction.py . Please have a look at https://github.com/coala/coala/pull/5198. Discussing this issue with @ishanSrt we decided to move the printing of the `apply_single` i.e which uses `format_lines()` from `coala_main.py`  into the consoleInteraction.py . So once this PR get's merged we wont we having a ` format_lines()`  function inside the `coala_main.py` . Then there will no redundancy in the code. :wink: . 

But maybe we can do that for `ShowAppliedPatchesAction.py`.

@Makman2  what is your opinion? Do you think this issue needs to be opened when it's getting fixed in the  https://github.com/coala/coala/pull/5198 .

Makman2:
[...]

PR #5215, opened 2018-02-27
Remove repeated definitions of format_lines()

nityeshaga:
Replaces the definitions of format_lines() in files- coala_main.py
and ShowAppliedPatchAction.py with import statements that import the
function from ConsoleInteraction.py.

Closes https://github.com/coala/coala/issues/5205

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


coalib/results/result_actions/ShowAppliedPatchesAction.py
```diff
@@ -32,6 +26,7 @@ def apply(self,
         FILE_DIFF_DICT_INDEX = 2
         SECTION_INDEX = 3
 
+        from coalib.output.ConsoleInteraction import format_lines
```
nityeshaga: The `import` statement had to be placed here instead of the usual - top of the file - because coala was encountering an error with circularly dependent import.

jayvdb: Then put the function in a different module which doesnt cause a circular dependency

nityeshaga: There's a circular dependency problem with `ShowAppliedPatches.py` regardless of whether I put the function in `ConsoleInteraction.py` or `coala_main.py`. Now, the function `format_lines()` is used only in these 3 modules-- `ShowAppliedPatches.py`, `ConsoleInteraction.py` and `coala_main.py`. 

Therefore, I IMHO I think that we need to put the function in one of these files only. Also, I don't think that it should be inside `ShowAppliedPatches.py`. 

Hence, the problem. :confused: 

Please suggest a way out of it @jayvdb

jayvdb: It is your task to find the solution.
I already know it.
And telling it to you doesn't help you learn.

nityeshaga:
I think that the `Interactions.py` module is the right home for the `format_lines()` function. ;-) So, I have moved its definition over there. 

I have also moved the definitions of `highlight_text()` and `color_letter()` in that file too as IMHO they would kind of fall in the same category as `format_lines()`. 

@jayvdb What do you think?

jayvdb:
get a green build. then we talk

nityeshaga:
Work In Progress.

jayvdb:
Please stop pushing until you have a version which passes the tests on your local machine.

nityeshaga:
Extremely sorry for the mess. :sweat_smile: 
> Please stop pushing until you have a version which passes the tests on your local machine.

I'll keep this in mind

coalib/output/Interactions.py
```diff
@@ -28,3 +45,35 @@ def fail_acquire_settings(log_printer, settings_names_dict):
 
         logging.error(msg)
         raise AssertionError(msg)
+
+
+def highlight_text(no_color, text, style, lexer=TextLexer()):
```
jayvdb: so now instead of `format_lines` being duplicated, `coalib/output/ConsoleInteraction.py:highlight_text` is being duplicated here.

How is this better?

nityeshaga: Sorry, I've fixed it now.
No duplicates of `highlight_text()` anymore :)

jayvdb: So, why the Interactions module?

coalib/results/result_actions/ShowAppliedPatchesAction.py
```diff
@@ -3,12 +3,7 @@
 from coalib.results.RESULT_SEVERITY import (
     RESULT_SEVERITY, RESULT_SEVERITY_COLORS)
 from pyprint.ConsolePrinter import ConsolePrinter
-
-
-def format_lines(lines, symbol='', line_nr=''):
-    def sym(x): return ']' if x is '[' else x
-    return '\n'.join('{}{:>5}{} {}'.format(symbol, sym(symbol), line_nr, line)
-                     for line in lines.rstrip('\n').split('\n'))
+from coalib.output.Interactions import format_lines
```
jayvdb: why do we even need to use format_lines in this module?

And are you sure that padding of 5 is not intentional, and 4 is good enough?

nityeshaga: > And are you sure that padding of 5 is not intentional, and 4 is good enough?

In ShowAppliedPatchesAction, `format_lines()` is used when [printing the `Severity`](https://github.com/coala/coala/blob/master/coalib/results/result_actions/ShowAppliedPatchesAction.py#L45) of the patch.
Here's how it looked (before this PR):
*Notice the line* `!    ![Severity: NORMAL]`
showappliedpatchaction_on_master
And here's how it looks now (after this PR):
showappliedpatchesaction_on_removeredun3
So, the version of `format_lines()` in this PR works like it was expected to IMO. It also passes all the original set of tests as they were, without any modification.

> why do we even need to use format_lines in this module?

Yes I agree, that was a bad decision on my part. :sweat_smile:
I wasn't sure if creating a new file to solve a *difficulty/low* issue was allowed or not. Now I think it would be better if I create a new file (`Formatting.py` maybe?) and put `format_lines()`, `color_letter()` and `highlight_text()` inside it. What do you think @jayvdb ?

nityeshaga: > what does format_lines do?

`format_lines` returns a formatted version of the string passed to it as a parameter. It formats the string to make it look elegantly in coala's CLI by adding a little prefix to it. 

> Why is that functionality needed in ShowAppliedPatchesAction.py?

`ShowAppliedPatchesAction` already uses it to format the line which displays the Severity of any applied action.

So, unless you are suggesting that we change how `ShowAppliedPatchesAction` shows its messages, I think that `format_lines` is the right tool for the job.

jayvdb: You still have not read `format_lines` and its usages carefully.
It is gobbledy-gook that people have copied and pasted without thinking.

nityeshaga: > You still have not read format_lines and its usages carefully.

:sweat_smile: I'm sorry, I didn't find anything else to read about it except the 4 line definition which was present in all 3 of the above mentioned files.

So, are you suggesting that we modify the definition of `format_lines()`? maybe make pass the amount of padding as a parameter? Or are you suggesting that we remove the function altogether?

jayvdb: Read those four lines very carefully.  Understand exactly what happens when it is called each time.

nityeshaga: Here's what I understand `format_lines()` is doing:
Each time the function is called with a string (`lines`), it splits that string at `\n`s, prefixes each split string with the `symbol` (passed to it as a parameter), adds a padding of 4, adds the `sym()` version of `symbol`, adds `line_nr` and finally joins them using `\n` and returns this joined version.
Also, `sym()` is a local function inside `format_lines()` which is just responsible to return the closing `]` in case the `symbol` that is passed is `[`. Otherwise, it returns the passed character itself.

Isn't that all?  :sweat_smile:

jayvdb: No I dont want you to tell me what it does in theory.  I already know the answers.

You need to see why in practise , in some of the real world scenarios where it is used, it isnt particularly useful.

Issue #5206, opened 2018-02-26
Tests fail to reach 100% code coverage locally, though CI build passes.


nvzard:
Here are the local logs: https://gist.github.com/nvzard/ef99b21d57a429b08760347f46edf12e

Makman2:
Task is to write a test filling the gap.

harshhx17:
Can this issue be assigned to me @Makman2 
:smile:

nvzard:
I was testing something on docker and found this,
Tests give 100% coverage when "coverage~=4.2" but fail to give 100% when "coverage~=4.3.4"
Here are the logs: https://pastebin.com/rzXyS6TM

harshhx17:
Thanks @nvzard :thinking:

adtac:
I swear there was another issue related to this from a long time ago (I introduced the bug, I should know). I tried searching and I can't find it. Guess this is the one now :D

PR #5216, opened 2018-02-27
CachingUtilitiesTest.py: Add test for local build

harshhx17:
This adds test to CachingUtilitiesTest.py for the code in
CachingUtilities.py which runs when coala is run for the first
time on the project

Fixes https://github.com/coala/coala/issues/5206

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


harshhx17:
Here are the new local build logs
https://gist.github.com/harshhx17/eef0dc45b04dec9a3bca931e8f97e72c

harshhx17:
@Makman2  can you please review it :smile:

tests/misc/CachingUtilitiesTest.py
```diff
@@ -66,6 +67,11 @@ def test_settings_change(self):
         settings_hash = get_settings_hash(sections)
         self.assertTrue(settings_changed(self.log_printer, settings_hash))
 
+        with patch('coalib.misc.CachingUtilities.pickle_load') as \
+                mocked_pickle_load:
+            mocked_pickle_load.return_value = {}
+            self.assertFalse(settings_changed(self.log_printer, settings_hash))
```
adtac: This does not solve the problem. It just patches the symptom.

To properly solve this, create a `tearDown` that deletes all side effects from the tests after the tests are done.

adtac: By this, I mean to say remove all keys related to tests from the pickle after the tests are done.

harshhx17:
I will look into it :smile:

harshhx17:
@adtac 
imo, Since I have not created any class variable, I do not think the variables related to the pickle tests will exist outside the scope of the patch call...
However, if i had used a patcher(and mocked CachingUtilities.pickle_load as self.mocked_pickle_load) in the setup(which would mock the pickle_loadd to all the tests), I would need to add that to the teardown so that it doesnot effect other tests....[which i have not done as it is not required by all the tests]

tests/misc/CachingUtilitiesTest.py
jayvdb: @harshhx17 replied at https://github.com/coala/coala/pull/5216#issuecomment-383381767

jayvdb: Replying to ^ , this isnt about what you have done or not done.

@adtac is asking you to re-create the *scenario* which this `if` is supposed to address.

Your code does force the use of this `if` branch, which does achieve 100% coverage, but it doesnt reproduce the scenario, so the objective of 100% coverage isnt obtained.

The line `settings_hash_db = pickle_load(None, 'settings_hash_db', {})` will already return `{}`, so you dont need to mock `pickle_load`.  You could mock something quite a bit deeper to cause `{}` to be returned.  That may not make @adtac happy, but at least you are more closely recreating the scenario.

@adtac is saying you should use `setUp` and `tearDown` to create the scenario, e.g. create a backup of the setting cache, remove the settings for this directory, and then restore the settings again after the test is completed.

As it is an invasive test process, I suggest putting it into its own test class.

Issue #4690, opened 2017-08-28
`'python' in Language['python 3']` should return True


adhikasp:
Title say it all. It is surprising otherwise.

gurpreetsingh00885:
Would you please care to explain a little?

andyy143:
would you elabrote it

rohanrk:
I'd like to work on this

PR #5303, opened 2018-04-01
Language.py: Fix __contains__ method

rohanrk:
The __contains__ method in the Language class only checks
if self.versions is a superset of item.versions. This change
adds a subset check so if self.versions is a subset
of item.versions, __contains__ will return True.

Closes https://github.com/coala/coala/issues/4690

### DIFF
--- a/coalib/bearlib/languages/Language.py
+++ b/coalib/bearlib/languages/Language.py
@@ -422,8 +422,11 @@ class Language(metaclass=LanguageMeta):
 
     def __contains__(self, item):
         item = Language[item]
+        item_versions = set(item.versions)
+        versions = set(self.versions)
         return (type(self) is type(item)
-                and set(item.versions).issubset(set(self.versions)))
+                and (item_versions.issubset(versions)
+                     or item_versions.issuperset(versions)))
 
     def __reduce__(self):
         return (Language.__getitem__, (str(self),))
--- a/tests/bearlib/languages/LanguageTest.py
+++ b/tests/bearlib/languages/LanguageTest.py
@@ -17,6 +17,18 @@ class LanguageTest(unittest.TestCase):
         cpp_unpickled = pickle.loads(cpp_str)
         self.assertEqual(str(cpp), str(cpp_unpickled))
 
+    def test_contains_method(self):
+        # Test alias
+        self.assertTrue('py' in Language[Language.Python])
+        # Test version
+        self.assertTrue('python' in Language[Language.Python == 3])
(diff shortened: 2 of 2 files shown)


coalib/bearlib/languages/Language.py
```diff
@@ -423,7 +423,8 @@ def __or__(self, other):
     def __contains__(self, item):
         item = Language[item]
         return (type(self) is type(item)
```
jayvdb: restructure all of this so that you can store `set(item.versions)` and `set(self.versions)` instead of building those sets multiple times.

jayvdb:
(also please rebase)

jayvdb:
(removing red cross, as it now looks good; but needs to be better performance)

coalib/bearlib/languages/Language.py
jayvdb: Still not done.

coalib/bearlib/languages/Language.py
```diff
@@ -422,8 +422,11 @@ def __or__(self, other):
 
     def __contains__(self, item):
         item = Language[item]
+        itemVers = set(item.versions)
```
jayvdb: we use `snake_case` variable names, per PEP8 guideline.

coalib/bearlib/languages/Language.py
```diff
@@ -422,8 +422,11 @@ def __or__(self, other):
...
+        versions = set(self.versions)
         return (type(self) is type(item)
-                and set(item.versions).issubset(set(self.versions)))
+                and (itemVers.issubset(versions)
```
jayvdb: The logic seems very strange:

x subset y or x superset y

what is the better way of saying that?

is that correct?

rohanrk: I think the logic is sound. 

I could also write it as ```x subset y or y subset x``` if it makes it clearer. Either all of the versions in the passed in item will be contained in self (e.g. 'python 3' in Language['python']) or all the versions in self will be contained in item (e.g. 'python' in Language['python 3']) 

if neither of these conditions are met then item isn't in self.

I can't think of a nicer way to write this. Maybe

```python
x | y == x or x | y == y
```

adhikasp:
Ack b33c192c 74693d59

jayvdb:
ack b33c192 74693d5

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

Issue #5188, opened 2018-02-19
persistent_hash: Missing support for collection types with not guaranteed iteration order


Makman2:
Like `dict` and `set`. `sorted` can be used to sort the data beforehand.

Makman2:
Blocked by https://github.com/coala/coala/issues/4350

palash25:
@Makman2 would like to work on this once it is unblocked

palash25:
although I'm not sure if it would work for nested dicts :confused:

palash25:
@Makman2 from what I can gather from the `generate_tasks()` method is that the task objects are tuples of a tuple of args and a dict of kwargs . Tuples preserve order so there won't be any problem hashing them the current method is sufficient for that. However in case of the kwargs dict since the order is not preserved maybe we can change the dict to a frozenset`frozenset(dict.items())` (assuming there are no duplicates) and then serialize it using `pickle.dumps()` and then find the sha1 hash.
I tried it, converted the dict to a frozentset and it didn't give me `unhashable type error`

palash25:
I am having trouble visualizing the task objects can you give me an example of one?

Makman2:
But how do you intend to pass kwargs with a non-mapping type? And even if our task's args and kwargs wouldn't be dicts, they could themselves contain dicts. And even if you use `frozenset`, the order isn't preserved across machines, and also not guaranteed to be preserved across coala runs :)

palash25:
I wasn't sure about the structure of kwargs and args. 

> And even if our task's args and kwargs wouldn't be dicts, they could themselves contain dicts.

This makes it even more difficult since now they could be nested dicts.

Makman2:
`pickle` has to do something similar to pickle special objects, you can likely get some inspiration from there :+1:

palash25:
Hey @Makman2 I found this module dill: https://github.com/uqfoundation/dill which is an extension of pickle and is under active development.

Your comment intrigued me about dill.

> And even if you use frozenset, the order isn't preserved across machines

 Their documentation states that 

>  In addition to pickling python objects, dill provides the ability to save the state of an interpreter session in a single command. Hence, it would be feasable to save a interpreter session, close the interpreter, ship the pickled file to another computer, open a new interpreter, unpickle the session and thus continue from the 'saved' state of the original interpreter session.

[...]

PR #5324, opened 2018-04-06
PersistentHash: Preserve Order

palash25:
Task objects are now sorted first before
computing their persistent hash

Closes https://github.com/coala/coala/issues/5188

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/core/PersistentHash.py
+++ b/coalib/core/PersistentHash.py
@@ -1,19 +1,44 @@
+from collections import Iterable
+from copy import deepcopy
 from hashlib import sha1
 import pickle
 
 
+def order(obj):
+    """
+    Recursively preserves the order of any sets or dicts
+    found in obj.
+    """
+    if isinstance(obj, set):
+        return sorted(obj)
+    if isinstance(obj, dict):
+        temp_obj = deepcopy(obj)
+        for k, v in temp_obj.items():
+            if isinstance(v, set) or isinstance(v, dict):
+                temp_obj[k] = order(v)
+        return sorted(temp_obj.items())
+    return obj
+
+
+def preserve_order(a, k):
+    arg_list = []
+    for ele in a:
+        arg_list.append(order(ele))
+
+    kwargs = sorted(({key: order(value) for key, value in k.items()}).items())
+    return tuple(arg_list), kwargs
+
+
 def persistent_hash(obj):
-    """
-    Calculates a persistent hash of the given object.
+    if isinstance(obj, Iterable):
+        if isinstance(obj, dict):
+            obj = sorted(obj.items())
+        if isinstance(obj, set):
+            obj = sorted(obj)
+        # Specific condition for identifying task objects
+        if (isinstance(obj[0], set) or isinstance(obj[1], dict)):
+            obj = preserve_order(obj[0], obj[1])
 
-    This hash method uses pickle (protocol 4) to serialize the given object and
-    hash the byte-stream. The hashing algorithm used is SHA-1.
-
-    :param obj:
-        The object to calculate the persistent hash for.
-    :return:
-        The persistent hash.
-    """
     fingerprint_generator = sha1()
     fingerprint_generator.update(pickle.dumps(obj, protocol=4))
     return fingerprint_generator.digest()
--- a/tests/core/PersistentHashTest.py
+++ b/tests/core/PersistentHashTest.py
@@ -14,3 +14,37 @@ class PersistentHashTest(unittest.TestCase):
         self.assertEqual(
             persistent_hash((1, 2, 3)),
             b'\xb5\xd6\xd7\xbeLD\x90\x9fz.\xae\xc4\xb9P\n\xf8\xf5\x03S\xb6')
+
+    def test_sets(self):
+        self.assertEqual(
+            persistent_hash({'a', '1', 'g'}),
+            b'\xbc\x99\x95\x97#F\x13<8\x1cK\x81\xf2KxQ@\xf6\x01%')
+
+    def test_plain_dicts(self):
+        self.assertEqual(
+            persistent_hash({'a': '1', 'g': '9'}),
+            b'D\xe8NL!\x8c\xec\xff\xf3\x93\xee=0K#\x1fVL\x06$')
+
+    def test_dicts(self):
+        self.assertEqual(
+            persistent_hash(((), {'a': '1', 'g': '9'})),
+            b'\x99E\xday\xd3\xc0;S\xb7\x01\xb1\xd6>F\xdb\xc3y\x12\xf7W')
+
+    def test_tuple_with_sets_and_dicts(self):
+        self.assertEqual(
+            persistent_hash((({'g', 'a'}, {'e': '7', 'a': '1'}), {})),
+            b'xvr\xa0\xd3\x9c\x125\x12H\xcf\x13\xc4\xba\xf5\x15Hz\xe2\x00')
+
+    def test_nested_dicts(self):
+        self.assertEqual(
+            persistent_hash(((), {'q': {'g': '1', 'b': '8'}, 'a': {},
+                                  'g': {'z', 'c', 'd'}, 'b': '8'})),
+            b'\xdaI\xb5S(wb\xd7\x82\xa0\x17cV-}\x1cL\x9d\xc1a')
+
+    def test_nested_tuples_and_dicts(self):
+        self.assertEqual(
+            persistent_hash((('a', 'b', {'g', 'b', '8'},
+                              {'g': '1', 'a': {}, 'b': '8'}),
+                             {'q': {'g': '1', 'a': '1'}, 'a': {},
+                              'g': {'z', 'd'}, 'b': '8'})),
+            b'\xa9z[U\xfa\xd1x\x95\x00\xf1,h%Y\xa2u\x87\xb0\xb2\x13')


coalib/core/PersistentHash.py
```diff
@@ -10,10 +55,14 @@ def persistent_hash(obj):
     hash the byte-stream. The hashing algorithm used is SHA-1.
 
     :param obj:
-        The object to calculate the persistent hash for.
+            The object to calculate the persistent hash for.
```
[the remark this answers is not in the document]
palash25: > create a newcomer issue.

I'm sorry I didn't get you, its not merged yet, how can I create a newcomer issue?
can't I just amend it, myself?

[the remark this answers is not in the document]
jayvdb: is this change useful?  if so, create a newcomer issue for someone else to fix it.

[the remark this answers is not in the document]
jayvdb: (if not, yes, remove it)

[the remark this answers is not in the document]
palash25: No it isn't really useful. I don't think a newcomer would benefit much from removing a couple of whitespaces. Amending it.

coalib/core/PersistentHash.py
```diff
@@ -1,7 +1,53 @@
+from collections import Iterable
 from hashlib import sha1
 import pickle
 
 
+def tuple_replace(tup, x, y):
+    tup_list = list(tup)
+    for element in tup_list:
+        if element == x:
+            tup_list[tup_list.index(element)] = y
```
Makman2: Use `list.index` instead to find elements. That's probably way faster since it should utilize C code.

coalib/core/PersistentHash.py
```diff
@@ -1,7 +1,53 @@
...
+    new_tuple = tuple(tup_list)
+    return new_tuple
```
Makman2: `return tuple(tup_list)` ;)

coalib/core/PersistentHash.py
```diff
@@ -1,7 +1,53 @@
...
+    sorted_args = args_tuple
+
+    # iterates over the tuple of positional arguments
+    # looks for sets and dicts, sort them and put them
+    # back into the tuple.
+    for element in sorted_args:
+        if isinstance(element, Iterable):
+            if any(element) and isinstance(element, dict):
+                sorted_args = tuple_replace(sorted_args, element,
+                                            dict(sorted(element.items())))
```
Makman2: `dict(sorted())`? Then you would effectively remove the sorting again :O

coalib/core/PersistentHash.py
```diff
@@ -1,7 +1,53 @@
...
+            if any(element) and isinstance(element, set):
```
Makman2: `any(element)` used twice.

Makman2: Should be put into a variable.

coalib/core/PersistentHash.py
```diff
@@ -1,7 +1,53 @@
...
+    # looks for sets and dicts within the kwargs dict
+    # sort each individual set or dict and puts them
+    # back them into the dict.
+    for key, value in sorted_kwargs.items():
+        if any(value) and isinstance(value, dict):
+            sorted_kwargs[key] = dict(sorted(value.items()))
+        if any(value) and isinstance(value, set):
+            sorted_kwargs[key] = sorted(value)
+
+    sorted_kwargs = dict(sorted(sorted_kwargs.items()))
```
Makman2: again you make the sorting obsolete again by converting the stuff back into a dict.

Makman2: Probably that's the cause why you receive different results on different computers in your test, I guess this is the crucial part here (and the other similar statement above)

coalib/core/PersistentHash.py
Makman2: Do you have to replace items actually? I mean just create a new tuple / list and further operate on that.

This has maybe some flaws, e.g. a tuple with `((1, 3),)` would have the same persistent hash as the dict `{1: 3}`, but maybe that's not so important for us? Or we could somehow modify the internal data so we can distinguish different object types. Anyway, not so important, but I wouldn't replace things inside instances.

coalib/core/PersistentHash.py
```diff
@@ -2,18 +2,50 @@
...
+
+    def order_dict(self):
+        for key, value in self.arg_dict.items():
+            self.dict[key] = order(value)
+        return sorted(self.dict.items())
+
+    def order_tuple(self):
+        for o in self.arg_tuple:
+            self.list.append(order(o))
+        return tuple(self.list)
```
Makman2: I guess a class isn't really necessary for those functions, is it? ;)

coalib/core/PersistentHash.py
```diff
@@ -2,18 +2,50 @@
...
+    Recursively preserves the order of any sets or dicts
+    found in obj.
     """
-    Calculates a persistent hash of the given object.
+    if isinstance(obj, set):
+        return sorted(obj)
+    elif isinstance(obj, dict):
+        for k, v in obj.items():
+            if isinstance(v, set) or isinstance(v, dict):
+                obj[k] = order(v)
```
Makman2: you modify the object in-place, that's not ok. Could you create a copy instead?

coalib/core/PersistentHash.py
```diff
@@ -2,18 +2,50 @@
...
-def persistent_hash(obj):
+def order(obj):
+    """
+    Recursively preserves the order of any sets or dicts
+    found in obj.
     """
-    Calculates a persistent hash of the given object.
+    if isinstance(obj, set):
+        return sorted(obj)
+    elif isinstance(obj, dict):
```
Makman2: since you return above, you can make this a normal `if` instead. (Btw: I've seen that many linters suggest to start a new `if` in case you return inside the first branch, so I tend to follow that recommendation :smile:)

coalib/core/PersistentHash.py
```diff
@@ -2,18 +2,50 @@
...
+        for k, v in obj.items():
+            if isinstance(v, set) or isinstance(v, dict):
+                obj[k] = order(v)
+        return sorted(obj.items())
+    else:
+        return obj
```
Makman2: Here too, `else` branch not necessary, `return obj` can be moved out.

coalib/core/PersistentHash.py
```diff
@@ -2,18 +2,50 @@
...
+        return sorted(self.dict.items())
+
+    def order_tuple(self):
+        for o in self.arg_tuple:
+            self.list.append(order(o))
+        return tuple(self.list)
+
+
+def persistent_hash(obj):
+    if isinstance(obj, tuple):
```
Makman2: why checking if it's a tuple before checking anything else? What if I directly pass it a dict? I would like to see it working for them too^^

Makman2: Hm just thinking, may this work?
```python
obj = order(obj)

# do pickle things
```

coalib/core/PersistentHash.py
```diff
@@ -1,19 +1,42 @@
...
+def order(obj):
+    """
+    Recursively preserves the order of any sets or dicts
+    found in obj.
     """
-    Calculates a persistent hash of the given object.
+    if isinstance(obj, set):
+        return sorted(obj)
+    if isinstance(obj, dict):
+        temp_obj = obj
```
Makman2: hehe nope, that doesn't create a copy ;) all python objects are references, so you need to do a first-level copy (aka shallow copy) in this case (because you already work recursively here), see `copy.copy`. However, you need to be careful with dicts referencing each other, that would cause an infinite loop (please add a test for that case!).
Maybe you can also work with deep-copies, behaviour for deep-copying is overridable as well. So you can only make it copy dictionaries itself, but not any other objects, then you totally avoid recursion problems.

Makman2: Or maybe order the dict beforehand into a list/tuple structure, and then iterate over it and replace dict/set values with their ordered counterparts.

coalib/core/PersistentHash.py
```diff
@@ -1,19 +1,42 @@
...
-        The object to calculate the persistent hash for.
-    :return:
-        The persistent hash.
-    """
+def preserve_order(a, k):
+    arg_list = []
+    kwarg_dict = {}
+
+    for ele in a:
+        arg_list.append(order(ele))
```
Makman2:
```python
args = tuple(order(ele))
kwargs = sorted({key: order(value) for key, value in k.items()})

return args, kwargs
```

coalib/core/PersistentHash.py
```diff
@@ -1,19 +1,42 @@
...
+    for key, value in k.items():
+        kwarg_dict[key] = order(value)
+
+    return (tuple(arg_list), sorted(kwarg_dict.items()))
+
+
+def persistent_hash(obj):
+    if isinstance(obj, Iterable):
+        if (isinstance(obj[0], set) or isinstance(obj[1], dict)):
+            obj = preserve_order(obj[0], obj[1])
```
Makman2: I still don't understand those conditions, why do you access `obj[0]` and `obj[1]`?

coalib/core/PersistentHash.py
palash25: I tried it but it gives me a wrong hash.

palash25: I have made the changes to the kwargs code but kept the args one as it is.

Makman2:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

Issue #5176, opened 2018-02-17
requirements.txt: Remove redundant version part of setuptools


Makman2:
`setuptools>=21.0.0` is equal to `setuptools>=21`.

chetasr:
I would like to take this issue

arpit7714:
can i work on this!

amitsin6h:
@Makman2 please review my PR #5181

monsij:
@amitsin6h Get yourself assigned first before you start working on an issue.

amitsin6h:
@Makman2 can you assign me.

Makman2:
@arpit7714 was commented first, so I have to assign him. Sorry @amitsin6h

himanshuc3:
@Makman2 I want to start working on this since no one's assigned!

jayvdb:
This is a bit more difficult because of `.misc/check_setuptools.py`

ananyaarun:
@jayvdb , @Makman2  can i work on this issue if no one is presently ?

utkarsh2102:
@jayvdb, can I take this up?
The one who was working on it seems to be inactive; been 8 months :P

li-boxuan:
@himanshuc3 Are you still working on your PR? If not we'll close your PR and assign to others.

jayvdb:
The low difficulty part of this is check_setuptools.py, and that has now been moved to the mobans repository.

A new issue should be created there to enhance it, or better integrate this check.  Until then, this is blocked.

jayvdb:
c.f. https://github.com/coala/coala/pull/5635 which will lock the sync of these files.

PR #5325, opened 2018-04-06
requirements.txt: setuptools>=21.0.0 changed to setuptools>=21.

himanshuc3:
``setuptools>=21.0.0`` was redundant.
So, it is changed to ``setuptools>=21`` in this commit.

Closes https://github.com/coala/coala/issues/5176


jayvdb:
You'll need to fix `.misc/check_setuptools.py` also

jayvdb:
Abandoned and now sort of redundant.

Issue #5348, opened 2018-04-09
Reactivate osx job on master only


jayvdb:
https://github.com/coala/coala/issues/3611 entirely removed osx.

Would be good to reactivate,  but not run on PRs

PR #5353, opened 2018-04-09
.travis.yml: Reactivate osx job on master only

Hori75:
Reactivates osx job, on master only, not to run on PRs.

Closes https://github.com/coala/coala/issues/5348

up your stuff for merge!

### Checklist

- [X] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [X] I ran coala over my code locally. (*All commits have to pass
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
```
jayvdb: we need `sudo: false` here , and then add `sudo: true` in the definition of the osx job

.travis.yml
```diff
@@ -7,8 +6,16 @@ python:
...
 jobs:
   include:
-    # OSX build removed due to long build startup delays
-    # Restore matrix job entry from d2d67fab to test OSX
+    - os: osx
+      osx_image: xcode6.4
+      language: bash
+      python: 3.6
+      env:
+        - DYLD_LIBRARY_PATH=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/
```
jayvdb: I think this is not needed any more.  It was for llvm support, which has been moved to bears only.

jayvdb:
Also can you show a build log with this job running (i.e. create a new fork on your travis-ci , and remove the restrictions so that it runs.

Hori75:
Osx build on test branch:
https://travis-ci.org/Hori75/coala/builds/396022044

.travis.yml
```diff
@@ -35,8 +35,12 @@ jobs:
       python: 3.6
       if: branch = master AND type = push
 
-    # OSX build removed due to long build startup delays
-    # Restore matrix job entry from d2d67fab to test OSX
+    - os: osx
+      sudo: true
+      python: 3.6
+      language: bash
```
bekicot: Why is it set to bash?

bekicot:
Looks like the problem is unrelated to your changes. It is coming from pre deactivation.

From the travis you shown. I think the error is because it tries to replace default python (python2) to brew installed python3. I tried several month ago in my mac to use python3 as default to call `python` and it causes problem. I would suggest to not doing it, but instead using `pip3 install coala --user`

It may fix the issue.

.travis.yml
Hori75: https://travis-ci.org/Hori75/coala/jobs/396015861

I have tried without setting it to bash and I got command not found errors

jayvdb: `pip` and `python` need to point to the python3 which is being installed.

I think you need to add `/usr/local/opt/python/libexec/bin` to the PATH
c.f.
https://travis-ci.org/Hori75/coala/jobs/396904486#L1411

then these lines shouldnt be necessary
```
sudo ln -sf $(which python3) $(which python)
sudo ln -sf $(which pip3) $(which pip 2>/dev/null || echo /usr/bin/pip)
```

jayvdb:
The problems are in the OSX specific lines.  They worked previously (c.f. travis links found at https://github.com/coala/coala/pull/3380), but now they do not.  This is because Travis doesnt support Python3 on OSX, so we need to assist the 'bash' environment look like the 'python' Linux environments.

Issue #3785, opened 2017-02-24
Docs: Let `setup.py docs` generate them from the current repo, not installation


Makman2:
As a dev I have often a stable version or some development version from pip installed. When executing `setup.py docs`, it grabs the docs from the installed package, not the repo.

jayvdb:
Another aspect of this is that the `docs` command is currently not dependent on the build phase, so it is possible to run `setup.py docs` without a built coalib, and see

```
Running Sphinx v1.4.9

Exception occurred:
  File "conf.py", line 64, in <module>
    from coalib.misc.Constants import VERSION
ModuleNotFoundError: No module named 'coalib'
```

jayvdb:
Yup, confirmed this is a problem.

jayvdb:
Solution is https://gitlab.com/coala/mobans/issues/8

PR #5426, opened 2018-04-27
Netlify docs

jayvdb:
Demo of https://gitlab.com/coala/mobans/merge_requests/1


RaiVaibhav:
Hello John why it is block ?, and I think this step should in Readme, a newcomer type issue :P

netlify.toml
```diff
@@ -0,0 +1,3 @@
+[build]
+  publish = "docs/_build/html"
+  command = "pip install -r docs-requirements.txt && export PYTHONPATH=$(pwd) && python setup.py docs"
```
virresh: I may be wrong, but shouldn't we not be having the need to export ```PYTHONPATH``` if sphinx has its own conf file and the repository is always relative to it ?

virresh: Spotted something,
We do [set the path in conf.py](https://github.com/coala/coala/blob/master/docs/conf.py#L25)
but since we do the [import before setting the path]((https://github.com/coala/coala/blob/master/docs/conf.py#L20)), the path being set is actually useless

virresh: Oh
So does setting that path serve any other purpose then ?
If not then should we remove that as well (possible the removal would be a non-standard inclusion, but we get rid of some un-used code maybe) ?

jayvdb: I suspect that the only reasons for it was so that "setup.py docs" would work.

It used to work.  See 687ec1bc8 which broke it.  And https://github.com/coala/coala/issues/3785 for the original issue.

Probably broken atm.  I've raised https://gitlab.com/coala/mobans/issues/8

Until we have a verified solution again, that line should stay in the local docs/conf.py .  Worst case it will disappear when we mobanise conf.py.

jayvdb:
@gitmate-bot rebase

netlify.toml
jayvdb: Well, turns out that there is a issue which should be resolved in mobans.  See https://gitlab.com/coala/mobans/merge_requests/11

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/coala/pull/5426/commits/9b3da26ec49c2428ca70df915208a0d50c299dc5

jayvdb:
(previously approved by @virresh)

jayvdb:
@gitmate-bot ff

Issue #5115, opened 2018-01-27
Throw error for conflicting cli args: `config` and `no_config`


khanchi97:
coala should throw error if we are using both cli options( -c and -I ) at same time, as these options are conflicting to each other.
`coala -c .confile --no-config`
or
`coala -c .confile -I`

Nosferatul:
unassigned due to inactivity.

raduklb:
I would like to work on this issue.

PR #5458, opened 2018-05-12
CliParsing.py: Throw error for conflicting arguments

raduklb:
Add conflict check for arguments `-config` and `--no-config`

Fixes https://github.com/coala/coala/issues/5115

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/tests/parsing/CliParsingTest.py
+++ b/tests/parsing/CliParsingTest.py
@@ -71,3 +71,8 @@ class CliParserTest(unittest.TestCase):
         with self.assertRaisesRegex(SystemExit, '2') as cm:
             check_conflicts(sections)
             self.assertEqual(cm.exception.code, 2)
+
+        sections = parse_cli(arg_list=['--no-config', '--config', '.coafile'])
+        with self.assertRaisesRegex(SystemExit, '2') as cm:
+            check_conflicts(sections)
+            self.assertEqual(cm.exception.code, 2)
--- a/coalib/parsing/CliParsing.py
+++ b/coalib/parsing/CliParsing.py
@@ -116,9 +116,11 @@ def check_conflicts(sections):
         if (
                 section.get('no_config', False) and
                 (section.get('save', False) or
-                 section.get('find_config', False))):
+                 section.get('find_config', False) or
+                 str(section.get('config', 'input')) != 'input')):
             ArgumentParser().error(
-                "'no_config' cannot be set together 'save' or 'find_config'.")
+                "'no_config' cannot be set together with 'save', "
+                "'find_config' or 'config'.")
 
         if (
                 not section.get('json', False) and


jayvdb:
Your commit message has `-config` ; needs to be `--config`

tests/parsing/CliParsingTest.py
```diff
@@ -71,3 +71,8 @@ def test_check_conflicts(self):
         with self.assertRaisesRegex(SystemExit, '2') as cm:
             check_conflicts(sections)
             self.assertEqual(cm.exception.code, 2)
+
+        sections = parse_cli(arg_list=['--no-config', '-config'])
```
tallicamike: I think that here as well it should be `--config` as, if I'm not mistaken, even though the test is passing it isn't actually checking that functionality. The command fails with an exception, but it fails because it doesn't recognize `-config`. I think it should be `--config` and a .coafile should be provided as argument to the `--config` option. I don't know which is the best way to do this: maybe create a temporary .coafile? As in: https://github.com/coala/coala/blob/ac5aa59e7fa27ce13fca57ed40218477e568307a/tests/parsing/ConfParserTest.py#L45-L47

pareksha: Yes, it should be `--config` but I don't think a temporary .coafile is required. The test is meant to only check whether conflicting arguments are given.

tallicamike: In this way I think it would fail because it doesn't have a configuration file provided before checking the conflict between `--config` and `--no-config`. So, if I am right the test will pass but will still not do the intended checking. On the other hand I don't know if a .coafile is needed, any file should do in the end indeed, @pareksha. What are your thoughts on this?

raduklb:
@pareksha Can you have a look now, please?

tests/parsing/CliParsingTest.py
anctartica: looking at the other tests I don't think it is required.

tests/parsing/CliParsingTest.py
```diff
@@ -71,3 +71,8 @@ def test_check_conflicts(self):
         with self.assertRaisesRegex(SystemExit, '2') as cm:
             check_conflicts(sections)
             self.assertEqual(cm.exception.code, 2)
+
+        sections = parse_cli(arg_list=['--no-config', '--config', '.coafile'])
```
anctartica: Look at other tests in `test_check_conflicts`. None of them specify `.coafile`. It is not needed

raduklb: I think a .coafile is needed because if you run coala just with --no-config and --config an error of usage will be generated.

anctartica: "error of usage" is also generated for this test case:
https://github.com/raduklb/coala/blob/d4e3198b8fb5f3e9a356a71b171cf2e54c3e48c9/tests/parsing/CliParsingTest.py#L57?

anctartica: ok the `--config` expects one argument so another argument is needed after this.

jayvdb:
ack https://github.com/coala/coala/pull/5458/commits/d4e3198b8fb5f3e9a356a71b171cf2e54c3e48c9

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

Issue #5466, opened 2018-05-17
Appveyor builds fail on pip install


jayvdb:
https://ci.appveyor.com/project/coala/coala/build/1.0.12921/job/7qk3gl78ax38vc3i

```
python --version
Python 3.4.4
python -c "import struct; print(struct.calcsize('P') * 8)"
32
%CMD_IN_ENV% pip install --cache-dir=C:\pip_cache -r requirements.txt -r test-requirements.txt
Using default MSVC build environment for 32 bit architecture
Executing: pip install --cache-dir=C:\pip_cache -r requirements.txt -r test-requirements.txt
ERROR: To modify pip, please run the following command:
c:\python34\python.exe -m pip install --cache-dir=C:\pip_cache -r requirements.txt -r test-requirements.txt
Command exited with code 1
```

I guess the problem is `pip>=9.0.0,<10` in `test-requirements.txt`

PR #5467, opened 2018-05-17
appveyor.yml: Run pip install as Python module

jayvdb:
pip now complains about being downgraded when being
executed via the `pip` command on Windows.

Closes https://github.com/coala/coala/issues/5466


jayvdb:
@gitmate-bot ff

Issue #5359, opened 2018-04-10
Add module docstrings


jayvdb:
coalib usually doesnt have any module docstrings.
They are used to describe the content & scope of the module succinctly.

palash25:
Do you want the docstrings to each module added as a separate commit or just one commit for all the module docstrings?

Also would like to be assigned to this (this will help me get a deeper understanding of all the core modules).

jayvdb:
@palash25 , I havent assigned this to you as you have some large chunks of work that we need to get closed.  This is another large chunk of work, and a PR will create a lot of thinking for Makman mostly.

We should probably split this up into segments so the PRs dont result in unicorns.
Also PRs for smaller areas could be reviewed by different 'module owners'.  e.g. there is an aspects team who could do the aspects modules.  etc.

jayvdb:
Created and assigned four issues for various components.

jayvdb:
a module docstring is in addition to the other docstrings. It summarises the objective of the module in the first line, and only one line, then has a blank line, and then can contain an overview of the contents of the module.

It shouldnt list all objects in it, as their documentation will appear underneath the module level docstring. Instead it should show how the module is intended to be used.

jayvdb:
In most cases, it is best to start this process by only adding the first line, but with the intention of adding more later, so that syntax should be:

```py
"""
Blah module to do foo.
"""
```

Then more can be added later.

jayvdb:
Will be useful to look for examples of this done well.

https://github.com/PyCQA/pydocstyle/tree/master/src/pydocstyle has good one liners for their modules, but lacks the details.

https://github.com/python/cpython/tree/master/Lib is the canonical example, but their style varies as each component owner can choose if/how implementation of pep257 is done, and very importantly they have sufficient manpower to build separate documentation not autogenerated only from the source.  We do not (yet?).

https://github.com/python/cpython/blob/master/Lib/operator.py is a good example of a basic module, with a bit of detail to help contextualise the reader.

https://github.com/python/cpython/blob/master/Lib/calendar.py is a another nice example of detail added.

https://github.com/python/cpython/blob/master/Lib/xml/__init__.py is a good example for a package which contains sub-packages.

Issue #2821, opened 2016-10-03
Option to skip files without code


jayvdb:
In Python 2, an empty `__init__.py` file is meaningful without any content. Some packages like to leave it completely empty. However this will cause errors with some linters, such as `pydocstyle` and the proposed `CopyrightBear` (`flake8-copyright` has a problematic file size strategy for handling this.).

There may be other languages where a file should be excluded from a linter based on whether the file is empty, or nearly empty. If so, a generic approach should be created for determining emptiness.

jayvdb:
I havent been able to find another language where an empty file can have a meaning.  I have added the Python meaning to https://www.rosettacode.org/wiki/Empty_program#Python

However that page shows many languages consider an empty file to be valid, in which case an empty file might be used in test data for parsers of those languages, and running static analysis tools on the empty files is best avoided.

Makman2:
To exclude files without code just exclude it from the file list.
```
ignore = **/__init__.py
```

As empty files in other languages have no meaning, they should be removed from codebase anyway then.

Imo there's no need to have some special checks (which need to be coded into the core which makes it even more complicated) for file emptiness. If linters have problems with empty files, imo it's their problem (I mean an empty file can happen, it's not some not-happening-at-all-case ;D), and to avoid buggy behaviour inside coala it's possible to ignore specific files like above.

adtac:
I think this should be a bear-specific flag. For instance, the `CopyrightBear` could verify if the file is empty pass it to `flake8-copyright` only if it has content. IMO this is too specific to be a coala wide option.

jayvdb:
"Source file has no code." is one of the simplest possible static analysis of a file. It would belong in `coala`. Specific use of that information might belong in the relevant bear only, but that is not relevant for this issue.

Makman2:
So this is a bear proposal?

Then yes agree, this would be a useful analysis to detect unnecessary empty files in a project.

sils:
LineCountBear could have an option for that and an ignore grep e.g. for
init and stuff

jayvdb:
> So this is a bear proposal?

No, it is support for very basic static analysis. A 'this file contains code' function is very generic, and provides a solid foundation for most other types of analysis to be built on.

Another off-shoot of this would be 'what language is this file', which is relevant as a `.php` file might contain only HTML or even XML, and indeed a `.html` file could be HTML or XML.

adtac:
[...]

PR #5475, opened 2018-05-22
Formatting: Add tastes to `FileLength` aspect

pareksha:
This adds `min_file_length` and `include_blank_lines`
as tastes to `FileLength` aspect.

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/aspects/Formatting.py
+++ b/coalib/bearlib/aspects/Formatting.py
@@ -107,6 +107,9 @@ class FileLength:
     max_file_length = Taste[int](
         'Maximum number of line for a file',
         (999,), default=999)
+    min_file_length = Taste[int](
+        'Minimum number of lines for a file.',
+        (1,), default=1)
 
 
 @Formatting.subaspect


jschwarzwalder:
The description of minimum line length is not clear.

In some languages will it always return 1 since it is  possible to combine multiple lines all into one line.

For example what would the min line length be of this code sample?

```javascript 
function fizzBuzz()
{ 	
//for each number between 1 &100
for(var i=1;i<=100;i++)
{
   //print Fizz if divisible by 3
 	 	if(i%3 === 0)
  {
 			 print('Fizz'); 	
  	}
   //print Buzz if divisible by 5
 		if(i%5 === 0)
  {
 		  print('Buzz');
 		}
   // otherwise print the number
 		if(i%3 !== 0 && i%5 !== 0)
  {
 			 print(i);
 		}
 	}
 }

pareksha:
@jschwarzwalder you are exactly on point. This is the thing me and John have been debating on, all this time. Most probably, implementation of line-counting algorithms will be required to correctly predict the complexity of the code and accordingly show the number of lines. 

I had proposed `min_line_length` and `include_blank_lines` tastes as part of the `FileLength` aspect considering the current design of the bear (`LineCountBear` https://github.com/coala/coala-bears/blob/master/bears/general/LineCountBear.py) and it feels now that the bear itself needs a redesign on this plus supporting aspects on top of it.

I think you should see https://github.com/coala/cEPs/pull/148#discussion_r191038337 and furthermore https://github.com/coala/cEPs/pull/148#discussion_r192719097 to know more.

Lay your opinions too on this. 
And IMO this PR needs to be closed.

pareksha:
Yes, `min_file_length` is compatible with the present usability of the bear and so, it is ok for now. I also agree that `include_blank_lines` is sort of problematic for now. Once the bear is migrated, we can think of stretching its functionality.

I have removed `include_blank_lines`. Please review :smiley:

jayvdb:
@pareksha , please remove "This is part of GSoC 2018 project." (or any other reference to GSoC) from commit message, and create issues for your PRs, and using them in your commit messages.

That information is not relevant to the reader of the commit.

Your issues should refer to the cEP PR, or to the cEP once it is merged.

jschwarzwalder:
If the only use case currently is to detect if the file length is 0, then why not make a taste for is_empty?

I would still have many questions about what `min_file_length` means. If we include it, it should qualify what would be counted. 

If the `min_file_length` ignores **comments** and **blank lines** that can be helpful to know how many actual lines of code exist, but without both those qualifiers It might be too ambiguous for developers to use. 

Does it minify the code? Does it simply count the current number of existing lines? Does it give suggestions for potential ways to reduce the number of lines with things the ternary operations?

jayvdb:
Python has a requirement of `__init__.py` files, and these files are often empty. c.f. https://github.com/coala/coala/issues/2821

A min length of 1 will break coala repos, but we could fix that by adding a module docstring to every empty `__init__.py`.  c.f. https://github.com/coala/coala/issues/5359

jayvdb:
`max_file_length` defines a rule for a hard *physical line* count maximum.

The `LineCountBear` would implement this taste by detecting files which exceed it, and producing an *error*.  That bear is a detection bear, not a fixing bear.

We do not have any bear which attempts to fix files which exceed a maximum line count.  That would be difficult.

We do not have any bear which attempts to fix files which have insufficient line count.  However this is more feasible to implement.  We could have a bear which automatically fixed a *minimum* line count **for Python**, by automatically adding a dummy docstring.  It would also be possible to fix it for any scripting language by adding a 'hashbang', but that would be dumb as the otherwise empty file probably isn't runnable.

jayvdb:
Also please investigate the `suggested_values`.  What is it used for?  Why are you using this specific value for it.  It is not intuitive.

pareksha:
`suggested_values` are more or less used to suggest values (not all values that the user can put, but what the user should put to get the desirable/better results) as I see here https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/Formatting.py#L253 and at many other places too. What I meant here by `(1,)` was that the values can be 1 or any other `int` but now I am also feeling that there should be more than one suggested value (and not just the default value). So, I think `suggested_values` in both `min_file_length` and `max_file_length` need to be changed to have more integers.  

How about `(499, 999, 4999, 9999,)` in `max_file_length`  
and `(1, 50, 100,)` in `min_file_length` as suggested_values?

jayvdb:
ack https://github.com/coala/coala/pull/5475/commits/bd23043f0242966d6d1c38753572fbbff11cf57d

jayvdb:
@gitmate-bot ff

Issue #3530, opened 2017-01-11
Add FileProxy


Makman2:
See https://github.com/coala/cEPs/blob/master/cEP-0002.md#the-file-proxy

The `FileProxy` shall be responsible for providing files and their contents to the bears and bear-writers. It shall cache contents as efficient as possible, and also provide different interfaces to the file (like utf8-decoded with and without line-endings or a binary version).

Mariatta:
Is this issue to track the implementation of it? Perhaps label accordingly. (Size? Difficulty?)
Thanks.

palash25:
@Makman2 if FileProxy is going to be used to cache contents of the file I think we should also add the last modified timestamp (using `os.path.getmtime`) as a property of the FileProxy so that it can be used for caching purposes.
I looked into the cEP and it doesn't mention timestamp as a property.

Makman2:
> I think we should also add the last modified timestamp (using os.path.getmtime) as a property of the FileProxy so that it can be used for caching purposes.

Exactly what I had in mind. This will significantly speed up the caching system. And yeah the cEP doesn't mention it, but it wasn't written with NextGen-Core in mind :)

palash25:
@Makman2 I had a question about the `__hash__()` function in the proposed `FileProxy` for what purpose do we need a hash of the file. For caching currently we use modification timestamps of files using `os.path.gettime`. If its not for caching what is it for then?

Makman2:
We actually don't need it for our purposes, but it still might be useful to implement the method. For our caching specifically what's important is to override pickling. The FileProxy is a persistent object and has to be treated carefully to make everything work properly.

PR #5485, opened 2018-05-28
Add FileFactory.py

palash25:
This adds the FileFactory class. Objects of this
class will be used to represent files and
replace the file contents in the file dict.

Closes https://github.com/coala/coala/issues/3530

cc @Makman2 @fneu @shreyans800755 

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/io/FileFactory.py
+++ b/coalib/io/FileFactory.py
@@ -1,0 +1,131 @@
+import os
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('name', 'timestamp')
+class FileFactory:
+    """
+    The ``FileFactory`` is a class to provide different views on a file.
+    It contains the following information about the file:
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in raw format.
+        * The file in the form of a string UTF-8 decoded from the raw format.
+        * The lines of the file in the form of a tuple.
+
+    Two ``FileFactory`` objects are considered equal if they have the
+    same filename and timestamp.
+
+    To initialize a ``FileFactory`` object for a file:
+
+    >>> import os
+    >>> import tempfile
+    >>> temp = tempfile.NamedTemporaryFile(delete=False)
+    >>> temp.write(bytes('This is a test file.', 'UTF-8'))
+    20
+    >>> temp.close()
+    >>> ff = FileFactory(temp.name)
+
+    File indices start with zero.
+    To retrieve a single line:
+
+    >>> ff.get_line(0)
+    'This is a test file.'
+
+    Get all the lines in the file at once:
+
+    >>> ff.lines
+    ('This is a test file.',)
+
+    Get the file contents in raw format:
+
+    >>> ff.raw
+    b'This is a test file.'
+
+    Get the file as a string:
+
+    >>> ff.string
+    'This is a test file.'
+
+    Get the filename:
+
+    >>> ff.name == temp.name
+    True
+
+    To conveniently iterate over the lines of a file:
+
+    >>> for line in ff:
+    ...     print(line)
+    This is a test file.
+
+    >>> os.remove(ff.name)
+    """
+
+    def __init__(self, filename):
+        """
+        :param filename:
+            The filepath.
+        """
+        self._filename = os.path.abspath(filename)
+        self._timestamp = os.path.getmtime(self._filename)
+
+    def get_line(self, line):
+        """
+        :param line:
+            The line number from which the string is to be retrieved.
+            The index starts from zero. An ``IndexError`` is raised if
+            the line number provided is out of range.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        return self.lines[line]
+
+    @property
+    def lines(self):
+        """
+        :return:
+            A tuple containing the lines of the file.
+        """
+        return tuple(self.string.splitlines())
+
+    @property
+    def raw(self):
+        """
+        :return:
+            The file contents as a byte sequence.
+        """
+        with open(self._filename, 'rb') as fp:
+            return fp.read()
+
+    @property
+    def string(self):
+        """
+        :return:
+            The file contents as a string UTF-8 decoded.
+        """
+        return self.raw.decode()
+
+    @property
+    def name(self):
+        """
+        :return:
+            The absolute filepath.
+        """
+        return self._filename
+
+    @property
+    def timestamp(self):
+        """
+        :return:
+            The last modified timestamp of the file.
+        """
+        return self._timestamp
+
+    def __iter__(self):
+        """
+        :return:
+            An iterator iterating over all lines in the file.
+        """
+        return iter(self.lines)
--- a/tests/io/FileFactoryTest.py
+++ b/tests/io/FileFactoryTest.py
@@ -1,0 +1,52 @@
+import os
+import unittest
+
+from coalib.io.FileFactory import FileFactory
+
+
+def get_path_components(filepath):
+    """
+    Splits the filepath into components to provide a unique
+    test results that passes on all CIs.
+    """
+    return os.path.normpath(filepath).split(os.sep)
+
+
+class FileFactoryTest(unittest.TestCase):
+
+    def setUp(self):
+        file_factory_test_dir = os.path.join(os.path.split(__file__)[0],
+                                             'FileFactoryTestFiles')
+
+        self.test_file = os.path.join(file_factory_test_dir, 'test1.txt')
+        self.other_test_file = os.path.join(file_factory_test_dir, 'test2.txt')
+        self.uut = FileFactory(self.test_file)
+        self.other_file_factory = FileFactory(self.other_test_file)
+
+    def test_equal(self):
+        self.assertEqual(self.uut, FileFactory(self.test_file))
+        self.assertNotEqual(self.uut, self.other_file_factory)
+
+    def test_iter(self):
+        self.assertEqual(list(self.uut), ['This is a test file.'])
+
+    def test_line(self):
+        self.assertEqual(self.uut.get_line(0), 'This is a test file.')
+        with self.assertRaises(IndexError):
+            self.uut.get_line(1)
+
+    def test_lines(self):
+        self.assertEqual(self.uut.lines, ('This is a test file.',))
+
+    def test_raw(self):
+        self.assertEqual(self.uut.raw, b'This is a test file.')
+
+    def test_string(self):
+        self.assertEqual(self.uut.string, 'This is a test file.')
+
+    def test_timestamp(self):
+        self.assertEqual(self.uut.timestamp, os.path.getmtime(self.test_file))
+
+    def test_name(self):
+        self.assertEqual(get_path_components(self.uut.name)[-4:],
+                         ['tests', 'io', 'FileFactoryTestFiles', 'test1.txt'])
--- a/tests/io/FileFactoryTestFiles/test1.txt
+++ b/tests/io/FileFactoryTestFiles/test1.txt
@@ -1,0 +1,1 @@
+This is a test file.
--- a/tests/io/FileFactoryTestFiles/test2.txt
+++ b/tests/io/FileFactoryTestFiles/test2.txt
@@ -1,0 +1,1 @@
+Another test file.


coalib/io/FileFactory.py
```diff
@@ -0,0 +1,120 @@
...
+    def line(self, line):
+        """
+        :param line:
+            The line number from which the string is to be retrieved.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        return linecache.getline(self._filename, line)
+
+    def lines(self):
```
Makman2: make it maybe a property?

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        To get all the lines in the file in form of a list.
+        """
+        self.lines = linecache.getlines(self._filename)
```
Makman2: you don't return something ;)

Makman2: Ah wait, now I understand. You try to cache things here. However, this is highly inconsistent because the first time you have to call it (without a return value even), and next time you can't call it and must access it like a field. Please create consistent API ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
+import linecache
+import os
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('_filename', '_stamp')
+class FileFactory:
+    """
+    The FileFactory is a wrapper object to replace the actual file contents
```
shreyans800755: wrapper `class`

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+    Make the object into an iterable when required
+
+    >>> for line in fp:
+    ...     print(line)
+    This is a test file.
+    <BLANKLINE>
+
+    Retrieve a single line:
+
+    >>> fp.line(1)
```
shreyans800755: Are line indices starting with 1 ? If so, why ?

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+    ('This is a test file.\\n',)
+    """
+
+    def __init__(self, filename):
+        """
+        :param filename:
+            The absolute filepath.
+        """
+        self._filename = os.path.abspath(filename)
+        self._parent = os.path.abspath(os.path.dirname(self._filename))
```
shreyans800755: Do we really need this variable as data member ?
This is easily deducible whenever required.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+        self._stamp = os.path.getmtime(self._filename)
```
shreyans800755: Why do we need this as data member ?
This might not be same when we want to fetch it. And we'll again have to call `getmtime` function.
I don't see any point in storing this.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+            The line number from which the string is to be retrieved.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        return linecache.getline(self._filename, line)
+
+    @property
+    def lines(self):
+        """
+        To get all the lines in the file in form of a list.
```
shreyans800755: return value seems to be a tuple and not a list. I guess you don't mean python list, do you ?

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        return tuple(linecache.getlines(self._filename))
+
+    def binary(self):
+        """
+        :return:
+            The file contents in binary format.
+        """
+        with open(self._filename, 'rb') as fp:
+            return fp.read()
```
shreyans800755: This doesn't really give file contents in binary format. It returns file contents in raw format.
binary doesn't seem to be the right word.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,115 @@
...
+    >>> fp.lines
+    ('This is a test file.\\n',)
+    """
+
+    def __init__(self, filename):
+        """
+        :param filename:
+            The absolute filepath.
+        """
+        self._filename = os.path.abspath(filename)
```
shreyans800755: We should take file name in any format - relative/absolute. And then convert it to absolute if its relative while storing it in self._filename

coalib/io/FileFactory.py
palash25: I see your point but @Makman2 also wanted to compare factory objects based on their filename and timestamp both

If we are going to compare the object based on the timestamp then we would need this as a data member.

Maybe I can move this to this data member to its own function that way whenever it is called it will store the most recent time stamp but I wonder whether moving this out of the `__init__` method would still make the `@generate_eq(_stamp)` work or not.

coalib/io/FileFactory.py
palash25: It was a list before but after the review I changed it to tuple and forgot to update the docstring. Will update it thanks.

coalib/io/FileFactory.py
palash25: I think its because of `linecache`'s it counts lines from 1

coalib/io/FileFactory.py
palash25: We might need it in phase 2 for the ignore directory functionality though I'm still not sure as I need to have a lot of discussion about its design with Makman so I was thinking about letting it stay there and if we don't need it in phase 2 then I will remove it.

coalib/io/FileFactory.py
palash25: So you want me to put a check whether the path is relative or not and then assign it to the data member?

coalib/io/FileFactory.py
palash25: I just tested moving the timestamp to its own function and it causes problems with `generate_eq('_filename', '_stamp')` because it cannot find `_stamp` in the `__init__` method

Makman2: The `timestamp` is essential to caching. It effectively encodes information if a file has been changed or not.

`timestamp` might not be up-to-date, but it will be at least the earliest state we can access. We can always have external programs modifying the file in the meanwhile, and by using the timestamp as early as possible, we have a safety measure that automatically causes our caches to revalidate.

Makman2: @palash25 regarding `generate_eq`: Just use the property names, I guess that works better

coalib/io/FileFactory.py
Makman2: 1. Choose a convention that you think makes sense. coala's `Result`s have a one-based convention, but usually lists, tuples, array-like-things are zero-based due to technical reasons (e.g. see for-loops).
2. Document it! :) This is extremely important for the user to know.

My feeling: I tend to zero-based notation.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def raw(self):
+        """
+        :return:
+            Opens the file in binary mode and returns its contents.
+        """
+        with open(self._filename, 'rb') as fp:
+            return fp.read()
+
+    def string(self):
```
Makman2: -> `@property`

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+            The line number from which the string is to be retrieved.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        return linecache.getline(self._filename, line+1)
+
+    @property
+    def lines(self):
+        """
+        To get all the lines in the file in form of a tuple.
```
Makman2: Be consistent with documentation, use `:return:` as well

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        return tuple(linecache.getlines(self._filename))
+
+    def raw(self):
```
Makman2: -> `@property`

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, filename):
+        """
+        :param filename:
+            The filepath.
+        """
+        self._filename = os.path.abspath(filename)
+        self._stamp = os.path.getmtime(self._filename)
+
+    def line(self, line):
```
Makman2: Maybe call it `getline`, like `linecache`?

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+class FileFactory:
+    """
+    The FileFactory is a wrapper class to replace the actual file contents
+    from the file-dict. The object contains the following information about
+    the file:
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in the form of a string.
+        * The file in the form of a list.
```
Makman2: not tuple? ;)
Btw, I would also add that it's `UTF-8` decoded, that's the difference between strings/lists and raw format ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+        * The file in raw format.
+
+    The equality of the object is checked just based on its filename and
+    timestamp.
+
+    To initialize a FileFactory object for a file:
```
Makman2: Enclose code terms like `FileFactory` into double backticks so they get rendered as code :+1:

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,111 @@
...
+
+    >>> fp = FileFactory("./tests/io/FileFactoryTestFiles/test.txt")
```
Makman2: Use single quotes, that's the preferred quotation for strings that don't contain single backticks

Makman2: eh single quotes* :3

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('name', 'timestamp')
+class FileFactory:
+    """
+    The ``FileFactory`` is a wrapper class to replace the actual file contents
+    from the ``file-dict``. The object contains the following information about
+    the file:
```
Makman2: Not sure we should mention the file-dict part here. I would rather keep it inside the cEPs, so once file-dicts are history, nobody will ever wonder what this means.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in the form of a string.
+        * The file (UTF-8 decoded) in the form of a tuple.
```
Makman2: you should say that these are all lines inside the tuple (not that people think each char :sweat_smile: )

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+@generate_eq('name', 'timestamp')
+class FileFactory:
+    """
+    The ``FileFactory`` is a wrapper class to replace the actual file contents
+    from the ``file-dict``. The object contains the following information about
+    the file:
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in the form of a string.
```
Makman2: Maybe move the "(UTF-8 decoded)" part here, the string also is decoded from the raw content, so it should be mentioned as soon as it's becoming relevant.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
+import os
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('name', 'timestamp')
+class FileFactory:
```
Makman2: I'm not sure if we should just call it `File` :sweat_smile: But let's shift that discussion after merge :)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    >>> fp = FileFactory('./tests/io/FileFactoryTestFiles/test.txt')
+
+    Make the object into an iterable when required
+
+    >>> for line in fp:
+    ...     print(line)
+    This is a test file.
+
+    File indicies start with zero in case of
```
Makman2: indic~~i~~es

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+        * The file in the form of a string.
+        * The file (UTF-8 decoded) in the form of a tuple.
+        * The file in raw format.
+
+    The equality of the object is checked just based on its filename and
+    timestamp.
+
+    To initialize a ``FileFactory`` object for a file:
+
+    >>> fp = FileFactory('./tests/io/FileFactoryTestFiles/test.txt')
```
Makman2: ping -> Tempfiles ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+    >>> fp.string
+    'This is a test file.'
+
+    Get the filename:
+
+    >>> import tempfile
+    >>> t = tempfile.NamedTemporaryFile()
+    >>> f = FileFactory(t.name)
+    >>> f.name == t.name
+    True
```
Makman2: that should be used above, see my "ping"-comment ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def get_line(self, line):
+        """
+        :param line:
+            The line number from which the string is to be retrieved.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        content = self.lines
+        return content[line]
```
Makman2: -->
```python
return self.lines[line]
```

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @property
+    def lines(self):
+        """
+        :return:
+            File contents in the form of a tuple.
+        """
+        with open(self._filename) as fp:
+            return tuple(fp.readlines())
```
Makman2: reuse other fields:
```python
return self.string.splitlines()
```

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+    following information about the file:
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in raw format.
+        * The file in the form of a string UTF-8 decoded from the raw format.
+        * The lines of the file in the form of a tuple.
+
+    The equality of the object is checked just based on its filename and
+    timestamp.
```
Makman2: --> Two `FileFactory` objects are conisdered equal if they have the same filename and timestamp.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
+import os
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('name', 'timestamp')
+class FileFactory:
+    """
+    The ``FileFactory`` is a wrapper class whose objects contain the
```
Makman2: What does it wrap actually? Maybe don't say it's a wrapper class, just say it provides different views on a file.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    >>> fp.get_line(0)
+    'This is a test file.'
+
+    Get all the lines in the file as a list:
+
+    >>> fp.lines
+    ('This is a test file.',)
+
+    Get the file contents in raw mode:
```
Makman2: > Get the *raw* file contents ~~in raw mode~~:

I wouldn't mention that it's a "mode", could confuse people^^

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    >>> fp.string
+    'This is a test file.'
+
+    Get the filename:
+
+    >>> fp.name == temp.name
+    True
+
+    Make the object into an iterable when required
```
Makman2: missing colon :)

Makman2: And:

~~when required~~

That's redundant ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+            Opens the file in binary mode and returns its contents.
+        """
+        with open(self._filename, 'rb') as fp:
+            return fp.read()
+
+    @property
+    def string(self):
+        """
+        :return:
+            The file contents in string format.
```
Makman2: as a string ~~in string format~~

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        :return:
+            The last modified timestamp of the file.
+        """
+        return self._stamp
+
+    def __iter__(self):
+        """
+        :return:
+            The contents of the file as an iterator.
```
Makman2: Mention that it iterates over lines.

coalib/io/FileFactory.py
Makman2: Maybe don't say "Make it an iterator", but say "Conveniently iterate over lines:"

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    >>> fp.name == temp.name
+    True
+
+    Make the object into an iterable when required
+
+    >>> for line in fp:
+    ...     print(line)
+    This is a test file.
+    """
```
Makman2: Delete the temp file again ;)

Makman2: If you delete it, leave another whitespace before last code, so the user sees that's just cleanup and not belonging to the iterator functionality:
```
>>> ...
... print(line)
This is a test file.

>>> temp.delete()
```

jayvdb:
Failure in AppVeyor.

palash25:
@Makman2 appveyor is failing due to some permission error while accessing the temporary file created in the docstring. Is there a work around for this? I have been tweaking the dir path of the temp file but couldn't get it to pass.

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class FileFactoryTest(unittest.TestCase):
+
+    def setUp(self):
+        file_factory_test_dir = os.path.join(os.path.split(__file__)[0],
+                                             'FileFactoryTestFiles')
+
+        self.test_file = os.path.join(file_factory_test_dir, 'test.txt')
+        self.file_factory_object = FileFactory(self.test_file)
```
Makman2: Optional: To keep things shorter, people refer to the thing to test as `uut` = unit under test. But I also have nothing against longer more descriptive names :)

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,47 @@
...
+        for f in self.file_factory_object:
+            self.assertEqual(type(f), str)
+
+    def test_line(self):
+        self.assertEqual(self.file_factory_object.get_line(0),
+                         'This is a test file.')
+
+    def test_lines(self):
+        file_tuple = self.file_factory_object.lines
+        self.assertEqual(file_tuple, ('This is a test file.',))
```
Makman2: I would inline `file_tuple`, then the test layout looks the same in all places ;)

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,47 @@
...
+                         FileFactory(self.test_file))
+
+    def test_iter(self):
+        self.file_factory_object.lines
+        for f in self.file_factory_object:
+            self.assertEqual(type(f), str)
+
+    def test_line(self):
+        self.assertEqual(self.file_factory_object.get_line(0),
+                         'This is a test file.')
```
Makman2: What if you do `get_line(1)`? Can you test that too?

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
+import os
+
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('name', 'timestamp')
+class FileFactory:
+    """
+    The ``FileFactory`` is a class to provide different views on a file.
+    The objects of this class contains the following information about the file:
```
Makman2: ~~The objects of this class~~ **It** contains [...]

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
...
+    >>> temp.close()
+    >>> fp = FileFactory(temp.name)
+
+    File indices start with zero.
+    To retrieve a single line:
+
+    >>> fp.get_line(0)
+    'This is a test file.'
+
+    Get all the lines in the file as a list:
```
Makman2: as a *tuple* ;)

Makman2: But maybe say
> Get all the lines in the file at once:

Then you don't specify the actual return type, it's some object where you have access to all lines :) In our case it's a tuple to guarantee immutability, but that is an implementation detail to provide a reliable and safe API.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
...
+
+    To initialize a ``FileFactory`` object for a file:
+
+    >>> import os
+    >>> import tempfile
+    >>> temp = tempfile.NamedTemporaryFile(delete=False)
+    >>> temp.write(bytes('This is a test file.', 'UTF-8'))
+    20
+    >>> temp.close()
+    >>> fp = FileFactory(temp.name)
```
Makman2: just wondering: Why is the variable called `fp`, and not `ff`, or maybe even `file`? Or `file_factory`? :P

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
...
+        :param filename:
+            The filepath.
+        """
+        self._filename = os.path.abspath(filename)
+        self._timestamp = os.path.getmtime(self._filename)
+
+    def get_line(self, line):
+        """
+        :param line:
+            The line number from which the string is to be retrieved.
```
Makman2: I would also mention here that the index is zero-based. If I am a newbie, I would look at the documentation of the function that I use directly, and usually not read the whole class-docs to see that there some detail is mentioned that applies here only ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
...
+        :return:
+            A tuple containing the lines of the file.
+        """
+        return tuple(self.string.splitlines())
+
+    @property
+    def raw(self):
+        """
+        :return:
+            Opens the file in binary mode and returns its contents.
```
Makman2: > Opens the file

That's not a description of the return value :P If you want to mention that (I wouldn't, because we will later improve caching here and then things get complicated :3), then as normal description :+1:

Makman2: Looking at `def string`, I would make the description similar:
> The file contents as a byte sequence.

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        with open(self._filename, 'rb') as fp:
+            return fp.read()
+
+    @property
+    def string(self):
+        """
+        :return:
+            The file contents as a string.
```
Makman2: You could mention that the string was UTF-8 decoded from the file.

coalib/io/FileFactory.py
palash25: I still sometimes get confused between proxy and factory :stuck_out_tongue:

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,133 @@
...
+    The ``FileFactory`` is a class to provide different views on a file.
+    It contains the following information about the file:
+
+        * The filename (absolute file path).
+        * The modification timestamp of the file.
+        * The file in raw format.
+        * The file in the form of a string UTF-8 decoded from the raw format.
+        * The lines of the file in the form of a tuple.
+
+    Two ``FileFactory`` objects are conisdered equal if they have the
```
Makman2: typo ;) --> `considered`

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,46 @@
...
+                                             'FileFactoryTestFiles')
+
+        self.test_file = os.path.join(file_factory_test_dir, 'test.txt')
+        self.uut = FileFactory(self.test_file)
+
+    def test_equal(self):
+        self.assertEqual(self.uut, FileFactory(self.test_file))
+
+    def test_iter(self):
+        self.uut.lines
```
Makman2: redundant statement

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,46 @@
...
+        for f in self.uut:
```
Makman2: `f` -> `line`

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,46 @@
...
+            self.assertEqual(type(f), str)
```
Makman2: I would rather do
```python
self.assertEqual(list(self.uut), ['This is a test file.'])
```
Easier and better test, because you assert for the result ;)

coalib/io/FileFactory.py
```diff
@@ -0,0 +1,133 @@
...
+        :param line:
+            The line number from which the string is to be retrieved.
+            The index starts from zero.
+        :return:
+            The retrieved string at the given line number from the file.
+        """
+        try:
+            return self.lines[line]
+        except IndexError:
+            return None
```
Makman2: Do you want to catch that? Maybe let the caller catch that^^ And document that an `IndexError` is raised in case of an invalid line.

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual(self.uut.raw, b'This is a test file.')
+
+    def test_string(self):
+        self.assertEqual(self.uut.string, 'This is a test file.')
+
+    def test_timestamp(self):
+        self.assertEqual(self.uut.timestamp, os.path.getmtime(self.test_file))
+
+    def test_name(self):
+        if os.name == 'posix':
```
Makman2: Hm I dislike such OS-dependent switches.
1. If you already do it, assume that the posix-way is the "right" way ;) Windows' backslashes have been always a pain. So revert the condition: `if os.name == 'windows': ...`
2. can we utilize `os.path` functions to get a representation of the path that is unique across all OSes? Something like invoking `os.path.split` recursively and retrieving a list of path components.
   ```python
   path = get_path_components(self.uut.path)
   self.assertEqual(path, ['tests', 'io', 'FileFactoryTestFiles', 'test.txt'])
   ```

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,53 @@
+import os
+import unittest
+
+from coalib.io.FileFactory import FileFactory
+
+
+def get_components(filepath):
+    filepath = filepath[-38:]  # extract a path common on all CIs
```
Makman2: `-38`? :O what's that for a hardcoded value :O

Makman2: rather assert below with
```
assertEqual(get_components(real)[-5:], [last, 5, elements, ..., ...])
```

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,53 @@
+import os
+import unittest
+
+from coalib.io.FileFactory import FileFactory
+
+
+def get_components(filepath):
```
Makman2: --> `get_filepath_components`

tests/io/FileFactoryTest.py
Makman2: The comment can be also made a simple docstring ;)

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def setUp(self):
+        file_factory_test_dir = os.path.join(os.path.split(__file__)[0],
+                                             'FileFactoryTestFiles')
+
+        self.test_file = os.path.join(file_factory_test_dir, 'test.txt')
+        self.uut = FileFactory(self.test_file)
+
+    def test_equal(self):
+        self.assertEqual(self.uut, FileFactory(self.test_file))
```
Makman2: and inequality? ;)

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_iter(self):
+        for line in self.uut:
```
Makman2: you don't need that `for`-loop ;)

tests/io/FileFactoryTest.py
Makman2: or `get_path_components`

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_string(self):
+        self.assertEqual(self.uut.string, 'This is a test file.')
+
+    def test_timestamp(self):
+        self.assertEqual(self.uut.timestamp, os.path.getmtime(self.test_file))
+
+    def test_name(self):
+        path_components = get_components(self.uut.name)
+        self.assertEqual(path_components,
```
Makman2: Optional: Inline `path_components`
```python
self.assertEqual(get_components(self.uut.name),
                 [...])
```

jayvdb:
Failures on all CI

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,58 @@
...
+
+from coalib.io.FileFactory import FileFactory
+
+
+def get_path_components(filepath):
+    """
+    Splits the filepath into components to provide a unique
+    test results that passes on all CIs.
+    """
+    filepath = filepath[-39:]
```
Makman2: why are doing `-39`? :sweat_smile:

Makman2: Ah to prevent the timeout. But really debug that function, and you'll see you have done some mistake here ;) `/path` recurses into nowhere ;)

tests/io/FileFactoryTest.py
```diff
@@ -0,0 +1,59 @@
...
+    test results that passes on all CIs.
+    """
+    if filepath[0] == '/' or filepath[0] == '\\':
+        filepath = filepath[1:]
+    components = []
+    while filepath != '':
+        head, tail = os.path.split(filepath)
+        components.append(tail)
+        filepath = head
+    return components
```
Makman2: Actually sorry for maybe not telling earlier, but I've googled a bit around, and there's actually this possibility:
```python
os.path.normpath(path).split(os.sep)
```

Makman2:
ack b06dfe44fa99ffeb997f6f0a4bf302938fb9734c

Makman2:
@gitmate-bot ff

PR #5492, opened 2018-06-03
aspects: Create `map_ambiguous_setting_to_aspect`

pareksha:
Create decorator `map_ambiguous_setting_to_aspect` for
mapping bear settings to aspects in cases where data_types
and values are different in bear settings and aspects.
A typical example is `use_spaces` (bear setting) and
`indent_type` (taste of `Indentation` aspect).

This is part of `Convert Bear To Aspects` GSoC 2018 project.

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/aspects/decorators.py
+++ b/coalib/bearlib/aspects/decorators.py
@@ -49,3 +49,56 @@ def map_setting_to_aspect(**aspectable_setting):
         return _new_func
 
     return _func_decorator
+
+
+def map_ambiguous_setting_to_aspect(**aspectable_settings):
+    """
+    Convert bear settings to use aspects in which there is some ambiguity.
+
+    The ambiguity is because of differing values (maybe either name or even
+    data-types) in bear settings and their corresponding aspect tastes.
+
+    This decorator is used by ``Bear.run()`` to automatically map and
+    override the value in the bear setting with their equivalent taste
+    as provided in the parameter.
+
+    The order of setting override from the highest to lowest is:
+    - Explicit setting
+    - Explicit aspect/taste default (if aspect is activated in Section)
+    - Aspect/taste default (if aspect is activated in Section)
+    - Setting default (in bear's run argument)
+
+    :param aspectable_settings:
+        A dictionary in which keys are the bear settings and the corresponding
+        value for each key is a tuple containing an equivalent taste as first
+        value and a list of tuples with matching bear settings' and aspects'
+        value as the second value.
+    """
+    def _func_decorator(func):
+        @wraps(func)
+        def _new_func(self, *args, **kwargs):
+            if not self.section.aspects:
+                return func(self, *args, **kwargs)
+
+            aspects = self.section.aspects
+            for arg, aspect_value in aspectable_settings.items():
+                # Explicit bear setting takes priority over aspects
+                if arg in self.section:
+                    continue
+
+                taste, aspect_settings = aspect_value
+                aspect_instance = aspects.get(taste.aspect_name)
+                if aspect_instance:
+                    value = aspect_instance.tastes[taste.name]
+                    for value_pair in aspect_value[1]:
+                        if value_pair[0] == value:
+                            kwargs[arg] = value_pair[1]
+
+            return func(self, *args, **kwargs)
+
+        # Keep metadata
+        _new_func.__metadata__ = FunctionMetadata.from_function(func)
+
+        return _new_func
+
+    return _func_decorator
--- a/tests/bearlib/aspects/DecoratorsTest.py
+++ b/tests/bearlib/aspects/DecoratorsTest.py
@@ -4,6 +4,7 @@ from coalib.bearlib.aspects import (
     AspectList,
     get as get_aspect,
     map_setting_to_aspect,
+    map_ambiguous_setting_to_aspect,
 )
 from coalib.bears.LocalBear import LocalBear
 from coalib.settings.Section import Section
@@ -16,14 +17,27 @@ class RunDecoratedBear(LocalBear):
         remove_unreachable_code=get_aspect('UnreachableCode'),
         minimum_clone_tokens=get_aspect('Clone').min_clone_tokens,
     )
+    @map_ambiguous_setting_to_aspect(
+        use_spaces=(get_aspect('Indentation').indent_type,
+                    [('space', True), ('tab', False)]),
+    )
     def run(self,
             remove_unreachable_code: bool = False,
             minimum_clone_tokens: int = 10,
+            use_spaces: bool = True,
             ):
-        return [remove_unreachable_code, minimum_clone_tokens]
+        return [('remove_unreachable_code', remove_unreachable_code),
+                ('minimum_clone_tokens', minimum_clone_tokens),
+                ('use_spaces', use_spaces),
+                ]
 
 
 class MapSettingToAspectTest(unittest.TestCase):
+
+    EXPECTED = {'remove_unreachable_code': False,
+                'minimum_clone_tokens': 10,
+                'use_spaces': True,
+                }
 
     def setUp(self):
         self.section = Section('aspect section')
@@ -35,7 +49,10 @@ class MapSettingToAspectTest(unittest.TestCase):
             get_aspect('Clone')('py', min_clone_tokens=30),
         ])
         result = self.bear.execute()
-        self.assertEqual([True, 30], result)
+        expected = self.EXPECTED.copy()
+        expected['remove_unreachable_code'] = True
+        expected['minimum_clone_tokens'] = 30
+        self.assertEqual(expected, dict(result))
 
     def test_setting_priority(self):
         self.section.aspects = AspectList([
@@ -47,21 +64,57 @@ class MapSettingToAspectTest(unittest.TestCase):
         self.section.append(
             Setting('minimum_clone_tokens', 40))
         result = self.bear.execute()
-        self.assertEqual([False, 40], result)
+        expected = self.EXPECTED.copy()
+        expected['minimum_clone_tokens'] = 40
+        self.assertEqual(expected, dict(result))
 
     def test_partial_mapping(self):
         self.section.aspects = AspectList([
             get_aspect('UnreachableCode')('py'),
         ])
         result = self.bear.execute()
-        self.assertEqual([True, 10], result)
+        expected = self.EXPECTED.copy()
+        expected['remove_unreachable_code'] = True
+        self.assertEqual(expected, dict(result))
 
     def test_no_mapping(self):
         self.section.aspects = AspectList([])
         result = self.bear.execute()
-        self.assertEqual([False, 10], result)
+        expected = self.EXPECTED.copy()
+        self.assertEqual(expected, dict(result))
 
     def test_skip_mapping(self):
         self.section.aspects = None
         result = self.bear.execute()
-        self.assertEqual([False, 10], result)
+        expected = self.EXPECTED.copy()
+        self.assertEqual(expected, dict(result))
+
+
+class MapAmbiguousSettingToAspectTest(unittest.TestCase):
+
+    EXPECTED = {'remove_unreachable_code': False,
+                'minimum_clone_tokens': 10,
+                'use_spaces': True,
+                }
+
+    def setUp(self):
+        self.section = Section('aspect section')
+        self.bear = RunDecoratedBear(self.section, None)
+
+    def test_mapping(self):
+        self.section.aspects = AspectList([
+            get_aspect('Indentation')('py', indent_type='tab'),
+        ])
+        result = self.bear.execute()
+        expected = self.EXPECTED.copy()
+        expected['use_spaces'] = False
+        self.assertEqual(expected, dict(result))
+
+    def test_setting_priority(self):
+        self.section.aspects = AspectList([
+            get_aspect('Indentation')('py', indent_type='tab'),
+        ])
+        self.section.append(Setting('use_spaces', True))
+        result = self.bear.execute()
+        expected = self.EXPECTED.copy()
+        self.assertEqual(expected, dict(result))
--- a/coalib/bearlib/aspects/__init__.py
+++ b/coalib/bearlib/aspects/__init__.py
@@ -6,7 +6,10 @@ from pkgutil import iter_modules
 from types import ModuleType
 
 from .base import aspectbase
-from .decorators import map_setting_to_aspect
+from .decorators import (
+    map_setting_to_aspect,
+    map_ambiguous_setting_to_aspect,
+)
 from .exceptions import (AspectTypeError, AspectNotFoundError,
                          MultipleAspectFoundError)
 from .meta import aspectclass
@@ -23,6 +26,7 @@ __all__ = ['Root', 'Taste', 'TasteError',
            'aspectclass', 'aspectbase', 'AspectTypeError',
            'AspectNotFoundError', 'MultipleAspectFoundError',
            'AspectList', 'map_setting_to_aspect',
+           'map_ambiguous_setting_to_aspect',
            ]
 
 


jayvdb:
tests are failing.  See Circle CI logs.

pareksha:
Spent two days looking into the failing tests, I found that tests were considering `Formatting` aspect as in the docstring here https://github.com/coala/coala/pull/5492/files#diff-5c296e616e678d90af0811c1971749c4L18 not the actual ones defined in `Formatting.py` which was the reason tests were failing with `AspectNotFound` error. It's a relief that I was able to detect the problem here, else it would have worried me while migrating bears. 

As the docstring is nothing but an example to illustrate the design and structure of aspects (the concept of aspects and subaspects), so I think changing the name won't be an issue here.

Tests are passing but I am not able to reach 100% test coverage yet and I am not able to figure out why (I think I included all possibilities). Anyways, I'll see to it.

coalib/bearlib/aspects/root.py
```diff
@@ -15,9 +15,9 @@ class Root(aspectbase, metaclass=aspectclass):
     >>> from coalib.bearlib.aspects import Taste
 
     >>> @Root.subaspect
-    ... class Formatting:
+    ... class Format:
```
jayvdb: why are you renaming this?

jayvdb: related to https://github.com/coala/coala/pull/5492#issuecomment-396062741

jayvdb: I am pretty confident with your summary of the problem there.

What I suspect is that the doctests are not cleaning up after themselves.

Please create a new branch for investigating this.  Revert this rename so the problem re-occurs, and then change the pytest configuration to disable doctests.

If the error goes away when the doctests are disabled, we have 100% confirmed the doctest are polluting the Root aspect.

Then you can create an issue about this.
It is a multi-faceted issue.
One part to investigate is what happens when two subaspects are registered with the same name..?
Why isnt there an error when the second one (i.e. doctests and the real one) is registered?

The real solution should include for the doctest in this module to delete the `Formatting` that they created. (I am not sure how that can be done -- deleting aspects may not have been planned, so you may need to delete objects from a dictionary somewhere.)
Then we know how to un-do damage caused by doctest.

But a workaround is to create a separate PR "related to" the issue , which gives this subaspect a different name.  `Format` isnt a good new name.  Give it a crazy name, so we will always recognise it when it appears in logs unexpectedly. ;-)

pareksha: @jayvdb error goes away when the doctests are disabled in pytest. Now the tests are passing (although ultimately failing due to incomplete coverage).

I have created the issue here https://github.com/coala/coala/issues/5518.

About the multi-aspect part, I think you need to see https://github.com/coala/coala/blob/936a5d82a7ce6df5fac30ef39f08f2d2e53e3205/coalib/bearlib/aspects/__init__.py#L99. We already have `MultipleAspectFoundError` but it's worth noting that it's not raised for the two `Formatting` aspects (one of doctests and the actual one). This needs to be investigated.

I like the idea of the workaround, issue lies here https://github.com/coala/coala/blob/master/coalib/bearlib/aspects/root.py#L18

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -16,11 +17,20 @@ class RunDecoratedBear(LocalBear):
         remove_unreachable_code=get_aspect('UnreachableCode'),
         minimum_clone_tokens=get_aspect('Clone').min_clone_tokens,
     )
+    @map_ambiguous_setting_to_aspect(
+        use_spaces=(get_aspect('Indentation').indent_type,
+                    [('space', True), ('tab', False)]),
+        blanklines_at_EOF=(get_aspect('NewlineAtEOF').newline_at_EOF,
```
jayvdb: blanklines != newline *token*.  They are orthogonal.  They can not be mapped sensibly, and doing it in your test code makes the test logic confusing.

A newline token at EOL is terminating the last line with a newline token.

The *meaning* of that newline token varies between context and GUI editors, but it is not a blank line.

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -16,11 +17,20 @@ class RunDecoratedBear(LocalBear):
...
+                           [(True, 1), (False, 0)]),
+    )
     def run(self,
             remove_unreachable_code: bool = False,
             minimum_clone_tokens: int = 10,
+            use_spaces: bool = True,
+            blanklines_at_EOF: int = 1,
             ):
-        return [remove_unreachable_code, minimum_clone_tokens]
+        return [remove_unreachable_code, minimum_clone_tokens,
```
jayvdb: return a `dict` of these, so that the tests become more self explanatory.

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -47,21 +57,60 @@ def test_setting_priority(self):
...
     def test_no_mapping(self):
         self.section.aspects = AspectList([])
         result = self.bear.execute()
-        self.assertEqual([False, 10], result)
+        self.assertEqual([False, 10, True, 1], result)
 
     def test_skip_mapping(self):
         self.section.aspects = None
+        self.section.append(
+            Setting('use_spaces', False))
```
jayvdb: this needs to be one the next line ??

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -47,21 +57,60 @@ def test_setting_priority(self):
...
+        self.section.aspects = AspectList([
+            get_aspect('Indentation')('py', indent_type='tab'),
+            get_aspect('NewlineAtEOF')('py', newline_at_EOF=False)
+        ])
+        self.section.append(
+            Setting('use_spaces', False))
+        result = self.bear.execute()
+        self.assertEqual([False, 10, False, 0], result)
 
     def test_no_mapping(self):
```
jayvdb: class `MapAmbiguousSettingToAspectTest` doesnt need `no_mapping` or `skip_mapping` tests :/

jayvdb:
commit message should be `aspects: ..`

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -16,15 +17,28 @@ class RunDecoratedBear(LocalBear):
...
+    )
     def run(self,
             remove_unreachable_code: bool = False,
             minimum_clone_tokens: int = 10,
+            use_spaces: bool = True,
             ):
-        return [remove_unreachable_code, minimum_clone_tokens]
+        return [('remove_unreachable_code', remove_unreachable_code),
+                ('minimum_clone_tokens', minimum_clone_tokens),
+                ('use_spaces', use_spaces)]
```
jayvdb: closing `]` on the next line so more can be added easily

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -16,15 +17,28 @@ class RunDecoratedBear(LocalBear):
...
 
 
 class MapSettingToAspectTest(unittest.TestCase):
 
+    def list_to_dict(self, tuple_list):
```
jayvdb: this doesnt use `self` so it should be a module level function, not inside the class

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -34,8 +48,10 @@ def test_mapping(self):
             get_aspect('UnreachableCode')('py'),
             get_aspect('Clone')('py', min_clone_tokens=30),
         ])
-        result = self.bear.execute()
-        self.assertEqual([True, 30], result)
+        result = self.list_to_dict(self.bear.execute())
+        self.assertEqual({'remove_unreachable_code': True,
+                          'minimum_clone_tokens': 30,
+                          'use_spaces': True}, result)
```
jayvdb: closing `}` on the next line so it is easy to add more

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -46,22 +62,63 @@ def test_setting_priority(self):
...
+        result = self.list_to_dict(self.bear.execute())
+        self.assertEqual({'remove_unreachable_code': False,
+                          'minimum_clone_tokens': 10,
+                          'use_spaces': True}, result)
 
     def test_skip_mapping(self):
         self.section.aspects = None
-        result = self.bear.execute()
-        self.assertEqual([False, 10], result)
+        result = self.list_to_dict(self.bear.execute())
```
jayvdb: no need to modify the original line.
```
result = list_to_dict(result)
```

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -46,22 +62,63 @@ def test_setting_priority(self):
...
+                          'use_spaces': True}, result)
 
     def test_partial_mapping(self):
         self.section.aspects = AspectList([
             get_aspect('UnreachableCode')('py'),
         ])
-        result = self.bear.execute()
-        self.assertEqual([True, 10], result)
+        result = self.list_to_dict(self.bear.execute())
+        self.assertEqual({'remove_unreachable_code': True,
```
jayvdb: this should be a CONSTANT in the class, used to initialise the expected test results.

```py
EXPECTED = {'remove_unreachable_code': True,
 'minimum_clone_tokens': 10,
 'use_spaces': True,
 }
...

expected = EXPECTED.copy()
expected['blah'] = override
self.assertEqual(expected, results)
```

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -46,22 +62,63 @@ def test_setting_priority(self):
...
+        result = self.list_to_dict(self.bear.execute())
+        self.assertEqual({'remove_unreachable_code': False,
+                          'minimum_clone_tokens': 10,
+                          'use_spaces': False}, result)
+
+    def test_setting_priority(self):
+        self.section.aspects = AspectList([
+            get_aspect('Indentation')('py', indent_type='tab'),
+        ])
+        self.section.append(
```
jayvdb: why split this line?

tests/bearlib/aspects/DecoratorsTest.py
jayvdb: you could even create a helper

```
def assertResult(self, expected, result):
   self.assertEqual(expected, list_to_dict(result))
```

tests/bearlib/aspects/DecoratorsTest.py
pareksha: Good idea.
I am changing 'remove_unreachable_code' to False to be consistent with the bear defaults.

tests/bearlib/aspects/DecoratorsTest.py
pareksha: Sorry, I just saw that you asked the same on the previous review too. Will be even more careful next time. :)

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -4,27 +4,52 @@
...
 )
 from coalib.bears.LocalBear import LocalBear
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 
 
+def list_to_dict(tuple_list):
+    # Converts list of tuples to dict
+    result_dict = {}
+    for pair in tuple_list:
```
jayvdb: `for key, value in ...`

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -35,33 +60,73 @@ def test_mapping(self):
...
         self.section.aspects = AspectList([
             get_aspect('UnreachableCode')('py'),
             get_aspect('Clone')('py', min_clone_tokens=30),
         ])
-        self.section.append(
-            Setting('remove_unreachable_code', 'False'))
-        self.section.append(
-            Setting('minimum_clone_tokens', 40))
+        self.section.append(Setting('remove_unreachable_code', 'False'))
+        self.section.append(Setting('minimum_clone_tokens', 40))
```
jayvdb: These lines are not part of your change. Do not modify their style unless you need to improve their functionality as well.
It just makes your patch large and 'steals' `git blame`

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -35,33 +60,73 @@ def test_mapping(self):
             get_aspect('Clone')('py', min_clone_tokens=30),
         ])
         result = self.bear.execute()
-        self.assertEqual([True, 30], result)
+        expected = self.EXPECTED.copy()
+        expected['minimum_clone_tokens'] = 30
+        expected['remove_unreachable_code'] = True
```
jayvdb: this one should go first to be consistent

jschwarzwalder:
I am not sure what actions are being taken here.

It looks like it returns a dictionary of tuples that include a string from bear and a string from a taste.

As the developer do I need to take an action to select the value I want as output? Or will a change be selected for me?

tests/bearlib/aspects/DecoratorsTest.py
pareksha: Yeah, I guess this will make a good newcomer issue :)

pareksha:
@jschwarzwalder this function is to map bear setting to aspects. The aspects will be provided by the user and the linting will be done as per the provided aspect. Mapping functions make the migration of bear settings to aspects much easier.

And as for the dictionary of tuples part, it is solely for testing purpose, we are comparing the output of bear (running it using aspects) and comparing the same with the expected output.

jayvdb:
ack https://github.com/coala/coala/pull/5492/commits/9b72d2d91f9aa8b678a20717d190ac62072e4653

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,51 @@ def _new_func(self, *args, **kwargs):
         return _new_func
 
     return _func_decorator
+
+
+def map_ambiguous_setting_to_aspect(**aspectable_setting):
+    """
+    Map those bear settings to aspects in which there is some ambiguity.
```
jschwarzwalder: Consider "Specify aspects to use instead of the default bear settings where there is some ambiguity.

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,51 @@ def _new_func(self, *args, **kwargs):
...
+    The ambiguity is because of differing values (maybe either name or even
+    data-types) in bear settings and their corresponding aspect tastes.
+
+    This decorator can be used by ``Bear.run()`` to automatically map and
+    override value of bear settings with their equivalent taste.
```
jschwarzwalder: This decorator is used by Bear.run() to override the value in the bear setting with an equivalent taste as provided in the parameter.

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,51 @@ def _new_func(self, *args, **kwargs):
...
+
+    The order of setting override from the lowest to highest is:
+    - Setting default (in bear's run argument)
+    - Aspect/taste default (if aspect is activated in Section)
+    - Explicit aspect/taste default (if aspect is activated in Section)
+    - Explicit setting
```
jschwarzwalder: The order of setting override would be clearer if the order is reversed to :
The order of setting override from the highest to lowest is:

Explicit setting
Explicit aspect/taste default (if an aspect is activated in Section)
Aspect/taste default (if an aspect is activated in Section)
Setting default (in bear's run argument)

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,51 @@ def _new_func(self, *args, **kwargs):
...
+
+    :param aspectable_setting:
+        A dictionary in which keys are bear settings and corresponding value
```
jschwarzwalder: A dictionary in which keys are a bear settings and the corresponding value

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,51 @@ def _new_func(self, *args, **kwargs):
...
+        for each key is a tuple containing equivalent taste as first value and
```
jschwarzwalder: for each key is a tuple containing an equivalent taste as the first value and

jayvdb:
unack 9b72d2d

coalib/bearlib/aspects/decorators.py
pareksha: I don't think it's relevant to add here as ambiguity is purely because of differing values (maybe either name or even data-types) in bear settings and their corresponding aspect tastes. 

Bear settings itself are not ambiguous. :)

jschwarzwalder:
ack 9b72d2d

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,53 @@ def _new_func(self, *args, **kwargs):
...
+    :param aspectable_setting:
+        A dictionary in which keys are the bear settings and the corresponding
+        value for each key is a tuple containing an equivalent taste as first
+        value and a list of tuples with matching bear settings' and aspects'
+        value as the second value.
+    """
+    def _func_decorator(func):
+        @wraps(func)
+        def _new_func(self, *args, **kwargs):
+            if self.section.aspects:
```
jayvdb: Always bail out early instead of using indentation which makes the code unreadable as more conditions are added.

```py
if not self.section.aspects:
    return

...
```

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,53 @@ def _new_func(self, *args, **kwargs):
...
+                aspects = self.section.aspects
+                for arg, aspect_value in aspectable_setting.items():
+                    # Explicit bear setting takes priority over aspects
+                    if arg in self.section:
+                        continue
+
+                    taste = aspect_value[0]
```
jayvdb: Avoid using `[0]` on tuples.  give the components names so your code is more readible.

```python
taste, foo = aspect_value
```

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -4,27 +4,52 @@
...
 
 class MapSettingToAspectTest(unittest.TestCase):
 
+    EXPECTED = {'remove_unreachable_code': False,
+                'minimum_clone_tokens': 10,
+                'use_spaces': True,
+                }
+
+    def assertResult(self, expected, result):
+        self.assertEqual(expected, list_to_dict(result))
```
jayvdb: `self.assertEqual(expected, dict(result))`

tests/bearlib/aspects/DecoratorsTest.py
```diff
@@ -35,7 +60,10 @@ def test_mapping(self):
             get_aspect('Clone')('py', min_clone_tokens=30),
         ])
         result = self.bear.execute()
-        self.assertEqual([True, 30], result)
+        expected = self.EXPECTED.copy()
+        expected['remove_unreachable_code'] = True
+        expected['minimum_clone_tokens'] = 30
+        self.assertResult(expected, result)
```
jayvdb: Oh, `dict` accepts lists of tuples, so this can be simply

```py
self.assertEqual(expected, dict(result))
```

coalib/bearlib/aspects/decorators.py
pareksha: I have to do `return func(self, *args, **kwargs)` instead of `return` so that if aspects aren't there, the already existing default bear settings can be applied, otherwise it will return me an empty dict :)

coalib/bearlib/aspects/decorators.py
```diff
@@ -49,3 +49,53 @@ def _new_func(self, *args, **kwargs):
         return _new_func
 
     return _func_decorator
+
+
+def map_ambiguous_setting_to_aspect(**aspectable_setting):
```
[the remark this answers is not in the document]
pareksha: Yeah, you are right, `pydocstyle` says `1 blank line required between summary line and description`.

jayvdb:
ack 7b3da2d

jayvdb:
@gitmate-bot ff

PR #5494, opened 2018-06-04
Shell: Use sys.executable in doctest

Makman2:
... instead of python which is not availabe on Linux systems without
having Python 2 installed (Python 3 is called python3 on command
line). Now Python 2 is not necessary for the modified doctest
to execute.

### DIFF
--- a/coalib/misc/Shell.py
+++ b/coalib/misc/Shell.py
@@ -15,7 +15,8 @@ class ShellCommandResult(tuple):
 
     It additionally stores the return ``.code``:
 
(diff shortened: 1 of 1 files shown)


jayvdb:
ack https://github.com/coala/coala/pull/5494/commits/9e0884b7fd838c327f06f417f522ae8b4e007a32

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

PR #5496, opened 2018-06-05
BaseTestHelper.py: Add `BaseTestHelper` class

sangamcse:
`observed_result` is result generated from a bear and `expected_result` is tester's expected result from the bears.

Related to https://gitlab.com/coala/GSoC/GSoC-2018/issues/307
Closes https://github.com/coala/coala/issues/3676

### DIFF
--- a/coalib/testing/BaseTestHelper.py
+++ b/coalib/testing/BaseTestHelper.py
@@ -1,0 +1,54 @@
+from coala_utils.Comparable import Comparable
+
+
+def _assert_comparable_equal(observed_result, expected_result):
+    """
+    Compares individual fields of the result objects using
+    `__compare_fields__` of `coala_utils.Comparable` class
+    and yields messages in case of an attribute mismatch.
+    """
+
+    if not len(observed_result) == len(expected_result):
+        assert observed_result == expected_result, '%s != %s' % (
+            observed_result, expected_result)
+
+    messages = ''
+    for observed, expected in zip(observed_result, expected_result):
+        if (isinstance(observed, Comparable) and
+            isinstance(expected, Comparable)) and (type(observed) is
+                                                   type(expected)):
+            for attribute in type(observed).__compare_fields__:
+                try:
+                    assert getattr(observed, attribute) == getattr(
+                        expected, attribute), (
+                        '{} mismatch: {}, {} != {}, {}'.format(
+                            attribute,
+                            observed.origin, observed.message,
+                            expected.origin, expected.message))
+                except AssertionError as ex:
+                    messages += (str(ex) + '\n\n')
+        else:
+            assert observed_result == expected_result, '%s != %s' % (
+                observed_result, expected_result)
+
+    if messages:
+        raise AssertionError(messages)
+
+
+class BaseTestHelper:
+    """
+    This is a base class for all Bears' tests of coala's testing API.
+    """
+
+    def assert_result_equal(self,
+                            observed_result,
+                            expected_result):
+        """
+        Asserts that an observed result from a bear is exactly same
+        as the expected result from the bear.
+
+        :param observed_result: The observed result from a bear
+        :param expected_result: The expected result from a bear
+        """
+
+        return _assert_comparable_equal(observed_result, expected_result)
--- a/tests/testing/BaseTestHelperTest.py
+++ b/tests/testing/BaseTestHelperTest.py
@@ -1,0 +1,31 @@
+import pytest
+
+from coalib.results.Result import Result
+from coalib.testing.BaseTestHelper import BaseTestHelper
+
+
+class BaseTestHelperTest(BaseTestHelper):
+    def test_good_assert_result_equal(self):
+        self.assert_result_equal(['a', 'b'], ['a', 'b'])
+
+    def test_inequality_assert_result_equal(self):
+        with pytest.raises(AssertionError) as ex:
+            self.assert_result_equal(['a', 'b'], ['a', 'b', None])
+        assert '[\'a\', \'b\'] != [\'a\', \'b\', None]' in str(ex.value)
+
+    def test_not_comparable_assert_result_equal(self):
+        with pytest.raises(AssertionError) as ex:
+            self.assert_result_equal(['a', 'b'], ['a', 'c'])
+        assert '[\'a\', \'b\'] != [\'a\', \'c\']' in str(ex.value)
+
+    def test_comparable_assert_result_equal(self):
+        expected = [Result.from_values(origin='AnyBea',
+                                       message='This file has 2 lines.',
+                                       file='anyfile')]
+        observed = [Result.from_values(origin='AnyBear',
+                                       message='This file has 2 lines.',
+                                       file='anyfile')]
+        with pytest.raises(AssertionError) as ex:
+            self.assert_result_equal(expected, observed)
+        assert ('origin mismatch: AnyBea, This file has 2 lines. != AnyBear, '
+                'This file has 2 lines.\n\n') == str(ex.value)


coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,64 @@
+from coala_utils.Comparable import Comparable
+
+
+class BaseTestHelper:
+    """
+    This is a base class for all tests of coala's testing API.
```
NiklasMM: for all tests *Bears*

sangamcse:
Updated:

- Docs on [line 6](https://github.com/coala/coala/pull/5496/files#diff-feb1a7b4c2139a13b1c7c4ee1dd84726R6).
- Removed `setUp`, `tearDown`, `setUpClass` and `tearDownClass`.

cc @NiklasMM

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,36 @@
...
+                                    observed.origin, observed.message,
+                                    expected.origin, observed.message)
+                        except AssertionError as ex:
+                            messages += (str(ex) + '\n\n')
+                else:
+                    assert observed_result == expected_result, '%s != %s' % \
+                        (observed_result, expected_result)
+            if messages:
+                raise AssertionError(messages)
+        else:
```
kriti21: You can write maybe like 
`if not len(observed_result) == len(expected_result):` This would eliminate one level of indentation atleast. There are so many nested if and for loops.

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,36 @@
...
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
+        """
+        Compares individual fields of the Result object and yields messages in
+        case of an attribute mismatch.
+        """
+        if len(observed_result) == len(expected_result):
+            messages = ''
+            for observed, expected in zip(observed_result, expected_result):
+                if (isinstance(observed, Comparable)
+                        and isinstance(expected, Comparable)) \
```
kriti21: I might be wrong here but line breaking using `\` is not recommended.

sangamcse: That gives syntax error

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,41 @@
+from coala_utils.Comparable import Comparable
+
+
+class BaseTestHelper:
+    """
+    This is a base class for all Bears' tests of coala's testing API.
+    """
+
+    def assertComparableObjectsEqual(self, observed_result, expected_result):
```
IpshitaC: Perhaps you could use snake case for the method name, to ensure consistency. `assert_comparable_objects_equal`

tests/testing/BaseTestHelperTest.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from coalib.results.Result import Result
+from coalib.testing.BaseTestHelper import BaseTestHelper
+
+
+class BaseTestHelperTest(BaseTestHelper):
+    def test_good_assertComparableObjectsEqual(self):
```
IpshitaC: A uniform case would be a lot better. `test_good_assert_comparable_objects_equal`

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        Compares individual fields of the Result object and yields messages in
```
IpshitaC: Change to `result objects`

tests/testing/BaseTestHelperTest.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertComparableObjectsEqual(['a', 'b'], ['a', 'b'])
+
+    def test_inequality_assertComparableObjectsEqual(self):
```
IpshitaC: Same. A uniform snake case for this method name would be a lot better

tests/testing/BaseTestHelperTest.py
```diff
@@ -0,0 +1,31 @@
...
+        with pytest.raises(AssertionError) as ex:
+            self.assertComparableObjectsEqual(['a', 'b'], ['a', 'b', None])
+        assert '[\'a\', \'b\'] != [\'a\', \'b\', None]' in str(ex.value)
+
+    def test_not_comparable_assertComparableObjectsEqual(self):
```
IpshitaC: Same here. Case of the method name

tests/testing/BaseTestHelperTest.py
```diff
@@ -0,0 +1,31 @@
...
+        with pytest.raises(AssertionError) as ex:
+            self.assertComparableObjectsEqual(['a', 'b'], ['a', 'c'])
+        assert '[\'a\', \'b\'] != [\'a\', \'c\']' in str(ex.value)
+
+    def test_comparable_assertComparableObjectsEqual(self):
```
IpshitaC: Here too

coalib/testing/BaseTestHelper.py
IpshitaC: The method name sounds a bit complicated. Is there some way it can be made a bit more intuitive perhaps?

RaiVaibhav: I just gone through the yash Nishar commit of `assertComparableObjectsEqual` and I didn't find any review by John or Makman on name casing, btw for me `assertComparableObjectsEqual` is ok here.

Sorry for nitpicking the comment :stuck_out_tongue:

tests/testing/BaseTestHelperTest.py
RaiVaibhav: same for me it should be `test_good_assertComparableObjectsEqual` as IMO `assertComparableObjectsEqual` is a method name which together means something

coalib/testing/BaseTestHelper.py
sangamcse: @RaiVaibhav, it is not necessary to keep same thing in the base class. We can think of better method name here as you can see the previous method name is similar to unittest method name type. But in this project, we are moving towards pytest.

RaiVaibhav:
>Try restarting your Travis build too

Rebuilding will help IMO it's failing on whole coala/coala repo

RaiVaibhav:
Ok, :+1:

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,44 @@
...
+                        assert getattr(observed, attribute) == getattr(
+                            expected, attribute), (
+                            '{} mismatch: {}, {} != {}, {}'.format(
+                                attribute,
+                                observed.origin, observed.message,
+                                expected.origin, expected.message))
+                    except AssertionError as ex:
+                        messages += (str(ex) + '\n\n')
+            else:
+                assert observed_result == expected_result, '%s != %s' % (
```
IpshitaC: I think this is good enough.

sangamcse:
@gitmate-bot rebase

coalib/testing/BaseTestHelper.py
Makman2: so no unittest style asserts any more?

sangamcse: Yes 😅

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    This is a base class for all Bears' tests of coala's testing API.
+    """
+
+    def assert_comparable_objects(self,
+                                  observed_result,
+                                  expected_result):
+        """
+        Compares individual fields of the result objects and yields messages
+        in case of an attribute mismatch.
```
jayvdb: 'attribute mismatch' is unclear.

Is it attribute *type* mismatch, or attribute *value* mismatch?

When I read `assert_comparable_objects`, I assume it means it checks that the *types* are comparable, but does not check that they are the same.

If it is about the values, then the name should include they key word `equals`, like [`assertCountEqual`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual)

However, by reading the code I can see that `comparable` here refers to `coala_utils.Comparable`.  That needs to be described in the docstring.

fwiw, most of this logic should be put into `Comparable.__eq__`, and similar methods.

Then the code here would simple use `!=` , and raise asserts accordingly.

coalib/testing/BaseTestHelper.py
```diff
@@ -0,0 +1,56 @@
+from coala_utils.Comparable import Comparable
+
+
+class BaseTestHelper:
+    """
+    This is a base class for all Bears' tests of coala's testing API.
+    """
+
+    def _assert_comparable_equal(self,
```
jayvdb: module level scope means move it outside the class.

sangamcse: Ohk. Got it.

jayvdb:
ack 936a5d8

jayvdb:
@gitmate-bot ff

Issue #4999, opened 2017-12-23
Bears operating on directories


Makman2:
Currently there's only `ProjectBear`, that works on all files at once. It's worth to have another bear base that only operates on directories, or maybe even both, files and directories! Maybe it can be even integrated into `ProjectBear`. In any way, a concept is needed to make it work properly, also with respect to caching it has to be developed carefully (once it's being implemented).

ishanSrt:
I would like to take this up but this requires too many changes.
Tell me things that I am not supposed to change at all.
Also point me out where does coala infer the directories from the file paths given as you mentioned this on gitter

Makman2:
>  Tell me things that I am not supposed to change at all.

I can't, because I don't know.

> Also point me out where does coala infer the directories from the file paths given as you mentioned this on gitter

You have to look out where the `file_dict` is filled up, there functions from `Globbing` should appear.

ishanSrt:
@Makman2 I see bears like `PythonPackageInitBear` resolving directories from files with:
```
dirs = {os.path.split(filename)[0]
                for filename in self.file_dict.keys()
                if filename.endswith('.py')}
```
although I don't see `GitCommitBear` doing such a thing
Also while having a look at other global bears, I am not able to run `DuplicateFileBear`
when I run it like this: `coala --bears=DuplicateFileBear --files=test.py -I`
It shows me
```
[ERROR][16:42:58] Bear DuplicateFileBear failed to run. Take a look at debug messages (`-V`) for further information.
``` 
and the debug message shows me
```
[WARNING][16:43:09] coala will not run any analysis. Did you forget to give the `--files` argument?
[WARNING][16:43:09] coala will not run any analysis. Did you forget to give the `--bears` or `--aspects` argument?
```

Makman2:
Problem is there might be directories without files, this is where this approach fails:

>  I see bears like PythonPackageInitBear resolving directories from files with:

Also one has to note that you can't commit directories inside git. Maybe we want to follow this "convention", in our case it means that our linting results only depend on *files*, not directories.

ishanSrt:
hmmm.. need some time to ponder on that.
BTW why is DuplicateFileBear behaving that way?

Makman2:
> Also while having a look at other global bears, I am not able to run DuplicateFileBear

Don't know, but that's a different issue. Please file one :+1:

ishanSrt:
@Makman2  I want to work on this

newbazz:
[...]

PR #5498, opened 2018-06-07
Add Directory class

palash25:
Directory class will act as an interface
to directories providing useful information
about them. The objects of this class can be
used by bears that operate only on directories
and perform analysis based on the things like
directory structure

Closes https://github.com/coala/coala/issues/4999

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/io/Directory.py
+++ b/coalib/io/Directory.py
@@ -1,0 +1,89 @@
+import os
+
+from coalib.parsing.Globbing import relative_recursive_glob
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('path', 'timestamp')
+class Directory:
+    """
+    The ``Directory`` class acts as an interface to directories and
+    provides useful information about them such as:
+
+        * The directory path.
+        * The parent directory path.
+        * The last modified timestamp of the directory.
+        * The children contained inside a directory.
+
+    >>> import os
+    >>> d = Directory('tests/io/DirectoryTestDir/')
+
+    Get the number of files and sub-directories at the top level:
+
+    >>> len(d.get_children())
+    3
+
+    Get the number of all the files and sub-directories recursively:
+    >>> len(d.get_children_recursively())
+    5
+
+    Get the path of the ``Directory`` object:
+
+    >>> os.path.basename(d.path).endswith('DirectoryTestDir')
+    True
+
+    Get the parent directory:
+
+    >>> os.path.basename(d.parent).endswith('io')
+    True
+
+    Get the last modified timestamp of the directory:
+
+    >>> d.timestamp == os.path.getmtime(d.path)
+    True
+    """
+
+    def __init__(self, path):
+        self._path = os.path.abspath(path)
+        self._parent = os.path.abspath(os.path.dirname(self._path))
+        self._timestamp = os.path.getmtime(self._path)
+
+    def get_children(self):
+        """
+        :return:
+            A list of all the files and sub-directories cotained at the
+            top level of the directory.
+        """
+        return os.listdir(self._path)
+
+    def get_children_recursively(self):
+        """
+        :return:
+            A list of all the files and sub-directories contained
+            inside a directory.
+        """
+        return list(relative_recursive_glob(self._path, '**'))[1:]
+
+    @property
+    def path(self):
+        """
+        :return:
+            The directory path.
+        """
+        return self._path
+
+    @property
+    def parent(self):
+        """
+        :return:
+            The parent directory.
+        """
+        return self._parent
+
+    @property
+    def timestamp(self):
+        """
+        :return:
+            The last modified timestamp of the directory.
+        """
+        return self._timestamp
--- a/tests/io/DirectoryTest.py
+++ b/tests/io/DirectoryTest.py
@@ -1,0 +1,34 @@
+import os
+import unittest
+
+from coalib.io.Directory import Directory
+
+
+class DirectoryTest(unittest.TestCase):
+
+    def setUp(self):
+        self.test_dir_path = os.path.join(os.path.dirname(__file__),
+                                          'DirectoryTestDir')
+        self.another_test_dir_path = os.path.join(os.path.dirname(__file__),
+                                                  'DirectoryTestDir', 'Dir1')
+        self.uut = Directory(self.test_dir_path)
+
+    def test_equal(self):
+        self.assertEqual(self.uut, Directory(self.test_dir_path))
+        self.assertNotEqual(self.uut, Directory(self.another_test_dir_path))
+
+    def test_get_children(self):
+        self.assertEqual(len(self.uut.get_children()), 3)
+
+    def test_get_children_recursively(self):
+        self.assertEqual(len(self.uut.get_children_recursively()), 5)
+
+    def test_path(self):
+        self.assertEqual(self.uut.path, self.test_dir_path)
+
+    def test_parent(self):
+        self.assertEqual(self.uut.parent, os.path.dirname(self.test_dir_path))
+
+    def test_timestamp(self):
+        self.assertEqual(self.uut.timestamp,
+                         os.path.getmtime(self.test_dir_path))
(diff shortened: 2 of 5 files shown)


coalib/io/Directory.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def children(self):
+        elements = relative_recursive_glob(self._path, '**')
+        children = []
+
+        for e in elements:
+            children.append(e)
+
+        return children[1:]
+
```
shreyans800755:
```
children = [e for e in relative_recursive_glob(self._path, '**')]
return children[1:]
```

Or may be just:
```
return [e for e in relative_recursive_glob(self._path, '**')][1:]
```

coalib/io/Directory.py
```diff
@@ -0,0 +1,28 @@
...
+
+@generate_eq('path')
+class Directory:
+    def __init__(self, path):
+        self._path = os.path.abspath(path)
+        self._parent = os.path.abspath(os.path.dirname(self._path))
+
+    def children(self):
+        elements = relative_recursive_glob(self._path, '**')
```
Makman2: can't we use `os.listdir`? I assume that's way faster because it's not evaluating globs^^

coalib/io/Directory.py
```diff
@@ -0,0 +1,28 @@
...
+
+from coalib.parsing.Globbing import relative_recursive_glob
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('path')
+class Directory:
+    def __init__(self, path):
+        self._path = os.path.abspath(path)
+        self._parent = os.path.abspath(os.path.dirname(self._path))
```
Makman2: Is it possible to maybe add a timestamp here as well? Since `Directory`s will be cached too.

coalib/io/Directory.py
palash25: `os.listdir` only gives us the children contained in the parent directories and does not go down into the sub directories.
Since the directory caching strategy is not clear I'm not sure whether to get all the children recursively or just the ones one level down the parent.

What do you think? Should I still change it to `listdir` or keep the current code as it is or have two different methods one recursive and one non-recursive?

Makman2: Ahh you recurse. Definitely make it top-level only right now, always recursing into subdirs is a performance penalty. I'm not sure how much our globbing is caching directory contents, but I wouldn't rely on that^^

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        return os.listdir(self._path)
+
+    def get_children_recursively(self):
+        """
+        :return:
+            A list of all the files and sub-directories contained
+            inside a directory.
+        """
+        return [e for e in relative_recursive_glob(self._path, '**')][1:]
```
Makman2: -->
```python
list(relative_recursive_glob(self._path, '**'))[1:]
```
when we are on it, we can again use a `tuple` ;)

Makman2: What is actually `[1:]` for? Is the first element something like `..` or the directory itself?

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
+import os
+
+from coalib.parsing.Globbing import relative_recursive_glob
+from coala_utils.decorators import generate_eq
+
+
+@generate_eq('path', 'timestamp')
+class Directory:
+    """
+    The `Directory` class acts as an interface to directories and
```
Makman2: double backticks, it's RST ;)

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
...
+    provides useful information about them such as:
+
+        * The directory path.
+        * The parent directory path.
+        * The last modified timestamp of the directory.
+        * The number of children contained inside a directory.
```
Makman2: Not just the number, the children itself ;)
--> "The children contained inside ~~a~~ *the* directory"

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
...
+
+    >>> import os
+    >>> d = Directory('tests/io/DirectoryTestDir/')
```
Makman2: Again temporary files/directories? ;)

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
...
+    >>> len(d.get_children())
+    3
+
+    Get the number of all the files and sub-directories recursively:
+    >>> len(d.get_children_recursively())
+    5
+
+    Get the path of the `Directory` object:
+
+    >>> os.path.relpath(d.path).endswith('DirectoryTestDir')
```
Makman2: do you have to do `relpath`?

Makman2: Maybe you can use `splitext` or similar `os.path` functions to get the last file path component. You can reuse the same below :+1:

coalib/io/Directory.py
palash25: > What is actually [1:] for? Is the first element something like .. or the directory itself?

Yep its the current directory itself

coalib/io/Directory.py
palash25: I wanted to show the different outputs given by the `get_children` and `get_children_recursively` methods.

I'm not sure if we can set up directories and files inside temporary  directories since they are all random. Even if we can do that using `tempfile` I would have to create more at least more than 2 temp files/dirs, I'm not sure if creating so many tempfiles or dirs would look good in the docstrings .
Using an already existing directory structure lets us avoid that and show that difference between different methods too.

Still if you want me to change that let me know.

Makman2: You can create a top-level directory, and then create inside the structure you need. But that's definitely more cumbersome. I'm also not sure if we should support executing pytest from different directories^^ Ideally yes, better than not, though makes things certainly harder.

coalib/io/Directory.py
```diff
@@ -0,0 +1,89 @@
...
+    5
+
+    Get the path of the ``Directory`` object:
+
+    >>> os.path.basename(d.path).endswith('DirectoryTestDir')
+    True
+
+    Get the parent directory:
+
+    >>> os.path.basename(d.parent).endswith('io')
```
Makman2: maybe `'tests/io'`

tests/io/DirectoryTest.py
```diff
@@ -0,0 +1,31 @@
+import os
+import unittest
+
+from coalib.io.Directory import Directory
+
+
+class DirectoryTest(unittest.TestCase):
+
+    def setUp(self):
+        self.test_dir_path = os.path.join(os.path.split(__file__)[0],
```
Makman2: you can use `os.path.dirname` for `os.path.split(...)[0]`

tests/io/DirectoryTest.py
```diff
@@ -0,0 +1,31 @@
...
+                                          'DirectoryTestDir')
+        self.uut = Directory(self.test_dir_path)
+
+    def test_equal(self):
+        self.assertEqual(self.uut, Directory(self.test_dir_path))
```
Makman2: test also inequality

coalib/io/Directory.py
palash25: cant do that or it would fail on windows

tests/io/DirectoryTest.py
```diff
@@ -0,0 +1,34 @@
...
+from coalib.io.Directory import Directory
+
+
+class DirectoryTest(unittest.TestCase):
+
+    def setUp(self):
+        self.test_dir_path = os.path.join(os.path.dirname(__file__),
+                                          'DirectoryTestDir')
+        self.another_test_dir_path = os.path.join(os.path.dirname(__file__),
+                                                  'DirectoryTestDir/Dir1')
```
Makman2: -->
```python
os.path.join(os.path.dirname(__file__), 'DirectoryTestDir', 'Dir1')
```
The `/` inside the path isn't platform independent ;) Let `os.path.join` do the task ;)

Makman2:
ack 32e3e462ecb940844af93350b5aefd81ce46282b

Makman2:
@gitmate-bot ff

Issue #5321, opened 2018-04-06
Test the man page


jayvdb:
#4894 added man/man1 to setup.py, but who knows if it works.  We should have some test logic which verifies that `man` can load the man page correctly

In the referenced issue, a much more complex solution was used by a team who generally verify their stuff works.
https://github.com/wkentaro/wstool/blob/68003a95c2bde7b26101a9621de814ff82a73b64/setup.py#L21

nikitavbv:
I am interested in working on this issue. Could you assign it to me please?

jayvdb:
Done. Im actually 90% confident the current code is busted.

Also setup.py has 'moved' to gitlab coala/mobans. Create an issue there if you find it has a bug.

nikitavbv:
> Im actually 90% confident the current code is busted.

You are right. Man pages are not installed properly. I created an issue for that: https://gitlab.com/coala/mobans/issues/33

jayvdb:
Remove the moban sync stuff in `.travis.yml` to work on building a tested solution in this repo, which we can then migrate to the mobans repo.

I have a pending patch which removes the sync stuff, and will let you know when that is merged.

nikitavbv:
Actually, I was wrong with some of my assumptions of how setup.py works here. Man pages are installed correctly. I closed the Gitlab issue.
Now I will add testing of man pages to the Travis build.

jayvdb:
As noted on the PR, this isnt really solved yet.

PR #5500, opened 2018-06-07
.travis.yml: Add man page test

nikitavbv:
Adds test to Travis CI to verify that `man` can load man page
correctly.

Closes https://github.com/coala/coala/issues/5321

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/.ci/check_man.sh
+++ b/.ci/check_man.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+if [ $CIRCLECI ]; then
+  sudo apt install man
+  export MANPATH=$MANPATH:~/.local/man
+elif [ $TRAVIS ]; then
+  export MANPATH=$MANPATH:$VIRTUAL_ENV/man
+fi
+
+man coala | grep "https://coala.io" > /dev/null 2>&1
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,6 +106,7 @@ script:
     fi
   - coala --non-interactive
   - python setup.py docs
+  - .ci/check_man.sh
 
 after_success:
   - codecov
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,10 @@ jobs:
           name: sphinx-build
           command: python setup.py docs
 
+      - run:
+          name: check-man
+          command: .ci/check_man.sh
+
       - save_cache:
           key: coala-cache-{{ checksum "requirements.txt" }}
           paths:


.travis.yml
```diff
@@ -84,6 +84,7 @@ script:
     fi
   - coala --non-interactive
   - python setup.py docs
+  - man $VIRTUAL_ENV/man/man1/coala.1 > /dev/null 2>&1
```
jayvdb: for missing manpages, I see exit code 16, so this looks good.

alternatively/additionally we could check the output for some string we know will exist if it is our manpage.

I suggest you move this to .ci/check_man.sh , and also get it working under Circle CI (which should be green, while Travis is red atm.

nikitavbv: Thank you for review!
Circle CI test is added now, check itself moved to `.ci/check_man.sh`, output check added (check if `https://coala.io` exists in output)

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot rebase

.ci/check_man.sh
```diff
@@ -0,0 +1,14 @@
...
+if [ $CIRCLECI ]; then
+    sudo apt install man
+    pagepath=~/.local/man/man1/coala.1
+elif [ $TRAVIS ]; then
+    pagepath=$VIRTUAL_ENV/man/man1/coala.1
+else
+    pagepath=coala
+fi
+
+man $pagepath | grep "https://coala.io" > /dev/null 2>&1
```
jayvdb: to know it is installed correctly, the command needs to be `man coala` without an explicit path, or a `.1` suffix.

If that isnt possible, I guess setting `MANPATH` is a bit better, but then we need to tell users to also set MANPATH, and that gets complicated when they have different OS, venvs, etc.

.travis.yml
```diff
@@ -106,6 +106,7 @@ script:
     fi
   - coala --non-interactive
   - python setup.py docs
+  - sh .ci/check_man.sh
```
jayvdb: make the script executable so you dont need to do `sh ..`

.ci/check_man.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+if [ $CIRCLECI ]; then
+    sudo apt install man
+    export MANPATH=$MANPATH:~/.local/man
```
jayvdb: You've verified it produces a correct manpage which can be read by `man` (this might even be a good addition to the real test suite, for the man page building command), but you've also verified that it can *not* be found by `man`.

The problem with this is that `man coala` does not actually work by default for users.

We would have to also add documentation for users to add this line somewhere in the startup scripts, or add a different line depending on how they installed coala.

Before adding documentation, ideally we find a better way so that it 'just works'.

also c.f. https://gitlab.com/coala/mobans/issues/1#note_80986917

jayvdb: So apparently we are using two space indents, but it isnt enforced yet.

nikitavbv: > We would have to also add documentation for users to add this line somewhere in the startup scripts, or add a different line depending on how they installed coala.

Actually, if virtualenv is not used then man pages are installed correctly. In case it is - then yes, users need to add directory in virtualenv to `MANPATH`. Do we need to create a separate issue for that? If so, I can work on adding required information to the docs.
Alternatively, we can think of a way to add required directory to `MANPATH` automatically, but I don't think it is a good idea.

> Before adding documentation, ideally we find a better way so that it 'just works'.

Unluckily, there is no way to specify absolute paths in `data_files`. So, if virtualenv is used, man files are placed relatively to virtualenv root.

> also c.f. https://gitlab.com/coala/mobans/issues/1#note_80986917

Yes, I am going to leave some short comment there as soon as Gitlab goes up again.

> So apparently we are using two space indents, but it isnt enforced yet.

Okay, I will change indentation now.

jayvdb: > Actually, if virtualenv is not used then man pages are installed correctly.

We need a test which proves this ;-)

Please change the commit message to "Related to https://github.com/coala/coala/issues/5321" and we'll merge this as an interim measure, as it does show what happens with a ~/.local install and with a virtualenv install.  But we need a test of non-virtual-env, real install.

Until we have that verified by CI, this issue isnt completely resolved.

nikitavbv: > We need a test which proves this ;-)

Hm, I took a fresh look at this and noticed that I was wrong before. If we use `/usr/share/man/man1` as data files path, then man pages are installed correctly and no changes to `MANPATH` are needed. So if we use that, we can remove `export MANPATH=...` for CircleCI, as it does not use virtualenv. There is permission problem though. If we use `/usr/share/man/man1` then man page can be only installed if running as root. If not - there will be `Permission denied` error during `pip install`. To solve that we need to run the whole CI job as root (not a good solution) or to use hack like `sudo chown $(whoami) -R /usr/share` (which I don't like too) to make installation possible without root access.
As for Travis - as I said there is no way to go outside virtualenv. If we need to install man pages correctly there, we will need to change CI job / add a separate so it does not use virtualenv.
What are your thoughts on this?

> But we need a test of non-virtual-env, real install.

Sure, as I said this can be done on CircleCI, but a bit problematic on Travis.

> Have you looked at what they are doing in https://github.com/wkentaro/wstool/blob/68003a95c2bde7b26101a9621de814ff82a73b64/setup.py#L21 ?

Yes, I have. They use `/usr/share` prefix for man page installation. That prefix allows to install man pages to `/usr/share/man/man1`. However, virtualenv users still need to add path to `MANPATH`.

jayvdb:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/coala/pull/5500/commits/466fc8446e9091550ea6b91f6175cca727a5915c

jayvdb:
@gitmate-bot ff

Issue #5501, opened 2018-06-07
Modify `LocalBearTestHelper` according to the base class


sangamcse:
blocked till https://github.com/coala/coala/pull/5496 gets merged.

status/blocked

PR #5505, opened 2018-06-08
LocalBearTestHelper.py: Inherit from `BaseTestHelper`

sangamcse:
This adds `BaseTestHelper` class, base class of all bears'
tests, to the `LocalBearTestHelper`.

Related to https://gitlab.com/coala/GSoC/GSoC-2018/issues/332
Closes https://github.com/coala/coala/issues/5501

### DIFF
--- a/coalib/testing/LocalBearTestHelper.py
+++ b/coalib/testing/LocalBearTestHelper.py
@@ -6,9 +6,9 @@ from unittest.mock import patch
 import pytest
 
 from coalib.bearlib.abstractions.LinterClass import LinterClass
+from coalib.testing.BaseTestHelper import BaseTestHelper
 from coalib.testing.BearTestHelper import generate_skip_decorator
 from coalib.bears.LocalBear import LocalBear
-from coala_utils.Comparable import Comparable
 from coala_utils.ContextManagers import prepare_file
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
@@ -101,7 +101,7 @@ def get_results(local_bear,
         return bear_output
 
 
-class LocalBearTestHelper(unittest.TestCase):
+class LocalBearTestHelper(BaseTestHelper, unittest.TestCase):
     """
     This is a helper class for simplification of testing of local bears.
 
@@ -110,28 +110,6 @@ class LocalBearTestHelper(unittest.TestCase):
 
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
-
-    def assertComparableObjectsEqual(self, observed_result, expected_result):
-        if len(observed_result) == len(expected_result):
-            messages = ''
-            for observed, expected in zip(observed_result, expected_result):
-                if (isinstance(observed, Comparable)
-                    and isinstance(expected, Comparable)) and (
-                            type(observed) is type(expected)):
-                    for attribute in type(observed).__compare_fields__:
-                        try:
-                            self.assertEqual(
-                                getattr(observed, attribute),
-                                getattr(expected, attribute),
-                                msg='{} mismatch.'.format(attribute))
-                        except AssertionError as ex:
-                            messages += (str(ex) + '\n\n')
-                else:
-                    self.assertEqual(observed_result, expected_result)
-            if messages:
-                raise AssertionError(messages)
-        else:
-            self.assertEqual(observed_result, expected_result)
 
     def check_validity(self,
                        local_bear,
@@ -193,6 +171,7 @@ class LocalBearTestHelper(unittest.TestCase):
         :param create_tempfile:  Whether to save lines in tempfile if needed.
         :param tempfile_kwargs:  Kwargs passed to tempfile.mkstemp().
         """
+        assert isinstance(self, BaseTestHelper)
         assert isinstance(self, unittest.TestCase)
         self.assertIsInstance(local_bear,
                               LocalBear,
@@ -242,6 +221,7 @@ class LocalBearTestHelper(unittest.TestCase):
                                  from which settings will be created that will
                                  be made available for the tested bear.
         """
+        assert isinstance(self, BaseTestHelper)
         assert isinstance(self, unittest.TestCase)
         self.assertIsInstance(local_bear,
                               LocalBear,
@@ -260,10 +240,10 @@ class LocalBearTestHelper(unittest.TestCase):
                                   tempfile_kwargs=tempfile_kwargs,
                                   settings=settings)
         if not check_order:
-            self.assertComparableObjectsEqual(
+            self.assert_result_equal(
                 sorted(bear_output), sorted(results))
         else:
-            self.assertComparableObjectsEqual(bear_output, results)
+            self.assert_result_equal(bear_output, results)
 
         return bear_output
 
--- a/tests/testing/LocalBearTestHelperTest.py
+++ b/tests/testing/LocalBearTestHelperTest.py
@@ -48,7 +48,7 @@ class LocalBearCheckResultsTest(Helper):
             self.check_results(self.uut, ['a', 'b'], ['a', 'b', None],
                                check_order=True)
 
-    def test_good_assertComparableObjectsEqual(self):
+    def test_good_check_results(self):
         self.uut = LineCountTestBear(Section('name'), Queue())
         file_content = 'a\nb\nc'
         with prepare_file(file_content.splitlines(), filename=None,
@@ -63,7 +63,7 @@ class LocalBearCheckResultsTest(Helper):
                                filename=fname,
                                create_tempfile=False)
 
-    def test_bad_assertComparableObjectsEqual(self):
+    def test_bad_check_results(self):
         with self.assertRaises(AssertionError) as cm:
             self.uut = LineCountTestBear(Section('name'), Queue())
             file_content = 'a\nb\nc'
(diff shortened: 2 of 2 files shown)


sangamcse:
Oops, I forgot to run `coala`.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -101,7 +100,7 @@ def get_results(local_bear,
         return bear_output
 
 
-class LocalBearTestHelper(unittest.TestCase):
+class LocalBearTestHelper(BaseTestHelper):
```
jayvdb: this is a backwards incompatible change.

all test classes in the bears, and anywhere else that has used LocalBearTestHelper, become broken if this is merged.  They all need to be rewritten, just like LocalBearTestHelper in this patch has needed to be completely rewritten.

Most features of pytest that we want to use co-exist nicely with unittest classes.
https://github.com/pytest-dev/pytest/blob/master/doc/en/unittest.rst#pytest-features-in-unittesttestcase-subclasses

https://pypi.org/project/pytest-restrict/ explicitly supports unittest classes; that is what it is used for ... enforcing unittest classes when Django developers are using pytest, and should be using their customised Django unittest sub-classes.

NiklasMM: Two of the features I find most useful, namely parametrization and fixtures, do *not* work with unittest. That's why we want to get rid of it. 

But I see the difficulty of introducing this change into an existing test suite. @sangamcse do you think it's feasible to keep the unittest dependency for now and still use the base class?

sangamcse: In this case, only class name has changed which is done by adding `Test` as suffix to the class name
> They all need to be rewritten, just like LocalBearTestHelper in this patch has needed to be completely rewritten.

sangamcse: I tried testing coala-bears locally. All tests passed except `MarkdownBear`.
> all test classes in the bears, and anywhere else that has used LocalBearTestHelper, become broken if this is merged.

NiklasMM: Another idea would be to deprecate the existing `LocalBearTestHelper`, leave it as it is now and create a new one, which is based on the new base class and can be rolled out step by step.

sangamcse: @jayvdb, you are right. I tried to test coala-bears by cloning this branch in another venv in my system. It collected 212 test items instead of 856 test items. So it will break coala-bears test if we go on directly with this changes.
> all test classes in the bears, and anywhere else that has used LocalBearTestHelper, become broken if this is merged. 

I agree with @NiklasMM. We can go on by creating new one, deprecating the existing `LocalBearTestHelper` if you all agree.

sangamcse: @Makman2, your opinion?

jayvdb: @NiklasMM , fixtures do work, with a slightly different mechanism for some types of fixtures.
see https://github.com/pytest-dev/pytest/blob/master/doc/en/unittest.rst#mixing-pytest-fixtures-into-unittesttestcase-subclasses-using-marks
and `git grep fixture tests` in coala-core repo.

wrt parameterisation, and even fixtures, I didnt see the rationale given in the cEP for how those will be used to improve bear testing.  I would argue that https://github.com/wolever/parameterized is as good a solution which works well for all test frameworks.  We already have a complex set of pytest plugins, and already use `testfixtures`, so an additional test dependency isnt a new complexity problem.

Anyway, my intention isnt to re-litigate that.  I am only commenting because of the huge breakage.

Leaving the existing unittest class as-is, and deprecating it, will reduce the risk substantially.  And then at the end of the project the huge rewrite of bear tests can happen, and it wont matter if there are some bear tests which are not migrated to pytest-only style for reasons not yet predicted.

If the immediate goal is only to enforce a common test base class, that can be achieved by changing this class to inherit from both:

```py
class LocalBearTestHelper(BaseTestHelper, unittest.TestCase):
    ...
```

coalib/testing/LocalBearTestHelper.py
```diff
@@ -193,13 +171,12 @@ def check_invalidity(self,
...
-        self.assertIsInstance(local_bear,
-                              LocalBear,
-                              msg='The given bear is not a local bear.')
-        self.assertIsInstance(lines,
-                              (list, tuple),
-                              msg='The given lines are not a list.')
+        assert isinstance(
+            local_bear, LocalBear), 'The given bear is not a local bear.'
+        assert isinstance(
+            lines, (list, tuple)), 'The given lines are not a list.'
```
NiklasMM: Due to pytest's assertion introspection we already get [very good failure reports](https://docs.pytest.org/en/latest/example/reportingdemo.html#tbreportdemo). This eliminates the need for custom messages in these assert statements. 

I'm not entirely sure, if that feature is enabled for unittest.TestCases. That's something you'll need to verify first.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -210,7 +187,7 @@ def check_invalidity(self,
                                   )
         msg = ("The local bear '{}' yields no result although it "
                'should.'.format(local_bear.__class__.__name__))
-        self.assertNotEqual(len(bear_output), 0, msg=msg)
+        assert not len(bear_output) == 0, msg
```
NiklasMM: `assert len(bear_output) != 0`

coalib/testing/LocalBearTestHelper.py
```diff
@@ -111,28 +111,6 @@ class LocalBearTestHelper(unittest.TestCase):
     If you miss some methods, get in contact with us, we'll be happy to help!
     """
 
-    def assertComparableObjectsEqual(self, observed_result, expected_result):
```
jayvdb: just noting this was unused by bears repo, so no problem removing this.

tests/testing/LocalBearTestHelperTest.py
```diff
@@ -78,19 +78,11 @@ def test_bad_assertComparableObjectsEqual(self):
...
-                         'lines.\'\n'
-                         '- This file has 3 lines.\n'
-                         '?               ^\n'
-                         '+ This file has 2 lines.\n'
-                         '?               ^\n'
-                         ' : message_base mismatch.\n\n',
-                         str(cm.exception))
+        self.assertEqual('origin mismatch: LineCountTestBear, This file has '
+                         '3 lines. != LineCountTestBea, This file has 2 '
+                         'lines.\n\nmessage_base mismatch: LineCountTestBear,'
```
jayvdb: please put `message_base ...` on a new line, so it is slightly easier to read.

this message change looks like it is harder to understand that the previous assertion message.

coalib/testing/LocalBearTestHelper.py
jayvdb: changing these to using `assert ...` is not part of "Inherit from ..." if this class also inherits from `unittest.TestCase`.

So you can investigate whether these are needed later.
Also rather than having custom asserts here, we can and should use arg annotations and our decororator which enforces those.  Also noted this in https://github.com/coala/coala/pull/5522#discussion_r195143624
But again, changes to these assertions is a functional change which doesnt need to be in the same commit as "Inherit from ...", and it does need a separate issue created where you *show* that the output can be improved.  As Niklas has said, pytest already gives very detailed information about any failures, and pytest does that even for `unittest.TestCase` subclasses, so any custom type error messages need to be better than pytest's output.

coalib/testing/LocalBearTestHelper.py
```diff
@@ -242,16 +219,14 @@ def check_results(self,
...
-        self.assertIsInstance(local_bear,
-                              LocalBear,
-                              msg='The given bear is not a local bear.')
-        self.assertIsInstance(lines,
-                              (list, tuple),
-                              msg='The given lines are not a list.')
-        self.assertIsInstance(results,
-                              list,
-                              msg='The given results are not a list.')
+        assert isinstance(
```
jayvdb: as above.  the way to improve these is via annotations.

coalib/testing/LocalBearTestHelper.py
sangamcse: @NiklasMM, when I removed the custom message, `unittest`'s `self.assertIsInstance(..)` gave 
```py
E   AssertionError: '' is not an instance of (<class 'list'>, <class 'tuple'>)
```
 
but `pytest`'s `assert` gave
```py
E       AssertionError
```

tests/testing/LocalBearTestHelperTest.py
```diff
@@ -78,19 +78,12 @@ def test_bad_assertComparableObjectsEqual(self):
...
-                         '- This file has 3 lines.\n'
-                         '?               ^\n'
-                         '+ This file has 2 lines.\n'
-                         '?               ^\n'
-                         ' : message_base mismatch.\n\n',
-                         str(cm.exception))
+        self.assertEqual('origin mismatch: LineCountTestBear, This file '
+                         'has 3 lines. != LineCountTestBea, This file '
+                         'has 2 lines.\n\n'
+                         'message_base mismatch: LineCountTestBear, This '
```
jayvdb: this is a bug somewhere, and a regression.

1. origin mismatch & message_base mismatch both say the same thing. that is confusing.  something needs to be improved.

2. the pre-existing shows the line differences, and your change doesnt.  That is a regression, making it more difficult for bear writers to see the problem, not more easy which your project was supposed to do.

sangamcse: Done in PR https://github.com/coala/coala/pull/5564

Issue #1101, opened 2015-12-02
Bear debugging: Add Pdb integration!


Makman2:
This would be really cool to tightly integrate bear debugging with the built-in debugger from python Pdb.

Thinking of something like 
- jumping directly into the bear code
- easy settings investigation
- result inspector (/similar)?

Pdb is documented here (for Python 3.4, other versions exist too obviously :D):
https://docs.python.org/3.4/library/pdb.html

Makman2:
For "jumping directly into bear code": If starting coala in debug mode for a specific bear, use the `pdb` module directly to start a debug session (via `pdb.run`) so the user doesn't need to load the module manually with python on startup (via `python3 -m pdb ...`)

sils1297:
I've put a little time into https://github.com/coala-analyzer/coala/issues/565 , which is related, and found that it's not easy ;) It's not hard but I wouldn't say this is something for somebody completely new, which is the easy tag for me.

stellargo:
@Makman2 So I was wondering how exactly do we want this, like adding an import module to every bear? That seems like not the best way to go about this i guess.

stellargo:
Also just tell me if I am right, as of now we have -V, along with a --debug which is different basically because it is using ipdb for exception handling. These do not debug bear code and now we want to use pdb for doing that

stellargo:
Also :P, we can get into the bear code with a pdb.set_trace(), but exiting is a little issue. pdb natively does not have a function defined in it for this which we would need. Having our own edited pdb would solve it though.

Makman2:
> like adding a pdb.set_trace() to every bear?

That's not bearable, we might add it to the bear base class, however then it steps into "internal code" again. Would be nice to do something like `pdf.set_trace(bear_instance.run)` or so.

> pdb natively does not have a function defined in it for this which we would need.

You sure? Some people already told me there should be something. Haven't looked at the API more detailed myself though.

> Also just tell me if I am right, as of now we have -V, along with a --debug which is different basically because it is using ipdb for exception handling. These do not debug bear code and now we want to use pdb for doing that

Yes `--debug` would be a bit confusing in this context. We might change CLI-API then.

Makman2:
>> pdb natively does not have a function defined in it for this which we would need.
>
> You sure? Some people already told me there should be something. Haven't looked at the API more detailed myself though.

Okay I just looked, it seems it really doesn't exist. Even defining custom breakpoints from code seems to be not possible that easily :/

[...]

PR #5511, opened 2018-06-10
Add: Debugging Interface

RaiVaibhav:
Moved from https://github.com/RaiVaibhav/coala/pull/5
Related to https://github.com/coala/coala/issues/1101

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -1,4 +1,6 @@
 import inspect
+import pdb
+import collections
 import traceback
 from functools import partial
 from os import makedirs
@@ -20,6 +22,33 @@ from coalib.settings.Section import Section
 from coalib.settings.ConfigurationGathering import get_config_directory
 
 from .meta import bearclass
+
+
+class Debugger(pdb.Pdb):
+
+    def do_quit(self, arg):
+        self.clear_all_breaks()
+        super().do_continue(arg)
+        return 1
+
+    do_q = do_quit
+    do_exit = do_quit
+
+
+def debug_run(func, dbg=None, *args, **kwargs):
+    dbg = Debugger() if dbg is None else dbg
+    bear_results = dbg.runcall(func, *args, **kwargs)
+    if isinstance(bear_results, collections.Iterable):
+        results = []
+        iterator = iter(bear_results)
+        try:
+            while True:
+                result = dbg.runcall(next, iterator)
+                results.append(result)
+        except StopIteration:
+            return results
+    else:
+        return bear_results
 
 
 class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
@@ -232,7 +261,8 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
     def __init__(self,
                  section: Section,
                  message_queue,
-                 timeout=0):
+                 timeout=0,
+                 debugger=False):
         """
         Constructs a new bear.
 
@@ -252,6 +282,7 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
         self.section = section
         self.message_queue = message_queue
         self.timeout = timeout
+        self.debugger = debugger
 
         self.setup_dependencies()
         cp = type(self).check_prerequisites()
@@ -287,8 +318,10 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
             self.warn('The bear {} cannot be executed.'.format(
                 self.name), str(err))
             return
-
-        return self.run(*args, **kwargs)
+        if self.debugger:
+            return debug_run(self.run, Debugger(), *args, **kwargs)
+        else:
+            return self.run(*args, **kwargs)
 
     def execute(self, *args, debug=False, **kwargs):
         name = self.name
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -241,6 +241,14 @@ To run coala without user interaction, run the `coala --non-interactive`,
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
+    outputs_group.add_argument(
+        '--debug-bears', nargs='?', const=True,
+        help='Enable bear debugging with pdb, that can help to identify and'
+        ' correct errors in bear code. Steps into bear code as soon as being'
+        ' executed. To specify which bears to debug, supply bear names as'
+        ' additional arguments. If used without arguments, all bears specified'
+        ' with --bears will be debugged (even implicit dependency bears).')
+
     misc_group = arg_parser.add_argument_group('Miscellaneous')
 
     misc_group.add_argument(
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -27,7 +27,7 @@ from coalib.results.result_actions.PrintDebugMessageAction import (
 from coalib.results.result_actions.ShowPatchAction import ShowPatchAction
 from coalib.results.RESULT_SEVERITY import RESULT_SEVERITY
 from coalib.results.SourceRange import SourceRange
-from coalib.settings.Setting import glob_list
+from coalib.settings.Setting import glob_list, typed_list
 from coalib.parsing.Globbing import fnmatch
 
 
@@ -297,7 +297,8 @@ def instantiate_bears(section,
                       file_dict,
                       message_queue,
                       console_printer,
-                      debug=False):
+                      debug=False,
+                      debug_bears=False):
     """
     Instantiates each bear with the arguments it needs.
 
@@ -311,13 +312,26 @@ def instantiate_bears(section,
     :param console_printer:  Object to print messages on the console.
     :return:                 The local and global bear instance lists.
     """
+    debug_bears = (False if debug_bears is False else
+                   [x.lower() for x in debug_bears])
+    for bear in local_bear_list + global_bear_list:
+        for deps_bear in bear.BEAR_DEPS:
+            if debug_bears is not False and (
+                bear.__name__.lower() in debug_bears and (
+                    deps_bear.__name__.lower() not in debug_bears)):
+                debug_bears.append(deps_bear.__name__.lower())
+
     instantiated_local_bear_list = []
     instantiated_global_bear_list = []
     for bear in local_bear_list:
         try:
+            debugger = True if debug_bears is not False and (
+                    debug_bears[0] == 'true' or (
+                        bear.__name__.lower() in debug_bears)) else False
             instantiated_local_bear_list.append(bear(section,
                                                      message_queue,
-                                                     timeout=0.1,))
+                                                     timeout=0.1,
+                                                     debugger=debugger))
         # RuntimeError it will be raised only when debug will be set to True.
         except RuntimeError:
             if debug:
@@ -325,10 +339,14 @@ def instantiate_bears(section,
 
     for bear in global_bear_list:
         try:
+            debugger = True if debug_bears is not False and (
+                    debug_bears[0] == 'true' or (
+                        bear.__name__.lower() in debug_bears)) else False
             instantiated_global_bear_list.append(bear(file_dict,
                                                       section,
                                                       message_queue,
-                                                      timeout=0.1,))
+                                                      timeout=0.1,
+                                                      debugger=debugger))
         # RuntimeError it will be raised only when debug will be set to True
         except RuntimeError:
             if debug:
@@ -345,7 +363,8 @@ def instantiate_processes(section,
                           log_printer,
                           console_printer,
                           debug=False,
-                          use_raw_files=False):
+                          use_raw_files=False,
+                          debug_bears=False):
     """
     Instantiate the number of processes that will run bears which will be
     responsible for running bears in a multiprocessing environment.
@@ -379,7 +398,7 @@ def instantiate_processes(section,
     complete_file_dict = get_file_dict(complete_filename_list,
                                        allow_raw_files=use_raw_files)
 
-    if debug:
+    if debug or debug_bears:
         from . import DebugProcessing as processing
     else:
         import multiprocessing as processing
@@ -399,7 +418,8 @@ def instantiate_processes(section,
         complete_file_dict,
         message_queue,
         console_printer=console_printer,
-        debug=debug)
+        debug=debug,
+        debug_bears=debug_bears)
     loaded_valid_local_bears_count = len(local_bear_list)
     # Note: the complete file dict is given as the file dict to bears and
     # the whole project is accessible to every bear. However, local bears are
@@ -541,7 +561,8 @@ def process_queues(processes,
                    log_printer,
                    console_printer,
                    debug=False,
-                   apply_single=False):
+                   apply_single=False,
+                   debug_bears=False):
     """
     Iterate the control queue and send the results received to the print_result
     method so that they can be presented to the user.
@@ -585,7 +606,7 @@ def process_queues(processes,
     ignore_ranges = list(yield_ignore_ranges(file_dict))
 
     # One process is the logger thread (if not in debug mode)
-    while local_processes > (1 if not debug else 0):
+    while local_processes > (1 if not (debug or debug_bears) else 0):
         try:
             control_elem, index = control_queue.get(timeout=0.1)
 
@@ -737,7 +758,12 @@ def execute_section(section,
                              results (bear names are key) as well as the
                              file dictionary.
     """
-    if debug:
+    debug_bears = (False
+                   if 'debug_bears' not in section or (
+                       section['debug_bears'].value == 'False') else
+                   typed_list(str)(section['debug_bears']))
+
+    if debug or debug_bears:
         running_processes = 1
     else:
         try:
@@ -773,11 +799,12 @@ def execute_section(section,
                                                 None,
                                                 console_printer=console_printer,
                                                 debug=debug,
-                                                use_raw_files=use_raw_files)
+                                                use_raw_files=use_raw_files,
+                                                debug_bears=debug_bears)
 
     logger_thread = LogPrinterThread(arg_dict['message_queue'])
     # Start and join the logger thread along with the processes to run bears
-    if not debug:
+    if not (debug or debug_bears):
         # in debug mode the logging messages are directly processed by the
         # message_queue
         processes.append(logger_thread)
@@ -797,12 +824,13 @@ def execute_section(section,
                                None,
                                console_printer=console_printer,
                                debug=debug,
-                               apply_single=apply_single),
+                               apply_single=apply_single,
+                               debug_bears=debug_bears),
                 arg_dict['local_result_dict'],
                 arg_dict['global_result_dict'],
                 arg_dict['file_dict'])
     finally:
-        if not debug:
+        if not (debug or debug_bears):
             # in debug mode multiprocessing and logger_thread are disabled
             # ==> no need for following actions
             logger_thread.running = False
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -291,24 +291,6 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
     return file_dict
 
 
-def filter_raising_callables(it, exception, *args, debug=False, **kwargs):
-    """
-    Filters all callable items inside the given iterator that raise the
-    given exceptions.
-
-    :param it:        The iterator to filter.
-    :param exception: The (tuple of) exception(s) to filter for.
-    :param args:      Positional arguments to pass to the callable.
-    :param kwargs:    Keyword arguments to pass to the callable.
-    """
-    for elem in it:
-        try:
-            yield elem(*args, **kwargs)
-        except exception:
-            if debug:
-                raise
-
-
 def instantiate_bears(section,
                       local_bear_list,
                       global_bear_list,
@@ -329,26 +311,30 @@ def instantiate_bears(section,
     :param console_printer:  Object to print messages on the console.
     :return:                 The local and global bear instance lists.
     """
-    local_bear_list = [bear
-                       for bear in filter_raising_callables(
-                           local_bear_list,
-                           RuntimeError,
-                           section,
-                           message_queue,
-                           timeout=0.1,
-                           debug=debug)]
+    instantiated_local_bear_list = []
+    instantiated_global_bear_list = []
+    for bear in local_bear_list:
+        try:
+            instantiated_local_bear_list.append(bear(section,
+                                                     message_queue,
+                                                     timeout=0.1,))
+        # RuntimeError it will be raised only when debug will be set to True.
+        except RuntimeError:
+            if debug:
+                raise
 
-    global_bear_list = [bear
-                        for bear in filter_raising_callables(
-                            global_bear_list,
-                            RuntimeError,
-                            file_dict,
-                            section,
-                            message_queue,
-                            timeout=0.1,
-                            debug=debug)]
+    for bear in global_bear_list:
+        try:
+            instantiated_global_bear_list.append(bear(file_dict,
+                                                      section,
+                                                      message_queue,
+                                                      timeout=0.1,))
+        # RuntimeError it will be raised only when debug will be set to True
+        except RuntimeError:
+            if debug:
+                raise
 
-    return local_bear_list, global_bear_list
+    return instantiated_local_bear_list, instantiated_global_bear_list
 
 
 def instantiate_processes(section,
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,3 +1,4 @@
+import pdb
 from collections import defaultdict
 import datetime
 from io import BytesIO
@@ -7,6 +8,7 @@ from os.path import abspath, exists, isfile, join, getmtime
 import shutil
 
 from freezegun import freeze_time
+from unittest.mock import patch
 
 import requests
 import requests_mock
@@ -14,7 +16,7 @@ import requests_mock
 from coalib.bearlib.aspects.collections import AspectList
 from coalib.bearlib.aspects.Metadata import CommitMessage
 from coalib.bearlib.languages.Language import Language, Languages
-from coalib.bears.Bear import Bear
+from coalib.bears.Bear import Bear, Debugger
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.bears.GlobalBear import GlobalBear
 from coalib.bears.LocalBear import LocalBear
@@ -462,6 +464,30 @@ class BearTest(BearTestBase):
         self.assertIsInstance(result, Language)
         self.assertEqual(str(result), 'Hypertext Markup Language 5.1')
         self.check_message(LOG_LEVEL.DEBUG)
+
+    # Mock test added to solve the coverage problem by DebugBearsTest
+    @patch('pdb.Pdb.do_continue')
+    def test_custom_continue(self, do_continue):
+        arg = {}
+        self.assertEqual(Debugger().do_quit(arg), 1)
+        pdb.Pdb.do_continue.assert_called_once_with(arg)
+
+    # side_effect effectively implements run() method of bear
+    @patch('coalib.bears.Bear.Debugger.runcall', side_effect=((1, 2), 1, 2))
+    def test_debug_run_with_return(self, runcall):
+        section = Section('name')
+        my_bear = Bear(section, self.queue, debugger=True)
+        args = ()
+        kwargs = {}
+        self.assertEqual(my_bear.run_bear_from_section(args, kwargs), [1, 2])
+
+    @patch('coalib.bears.Bear.Debugger.runcall', return_value=None)
+    def test_debug_run_with_no_return(self, runcall):
+        section = Section('name')
+        my_bear = Bear(section, self.queue, debugger=True)
+        args = ()
+        kwargs = {}
+        self.assertIsNone(my_bear.run_bear_from_section(args, kwargs))
 
 
 class BrokenReadHTTPResponse(BytesIO):
--- a/tests/bears/DebugBearsTest.py
+++ b/tests/bears/DebugBearsTest.py
@@ -1,0 +1,62 @@
+import unittest
+import sys
+
+from io import StringIO
+
+from coalib.bears.Bear import Debugger, debug_run
+
+
+def func1(*args, **kwargs):
+    yield 1
+    yield 2
+    yield 3
+
+
+def func2(*args, **kwargs):
+    return [1, 2]
+
+
+def func3(*args, **kwargs):
+    return func1(*args, **kwargs)
+
+
+def execute_debugger(debugger_commands, func, *args, **kwargs):
+    input = StringIO('\n'.join(debugger_commands))
+    output = StringIO()
+    dbg = Debugger(stdin=input, stdout=output)
+    return debug_run(func, dbg, *args, **kwargs), output.getvalue()
+
+
+class DebugBearsTest(unittest.TestCase):
+    def setUp(self):
+        # restore the coverage settrace to prevent the coverage breakage
+        # on project because we can't chain coverage trace to run parallel
+        # with debugger. To increase the coverage Mock test has been added in
+        # BearTest file.
+        # https://goo.gl/sKaJfh
+        self.trace = sys.gettrace()
+
+    def tearDown(self):
+        sys.settrace(self.trace)
+
+    def test_run_return_yield_with_debugger(self):
+        result, output = execute_debugger('qcqc', func1)
+        self.assertEqual(result, [1, 2, 3])
+        lines = output.splitlines()
+        self.assertEqual(lines[1], '-> yield 1')
+        self.assertEqual(lines[3], '-> yield 2')
+        self.assertEqual(lines[5], '-> yield 3')
+
+    def test_run_return_list_with_debugger(self):
+        result, output = execute_debugger('q', func2)
+        self.assertEqual(result, [1, 2])
+        lines = output.splitlines()
+        self.assertEqual(lines[1], '-> return [1, 2]')
+
+    def test_run_return_generator_with_debugger(self):
+        result, output = execute_debugger('qcqcq', func3)
+        self.assertEqual(result, [1, 2, 3])
+        lines = output.splitlines()
+        self.assertEqual(lines[3], '-> yield 1')
+        self.assertEqual(lines[5], '-> yield 2')
+        self.assertEqual(lines[7], '-> yield 3')
--- a/tests/processes/ProcessingTest.py
+++ b/tests/processes/ProcessingTest.py
@@ -566,8 +566,8 @@ class ProcessingTest(unittest.TestCase):
     def test_loaded_bears_with_error_result(self):
         class BearWithMissingPrerequisites(Bear):
 
-            def __init__(self, section, queue, timeout=0.1):
-                Bear.__init__(self, section, queue, timeout)
+            def __init__(self, section, queue, timeout=0.1, debugger=False):
+                Bear.__init__(self, section, queue, timeout, debugger)
 
             def run(self):
                 return []
@@ -606,28 +606,33 @@ class ProcessingTest(unittest.TestCase):
     def test_global_instantiation(self):
         class TestOneBear(Bear):
 
-            def __init__(self, file_dict, section, queue, timeout=0.1):
+            def __init__(self, file_dict, section, queue, timeout=0.1,
+                         debugger=False):
                 raise RuntimeError
 
         class TestTwoBear(Bear):
 
             BEAR_DEPS = {TestOneBear}
 
-            def __init__(self, file_dict, section, queue, timeout=0.1):
+            def __init__(self, file_dict, section, queue, timeout=0.1,
+                         debugger=False):
                 raise RuntimeError
 
         class TestThreeBear(Bear):
 
             BEAR_DEPS = {TestTwoBear}
 
-            def __init__(self, file_dict, section, queue, timeout=0.1):
-                Bear.__init__(self, section, queue, timeout)
+            def __init__(self, file_dict, section, queue, timeout=0.1,
+                         debugger=False):
+                Bear.__init__(self, section, queue, timeout, debugger)
 
         global_bear_list = [TestTwoBear, TestThreeBear]
         list1, list2 = instantiate_bears(self.sections['cli'], [],
                                          global_bear_list, {}, self.queue,
                                          console_printer=self.console_printer,
-                                         debug=False)
+                                         debug=False,
+                                         debug_bears=['TestTwoBear',
+                                                      'TestThreeBear'])
 
         self.assertEqual(len(list1), 0)
         self.assertEqual(len(list2), 1)
@@ -635,7 +640,7 @@ class ProcessingTest(unittest.TestCase):
             global_bear_list = [TestOneBear]
             instantiate_bears(self.sections['cli'], [], global_bear_list, {},
                               self.queue, console_printer=self.console_printer,
-                              debug=True)
+                              debug=True, debug_bears=False)
 
 
 class ProcessingTest_GetDefaultActions(unittest.TestCase):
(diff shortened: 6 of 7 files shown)


coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
 from .meta import bearclass
 
 
+class db(pdb.Pdb):
+    def do_continue(self, arg):
+        self.clear_all_breaks()
+        super().do_continue(arg)
+        return(1)
```
Makman2: `return 1`, redundant braces

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
...
+
+
+dbg = db()
+dbg.do_q = dbg.do_continue
+dbg.do_quit = dbg.do_continue
+dbg.do_exit = dbg.do_continue
```
Makman2: what about assigning those to `db`?

```python
class dbg(pdb.Pdb):
    def __init__(self, ...):
        self.do_q = self.do_continue
        ...
    def do_continue(self, arg):
        self.clear_all_breaks()
        super().do_continue(arg)
        return 1
```

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
 from .meta import bearclass
 
 
+class db(pdb.Pdb):
+    def do_continue(self, arg):
```
Makman2: Shouldn't it be called somehow differently? Otherwise you override `do_c` / `do_continue` implicitly, and it will now always clear breakpoints, regardless of doing `q` or `c`.

tests/bears/BearTest.py
```diff
@@ -463,6 +473,30 @@ def test_bear_with_specific_language(self):
...
+    def test_do_continue(self, do_continue):
+        arg = {}
+        self.assertEqual(dbg.do_continue(arg), 1)
+
+    @patch('coalib.bears.Bear.dbg.runcall', side_effect=((1, 2), 3, 4))
+    def test_debug_mode_func_with_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
+        args = {}
+        kwargs = {}
```
Makman2: please read my comments on the old PR.

coalib/processes/Processing.py
```diff
@@ -329,26 +324,32 @@ def instantiate_bears(section,
...
+    for bear in local_bear_list:
+        try:
+            debugger = debug_flag_checker(bear, debug_bears=debug_bears)
+            instantiated_local_bear_list.append(bear(section,
+                                                     message_queue,
+                                                     timeout=0.1,
+                                                     debugger=debugger))
+        except RuntimeError:
+            if debug:
+                raise
```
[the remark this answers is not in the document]
RaiVaibhav: No not `re-raising` the `RuntimeError` it will be raised only when `debug` will set `True`. (If there is any RuntimeError then only)
By default `debug` is `False`. If user passed the arg `--debug` then this `debug` will set `True` and after that, it will catch the `RuntimeError` and raise the `RuntimeError` and `ipdb` will gets raised. If I removed the `debug == True` and let the exception gets `rasied` directly, then it will break the coala every time if any bear raises `RuntimeError` during its initialization, and why `RuntimeError` as the same process was happening on previously also.

And don't want to remove `debug==True` because here I like the concept of @userzimmermann  to raise a `ipdb` only when `--debug` is getting called.

coalib/processes/Processing.py
```diff
@@ -291,22 +292,15 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
...
-                raise
+def debug_flag_checker(bear, debug_bears=False):
+    if debug_bears is False:
+        debugger = False
+    elif ((str(debug_bears[0]).lower().strip()) == 'true') or (
+            (bear.__name__) in debug_bears):
+        debugger = True
+    else:
+        debugger = False
+    return debugger
```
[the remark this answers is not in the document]
RaiVaibhav: I think I have already explained here https://github.com/RaiVaibhav/coala/pull/5#discussion_r192411565
please take a look, why this function is needed and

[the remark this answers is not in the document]
RaiVaibhav: I am using this function only to initialize the `debugger` flag to `True` and this function checks whether  name of bears is in `--debug-bears` argument and accordinly it set the `debbuger` flag and return the `debugger` flag to `instantiate_bears` which will initalize the bear with `debugger` flag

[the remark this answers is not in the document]
RaiVaibhav: then above `if-else` clause places accordingly to `prevent` any error during checking the condition :)

coalib/bears/Bear.py
RaiVaibhav: No it will not IMO thats how `cmd.Cmd()` is working, if we don't explicitly  which command do what then it will not interfere those commands which is defined in `super()`

coalib/processes/Processing.py
```diff
@@ -751,7 +755,12 @@ def execute_section(section,
                              results (bear names are key) as well as the
                              file dictionary.
     """
-    if debug:
+    debug_bears = (False
+                   if 'debug_bears' not in section or (
+                       section['debug_bears'].value == 'False') else (
+                       typed_list(str)(section['debug_bears'])))
```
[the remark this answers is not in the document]
RaiVaibhav: I didn't get  If I don't do something like `typed_list(str)(section['debug_bears'])` then it will throw error :sweat_smile:

coalib/bears/Bear.py
RaiVaibhav: Actually `do_c` or `do_cont` will not but ya `do_continue` will, so making it `custom_quit`

coalib/processes/Processing.py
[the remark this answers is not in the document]
RaiVaibhav: As `debug_bears` in `Processing.py` can contain `Bool` Value i.e, `False` and it can contain a list like `['True']`, ['ABear', 'BBear', 'CBear']`

coalib/processes/Processing.py
```diff
@@ -291,22 +292,15 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
...
-    :param exception: The (tuple of) exception(s) to filter for.
-    :param args:      Positional arguments to pass to the callable.
-    :param kwargs:    Keyword arguments to pass to the callable.
-    """
-    for elem in it:
-        try:
-            yield elem(*args, **kwargs)
-        except exception:
-            if debug:
-                raise
```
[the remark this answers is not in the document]
RaiVaibhav: It now will be difficult :sweat_smile: , ahh... it now soo messy for me to repaste all :sweat_smile:

tests/bears/BearTest.py
RaiVaibhav: I know you said arg is a tuple and you said -->`Do you know you are passing empty dictionary` but on the `gitter` channel :P, actually these `arg and `kwargs` will not get used as I am using `mock.patch`.

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
...
+        return(1)
+
+
+dbg = db()
+dbg.do_q = dbg.do_continue
+dbg.do_quit = dbg.do_continue
+dbg.do_exit = dbg.do_continue
+
+
+def debug_mode_function(func, dbg, *args, **kwargs):
```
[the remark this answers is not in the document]
RaiVaibhav: debug_run is best, a small one :P

RaiVaibhav:
wip

coalib/parsing/DefaultArgParser.py
```diff
@@ -241,6 +241,14 @@ def default_arg_parser(formatter_class=None):
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
+    outputs_group.add_argument(
+        '--debug-bears', nargs='?', const=True, help='Debugging Bears will'
+        'help user in identifying and removing errors from  Bear code'
+        'with the help of Pdb interface, additionally user can pass'
+        'bear name as a additional argument to debug those specific bear,'
+        'if no additional arguments passed then it will debug all the bears'
+        'passed to a --bears or -b argument')
```
Makman2: A bit rephrasing to make it more clear.

> Enable bear debugging with pdb, that can help to identify and correct errors in bear code. Steps into bear code as soon as being executed.
> To specify which bears to debug, supply bear names as additional arguments.
> If used without arguments, all bears specified with --bears will be debugged (even implicit dependency bears).

1. First explaining what the flag actually does, not what it is for.
2. Explain shortly why that's useful.
3. Then explain how to use the flag. :)
4. It seems the file uses the term "bear" lower case, so I would do that too.

coalib/processes/Processing.py
```diff
@@ -751,7 +755,12 @@ def execute_section(section,
                              results (bear names are key) as well as the
                              file dictionary.
     """
-    if debug:
+    debug_bears = (False
+                   if 'debug_bears' not in section or (
+                       section['debug_bears'].value == 'False') else (
```
Makman2: redundant braces after `else`

coalib/processes/Processing.py
```diff
@@ -751,7 +755,12 @@ def execute_section(section,
...
+                            typed_list(str)(section['debug_bears'])))
```
Makman2: Indentation: 8 spaces back

coalib/processes/Processing.py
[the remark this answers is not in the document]
Makman2: 1. Interactive rebase
2. mark this commit as `edit`
3. `git reset HEAD~1`
4. `git add -p` and stage only things belonging to removing `filter_raising_callables`
5. create commit
6. `git add -p` and stage remaining code
7. create commit
8. Continue rebase: `git rebase --continue`

Then you have split up the commit.

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
 from .meta import bearclass
 
 
+class db(pdb.Pdb):
+    def __init__(self, *args, **kwargs):
+        self.do_q = self.custom_quit
+        self.do_quit = self.custom_quit
+        self.do_exit = self.custom_quit
+        super(db, self).__init__(*args, **kwargs)
```
Makman2: `__init__` of superclasses should be invoked at top if possible.

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
 from .meta import bearclass
 
 
+class db(pdb.Pdb):
+    def __init__(self, *args, **kwargs):
+        self.do_q = self.custom_quit
+        self.do_quit = self.custom_quit
+        self.do_exit = self.custom_quit
```
Makman2: I think you change this per-instance-patching to a per-class-patching (then you can completely leave out `__init__`, as you've nothing more to do there):
```python
class db(pdb.Pdb):
    def do_quit(self, arg):
        self.clear_all_bears()
        super().do_continue(arg)  # <-- super is actually redundant, but you can keep it since it makes the intention clear that we want to use the old, known "continue"
        return 1

    do_q = do_quit
    do_exit = do_quit
```

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
...
+        super(db, self).__init__(*args, **kwargs)
+
+    def custom_quit(self, arg):
+        self.clear_all_breaks()
```
Makman2: We keep that, it's definitely good, but I've thought about the scenario of setting breakpoints across bears or utilities. Imagine this:

Some bear fails. You now want to debug. You step into code, experiment around, and you have set some breakpoints. The bear that failed is a local bear, so you step each time a new file arises into `run`. But since `runcall` has ended already for the last `run`, breakpoints you've set are gone.

It would be cool to disable breakpoints on exit, but remember them until the complete coala session ends (or some other point we thing breakpoints shall be kept alive).

But: We do that after merge, that's something to discuss later since it requires more design work.

coalib/bears/Bear.py
```diff
@@ -22,6 +24,34 @@
...
+        super().do_continue(arg)
+        return 1
+
+
+def debug_run(func, dbg=pdb.Pdb(), *args, **kwargs):
```
Makman2: ah sorry, I meant `db` as a default value for `dbg`, my bad :sweat_smile:

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - run:
           name: pytest
           when: always
-          command: python -m pytest
+          command: bash .misc/pytest_run.sh
```
Makman2: Maybe call the script `tests.sh` instead?

coalib/processes/Processing.py
Makman2: `section` offers a defaulted `get` function, so you can shorten code here I believe.
```python
typed_list(str)(section(...))
if bool(section.get('debug_bears', False)) else
False

File: coalib/processes/Processing.py (hunk 12 above)
Comment (Makman2) [inline coalib/processes/Processing.py]: Does it have to be a `typed_list`? That's intended for bear-use only^^

Comment (Makman2) [inline coalib/processes/Processing.py]: Just thinking, what if you now specify `debug-bears = True` in a coafile. Won't you now get a typed list with `['True']` as value?

Comment (RaiVaibhav) [reply coalib/processes/Processing.py]: No it will split it up like `['T', 'r', 'u', 'e']`

Comment (RaiVaibhav) [reply coalib/processes/Processing.py]: Yes `typed_list` is used for the case like `--debug-bears ABear, BBear..` it will split up the the value like `['ABear' 'BBear']`

File: coalib/processes/Processing.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (RaiVaibhav) [reply coalib/processes/Processing.py]: I think I have already make it as a different commit

File: coalib/processes/Processing.py (hunk 11 above)
Comment (RaiVaibhav) [reply coalib/processes/Processing.py]: `bool(section.get('debug_bears', False))` will throw ValueError for the case `--debug-bears ABear` and If I do something like `bool(section.get('debug_bears', False).value)` then for the case `--debug-bears False` it will set the `debug_bears` value to `['False']` then in above code  I will have to check this case also. :sweat_smile:

File: coalib/processes/Processing.py (hunk 12 above)
Comment (RaiVaibhav) [reply coalib/processes/Processing.py]: Every time I am trying to make the `8 spaces back`  `PEP8Bear` made the change during running coala  :sweat_smile:

File: coalib/bears/Bear.py (hunk 15 above)
Comment (RaiVaibhav) [reply coalib/bears/Bear.py]: Point noted Bookmarking this comment. :+1:

**Comment by RaiVaibhav** [comment]
@gitmate-bot rebase

File: coalib/parsing/DefaultArgParser.py [hunk 18]
```diff
@@ -241,6 +241,15 @@ def default_arg_parser(formatter_class=None):
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
+    outputs_group.add_argument(
+        '--debug-bears', nargs='?', const=True,
+        help='Enable bear debugging with pdb, that can help to identify and'
+        'correct errors in bear code. Steps into bear code as soon as being'
```
jayvdb: have a look at this on the cli. Terrible :)

coalib/processes/Processing.py
```diff
@@ -751,7 +757,12 @@ def execute_section(section,
                              results (bear names are key) as well as the
                              file dictionary.
     """
-    if debug:
+    debug_bears = (False
```
jayvdb: This should be a normal if

tests/processes/ProcessingTest.py
```diff
@@ -347,39 +347,6 @@ def test_create_process_group(self):
             # python modules subprocess and os
             self.assertEqual(p.pid, pgid)
 
-    def test_filter_raising_callables(self):
```
jayvdb: Why is this removed?

coalib/processes/Processing.py
RaiVaibhav: didn't get why?

tests/processes/ProcessingTest.py
Makman2: It would have gotten too complicated keeping this as a dedicated function. It aimed to be simple, expecting a fixed set of arguments for functions/classes being invoked. But with the new debugger changes, we instantiate bears differently, depending on whether we have switched on debugging for them or not. I recommended @RaiVaibhav to inline it, so future modifications are easier + better to understand.

coalib/bears/Bear.py
```diff
@@ -22,6 +24,32 @@
...
+    def do_quit(self, arg):
+        self.clear_all_breaks()
+        super().do_continue(arg)
+        return 1
+
+    do_q = do_quit
+    do_exit = do_quit
+
+
+def debug_run(func, dbg=db(), *args, **kwargs):
```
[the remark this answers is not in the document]
Makman2: Okay good to know that you have a reasonable preference, it was actually my suggestion to do it like this :3

setup.cfg
```diff
@@ -48,6 +48,8 @@ branch = True
 cover_pylib = False
 source =
     coalib
+omit =
+    tests/bears/DebugBearsTest.py
```
Makman2: Why do you exclude the test file? Its coverage isn't measured anyway^^

coalib/bears/Bear.py
```diff
@@ -22,6 +24,32 @@
 from .meta import bearclass
 
 
+class db(pdb.Pdb):
```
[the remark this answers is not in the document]
Makman2: We could maybe just write it out -> `Debugger`

coalib/processes/Processing.py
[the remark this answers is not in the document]
Makman2: Ah understood. Could you please write a comment there then? That makes it clear for others as well :)

(sorry for responding so late, missed this comment)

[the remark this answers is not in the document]
Makman2: Also for the global bear. Even if it's duplicated, people see it immediately :+1:

coalib/processes/Processing.py
```diff
@@ -733,7 +757,12 @@ def execute_section(section,
...
+                   if 'debug_bears' not in section or (
+                       section['debug_bears'].value == 'False') else
+                   typed_list(str)(section['debug_bears']))
```
Makman2: does it have to be a typed list?

coalib/bears/Bear.py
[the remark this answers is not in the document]
RaiVaibhav: Yeah will be better :+1:

coalib/processes/Processing.py
Makman2: me neither, imo that's fine

coalib/processes/Processing.py
RaiVaibhav: I use the typed_list so that it will split the comma separated value.

RaiVaibhav: Any other suggestion will be helpfull for me :grinning:

tests/processes/ProcessingTest.py
RaiVaibhav: help me in getting `inline` I moved it into separate commit :sweat_smile:  (english problem sorry)

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assertEqual(debug_run(fun, dbg, *args, **kwargs), [1, 2, 3])
+        output = output.getvalue()
+        sys.settrace(old_trace)
+        # restore the coverage settrace to prevent the coverage breakage
+        # on project because  we can't chain coverage trace to run parallel
+        # with debugger. To increase the coverage Mock test has been added in
+        # BearTest file.
+        self.assertEqual(output.split('\n')[1], '-> yield 1')
+        self.assertEqual(output.split('\n')[3], '-> yield 2')
+        self.assertEqual(output.split('\n')[5], '-> yield 3')
```
Makman2: more tests please :)
- `run()` that doesn't yield, but returns an `Iterable`
- `run()` that returns a generator.

tests/processes/ProcessingTest.py
```diff
@@ -602,6 +603,71 @@ def check_prerequisites(cls):
...
+
+            BEAR_DEPS = {TestThreeBear}
+
+            def __init__(self,
+                         file_dict,
+                         section,
+                         queue, timeout=0.1,
+                         debugger=False):
+                raise RuntimeError
+                Bear.__init__(self, section, queue, timeout, debugger)
```
Makman2: redundant statement :D

tests/processes/ProcessingTest.py
```diff
@@ -602,6 +603,71 @@ def check_prerequisites(cls):
...
+
+            def run(self, filename, file):
+                pass
```
Makman2: If you don't run the bear, you don't have to override `run`. It raises by default `NotImplemented`, but that won't happen anyway :+1:

setup.cfg
Makman2: tests are by default omitted for coverage, measuring test coverage usually doesn't make sense :P

Makman2: test coverage of tests* :P

coalib/processes/Processing.py
Makman2: Ahhh nvm... without `typed_list` you don't get a solid list type you can use in Python like a normal list... Nvm, you're doing it totally correct :+1:

coalib/parsing/DefaultArgParser.py
RaiVaibhav:
```
  --debug-bears         Debugging Bears willhelp user in identifying and
                        removing errors from Bear codewith the help of Pdb
                        interface, additionally user can passbear name as a
                        additional argument to debug those specific bear,if no
                        additional arguments passed then it will debug all the
                        bearspassed to a --bears or -b argument
```

Its not fine? I know there are some spacing problem.

RaiVaibhav: @jayvdb  this is what I am getting in my terminal.

RaiVaibhav: were you talking about spacing problem?

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,65 @@
...
+        dbg = Debugger(stdin=input, stdout=self.output)
+        self.assertEqual(debug_run(func1, dbg, *args, **kwargs), [1, 2, 3])
+        output = self.output.getvalue()
+        # restore the coverage settrace to prevent the coverage breakage
+        # on project because  we can't chain coverage trace to run parallel
+        # with debugger. To increase the coverage Mock test has been added in
+        # BearTest file.
+        self.assertEqual(output.split('\n')[1], '-> yield 1')
+        self.assertEqual(output.split('\n')[3], '-> yield 2')
+        self.assertEqual(output.split('\n')[5], '-> yield 3')
```
Makman2: you can do `splitlines()` instead, and reuse the value:
```python
lines = output.splitlines()
self.assertEqual(lines[1], '-> yield 1')
``´

File: tests/bears/DebugBearsTest.py [hunk 29]
```diff
@@ -0,0 +1,65 @@
...
+    def tearDown(self):
+        sys.settrace(self.trace)
+
+    def test_debug_run_yield(self):
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        input = StringIO('q\nc\nq\nc')
+        dbg = Debugger(stdin=input, stdout=self.output)
+        self.assertEqual(debug_run(func1, dbg, *args, **kwargs), [1, 2, 3])
+        output = self.output.getvalue()
```
Makman2: maybe create a helper function in the test that does exactly what you are doing here:
1. Create a debugger instance, with a list of commands to execute
2. run `debug_run`
3. return the text the debugger generated in the output

So it looks in the end like
```python
output = self.execute_debugger(['q', 'c', 'q', 'c'])
```

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def func3(*args, **kwargs):
+    kwargs['Makman'] = 'Admin'
+    return func1(*args, **kwargs)
+
+
+class DebugBearsTest(unittest.TestCase):
+    def setUp(self):
+        self.trace = sys.gettrace()
```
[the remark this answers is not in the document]
Makman2: Btw, please include a link to coverage.py's issue where Ned tells that he doesn't know how to make coverage work together with pdb.

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_debug_run_list(self):
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        input = StringIO('q\nc')
+        dbg = Debugger(stdin=input, stdout=self.output)
+        self.assertEqual(debug_run(func2, dbg, *args, **kwargs), [1, 2])
+        output = self.output.getvalue()
+        self.assertEqual(output.split('\n')[1], '-> return([1, 2])')
+
+    def test_debug_run_generator(self):
```
Makman2: this name sounds technically equal to `test_debug_run_yield`. Can you call it differently / more precise?

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,65 @@
...
+    yield 2
+    yield 3
+
+
+def func2(*args, **kwargs):
+    return([1, 2])
+
+
+def func3(*args, **kwargs):
+    kwargs['Makman'] = 'Admin'
```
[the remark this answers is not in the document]
Makman2: Is it actually necessary? :O
I don't see you using it anywhere in the test :O

coalib/processes/Processing.py
```diff
@@ -330,26 +312,47 @@ def instantiate_bears(section,
...
-
-    return local_bear_list, global_bear_list
+    debug_bears = (False if debug_bears is False else
+                   [x.lower() for x in debug_bears])
+    for bear in local_bear_list + global_bear_list:
+        for deps_bear in bear.BEAR_DEPS:
+            if debug_bears is not False and (
+                bear.__name__.lower() in debug_bears and (
+                    deps_bear.__name__.lower() not in debug_bears)):
+                debug_bears.append(deps_bear.__name__.lower())
```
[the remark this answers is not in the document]
Makman2: So just move this into a different commit :+1:

tests/bears/DebugBearsTest.py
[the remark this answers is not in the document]
RaiVaibhav: Forgot to remove :disappointed:

tests/bears/DebugBearsTest.py
RaiVaibhav: Do I need to implement helper function now, can I do it after PR gets merge, I am already behind schedule
First phase work isn't completed yet

Makman2: Imo that's quite simple to do now:
```
def execute_debugger(debugger_commands, func, *args, **kwargs):
    input = StringIO('\n'.join(commands))
    output = StringIO()
    dbg = Debugger(stdin=input, stdout=output)
    debug_run(func, *args, **kwargs)
    return output.getvalue()
```
and then use it:
```
self.assertEqual(self.execute_debugger('q\nc\nq\nc', func1, *args, **kwargs))
```

Makman2: That's something you can reuse for all other test cases as well^^

coalib/bears/Bear.py
```diff
@@ -22,6 +24,32 @@
...
+    def do_quit(self, arg):
+        self.clear_all_breaks()
+        super().do_continue(arg)
+        return 1
+
+    do_q = do_quit
+    do_exit = do_quit
+
+
+def debug_run(func, dbg=Debugger(), *args, **kwargs):
```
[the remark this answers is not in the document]
RaiVaibhav: @Makman2  Let's suppose I do that then I also have to add the arguments `input` and `output` ( StringIO) to create a ideal test and simply said I will have to include the every argument of `pdb.Pdb.__init__` as a argument to `debug_run`.

A none will be great to preserve both condition.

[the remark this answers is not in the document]
Makman2: yes good point, nvm my comment then :+1: But I would also recommend to use `None` as a default, and if `None`:
```
if dbg is None:
    dbg = Debugger()
```

coalib/bears/Bear.py
RaiVaibhav: Making sense, I have to take time to think about it,:+1: yes, of course, will discuss  :smiley:

RaiVaibhav:
@gitmate-bot  rebase

RaiVaibhav:
@gitmate-bot  rebase

RaiVaibhav:
Rebased using gitmate because settings toolset commit is on top of this PR.

tests/bears/BearTest.py
```diff
@@ -463,6 +474,32 @@ def test_bear_with_specific_language(self):
...
+        pdb.Pdb.do_continue.assert_called_once_with(arg)
+        self.assertEqual(Debugger().do_quit(arg), 1)
+
+    @patch('coalib.bears.Bear.Debugger.runcall', side_effect=((1, 2), 1, 2))
+    def test_debug_run_with_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        self.assertEqual(debug_run(my_bear.run, Debugger, *args, **kwargs),
```
Makman2: Shouldn't it be `Debugger()` instead of just `Debugger`?

tests/bears/BearTest.py
```diff
@@ -463,6 +474,32 @@ def test_bear_with_specific_language(self):
...
+    def test_custom_continue(self, do_continue):
+        arg = {}
+        Debugger().do_quit(arg)
+        pdb.Pdb.do_continue.assert_called_once_with(arg)
+        self.assertEqual(Debugger().do_quit(arg), 1)
+
+    @patch('coalib.bears.Bear.Debugger.runcall', side_effect=((1, 2), 1, 2))
+    def test_debug_run_with_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
```
Makman2: do you actually need `TestOneBear`? Since you never invoke its `run` method, because it's stubbed away with `side_effect`. Maybe just use a plain `Bear` and write a comment that `side_effect` effectively "implements `run`".

tests/bears/BearTest.py
```diff
@@ -463,6 +474,32 @@ def test_bear_with_specific_language(self):
...
+                         [1, 2])
+
+    @patch('coalib.bears.Bear.Debugger.runcall', return_value=1)
+    def test_debug_run_with_no_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        self.assertEqual(debug_run(my_bear.run, Debugger, *args, **kwargs), 1)
+        self.assertIsNotNone(my_bear.run_bear_from_section(args, kwargs))
```
Makman2: why this double invocation?

tests/bears/BearTest.py
```diff
@@ -463,6 +474,32 @@ def test_bear_with_specific_language(self):
...
+        kwargs = {'d': 'e'}
+        self.assertEqual(debug_run(my_bear.run, Debugger, *args, **kwargs),
+                         [1, 2])
+
+    @patch('coalib.bears.Bear.Debugger.runcall', return_value=1)
+    def test_debug_run_with_no_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
```
Makman2: args necessary? Maybe just leave them empty ^^

Makman2: same above

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,68 @@
...
+    input = StringIO('\n'.join(debugger_commands))
+    output = StringIO()
+    dbg = Debugger(stdin=input, stdout=output)
+    return debug_run(func, dbg, *args, **kwargs), output.getvalue()
+
+
+class DebugBearsTest(unittest.TestCase):
+    def setUp(self):
+        # restore the coverage settrace to prevent the coverage breakage
+        # on project because  we can't chain coverage trace to run parallel
```
Makman2: since we are on it: double space between "because" and "we" ;)

tests/bears/BearTest.py
```diff
@@ -463,6 +474,32 @@ def test_bear_with_specific_language(self):
...
+    @patch('coalib.bears.Bear.Debugger.runcall', side_effect=((1, 2), 1, 2))
+    def test_debug_run_with_return(self, runcall):
+        section = Section('name')
+        my_bear = TestOneBear(section, self.queue, debugger=True)
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        self.assertEqual(debug_run(my_bear.run, Debugger, *args, **kwargs),
+                         [1, 2])
+
+    @patch('coalib.bears.Bear.Debugger.runcall', return_value=1)
```
Makman2: If this is testing a `run` function without return, the `return_value` for the mock should be `None` :) Otherwise it's technically identical to `test_debug_run_with_return`

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def func1(*args, **kwargs):
+    yield 1
+    yield 2
+    yield 3
+
+
+def func2(*args, **kwargs):
+    return([1, 2])
```
Makman2: just `return [1, 2]`, or `return 1, 2` (becomes a tuple then)

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(result, [1, 2, 3])
+        lines = output.splitlines()
+        self.assertEqual(lines[1], '-> yield 1')
+        self.assertEqual(lines[3], '-> yield 2')
+        self.assertEqual(lines[5], '-> yield 3')
+
+    def test_run_return_list_with_debugger(self):
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
+        result, output = execute_debugger('qc', func2, *args, **kwargs)
```
Makman2: Is a single `q` sufficient as a command?

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(result, [1, 2])
+        lines = output.splitlines()
+        self.assertEqual(lines[1], '-> return([1, 2])')
+
+    def test_run_return_generator_with_debugger(self):
+        args = ('a', ('b', 'c'))
+        kwargs = {'d': 'e'}
```
Makman2: Since you don't use args, I think you can remove them from the functions and from the tests :)

tests/bears/DebugBearsTest.py
```diff
@@ -0,0 +1,68 @@
...
+        result, output = execute_debugger('qcqcqc', func3, *args, **kwargs)
```
Makman2: shouldn't it be just 4 commands to make the debugger quit completely? Please check that up, it's good to use only as many commands as necessary, to catch possible bugs (debugger runs further actually and displays more lines of code, but we don't check additional lines)

Makman2: Same for the other tests above, please revise the number of commands necessary to let the debugger finish completely  :)

tests/processes/ProcessingTest.py
```diff
@@ -636,6 +603,46 @@ def check_prerequisites(cls):
...
+
+        class TestTwoBear(Bear):
+
+            BEAR_DEPS = {TestThreeBear}
+
+            def __init__(self, file_dict, section, queue, timeout=0.1,
+                         debugger=False):
+                raise RuntimeError
+
+        class TestOneBear(Bear):
```
Makman2: imo the `TestOneBear` should come first and not last :sweat_smile:

tests/processes/ProcessingTest.py
```diff
@@ -636,6 +603,46 @@ def check_prerequisites(cls):
...
+
+            BEAR_DEPS = {TestTwoBear}
+
+            def __init__(self, file_dict, section, queue, timeout=0.1,
+                         debugger=False):
+                Bear.__init__(self, section, queue, timeout, debugger)
```
Makman2: redundant `__init__`

tests/bears/BearTest.py
RaiVaibhav: previously you said it shouldn't have to be an empty :sweat_smile: so I make it like so that everyone gets it a tuple

tests/bears/DebugBearsTest.py
RaiVaibhav: actually, with generators, it will need one more so that in that it will throw `StopIterationError` and after that It will return a list (if I don't add extra command it will return something like `[1,2,3,None]`
without even catching the  `StopIterationError`

Makman2: Makes sense, nvm then :+1:

tests/processes/ProcessingTest.py
RaiVaibhav: Actually `__init__` is not redundant here as I am checking whether Bear gets initialized correctly without any error, If I remove the `__init__` then it will throw error

`TypeError: section must be an instance of one of (<class 'coalib.settings.Section.Section'>,) (provided value: {})`

tests/bears/DebugBearsTest.py
RaiVaibhav: With Mock test it args and kwargs may not  be useful but during testing of `Settings` toolset `args and kwargs` can be helpful in ideal test to IMO `args and kwargs` can be either empty when it is not getting used.

RaiVaibhav: `execute_debugger` function should be ideal for all the cases I think

Makman2: okay that's fine with me :+1:

tests/processes/ProcessingTest.py
Makman2: yeah, because you pass a file-dict where you should pass the section :) You just changed the signature, and forgot to use `file_dict` as well ;)

Makman2:
ack e138e29dbe40926e23f5d406e20e41fab72b746e

tests/bears/BearTest.py
```diff
@@ -478,27 +469,25 @@ def test_bear_with_specific_language(self):
...
         section = Section('name')
-        my_bear = TestOneBear(section, self.queue, debugger=True)
-        args = ('a', ('b', 'c'))
-        kwargs = {'d': 'e'}
-        self.assertEqual(debug_run(my_bear.run, Debugger, *args, **kwargs), 1)
-        self.assertIsNotNone(my_bear.run_bear_from_section(args, kwargs))
+        my_bear = Bear(section, self.queue, debugger=True)
+        args = ()
+        kwargs = {}
+        self.assertEqual(my_bear.run_bear_from_section(args, kwargs), None)
```
Makman2: --> `assertIsNone`

tests/bears/DebugBearsTest.py
```diff
@@ -24,13 +24,13 @@ def execute_debugger(debugger_commands, func, *args, **kwargs):
     input = StringIO('\n'.join(debugger_commands))
     output = StringIO()
     dbg = Debugger(stdin=input, stdout=output)
-    return debug_run(func, dbg, *args, **kwargs), output.getvalue()
```
Makman2: why removing `*args` here?

tests/bears/DebugBearsTest.py
```diff
@@ -50,17 +50,17 @@ def test_run_return_yield_with_debugger(self):
         self.assertEqual(lines[5], '-> yield 3')
 
     def test_run_return_list_with_debugger(self):
-        args = ('a', ('b', 'c'))
-        kwargs = {'d': 'e'}
-        result, output = execute_debugger('qc', func2, *args, **kwargs)
+        args = ()
+        kwargs = {}
```
Makman2: Imo it's okay if you completely omit those now. `execute_debugger` is designed to optionally accept keyword-params and positional params, everybody understands what it means if you don't pass any arguments at all ;)
```
execute_debugger('q', func2)
```

tests/bears/DebugBearsTest.py
RaiVaibhav: my fault :disappointed:

Makman2:
ack ec40d9750e31990f8eb6dc5be28eada7744809d6

Makman2:
@gitmate-bot rebase

Makman2:
@gitmate-bot ff

PR #5523, opened 2018-06-13
Add decorator add_value_checks

ishanSrt:
The decorator may work on any function but made for the
specific use of the bears' `run()` methods or the
`create_arguments()` methods. This decorator compares the value
given to the bear setting against a list of acceptable values
provided by the bear writer. The decorator accepts a list of
tuples with each tuple containing two values, the bear setting
name as a string and a list of acceptable values.

### DIFF
--- a/coalib/bearlib/type_annotations/Annotations.py
+++ b/coalib/bearlib/type_annotations/Annotations.py
@@ -1,0 +1,42 @@
+import inspect
+from functools import wraps
+
+from coalib.settings.FunctionMetadata import FunctionMetadata
+# The following ignore is added since the source code of some
+# bears use type_list as type annotations to arguments.
+# ignore PyUnusedCodeBear
+from coalib.settings.Setting import typed_list
+
+
+def add_value_checks(*settings):
+    """
+    Decorator for bears that checks a bear setting against a list
+    of acceptable values.
+
+    :param settings:
+        A list of tuples containing bear setting as a string, and
+        a list of acceptable values as the second value to the tuple.
+    """
+    def wrapping_function(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            args[0].ACCEPTABLE_VALUES_FOR_SETTINGS = settings[0]
+            if hasattr(args[0], 'CALL_TO_POPULATE') and (
+                    args[0].CALL_TO_POPULATE is True):
+                return
+            code = inspect.getsource(func)
+            hoax_code = code[code.find('def'):code.find('):')+3].strip()
+            hoax_code += '\n\treturn inspect.currentframe()\n'
+            hoax_code = hoax_code.replace(func.__name__, 'hoax_method', 1)
+            exec(hoax_code, globals(), globals())
+            frame = hoax_method(*args, **kwargs)
+            arguments, _, _, values = inspect.getargvalues(frame)
+            for setting_name, acceptable_values in settings[0]:
+                if values[setting_name] not in acceptable_values:
+                    raise ValueError('Invalid value "%s" given to the bear'
+                                     ' setting "%s"' % (values[setting_name],
+                                                        setting_name))
+        new_metadata = FunctionMetadata.from_function(func)
+        wrapper.__metadata__ = new_metadata
+        return wrapper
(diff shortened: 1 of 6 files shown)


coalib/bears/Bear.py
```diff
@@ -188,6 +188,8 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
     SEE_MORE = ''
     BEAR_DEPS = set()
     USE_RAW_FILES = False
+    ACCEPTABLE_VALUES_FOR_SETTINGS = []
```
jayvdb: this information should not be stored in the bear like this.  

it needs to be per setting, and using type annotations.  This was already discussed in the cEP.

that allows re-use of the type annotation.

ishanSrt:
A little bit of explanation: the Green Mode Project requires to differentiate between settings which can attain a specific set of values. A type of these setting which can have some discrete set of values is tackled in the PR. 

Most of the additions are self explanatory except for the variable `CALL_FROM_QUICKSTART`. The Green Mode project requires the valid values of these settings so that it can run coala over and over again by giving the bears all these values. Since we are also adding value checks in this PR, the field `ACCEPTABLE_VALUES_FOR_SETTINGS` won't get populated until the call to run method is made which contradicts the fact that we still don't have the acceptable values. The introduced variable uses the fact that even an incomplete call to the decorator populates the variable and thus saves execution time and running into errors.

Alternate approach could have been the bear writer populating a field else where and calling the decorator with those fields that quickstart can utilise directly but it just felt inelegant and contradicting the process of making bear development easier and easier by making changes at two different places to achieve a single task.

coalib/bears/Bear.py
ishanSrt: can't find an elegant way to use type annotations for value checks https://github.com/coala/coala-bears/issues/2532

ishanSrt: if what you mean by this:
>it needs to be per setting,

is
>Alternate approach could have been the bear writer populating a field else where and calling the decorator with those fields that quickstart can utilise directly but it just felt inelegant and contradicting the process of making bear development easier and easier by making changes at two different places to achieve a single task

Then I can make the necessary changes

ishanSrt: >this information should not be stored in the bear like this.

where exactly should it be stored then?

jayvdb: > where exactly do you want it to be stored?

inside the annotation!

An annotation can be a class, with any amount of logic and data inside it.

ishanSrt: but we can't use that unless coala is used with the `mypy` package loaded

ishanSrt: https://github.com/coala/coala/pull/5523#discussion_r195269799

jayvdb: have you seen my comment at https://github.com/coala/coala-bears/issues/2532 ?

ishanSrt: yes

jayvdb:
An interesting solution, but argument type annotations are the way that the Settings module works.  Not working with the Settings module will result in lots of other parts of coala being unable to understand these, and needing to be rewritten to use this functionality, when the capacity to do it already exists in the Settings module.

Example shown at https://github.com/coala/coala-bears/issues/2532 .  If it doesnt work with only minor implementation issues, I will implement it myself as a demo.
If I cant get it to work, we may revisit this decorator approach.

Note also that this decorator would not be usable in coala-bears repo until we do a release of coala, and that is a huge risk to take for any project, especially given that I uncovered that there are ~20 bears tests broken when using coala master (and @sangamcse has reproduced it, and was going to investigate further; https://gitter.im/coala/coala?at=5b274d604755974910317329), so that is another task someone needs to do before we can do a pre-release and use anything in coalib from coala-bears.

jayvdb:
It was demostrated Settings do work, even if a bit ugly without a bit of extra glue in StringConverter.

https://gitlab.com/coala/coala-utils/issues/73 is highly desirable and https://gitlab.com/coala/coala-utils/issues/77 also avoids many problems with the current `__path__` approach.

PR #5538, opened 2018-06-17
__init__.py: Add module docstring which confirm to PEP 257

ishanSrt:
Closes https://github.com/coala/coala/issues/5536#event-1684849637

>Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).
**The docstring for a module should generally list the classes, exceptions and functions (and any other objects) that are exported by the module, with a one-line summary of each. (These summaries generally give less detail than the summary line in the object's docstring.)**


ishanSrt:
ping @jayvdb I have tried to follow the PEP 257, but one line description means fuzzy explanations.

ishanSrt:
not added the functions as their docstrings have short and precise explanation there itself

coalib/settings/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+Has Setting, Section class; indexing tools; input checks; loads bears, files.
```
jayvdb: It must be prose. High level summary. See pep 257. It is hard .. ;)

coalib/settings/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+
+- **class Setting:** Aggregates the settings passed to coala.
```
jayvdb: No listing members. That is redundant to the TOC autogenerated from members, and it gets out of date quickly

ishanSrt: So remove the listing?

coalib/settings/__init__.py
ishanSrt: if it is *prose* then it is not detailed

jayvdb: The first line should not be detailed.  It is a summary.

After the blank line, you add as many lines of prose as you want, where it can be quite detailed.

coalib/settings/__init__.py
jayvdb: yes, replace with prose

coalib/settings/__init__.py
ishanSrt: ok I get you

coalib/settings/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+- **class Section:** Parses the Settings given in each section in config file.
+- **class FunctionMetadata:** Provides indexing tools from python docstrings.
```
jayvdb: this is not correct.  This has all the stuff needed to make bear arg annotation work

PR #5562, opened 2018-06-23
Add memoization to FileFactory

palash25:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/io/FileFactory.py
+++ b/coalib/io/FileFactory.py
@@ -1,6 +1,7 @@
 import os
 
 from coala_utils.decorators import generate_eq
+from coalib.io.memoized_property import memoized_property
 
 
 @generate_eq('name', 'timestamp')
@@ -70,6 +71,7 @@ class FileFactory:
         """
         self._filename = os.path.abspath(filename)
         self._timestamp = os.path.getmtime(self._filename)
+        self.cache = {}
 
     def get_line(self, line):
         """
@@ -82,7 +84,7 @@ class FileFactory:
         """
         return self.lines[line]
 
-    @property
+    @memoized_property
     def lines(self):
         """
         :return:
@@ -90,7 +92,7 @@ class FileFactory:
         """
         return tuple(self.string.splitlines())
 
-    @property
+    @memoized_property
     def raw(self):
         """
         :return:
@@ -99,13 +101,13 @@ class FileFactory:
         with open(self._filename, 'rb') as fp:
             return fp.read()
 
-    @property
+    @memoized_property
     def string(self):
         """
         :return:
             The file contents as a string UTF-8 decoded.
         """
-        return self.raw.decode()
+        return self.raw.decode(encoding='utf-8')
 
     @property
     def name(self):
--- a/coalib/io/memoized_property.py
+++ b/coalib/io/memoized_property.py
@@ -1,0 +1,21 @@
+import functools
+
+
+class memoized_property:
+    """
+    ``memoized_property`` serves as a decorator to cache
+    the results of the properties like ``raw``, ``string``
+    and ``lines`` provided by ``FileFactory``.
+    """
+
+    def __init__(self, method):
+        self._method = method
+        functools.update_wrapper(self, method)
+
+    def __get__(self, instance, owner):
+        try:
+            result = instance.cache[self._method]
+        except KeyError:
+            result = self._method(instance)
+            instance.cache[self._method] = result
+        return result
--- a/tests/io/FileFactoryTest.py
+++ b/tests/io/FileFactoryTest.py
@@ -37,11 +37,14 @@ class FileFactoryTest(unittest.TestCase):
 
     def test_lines(self):
         self.assertEqual(self.uut.lines, ('This is a test file.',))
+        self.assertEqual(self.uut.lines, ('This is a test file.',))
 
     def test_raw(self):
         self.assertEqual(self.uut.raw, b'This is a test file.')
+        self.assertEqual(self.uut.raw, b'This is a test file.')
 
     def test_string(self):
+        self.assertEqual(self.uut.string, 'This is a test file.')
         self.assertEqual(self.uut.string, 'This is a test file.')
 
     def test_timestamp(self):


coalib/io/FileCache.py
```diff
@@ -0,0 +1,137 @@
...
+                    # for this module.
+                    return None
+                cache[filename] = (
+                    len(data), None,
+                    data, fullname
+                )
+                return cache[filename][2]
+
+    try:
+        with open(fullname, 'rb') as fp:
```
palash25: This seemed to have done the trick. Now instead of a list the file contents are stored as bytes.

coalib/io/FileFactory.py
```diff
@@ -96,16 +98,15 @@ def raw(self):
...
+        return get_content(self._filename)
 
     @property
     def string(self):
         """
         :return:
             The file contents as a string UTF-8 decoded.
         """
-        return self.raw.decode()
+        return self.raw.decode(encoding=detect_encoding(self._filename))
```
Makman2: not UTF-8 for all cases? ;)

coalib/io/FileCache.py
```diff
@@ -0,0 +1,137 @@
...
+import sys
+import os
+import tokenize
+
+__all__ = ["get_content", "clearcache", "checkcache"]
+
+
+# The cache. Maps filenames to either a thunk which will provide source code,
+# or a tuple (size, mtime, contents, fullname) once loaded.
+cache = {}
```
Makman2: The cache should be tied to a `FileFactory` object, and not stay in global scope. That makes the cache cleaning up easier because it automatically cleans up as soon as the file factories are destructed. If someone wants to keep the cache alive, the relvant  file factory objects should be kept alive.

palash25: So do you want a `self.cache` in `FileFactory` to tie it to a Factory object?

coalib/io/FileCache.py
```diff
@@ -0,0 +1,41 @@
+import os
+
+__all__ = ['get_content', 'clearcache']
+
+
+# The cache. Maps filenames to either a thunk which will provide source code,
```
palash25: I forgot to edit this comment will do it in the next push

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+    cache[filename] = size, mtime, lines, filename
+    return lines
+
+
+class Memoize:
+    """
+    ``Memoize`` serves as a decorator for the properties
+    of ``FileFactory`` and caches the results of the property
+    for a particular ``FileFactory`` object.
+    """
```
Makman2: We should move that asap to coala-utils ;) Maybe even extend it to something like `cached_property`:
```python
@cached_property
def myprop(self):
    ...
```

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, method):
+        self._method = method
```
Makman2: Can we add here
```
functools.update_wrapper(self, func)
```
then the signatures look a bit nicer ;)

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+    stat = os.stat(filename)
+
+    with open(filename, 'rb') as fp:
+        lines = fp.read()
+    size, mtime = stat.st_size, stat.st_mtime
+    cache[filename] = size, mtime, lines, filename
+    return lines
+
+
+class Memoize:
```
Makman2: Did you copy parts of this class from somewhere? In case you did, please include a link :+1:

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+    If something's wrong, print a message, discard the cache entry,
+    and return an empty list.
+    """
+    stat = os.stat(filename)
+
+    with open(filename, 'rb') as fp:
+        lines = fp.read()
+    size, mtime = stat.st_size, stat.st_mtime
+    cache[filename] = size, mtime, lines, filename
+    return lines
```
Makman2: do we still need all that?

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+    of ``FileFactory`` and caches the results of the property
+    for a particular ``FileFactory`` object.
+    """
+
+    def __init__(self, method):
+        self._method = method
+
+    def __call__(self, instance):
+        self._instance = instance
+        return self.__get__(self._instance, self._instance.__class__)
```
Makman2: use `type(self._instance)`

coalib/io/FileCache.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __get__(self, instance, owner):
```
Makman2: actually, do you need to override `__get__`? Can you explain me how it works exactly? I believe `__get__` will be invoked when you do `self.memoized_func.attribute_you_want_to_access` (and that will only happen if it's not a property, as a property will always *call* the Memoize class).

coalib/io/FileCache.py
palash25: Yup. I will include the link.

coalib/io/FileCache.py
palash25: No we don't I forgot to update this function :sweat_smile:

coalib/io/FileCache.py
palash25: The results of the properties are stored in `self._cache` which is tied to the factory object. So I override `__get__` to access that attribute to store the results in the cache. And then `self.__get__` is returned on every `__call__`

palash25:
@Makman2 please take a look at this comment. https://github.com/coala/coala/pull/5562/#discussion_r200094124

coalib/io/FileCache.py
Makman2: But isn't it possible to call `__get__` differently? Like just `get_cache`? Why do we have to override the magic method `__get__`?

coalib/io/FileCache.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    with open(filename, 'rb') as fp:
+        contents = fp.read()
+    cache[filename] = contents
+    return contents
+
+
+class Memoize:
+    """
+    This is an edited version of the ``cached_property`` decorator given here:
```
Makman2: wait this is already a cached property? Why do we have to add `@Memoize` together with `@property`?

coalib/io/FileCache.py
```diff
@@ -0,0 +1,63 @@
...
+def update_cache(filename):
+    """
+    Update a cache entry and return its list of lines.
+    If something's wrong, print a message, discard the cache entry,
+    and return an empty list.
+    """
+    with open(filename, 'rb') as fp:
+        contents = fp.read()
+    cache[filename] = contents
+    return contents
```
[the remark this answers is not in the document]
palash25: Yes of course. Thats the main caching algorithm which saves the contents in raw format.

coalib/io/Memoize.py
```diff
@@ -0,0 +1,31 @@
...
+        except KeyError:
+            result = self._method(instance)
+            instance.cache[self._method] = result
+        return result
+
+    def get_content(file_factory, filename):
+        with open(filename, 'rb') as fp:
+            contents = fp.read()
+        file_factory.cache[function] = contents
+        return file_factory.cache[function]
```
Makman2: what is that? :O sounds like it should be in `FileFactory`, not in `Memoize` :3

coalib/io/Memoize.py
```diff
@@ -0,0 +1,31 @@
+import functools
+
+
+class Memoize:
```
Makman2: Since it's now also turning your functions into properties, you should call it somehow `*_property`, e.g. `cached_property`, `memoized_property`, ... Exact name is up to :P

coalib/io/FileFactory.py
```diff
@@ -82,30 +84,29 @@ def get_line(self, line):
...
-    @property
+    @Memoize
     def raw(self):
         """
         :return:
             The file contents as a byte sequence.
         """
-        with open(self._filename, 'rb') as fp:
-            return fp.read()
+        return Memoize.get_content(self, self._filename)
```
Makman2: Ah... don't implement `get_content` in `Memoize`, but here, and just memoize `raw` directly:
```
@Memoize
def raw(self):
    with open(...).....
```

coalib/io/Memoize.py
Makman2: Actually, have you seen this? --> https://pypi.org/project/cached-property/
:sweat_smile: 

In the end you learned something though :sweat_smile:

palash25:
> Actually, have you seen this? --> https://pypi.org/project/cached-property/

I have seen something similar but not this package :laughing: . But if I had just copy pasted this I wouldn't get to learn so much about the dunder functions, so I'm glad I missed this.

Makman2:
So do we want to use this package?

coalib/io/FileFactory.py
```diff
@@ -70,6 +71,7 @@ def __init__(self, filename):
         """
         self._filename = os.path.abspath(filename)
         self._timestamp = os.path.getmtime(self._filename)
+        self.cache = {}
```
Makman2: is that necessary? I would expect from `memoized_property` to create the cache field itself^^

coalib/io/memoized_property.py
```diff
@@ -0,0 +1,21 @@
...
+        self._method = method
+        functools.update_wrapper(self, method)
+
+    def __get__(self, instance, owner):
+        try:
+            result = instance.cache[self._method]
+        except KeyError:
+            result = self._method(instance)
+            instance.cache[self._method] = result
+        return result
```
Makman2: we should add a test for it ;) Or we use the existing package, then we save the tests :sweat_smile:

Makman2:
ack fc568a01ba489376b4d6efac12ea772cb3cb1573

Makman2:
@gitmate-bot ff

Issue #5533, opened 2018-06-16
Improve AssertionError message of `assert_result_equal(..)`


sangamcse:
After changing to `pytest`, the AssertionError message is changed to custom message see --> https://github.com/coala/coala/blob/master/coalib/testing/BaseTestHelper.py#L22 .

Previous AssertionError message is better --> https://github.com/coala/coala/blob/master/tests/testing/LocalBearTestHelperTest.py#L81

https://github.com/coala/coala/pull/5505#discussion_r195896051

sangamcse:
cc @jayvdb

jayvdb:
Also the equality comparison should be done in the relevant class.
https://gitlab.com/coala/coala-utils/issues/60

PR #5564, opened 2018-06-24
BaseTestHelper.py: Improve AssertionError message

sangamcse:
This improves the `AssertionError` message generated from
`assert_result_equal(...)` method by inheriting unittest
in `BaseTestHelper` class.

Closes https://github.com/coala/coala/issues/5533

### DIFF
--- a/coalib/testing/BaseTestHelper.py
+++ b/coalib/testing/BaseTestHelper.py
@@ -1,44 +1,42 @@
+import unittest
+
 from coala_utils.Comparable import Comparable
 
 
-def _assert_comparable_equal(observed_result, expected_result):
-    """
-    Compares individual fields of the result objects using
-    `__compare_fields__` of `coala_utils.Comparable` class
-    and yields messages in case of an attribute mismatch.
-    """
-
-    if not len(observed_result) == len(expected_result):
-        assert observed_result == expected_result, '%s != %s' % (
-            observed_result, expected_result)
-
-    messages = ''
-    for observed, expected in zip(observed_result, expected_result):
-        if (isinstance(observed, Comparable) and
-            isinstance(expected, Comparable)) and (type(observed) is
-                                                   type(expected)):
-            for attribute in type(observed).__compare_fields__:
-                try:
-                    assert getattr(observed, attribute) == getattr(
-                        expected, attribute), (
-                        '{} mismatch: {}, {} != {}, {}'.format(
-                            attribute,
-                            observed.origin, observed.message,
-                            expected.origin, expected.message))
-                except AssertionError as ex:
-                    messages += (str(ex) + '\n\n')
-        else:
-            assert observed_result == expected_result, '%s != %s' % (
-                observed_result, expected_result)
-
-    if messages:
-        raise AssertionError(messages)
-
-
-class BaseTestHelper:
+class BaseTestHelper(unittest.TestCase):
     """
     This is a base class for all Bears' tests of coala's testing API.
     """
+
+    def _assert_comparable_equal(self,
+                                 observed_result,
+                                 expected_result):
+        """
+        Compares individual fields of the result objects using
+        `__compare_fields__` of `coala_utils.Comparable` class
+        and raises messages in case of an attribute mismatch.
+        """
+
+        if not len(observed_result) == len(expected_result):
+            self.assertEqual(observed_result, expected_result)
+
+        messages = []
+        for observed, expected in zip(observed_result, expected_result):
+            if (isinstance(observed, Comparable) and
+                isinstance(expected, Comparable)) and (
+                    Comparable.__eq__(expected, observed)):
+                for attribute in observed.__compare_fields__:
+                    try:
+                        self.assertEqual(getattr(observed, attribute),
+                                         getattr(expected, attribute),
+                                         msg='{} mismatch.'.format(attribute))
+                    except AssertionError as ex:
+                        messages.append(str(ex))
+            else:
+                self.assertEqual(observed, expected)
+
+        if messages:
+            raise AssertionError('\n\n'.join(message for message in messages))
 
     def assert_result_equal(self,
                             observed_result,
@@ -51,4 +49,4 @@ class BaseTestHelper:
         :param expected_result: The expected result from a bear
         """
 
-        return _assert_comparable_equal(observed_result, expected_result)
+        return self._assert_comparable_equal(observed_result, expected_result)
--- a/tests/testing/BaseTestHelperTest.py
+++ b/tests/testing/BaseTestHelperTest.py
@@ -1,5 +1,3 @@
-import pytest
-
 from coalib.results.Result import Result
 from coalib.testing.BaseTestHelper import BaseTestHelper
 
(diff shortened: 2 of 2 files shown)


tests/testing/BaseTestHelperTest.py
```diff
@@ -9,23 +7,38 @@ def test_good_assert_result_equal(self):
...
+                         'None\n\n'
+                         "- ['a', 'b']\n"
+                         "+ ['a', 'b', None]\n"
+                         '?          ++++++\n', str(ex.exception))
 
     def test_not_comparable_assert_result_equal(self):
-        with pytest.raises(AssertionError) as ex:
+        with self.assertRaises(AssertionError):
             self.assert_result_equal(['a', 'b'], ['a', 'c'])
-        assert '[\'a\', \'b\'] != [\'a\', \'c\']' in str(ex.value)
```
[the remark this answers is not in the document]
sangamcse: It was differing in different python version. I tried [here](https://github.com/sangamcse/coala/commit/89b810b0ca4488797256a58fd1067e5f92fcffa3#diff-ef7e7d5c850300aa98477496e3c4e364R23). Still, I am trying it once again.
> why was this removed? the assertion might change slightly, but we need to check the assertion message still.

I will do that. 😄 
> You can also add Related to `https://github.com/coala/coala/issues/3808` in your commit message as that is what this PR is also doing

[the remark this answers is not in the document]
sangamcse: It is done now @jayvdb. Take a look. Thanks.

sangamcse:
Ping @jayvdb, need your review.

coalib/testing/BaseTestHelper.py
```diff
@@ -1,45 +1,43 @@
...
     This is a base class for all Bears' tests of coala's testing API.
     """
 
+    def _assert_comparable_equal(self,
+                                 observed_result,
+                                 expected_result):
+        """
+        Compares individual fields of the result objects using
+        `__compare_fields__` of `coala_utils.Comparable` class
+        and yields messages in case of an attribute mismatch.
```
jayvdb: yields?  -> raises ...

coalib/testing/BaseTestHelper.py
```diff
@@ -1,45 +1,43 @@
...
+        """
+
+        if not len(observed_result) == len(expected_result):
+            self.assertEqual(observed_result, expected_result)
+
+        messages = ''
```
jayvdb: put them into an array, and then format them at the end when raising the exception

coalib/testing/BaseTestHelper.py
sangamcse: Changes are done @jayvdb

sangamcse:
@jayvdb, I think all the changes are done. Have a look, please. 😄

coalib/testing/BaseTestHelper.py
```diff
@@ -1,45 +1,43 @@
...
+                        self.assertEqual(getattr(observed, attribute),
+                                         getattr(expected, attribute),
+                                         msg='{} mismatch.'.format(attribute))
+                    except AssertionError as ex:
+                        messages.append(str(ex))
+            else:
+                self.assertEqual(observed, expected)
+
+        if messages:
+            raise AssertionError('\n\n'.join(message for message in messages))
```
jayvdb: When running `test_jinja2_templating` (due to https://github.com/coala/coala-bears/issues/2678) which isnt the best of test approaches, but far from the worst, here are my error results

```pytest
>               raise AssertionError(messages)
E               AssertionError: 'Usin[68 chars]o XSS. Ensure autoescape=True to mitigate XSS vulnerabilities.' != 'Usin[68 chars]o XSS. Use autoescape=True to mitigate XSS vulnerabilities.'
E               - Using jinja2 templates with autoescape=False is dangerous and can lead to XSS. Ensure autoescape=True to mitigate XSS vulnerabilities.
E               ?                                                                                ^^ --
E               + Using jinja2 templates with autoescape=False is dangerous and can lead to XSS. Use autoescape=True to mitigate XSS vulnerabilities.
E               ?                                                                                ^
E                : message_base mismatch.

../coala/coalib/testing/LocalBearTestHelper.py:136: AssertionError
```

Now how can I work out which of the five `Result` that was about?

This entire test system is fundamentally flawed.

test helpers should help diagose, not help write tests which are unable to be diagnosed.

If I change it to `assert sorted(bear_output) == sorted(results)`, I see *nothing* because these `Comparable` objects are impossible for pytest to automatically inspect dynamically.  And `self.assertCountEqual` has also been busted: the output is a mess.  We are breaking the standard Python testing tools, and replacing them with something less helpful.

So at the moment, with pytest and unittest builtin systems not working on `Result` objects, we are reliant on this `Comparable` class and this helper to actually help decode the problem.  It doesnt do that before this patch, and this patch doesnt make it much better.

coalib/testing/BaseTestHelper.py
```diff
@@ -1,45 +1,43 @@
...
+            if (isinstance(observed, Comparable) and
+                isinstance(expected, Comparable)) and (
+                    Comparable.__eq__(expected, observed)):
+                for attribute in observed.__compare_fields__:
+                    try:
+                        self.assertEqual(getattr(observed, attribute),
+                                         getattr(expected, attribute),
+                                         msg='{} mismatch.'.format(attribute))
+                    except AssertionError as ex:
+                        messages.append(str(ex))
```
jayvdb: this makes the test suite reliant on pytest output, which differs whenever pytest finds a way to improve their messages.

tests/testing/BaseTestHelperTest.py
```diff
@@ -9,23 +7,41 @@ def test_good_assert_result_equal(self):
...
         observed = [Result.from_values(origin='AnyBear',
-                                       message='This file has 2 lines.',
+                                       message='This file has 3 lines.',
                                        file='anyfile')]
-        with pytest.raises(AssertionError) as ex:
+        with self.assertRaises(AssertionError) as ex:
             self.assert_result_equal(expected, observed)
-        assert ('origin mismatch: AnyBea, This file has 2 lines. != AnyBear, '
-                'This file has 2 lines.\n\n') == str(ex.value)
+        self.assertEqual("'AnyBea' != 'AnyBear'\n"
```
jayvdb: use `textwrap.dedent` to make this more readable.

tests/testing/BaseTestHelperTest.py
```diff
@@ -9,23 +7,41 @@ def test_good_assert_result_equal(self):
         self.assert_result_equal(['a', 'b'], ['a', 'b'])
 
     def test_inequality_assert_result_equal(self):
-        with pytest.raises(AssertionError) as ex:
+        with self.assertRaises(AssertionError) as ex:
             self.assert_result_equal(['a', 'b'], ['a', 'b', None])
-        assert '[\'a\', \'b\'] != [\'a\', \'b\', None]' in str(ex.value)
+        self.assertEqual("Lists differ: ['a', 'b'] != ['a', 'b', None]\n\n"
```
jayvdb: these tests now rely on the output of pytest - see note above about capturing the assertion and converting it into a string.

That isnt helpful.

And these test cases are not helpful.

What we need are tests of multiple complicated `Result` objects, and showing that the error message is helpful for real scenarios.

And our messages should not use pytest's messages, because they are dumb -- they are automatic and can not know the context of the situation, or know what the bear debugger needs to see in order to be able to diagnose the problem.

Issue #565, opened 2015-05-21
add profiling options


sils:
we should make it bear writers easy to profile their bears. To do that we could use `cProfile.runctx("self.run(*args, **kwargs)", globals(), locals())` for bear execution and then show customized profiling results to the user.

I think this would be a cool thing. I'm playing around with this right now so maybe I get a half baked solution.

sils1297:
so essentially we can do this like this hacky thing: http://pastebin.com/ucEPS0dA

However this would mean we'd have to execute it several times because we can't reuse the result. So we'd need to find a way to get the result and then add custom printing plus only do this conditionally. (I.e. check if `profiling` setting is True :))

sils1297:
so to close this bug:

 * [ ] Add profile option to CLI
 * [ ] Add profiling setting
 * [ ] Give profiler results to the interactor/back via a callback (we're going to purge interactors)
 * [ ] Add profiling information to docs

sils1297:
according to http://stackoverflow.com/questions/1584425/return-value-while-using-cprofile we can just do

```
import cProfile, pstats, sys
pr = cProfile.Profile()
pr.enable()
...
my_return_val = my_func(my_arg)
...
pr.disable()
ps = pstats.Stats(pr, stream=sys.stdout)
ps.print_stats()
```

sils1297:
damn that actually works, here's prototype no 2: http://pastebin.com/3nF5RGxQ

Makman2:
Just for interest: What kind of information the profiles supplies?

sils1297:
Statistics mainly, how much time is spent in which function and how often
they are called.

2015-05-22 12:18 GMT+02:00 Makman2 <notifications@github.com>:

AbdealiJK:
Where would we display this ? 
So, Im assuming we have a `profile=True` option in the .coafile
Now, when the bears in this section are run you can profile them.

Do we show all the profiles for all the bears at the end of coala's execution ?
As soon as the bear completes running ?

I think the best option is to save the profilers output into a file and mention that in the terminal.
Its really difficult to read that data on a terminal IMO and this makes the data handling easier as we can just dump into a file as soon as the bear.run finishes

sils1297:
A file seems sane, rather than a `profile=True` setting we could have a `profile=/path/to/dir` and create a file for each bear there with the data

sils1297:
@Makman2 sayd possibly "prometheus" can help

Makman2:
Reassigning due to inactivity.

PR #5599, opened 2018-07-05
Add: Profiling Interface

RaiVaibhav:
Add Profiler Interface.

Related to https://github.com/coala/coala/issues/565

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -1,10 +1,12 @@
 import inspect
+import cProfile
 import pdb
 import collections
 import traceback
+import logging
 from functools import partial
-from os import makedirs
-from os.path import join, abspath, exists
+from os import makedirs, getcwd
+from os.path import join, abspath, exists, isdir
 import requests
 from appdirs import user_data_dir
 
@@ -19,9 +21,34 @@ from coalib.results.Result import Result
 from coalib.results.TextPosition import ZeroOffsetError
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
 from coalib.settings.ConfigurationGathering import get_config_directory
 
 from .meta import bearclass
+
+
+def _setting_is_enabled(bear, key):
+    """
+    Check setting key is in section.
+
+    :param bear: Bear object.
+    :param key:  Setting key.
+    :return:     ``True`` if setting value is ``True``. Setting object if
+                 setting key is in section else ``False``.
+    """
+    if not isinstance(bear, Bear):
+        raise ValueError(
+            'Positional argument bear is not an instance of Bear class.')
+    if key is None:
+        raise ValueError('No setting key passed.')
+
+    if key not in bear.section:
+        return False
+    try:
+        return bool(bear.section[key])
+    except ValueError:
+        pass
+    return bear.section[key]
 
 
 def _is_debugged(bear):
@@ -32,17 +59,27 @@ def _is_debugged(bear):
     :return:     True if ``debug_bears`` is ``True`` or if bear name specified
                  in ``debug_bears`` setting match with the bear parameter.
     """
-    if not isinstance(bear, Bear):
-        raise ValueError('Positional argument bear is not an instance of '
-                         'Bear class.')
-    if 'debug_bears' not in bear.section:
-        return False
-    try:
-        return bool(bear.section['debug_bears'])
-    except ValueError:
-        pass
-    return bear.name.lower() in (
-            map(str.lower, bear.section['debug_bears']))
+    setting = _setting_is_enabled(bear, key='debug_bears')
+    if isinstance(setting, bool):
+        return setting
+    return bear.name.lower() in map(str.lower, setting)
+
+
+def _is_profiled(bear):
+    """
+    Check whether the bear is in profile mode according to its section-settings.
+
+    :param bear: Bear object.
+    :return:     current working directory if ``profile`` is ``True``, False
+                 if ``profile`` is ``False`` else return directory path
+                 specified in ``profile``.
+    """
+    setting = _setting_is_enabled(bear, key='profile')
+    if setting is True:
+        return getcwd()
+    if isinstance(setting, Setting):
+        return setting.value
+    return False
 
 
 class Debugger(pdb.Pdb):
@@ -322,6 +359,11 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
         self.message_queue = message_queue
         self.timeout = timeout
         self.debugger = _is_debugged(bear=self)
+        self.profile = _is_profiled(bear=self)
+
+        if self.profile and self.debugger:
+            raise ValueError(
+                'Cannot run debugger and profiler at the same time.')
 
         self.setup_dependencies()
         cp = type(self).check_prerequisites()
@@ -344,6 +386,35 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def _dump_bear_profile_data(self, profiler):
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        path = join(self.profile, filename)
+        if not isdir(self.profile):
+            try:
+                makedirs(self.profile)
+            except FileExistsError:
+                logging.error('File exists :'.format(self.profile))
+                raise SystemExit(2)
+
+        profiler.dump_stats(path)
+
+    def profile_run(self, *args, profiler=None, **kwargs):
+        profiler = cProfile.Profile() if profiler is None else profiler
+        bear_results = profiler.runcall(self.run, *args, **kwargs)
+        if isinstance(bear_results, collections.Iterable):
+            results = []
+            iterator = iter(bear_results)
+            while True:
+                try:
+                    result = profiler.runcall(next, iterator)
+                    results.append(result)
+                except StopIteration:
+                    break
+        else:
+            results = bear_results
+        self._dump_bear_profile_data(profiler)
+        return results
+
     def run_bear_from_section(self, args, kwargs):
         try:
             # Don't get `language` setting from `section.contents`
@@ -359,6 +430,8 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
             return
         if self.debugger:
             return debug_run(self.run, Debugger(bear=self), *args, **kwargs)
+        elif self.profile:
+            return self.profile_run(*args, **kwargs)
         else:
             return self.run(*args, **kwargs)
 
--- a/coalib/parsing/DefaultArgParser.py
+++ b/coalib/parsing/DefaultArgParser.py
@@ -243,13 +243,25 @@ To run coala without user interaction, run the `coala --non-interactive`,
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
-    outputs_group.add_argument(
+    devtool_exclusive_group = arg_parser.add_mutually_exclusive_group()
+
+    devtool_exclusive_group.add_argument(
         '--debug-bears', nargs='?', const=True,
         help='Enable bear debugging with pdb, that can help to identify and'
         ' correct errors in bear code. Steps into bear code as soon as being'
         ' executed. To specify which bears to debug, supply bear names as'
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
+
+    devtool_exclusive_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile. To specify where to dump the'
+        ' profiled files, supply the directory path. If specified directory'
+        ' does not exist it will be created. If the specified path points to an'
+        ' already existing file a error is raised. All bears (even'
+        ' implicit dependency bears) in a section will be profiled. Profiled'
+        ' data files will have a name format'
+        ' ``{section.name}_{bear.name}.prof``.')
 
     misc_group = arg_parser.add_argument_group('Miscellaneous')
 
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,10 +1,12 @@
 import pdb
+import os
 from collections import defaultdict
 import datetime
 from io import BytesIO, StringIO
 import multiprocessing
 import unittest
 from os.path import abspath, exists, isfile, join, getmtime
+from tempfile import TemporaryDirectory, NamedTemporaryFile
 import shutil
 
 from freezegun import freeze_time
@@ -13,19 +15,26 @@ from unittest.mock import patch
 import requests
 import requests_mock
 
+from coalib.coala_main import run_coala
 from coalib.bearlib.aspects.collections import AspectList
 from coalib.bearlib.aspects.Metadata import CommitMessage
 from coalib.bearlib.languages.Language import Language, Languages
-from coalib.bears.Bear import Bear, Debugger, _is_debugged
+from coalib.bears.Bear import (
+    Bear, Debugger, _setting_is_enabled, _is_debugged, _is_profiled)
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.bears.GlobalBear import GlobalBear
 from coalib.bears.LocalBear import LocalBear
 from coalib.results.Result import Result
 from coalib.results.TextPosition import ZeroOffsetError
 from coalib.output.printers.LOG_LEVEL import LOG_LEVEL
+from coalib.output.printers.LogPrinter import LogPrinter
 from coalib.processes.communication.LogMessage import LogMessage
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting, language
+from pyprint.ConsolePrinter import ConsolePrinter
+from coala_utils.ContextManagers import prepare_file
+
+from tests.TestUtilities import bear_test_module
 
 
 class BadTestBear(Bear):
@@ -65,6 +74,20 @@ class TestOneBear(LocalBear):
     def run(self, x: int, y: str, z: int = 79, w: str = 'kbc'):
         yield 1
         yield 2
+
+
+class TestTwoBear(Bear):
+
+    def run(self, *args, **kwargs):
+        yield 1
+        yield 2
+        yield 3
+
+
+class TestThreeBear(Bear):
+
+    def run(self, *args, **kwargs):
+        pass
 
 
 class TypedTestBear(Bear):
@@ -536,6 +559,130 @@ class BearTest(BearTestBase):
         self.assertEqual(_is_debugged(uut), True)
         section.append(Setting('debug_bears', 'abc, xyz'))
         self.assertEqual(_is_debugged(uut), False)
+
+    @patch('cProfile.Profile.dump_stats')
+    def test_profiler_with_no_directory_exists(self, dump_stats):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        section.append(Setting('profile', 'tRuE'))
+        bear = TestTwoBear(section, self.queue)
+        self.assertEqual(bear.run_bear_from_section(args, kwargs), [1, 2, 3])
+        dump_stats.assert_called_once_with(join(os.getcwd(),
+                                                'name_TestTwoBear.prof'))
+        section.append(Setting('profile', 'abc'))
+        bear = TestTwoBear(section, self.queue)
+        self.assertEqual(bear.run_bear_from_section(args, kwargs), [1, 2, 3])
+        dump_stats.assert_called_with(os.path.join(
+            bear.profile, 'name_TestTwoBear.prof'))
+        os.rmdir('abc')
+        section.append(Setting('profile', '1'))
+        bear = TestThreeBear(section, self.queue)
+        self.assertIsNone(bear.run_bear_from_section(args, kwargs))
+        dump_stats.assert_called_with(join(os.getcwd(),
+                                           'name_TestThreeBear.prof'))
+
+    @patch('cProfile.Profile.dump_stats')
+    def test_profiler_with_directory_exists(self, dump_stats):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        with TemporaryDirectory() as temp_dir:
+            section.append(Setting('profile', temp_dir))
+            bear = TestTwoBear(section, self.queue)
+            self.assertEqual(bear.run_bear_from_section(args, kwargs),
+                             [1, 2, 3])
+            dump_stats.assert_called_once_with(os.path.join(
+                bear.profile, 'name_TestTwoBear.prof'))
+
+    def test_profiler_with_file_path(self):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        with NamedTemporaryFile() as temp_file:
+            section.append(Setting('profile', temp_file.name))
+            bear = TestTwoBear(section, self.queue)
+            with self.assertRaises(SystemExit) as context:
+                bear.run_bear_from_section(args, kwargs)
+                self.assertEqual(context.exception.code, 2)
+
+    def test_profiler_with_debugger(self):
+        section = Section('name')
+        section.append(Setting('debug_bears', 'tRuE'))
+        section.append(Setting('profile', 'tRuE'))
+        with self.assertRaisesRegex(
+                ValueError,
+                'Cannot run debugger and profiler at the same time.'):
+            Bear(section, self.queue)
+
+    @patch('coalib.bears.Bear.Bear.profile_run')
+    def test_profiler_with_false_setting(self, profile_run):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        section.append(Setting('profile', '0'))
+        bear = TestThreeBear(section, self.queue)
+        self.assertIsNone(bear.run_bear_from_section(args, kwargs))
+        assert not profile_run.called
+
+    def test_is_profiled(self):
+        with self.assertRaisesRegex(
+                ValueError,
+                'Positional argument bear is not an instance of Bear class.'):
+            _is_profiled(bear=None)
+
+        section = Section('name')
+        uut = Bear(section, self.queue)
+        self.assertEqual(_is_profiled(uut), False)
+        section.append(Setting('profile', 'tRuE'))
+        self.assertEqual(_is_profiled(uut), os.getcwd())
+        section.append(Setting('profile', '0'))
+        self.assertEqual(_is_profiled(uut), False)
+        section.append(Setting('profile', 'dirpath'))
+        self.assertEqual(_is_profiled(uut), 'dirpath')
+
+    def test_setting_is_enabled(self):
+        with self.assertRaisesRegex(
+                ValueError,
+                'Positional argument bear is not an instance of Bear class.'):
+            _setting_is_enabled(bear=None, key='key')
+
+        section = Section('name')
+        uut = Bear(section, self.queue)
+        with self.assertRaisesRegex(ValueError, 'No setting key passed.'):
+            _setting_is_enabled(bear=uut, key=None)
+
+        self.assertFalse(_setting_is_enabled(bear=uut, key='key'))
+
+        section.append(Setting('key', 'value'))
+        self.assertEqual(_setting_is_enabled(bear=uut, key='key'),
+                         uut.section['key'])
+        section.append(Setting('key', 'tRuE'))
+        self.assertEqual(_setting_is_enabled(bear=uut, key='key'), True)
+        section.append(Setting('key', '0'))
+        self.assertEqual(_setting_is_enabled(bear=uut, key='key'), False)
+
+    def test_profiler_dependency(self, debug=False):
+        with bear_test_module(), \
+                prepare_file(['#fixme  '], None) as (lines, filename):
+            results = run_coala(console_printer=ConsolePrinter(),
+                                log_printer=LogPrinter(),
+                                arg_list=(
+                                    '-c', os.devnull,
+                                    '-f', filename,
+                                    '-b', 'DependentBear',
+                                    '-S', 'use_spaces=yeah',
+                                    '--profile', 'profiled_bears',
+                                ),
+                                autoapply=False,
+                                debug=debug)
+            cli_result = results[0]['cli']
+            self.assertEqual(len(cli_result), 1)
+
+        profiled_files = os.listdir('profiled_bears')
+        self.assertEqual(len(profiled_files), 1)
+        self.assertEqual(profiled_files[0], 'cli_SpaceConsistencyTestBear.prof')
+        shutil.rmtree('profiled_bears')
 
 
 class BrokenReadHTTPResponse(BytesIO):


coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def setup_profile(self, prof=None):
+        path = os.getcwd()
+        if self.profile == 'True':
```
Makman2: check against boolean values, not string:
```
if self.profile is True:
```

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+            filename = '{}_{}.prof'.format(self.section.name, self.name)
+            open(filename, 'wb')
```
Makman2: superfluous

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+                prof.dump_stats(os.path.join(filepath, filename))
+
+            else:
+                os.mkdir(os.path.join(path, subdir))
+                filename = '{}_{}.prof'.format(self.section.name, self.name)
+                open(os.path.join(filepath, filename), 'wb')
+                prof.dump_stats(os.path.join(filepath, filename))
+
+    def profile_run(self, *args, **kwargs):
+        prof1 = cProfile.Profile()
```
Makman2: please name them properly ;) proper naming automatically makes code documented :)

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        bear_results = prof1.runcall(self.run, *args, **kwargs)
+        if isinstance(bear_results, types.GeneratorType):
```
Makman2: not `Iterable` instead of `GeneratorType`?

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+            filepath = os.path.join(path, subdir)
+            if os.path.isdir(filepath):
+                filename = '{}_{}.prof'.format(self.section.name, self.name)
+                open(os.path.join(filepath, filename), 'wb')
+                prof.dump_stats(os.path.join(filepath, filename))
+
+            else:
+                os.mkdir(os.path.join(path, subdir))
+                filename = '{}_{}.prof'.format(self.section.name, self.name)
+                open(os.path.join(filepath, filename), 'wb')
```
Makman2: redundant

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        if self.profile == 'True':
+            filename = '{}_{}.prof'.format(self.section.name, self.name)
+            open(filename, 'wb')
+            prof.dump_stats(filename)
+        else:
+            subdir = self.profile
+            filepath = os.path.join(path, subdir)
+            if os.path.isdir(filepath):
+                filename = '{}_{}.prof'.format(self.section.name, self.name)
+                open(os.path.join(filepath, filename), 'wb')
```
Makman2: redundant

coalib/bears/Bear.py
```diff
@@ -262,7 +265,8 @@ def __init__(self,
                  section: Section,
                  message_queue,
                  timeout=0,
-                 debugger=False):
+                 debugger=False,
+                 profile=False):
```
Makman2: documentation missing ;)

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def setup_profile(self, prof=None):
+        path = os.getcwd()
```
Makman2: path is only used in the `else` branch, please move it there

coalib/bears/Bear.py
```diff
@@ -305,6 +310,45 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        if self.profile == 'True':
+            filename = '{}_{}.prof'.format(self.section.name, self.name)
+            open(filename, 'wb')
+            prof.dump_stats(filename)
+        else:
+            subdir = self.profile
```
Makman2: you should check if it's actually an absolute path. So
1. it's a relative path: append it to os.getcwd
2. it's an absolute path: ignore os.getcwd

Makman2: I guess the API saving the dumps already respect that, so in fact you don't need the `join` below, just pass `self.profile` to `prof.dump_stats` :)

coalib/bears/Bear.py
Makman2: Looks like a bit copied from above. How about doing
```
self.filename = '{}_{}.prof'.format(self.section.name, self.name) if self.profile is True else self.profile 
```

coalib/bears/Bear.py
Makman2: Why not actually running everything in the same profile object?

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,9 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True, help='Profile Bears')
```
Makman2: what arguments does it take?

coalib/bears/Bear.py
```diff
@@ -273,6 +277,10 @@ def __init__(self,
                               time limit, use 0.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
+        :param debugger       A flag which enable debugger on those bears which
+                              are specified in ``debug_bears`` settings.
```
Makman2: why do you introduce documentation about the debugger in the commit for the profiler?

coalib/bears/Bear.py
```diff
@@ -273,6 +277,10 @@ def __init__(self,
...
+        :param profiler       The Path or name of directory where profiler file
```
Makman2: mismatch with the actual parameter name `profile`. But I guess you should call it `profiler` in the signature, analogous to `debugger`

Makman2: ~~P~~path

coalib/bears/Bear.py
```diff
@@ -305,6 +314,36 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+            while True:
+                try:
+                    result = profile.runcall(next, iterator)
+                    results.append(result)
+                except StopIteration:
+                    break
+            self.setup_profile(profiler=profile)
+            return results
+        else:
+            self.setup_profile(profiler=profile)
```
Makman2: please deduplicate the code, you use `self.setup_profile` twice in both branches.

coalib/bears/Bear.py
```diff
@@ -305,6 +314,36 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
 
+    def setup_profile(self, profiler=None):
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        if self.profile is True:
+            profiler.dump_stats(filename)
+        elif os.path.exists(self.profile) and os.path.isdir(self.profile):
+            profiler.dump_stats(os.path.join(self.profile, filename))
+        elif not os.path.isfile(self.profile):
+            os.mkdir(os.path.join(self.profile))
+            profiler.dump_stats(os.path.join(self.profile, filename))
```
Makman2: not sure this behaviour is too complicated... why not just allowing `True`/`False` and put it either into the working directory or a subdirectory like `profile`/`profile-data`/`bear-profiling`/... ?

coalib/bears/Bear.py
```diff
@@ -305,6 +314,36 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        else:
+            print('directory and filename cannot be same', file=sys.stdout)
```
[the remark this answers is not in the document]
RaiVaibhav: @Makman2  
1. first it will check whether it will check the value of `self.profile` if its `True` then it will dump in the current working directory,
2. else  it will check whether the path given or `directory name` or in `self.profile`  already exists and also check it is a directory if yes then it will dump the profiler files in that directory.

3. else it will check whether the path given or `directory name` is already a `file` (not directory) if it is not a file then it will create a `directory` with the name passed in `self.profile` or the `path passed` and dump the profiler files in this directory.
3. else last case which left is user is passing the `filename` or `filepath` in `self.profile` then it becomes obvious that it is not possible to dump profiler files in `file` (I will raise the exception in this case instead of printing)

coalib/bears/Bear.py
RaiVaibhav: What it the bear return is not `iterable` I know you already said a bear which return none is a buggy bear but that bear can be profiled IMO

coalib/bears/Bear.py
RaiVaibhav: No its not, If I understand here correctly then you are talking about this `filename = '{}_{}.prof'.format(self.section.name, self.name)` this part was added in cEP and also it is also giving a useful information like suppose some bear `ABCBear` in 2 different sections of `.coafile` then it will create a files with name `sectionOne_ABCBear` and `sectionTwo_ABCBear`. (here section name is my assumption)

RaiVaibhav: >not sure this behaviour is too complicated

I don't think its complicated :sweat_smile:

coalib/bears/Bear.py
RaiVaibhav: Yeah I accidentally squashed in my last update

coalib/bears/Bear.py
RaiVaibhav: actually, the *cli* argument is `--profile` so I am directly using the `profile` keyword and passing same to bear __init__ , do you want me to change it to `--profiler` or make it `self.profiler = profile`?

coalib/bears/Bear.py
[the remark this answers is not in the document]
RaiVaibhav: IMO this also get the ouput we want and also cover the case where user pass the path or filename which is actually a file not a directory it will give a proper error which user can understand easily
i.e., `FileExistsError: [Errno 17] File exists: '/home/vaibhav/test/my.txt'`
But if `/home/vaibhav/test/my.txt` is a directory then it will not give any error 
, what I was not wanted was to halt the coala normal execution and I am hoping this is the thing you wanted?
```
    def setup_profile(self, profiler=None):
        filename = '{}_{}.prof'.format(self.section.name, self.name)
        if self.profile is True:
            profiler.dump_stats(filename)
        else:
            if os.path.exists(self.profile) and os.path.isdir(self.profile):
                profiler.dump_stats(os.path.join(self.profile, filename))
            else:
                os.mkdir(os.path.join(self.profile))
                profiler.dump_stats(os.path.join(self.profile, filename))
```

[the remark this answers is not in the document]
Makman2: It's okay to halt coala's execution, if the user requested to profile code, but you can't profile because of such filename issues, you want to rather stop and make the user fix it. Otherwise you do a full coala run (that depending on the setup could take minutes) without profiling which wastes time.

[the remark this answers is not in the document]
Makman2: And yeah, your code looks better :+1: However one doubt: So if the user specifies a path in `self.profile`, we want to make it a directory, right?
Two improvements:
1. In the first else branch, you execute `profiler.dump_stats(os.path.join(...))` regardless on which branch, that can be moved outside the if ;)
2. since in the end you will always do `dump_stats`, and the only thing that changes is the filename, how about doing:

```python
    def setup_profile(self, profiler=None):
        filename = '{}_{}.prof'.format(self.section.name, self.name)
        if self.profile is True:
            path = filename
        else:
            if not (os.path.exists(self.profile) and os.path.isdir(self.profile)):
                os.mkdir(os.path.join(self.profile))
            
            path = os.path.join(self.profile, filename)            

    profiler.dump_stats(path)
```

coalib/bears/Bear.py
Makman2: that's not what I mean ;) In both branches you use `self.setup_profile`, so in any case `self.setup_profile` will be executed. You can make the code a bit more compact by moving that out of the `if` clause:

```python
if ...
   results = ... # I know this name is already used here, but choose some other good name or so.
else ...
   results = ...

self.setup_profile(profiler=profile)
return results
```

coalib/bears/Bear.py
Makman2: > you are talking about this filename = '{}_{}.prof'.format(self.section.name, self.name)

Nope, that is totally okay :) However, with the above code changes by you, my doubts may do not apply any more. So nvm for now :3

coalib/bears/Bear.py
Makman2: Ah right. Then keep the name `profile`, however, the docstring has to be corrected ;)

coalib/bears/Bear.py
```diff
@@ -276,6 +276,8 @@ def __init__(self,
                               time limit, use 0.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
+        :param debugger       A flag which enable debugger on those bears which
+                              are specified in ``debug_bears`` settings.
```
Makman2: The docstring doesn't really apply here. Since this is working directly on a bear, it's not tied to the `debug_bears` setting in the section. The step where `debug_bears` is parsed and bears are selected to be debugged happens before. Just say simply that if `True`, a debugger is started.

coalib/bears/Bear.py
```diff
@@ -273,6 +276,10 @@ def __init__(self,
                               time limit, use 0.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
+        :param profile        A flag which enable profiler on bears of a
```
Makman2: missing colon after `:param profile`

coalib/bears/Bear.py
```diff
@@ -273,6 +276,10 @@ def __init__(self,
...
+                              section. Contains path or name of directory where
+                              profiled files will dump if none provided then
+                              profiled files will dump in current directory.
```
Makman2: 1. --> A flag which enable**s** profile**ing** on **this bear**.
2. It's not really a flag then, is it? Since it accepts paths as well.
3. Nothing is mentioned that `True` is a valid value (a bit confusing is that you mention "if none provided then...", which suggests that `None` will do the trick)

coalib/bears/Bear.py
```diff
@@ -283,6 +291,7 @@ def __init__(self,
         self.message_queue = message_queue
         self.timeout = timeout
         self.debugger = debugger
+        self.profile = profile
```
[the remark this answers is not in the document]
Makman2: bump

coalib/bears/Bear.py
```diff
@@ -305,6 +313,34 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def _dump_profiled_files(self, profiler):
```
Makman2: which files are profiled? ;) --> the name has to become better still :)

coalib/bears/Bear.py
```diff
@@ -305,6 +313,34 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        if self.profile is True:
+            profiler.dump_stats(filename)
+        else:
+            if not (os.path.exists(self.profile) and
+                    os.path.isdir(self.profile)):
+                os.mkdir(os.path.join(self.profile))
+            path = os.path.join(self.profile, filename)
+            profiler.dump_stats(path)
```
Makman2: since you use `path` only here, you could inline it

coalib/bears/Bear.py
```diff
@@ -318,8 +354,13 @@ def run_bear_from_section(self, args, kwargs):
             self.warn('The bear {} cannot be executed.'.format(
                 self.name), str(err))
             return
-        if self.debugger:
+        if self.profile and self.debugger:
+            raise ValueError('Invalid argument passed, run debugger or '
+                             'profiler one at a time.')
```
Makman2: the exception should be inside `__init__`. Don't allow to create bears that can't run.

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile, that can help to create'
+        ' a Performant Bears by profiling the Bear code. To specify where to'
```
Makman2: ~~a Performant Bears~~ --> performant bears

Never use "a" for plurals. A plural form without "a" is equivalent and a grammatical correct sentence.

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' to dump the profile files, supply the path or the directory name, all'
```
Makman2: to ~~to~~

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' the bear (even implicit dependency bears) in a section will get'
```
Makman2: all ~~the~~ bear**s**

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' profile and there profile data will save in the specified directory'
```
Makman2: will ~~get profile~~ be profiled

Makman2: I would place a period here so it's easier to read:
"In a section will be profiled. Profile data will **be** saved in the specified directory or path."

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' or the path with name format {section.name}_{bear.name}.prof.')
```
Makman2: Up to you: Maybe we shouldn't document that, people will understand what the naming scheme means once they see a profile dump.

coalib/processes/Processing.py
```diff
@@ -312,6 +312,11 @@ def instantiate_bears(section,
     :param console_printer:  Object to print messages on the console.
     :return:                 The local and global bear instance lists.
     """
+    profile = (False if 'profile' not in section or (
+                section['profile'].value == 'False') else (
```
Makman2: and what if it's `false`? ;) This is not so bullet proof, especially as coala allows words like "yes", "nope" ... as boolean values. I know we have done it for the debugger feature like this, but maybe now is the time to start thinking about that.

coalib/bears/Bear.py
RaiVaibhav: Makman we are not handeling the `ValueError` in [processing.py](https://github.com/coala/coala/blob/master/coalib/processes/Processing.py#L336) should I handle the `ValueError` so that other bear can initialize and  run?

coalib/parsing/DefaultArgParser.py
RaiVaibhav: Although this will be included in `Profiling Interface` doc but I want to be here as sometimes people like me use `--help` to the information about flag :stuck_out_tongue:

coalib/processes/Processing.py
RaiVaibhav: [FALSE_STRINGS](https://github.com/coala/coala/blob/8ffd2237f84ddaca50327a89ce0132dc9999357f/coalib/misc/Constants.py#L71) doesn't included "false" ?

coalib/bears/Bear.py
RaiVaibhav: >Don't allow to create bears that can't run.

Nvm IMO halting the further execution will be good

tests/bears/BearTest.py
```diff
@@ -124,6 +127,20 @@ def run(self, x: int, y: int, z: int = 33):
         yield z
 
 
+class TestOneBear(Bear):
```
jayvdb: Can we put these bears in `tests/test_bears/` to be re-usable.
(and add a note that the other bears here should be moved to files in that directory , and/or create an issue for someone to move them .)

tests/processes/ProcessingTest.py
```diff
@@ -607,24 +608,24 @@ def test_global_instantiation(self):
         class TestOneBear(Bear):
 
             def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
+                         debugger=False, profile=False):
```
jayvdb: use `*args` and `**kwargs`

tests/processes/ProcessingTest.py
```diff
@@ -566,7 +566,8 @@ def test_yield_ignore_ranges(self):
     def test_loaded_bears_with_error_result(self):
         class BearWithMissingPrerequisites(Bear):
 
-            def __init__(self, section, queue, timeout=0.1, debugger=False):
+            def __init__(self, section, queue, timeout=0.1, debugger=False,
```
jayvdb: use `*args` and `**kwargs`

tests/processes/ProcessingTest.py
```diff
@@ -607,24 +608,24 @@ def test_global_instantiation(self):
...
                 raise RuntimeError
 
         class TestThreeBear(Bear):
 
             BEAR_DEPS = {TestTwoBear}
 
             def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
-                Bear.__init__(self, section, queue, timeout, debugger)
+                         debugger=False, profile=False):
```
jayvdb: use `*args` and `**kwargs`

tests/processes/ProcessingTest.py
```diff
@@ -607,24 +608,24 @@ def test_global_instantiation(self):
         class TestOneBear(Bear):
 
             def __init__(self, file_dict, section, queue, timeout=0.1,
```
jayvdb: use `*args` and `**kwargs`

And also these bears can be moved into `tests/test_bears/runtime_error_bears/`

tests/processes/ProcessingTest.py
```diff
@@ -607,24 +608,24 @@ def test_global_instantiation(self):
...
+                         debugger=False, profile=False):
                 raise RuntimeError
 
         class TestTwoBear(Bear):
 
             BEAR_DEPS = {TestOneBear}
 
             def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
+                         debugger=False, profile=False):
```
jayvdb: use `*args` and `**kwargs`

tests/processes/ProcessingTest.py
```diff
@@ -566,7 +566,8 @@ def test_yield_ignore_ranges(self):
     def test_loaded_bears_with_error_result(self):
         class BearWithMissingPrerequisites(Bear):
 
-            def __init__(self, section, queue, timeout=0.1, debugger=False):
```
jayvdb: Why do these need to be changed?

Is it mandatory for bears `__init__` to have a new `debugger` and `profile` args?
That would be a pretty serious breakage for an enhancement which is suppose to happen transparently to bears.

Why specify defaults at all?  core will require that bears add new parameters and copy those defaults or risk a breakage because the `__init__` is called without specifying values for args.

tests/processes/ProcessingTest.py
RaiVaibhav: I will create an issue after this code gets merged, If no one takes the issue within a week or two I will solve that issue :+1:

tests/processes/ProcessingTest.py
RaiVaibhav: @Makman2  ^^ ping 
Discussion happend here need your suggestion :point_right: : [July 20, 2018 3:50 PM](https://gitter.im/coala/coala?at=5b51b75163cf1636bdde13bd)

tests/bears/BearTest.py
RaiVaibhav: If no one solves takes the issue I will solve that :+1:

Makman2: Let's do that later, it's been done for other bears until now as well :+1:

tests/processes/ProcessingTest.py
jayvdb: @Makman2 , lots of small breaking changes over 1.5 years means we loose users in the upgrade.  They can handle one or two minor changes affecting them , but they give up after a while, especially when things break in `__init__`.

Makman2: Makes sense :+1: Then let's use the `section` directly then :+1: CC @RaiVaibhav

Makman2: However, let's merge this first analogous to the `debugger` feature which is already merged, and then push a commit to enforce API compatibility again.

RaiVaibhav: :+1: My first job will be to solve the API breaking problem, just after this code gets merge and fix

coalib/bears/Bear.py
```diff
@@ -273,16 +276,24 @@ def __init__(self,
                               time limit, use 0.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
+        :param profile:       Enable profiling on this bear. Contains path or
+                              name of directory where profiled files will dump,
+                              if a value ``True`` is provided then profiled
```
Makman2: if ``True`` is provided files will be saved to the current directory.

coalib/bears/Bear.py
```diff
@@ -273,16 +276,24 @@ def __init__(self,
                               time limit, use 0.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
+        :param profile:       Enable profiling on this bear. Contains path or
+                              name of directory where profiled files will dump,
```
Makman2: will be dumped

coalib/bears/Bear.py
```diff
@@ -305,6 +316,33 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def _dump_bear_profiled_file(self, profiler):
```
Makman2: --> `_dump_bear_profile_data`

coalib/bears/Bear.py
```diff
@@ -305,6 +316,33 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        if self.profile is True:
+            profiler.dump_stats(filename)
+        else:
+            if not (os.path.exists(self.profile) and
+                    os.path.isdir(self.profile)):
```
Makman2: is `exists` necessary if you check with `isdir`?

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile, that can help to create'
+        ' performant bears by profiling the Bear code. To specify where to'
```
Makman2: ~~the~~ Bear code.

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' dump the profile files, supply the path or the directory name, all'
```
Makman2: ... supply the path or ~~the~~ **a** directory name. All ...

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' bears (even implicit dependency bears) in a section will be profiled.'
+        ' Profile data will be save in the specified directory or the path'
```
Makman2: will be save**d**

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,15 @@ def default_arg_parser(formatter_class=None):
...
+        ' with name format {section.name}_{bear.name}.prof.')
```
Makman2: Needs a bit rephrasing, since what is specified here has to be always a directory (a path in general can also be a file). Especially worth to mention is that if it doesn't exist, the directory will be created.

coalib/bears/Bear.py
```diff
@@ -305,6 +316,33 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+                os.mkdir(os.path.join(self.profile))
```
Makman2: what if a file exists with that name? won't that raise an exception?

coalib/processes/Processing.py
```diff
@@ -312,6 +313,11 @@ def instantiate_bears(section,
     :param console_printer:  Object to print messages on the console.
     :return:                 The local and global bear instance lists.
     """
+    profile = (False if 'profile' not in section or (
+                section['profile'].value.lower() in FALSE_STRINGS) else (
+                    True if section['profile'].value.lower() in TRUE_STRINGS
+        else section['profile'].value))
```
Makman2: indentation is off

coalib/bears/Bear.py
RaiVaibhav: If I understand full then then `FileExitsError` will raised, when user give `filepath`, and IMO that error should have to be happen here.

coalib/bears/Bear.py
RaiVaibhav: NO :+1:  `isdir` first check it exits :+1:

coalib/bears/Bear.py
```diff
@@ -273,16 +276,24 @@ def __init__(self,
...
+                              dumped, if ``True`` is provided files will be
+                              saved to the current directory.
         """
         Printer.__init__(self)
 
         if message_queue is not None and not hasattr(message_queue, 'put'):
             raise TypeError('message_queue has to be a Queue or None.')
+        if profile and debugger:
+            raise ValueError('Invalid argument passed, run debugger or '
+                             'profiler one at a time on {}'.format(self.name))
```
Makman2: Would rephrase that:
> Can't run debugger and profiler at the same time.

(btw it's imo not necessary to include the bear name, since you see it in the traceback which bear raised the error)

coalib/parsing/DefaultArgParser.py
```diff
@@ -249,6 +249,17 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile, that can help to create'
+        ' performant bears by profiling Bear code. To specify where to dump the'
+        ' profile files, supply the path of a directory or a directory name.'
```
Makman2: see above, "path of a directory or a directory name"

coalib/bears/Bear.py
```diff
@@ -324,6 +335,32 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def _dump_bear_profile_data(self, profiler):
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        if self.profile is True:
+            profiler.dump_stats(filename)
```
jayvdb: dont have two invocations of `profiler.dump_stats` in different branches unless they are doing different things.

your logic should be

1. determine `filename`
2. invoke `dump_stats`

dont mix them together.

coalib/parsing/DefaultArgParser.py
```diff
@@ -251,6 +251,17 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True,
```
jayvdb: you need to create an arg group for `--profile` and `--debug` and set that group to only allow one of them.
See argparse docs

coalib/bears/Bear.py
```diff
@@ -290,18 +293,26 @@ def __init__(self,
...
+                              saved to the current directory.
         :raises TypeError:    Raised when ``message_queue`` is no queue.
         :raises RuntimeError: Raised when bear requirements are not fulfilled.
         """
         Printer.__init__(self)
 
         if message_queue is not None and not hasattr(message_queue, 'put'):
             raise TypeError('message_queue has to be a Queue or None.')
+        if profile and debugger:
+            raise ValueError("Can't run debugger and profiler at the same "
```
jayvdb: avoid "Can't", and unnecessary use of `'`  Use simpler English.

and start long strings on the next line, so you dont need to break the string across lines

coalib/bears/Bear.py
```diff
@@ -1,5 +1,7 @@
 import inspect
 import pdb
+import os
+import cProfile
```
jayvdb: put your items in a sensible sort order.

jayvdb: put your new items in a sensible sort order.

coalib/bears/Bear.py
RaiVaibhav: @jayvdb  didn't get, don't know how am I going to tackle this condition where all the things get satisfied because if `self.profile` value is type `bool` so `os.path.join` will throw error and can create further problem which I have to handle just because to make `one invocations of profiler.dump_stats`

>determine filename

This is not the only case if self.profile will be `True` it will dump the profiled files in the current working directory

coalib/bears/Bear.py
RaiVaibhav: Any suggestions here?

coalib/bears/Bear.py
Makman2: What @jayvdb means is that you have duplicated the logic regarding the invocation of `profiler.dump_stats`: In any case, you are dumping things here. So to avoid redundancy, do it like suggested:

> 1. determine filename
> 2. invoke dump_stats

means

```python
if self.profile is True:
    path = ...
else:
    path = ...

profiler.dump_stats(path)
```

because the difference is the parameter you pass to `dump_stats`, not the call.

I believe we already discussed that somewhen ^^

coalib/bears/Bear.py
```diff
@@ -290,18 +293,26 @@ def __init__(self,
         :param message_queue: The queue object for messages. Can be ``None``.
         :param timeout:       The time the bear is allowed to run. To set no
                               time limit, use 0.
+        :param profile:       Enable profiling on this bear. Contains directory
+                              name or directory path where profiled files will
```
Makman2: Dir name or path, not both. Both are rather synonyms, so choose one.

coalib/bears/Bear.py
RaiVaibhav: That was different code but same code duplication was there :+1:

coalib/bears/Bear.py
```diff
@@ -344,6 +371,33 @@ def log_message(self, log_message, timestamp=None, **kwargs):
     def run(self, *args, dependency_results=None, **kwargs):
         raise NotImplementedError
 
+    def _dump_bear_profile_data(self, profiler):
+        filename = '{}_{}.prof'.format(self.section.name, self.name)
+        if self.profile is True:
+            path = filename
+        else:
+            path = os.path.join(self.profile, filename)
```
jayvdb: `join` is already imported.  dont need to use `os.path.join`

coalib/bears/Bear.py
```diff
@@ -344,6 +371,33 @@ def log_message(self, log_message, timestamp=None, **kwargs):
...
+            if not os.path.isdir(self.profile):
```
jayvdb: add import for `isdir` so you dont use `os.path.isdir` here

coalib/parsing/DefaultArgParser.py
```diff
@@ -243,14 +243,27 @@ def default_arg_parser(formatter_class=None):
...
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    exclusive_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile, that can help to create'
+        ' performant bears by profiling Bear code. To specify where to dump the'
+        ' profiled files, supply the directory path. If specified directory'
+        ' does not exist it will be created. If the specified path points to an'
+        ' already existing file a ``FileExitsError`` is thrown . All bears'
```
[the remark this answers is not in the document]
RaiVaibhav: If user provide a `existing file` path, doesn't `FileExitsError` have to raise?
In verbose mode it will traceback information like this 
```
Traceback information is provided below:

  File "/home/vaibhav/4/lib/python3.6/site-packages/coalib/bears/Bear.py", line 436, in execute
  File "/home/vaibhav/4/lib/python3.6/site-packages/coalib/bears/Bear.py", line 417, in run_bear_from_section
    return self.profile_run(*args, **kwargs)
  File "/home/vaibhav/4/lib/python3.6/site-packages/coalib/bears/Bear.py", line 398, in profile_run
    self._dump_bear_profile_data(profiler)
  File "/home/vaibhav/4/lib/python3.6/site-packages/coalib/bears/Bear.py", line 381, in _dump_bear_profile_data
    os.makedirs(self.profile)
  File "/home/vaibhav/4/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: 'abc.txt'
```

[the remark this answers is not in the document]
jayvdb: 1. `FileExistsError` not `FileExitsError`
2. 'thrown' in the context of `--help` means that the application will *not* catch the exception and will fail with a Python backtrace.  If an exception is handled within the application, it is not relevant within `--help` and you need to revise `--help` to explain what the *application* will do, i.e. how does it handle the `FileExistsError` inside the application.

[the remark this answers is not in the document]
RaiVaibhav: >FileExistsError not FileExitsError

:+1: 

>'thrown' in the context of --help means that the application will not catch the exception and will fail with a Python backtrace

Yeah this is the current scenario, we are agreed on not handling the exception instead failing with traceback.

tests/bears/BearTest.py
```diff
@@ -537,6 +559,107 @@ def test_is_debugged(self):
...
+                bear.profile, 'name_TestTwoBear.prof'))
+
+    def test_profiler_with_file_path(self):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        _, filename = mkstemp()
+        section.append(Setting('profile', filename))
+        my_bear = TestTwoBear(section, self.queue)
+        with self.assertRaisesRegex(FileExistsError, filename):
```
RaiVaibhav: Error on appveyor https://ci.appveyor.com/project/coala/coala/build/1.0.13785/job/v3m7rrwwbjmfymm2#L388

jayvdb:
@RaiVaibhav , you might need the Windows trick that is being used at https://github.com/coala/coala-bears/pull/2523/files#diff-c3fb76c94bc425a8cdc7577d5d0e3e31R60

RaiVaibhav:
`shutil.rmtree` will throw `NotADirectoryError` as I am checking for the `file` :)

RaiVaibhav:
I findout the reason of previous problem here  https://www.logilab.org/blogentry/17873
So instead of hardcoding I used `NamedTemporaryFile` + `with` so solve the problem.

tests/bears/BearTest.py
```diff
@@ -537,6 +559,107 @@ def test_is_debugged(self):
...
+        os.rmdir('abc')
+        section.append(Setting('profile', '1'))
+        bear = TestThreeBear(section, self.queue)
+        self.assertIsNone(bear.run_bear_from_section(args, kwargs))
+        dump_stats.assert_called_with('name_TestThreeBear.prof')
+
+    @patch('cProfile.Profile.dump_stats')
+    def test_profiler_with_directory_exists(self, dump_stats):
+        with TemporaryDirectory() as temp_dir:
+            args = ()
```
jayvdb: the other three methods declare `args = ()` and `kwargs = {}` outside the `with`.  why not here too for consistency?

tests/bears/BearTest.py
```diff
@@ -537,6 +559,107 @@ def test_is_debugged(self):
...
+                        '-b', 'DependentBear',
+                        '-S', 'use_spaces=yeah',
+                        '--profile',
+                    ),
+                    autoapply=False,
+                    debug=debug
+                )[0]['cli'])
+            )
+        assert os.path.isfile('cli_SpaceConsistencyTestBear.prof')
+        os.remove('cli_SpaceConsistencyTestBear.prof')
```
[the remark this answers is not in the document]
RaiVaibhav: No actually I actually used the `DependentBear` which have `BEAR_DEPS` `SpaceConsistencyTestBear` also `DependentBear` doesn't have any run so only one file gets created with `SpaceConsistencyTestBear`

tests/bears/BearTest.py
RaiVaibhav: Yeah I can do that :+1: Its not like something ambiguous :)

tests/bears/BearTest.py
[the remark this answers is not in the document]
jayvdb: Excellent.  Then you can assert that there are no other `*.prof` files created, thereby ensuring you know exactly what happened.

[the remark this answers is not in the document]
RaiVaibhav: No actually during the test many `prof` files getting created becuase of `pytest-porfiler` and it is getting ignored in `gitignore`.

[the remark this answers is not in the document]
RaiVaibhav: Nvm above, added the test

coalib/bears/Bear.py
```diff
@@ -45,6 +46,26 @@ def _is_debugged(bear):
...
+    if not isinstance(bear, Bear):
+        raise ValueError('Positional argument bear is not an instance of '
+                         'Bear class.')
+    if 'profile' not in bear.section:
+        return False
+    try:
+        return bool(bear.section['profile'])
+    except ValueError:
+        pass
+    return bear.section['profile'].value
```
jayvdb: If you squint carefully, you will see this is identical to `_is_debugged` except for this last line.

If your helper function returns the `Setting`, i.e. `bear.section['x']` at the end, you can move the custom logic of looking inside the setting to the section of code which needs to understand the setting value.  the caller would then look for True, False, or Setting.

coalib/parsing/DefaultArgParser.py
```diff
@@ -243,14 +243,26 @@ def default_arg_parser(formatter_class=None):
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
-    outputs_group.add_argument(
+    exclusive_group_debug_profile = arg_parser.add_mutually_exclusive_group()
```
jayvdb: `devtool_exclusive_group`

coalib/bears/Bear.py
RaiVaibhav: I initially thought of that but intentionally I dropped it because not only the last line which are different, but if you see this logic
```python
    if 'debug_bears' not in bear.section:
        return False
```
and 
```python
    if 'profile' not in bear.section:
        return False
```
If I try to make a helper function which work for both then it make both the condition dependent as if any of the conditon will be true both will run, so yeah its little bit complicated and it will become somewhat more complex.

Makman2: I don't see a problem unifying the code ^^

> If I try to make a helper function which work for both then it make both the condition dependent as if any of the conditon will be true both will run

Why should that happen? You still have an `_is_profiled` function, but it will call some code both common for `_is_debugged` and `_is_profiled`^^

coalib/parsing/DefaultArgParser.py
```diff
@@ -243,14 +243,26 @@ def default_arg_parser(formatter_class=None):
...
         ' executed. To specify which bears to debug, supply bear names as'
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    devtool_exclusive_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile. To specify where to dump the'
+        ' profiled files, supply the directory path. If specified directory'
+        ' does not exist it will be created. If the specified path points to an'
+        ' already existing file a ``FileExistsError`` is thrown . All bears'
```
Makman2: `FileExistsError` is not relevant from user perspective, just say that it's an error.
+
No space before period

coalib/bears/Bear.py
jayvdb: I do see this logic
```py
    if 'debug_bears' not in bear.section:
        return False
```
and
```py
    if 'profile' not in bear.section:
        return False
```
.. and I see they are absolutely identical.  If you are seeing a difference, you are not squinting correctly.

To make sure there is no confusion, the new function would have a signature like `_setting_is_enabled(setting_name)` .  that helper would only look at one setting name; the name provided in the args

coalib/parsing/DefaultArgParser.py
jayvdb: IMO it is stupid, but I was told at https://github.com/coala/coala/pull/5599#discussion_r207102115 that it was part of the agreed upon design 

A crash is never appropriate for a known problem.  A full backtrace can go to the log/console and the application should exit with an appropriately chosen exit code.

This is a simple error scenario.  how does the backtrace even help the user who choose a directory which already had profile data in it?

RaiVaibhav: >how does the backtrace even help the user who chooses a directory which already had profile data in it?
No error for this case, profiled files will get overwritten.

An error will be generated when a user try to dump profiled files in the `filepath`

Makman2: > A crash is never appropriate for a known problem. A full backtrace can go to the log/console and the application should exit with an appropriately chosen exit code.

yeah that's right. I previously thought it's working bad in coala to exit out with proper error messages, but I have just seen `get_exitcode` and that is has a branch for `SystemExit`.

@RaiVaibhav could you reraise a `SystemExit` with some error code you think that's suitable (maybe reuse an existing error code from our list), and do an error log?

coalib/bears/Bear.py
```diff
@@ -32,17 +50,29 @@ def _is_debugged(bear):
...
+    setting = _setting_is_enabled(bear, key='debug_bears')
     try:
-        return bool(bear.section['debug_bears'])
+        value = bool(setting)
     except ValueError:
-        pass
-    return bear.name.lower() in (
-            map(str.lower, bear.section['debug_bears']))
+        value = bear.name.lower() in (
+                    map(str.lower, setting))
```
Makman2: can be put into a single line, reads a bit better ;)

Makman2: Don't forget to remove the redundant braces after that

coalib/bears/Bear.py
```diff
@@ -32,17 +50,29 @@ def _is_debugged(bear):
...
+    return value
+
+
+def _is_profiled(bear):
+    """
+    Check whether the bear is in profile mode according to its section-settings.
+
+    :param bear: Bear object.
+    :return:     True if ``profile`` is ``True`` or False if ``profile`` is
+                 ``False`` else return directory path specified in ``profile``.
```
Makman2: Maybe always return a directory name, and if it's just `True´, return `os.getcwd()` :+1: Easier to understand.

coalib/bears/Bear.py
```diff
@@ -24,6 +25,23 @@
...
+    :param key:  Setting key.
+    :return:     Setting object if setting key is in section else ``False``.
+    """
+    if not isinstance(bear, Bear):
+        raise ValueError('Positional argument bear is not an instance of '
+                         'Bear class.')
+    if key is None:
+        raise ValueError('No setting key passed.')
+
+    return (False if key not in bear.section else bear.section[key])
```
Makman2: redundant braces

coalib/bears/Bear.py
Makman2: And if the bear is not profiled, return an empty string. (document all that properly)

coalib/parsing/DefaultArgParser.py
Makman2: > No error for this case, profiled files will get overwritten.

That should be documented as well btw :+1:

coalib/parsing/DefaultArgParser.py
```diff
@@ -243,14 +243,25 @@ def default_arg_parser(formatter_class=None):
         '-r', '--relpath', nargs='?', const=True,
         help='return relative paths for files (must be called with --json)')
 
-    outputs_group.add_argument(
+    devtool_exclusive_group = arg_parser.add_mutually_exclusive_group()
+
+    devtool_exclusive_group.add_argument(
```
Makman2: unrelated change, different commit (even PR I would say so we can merge things quicker)

tests/bears/BearTest.py
```diff
@@ -537,6 +560,125 @@ def test_is_debugged(self):
...
+            self.assertEqual(bear.run_bear_from_section(args, kwargs),
+                             [1, 2, 3])
+            dump_stats.assert_called_once_with(os.path.join(
+                bear.profile, 'name_TestTwoBear.prof'))
+
+    def test_profiler_with_file_path(self):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        with NamedTemporaryFile() as temp_file:
```
Makman2: yeah like mentioned, please close the file immediately, otherwise you can't access it on windows. (please be sure that it also gets deleted at the end of the test)

RaiVaibhav: I think I have made the requested changes by John

coalib/parsing/DefaultArgParser.py
RaiVaibhav: Different commit is fine IMO :+1:

coalib/bears/Bear.py
RaiVaibhav: I think False is better instead of `empty string`

coalib/parsing/DefaultArgParser.py
RaiVaibhav: >That should be documented as well btw

Will create a new PR for the changes in doc

coalib/bears/Bear.py
```diff
@@ -24,6 +26,23 @@
 from .meta import bearclass
 
 
+def _setting_is_enabled(bear, key):
+    """
+    Check setting key is in section.
+
+    :param bear: Bear object.
+    :param key:  Setting key.
+    :return:     Setting object if setting key is in section else ``False``.
```
jayvdb: I said this should return False, True or Setting .

You have a _setting_is_enabled which doesnt return True.  How nuts is that?

coalib/bears/Bear.py
```diff
@@ -32,17 +51,29 @@ def _is_debugged(bear):
...
+    Check whether the bear is in profile mode according to its section-settings.
+
+    :param bear: Bear object.
+    :return:     current working directory if ``profile`` is ``True``, False
+                 if ``profile`` is ``False`` else return directory path
+                 specified in ``profile``.
+    """
+    setting = _setting_is_enabled(bear, key='profile')
+    try:
+        value = getcwd() if bool(setting) else False
```
jayvdb: sheet.  No.  Move the try: except around bool() into _setting_is_enabled, *exactly* as it was within _is_debugged except using `key`.  that part is *common* to both, and puts the bool exception handling in one spot.

Putting `getcwd()` inside the same exception handler for bool(setting) is *wrong*.

It is not more efficient.

It is ugly.

And `getcwd` can throw its own exception, and when someone gets that exception here you are forcing the person debugging this to wonder if it was caused by bool()  (they will probably mentally eliminate that because it will be an OSError), but you are also making it hard for them to use pdb because then cant directly step into getcwd.  They will need to step into bool(setting) first which will take them down the rabbit warren which is StringConverter, which they probably know they dont need to be in.

All because you tried to be 'efficient'.

Save the value from `_setting_is_enabled`, and use `if rv is True:` .  That is the fastest `if` that exists (except 1 == 1), as it is using identity (memory address comparision).  And it is readable.  It tells a story slowly.

Simple code should be simple.

Let the reader keep their brain power unused, so that their brain isnt sore before they reach the really hard code.

coalib/bears/Bear.py
```diff
@@ -322,6 +353,11 @@ def __init__(self,
         self.message_queue = message_queue
         self.timeout = timeout
         self.debugger = _is_debugged(bear=self)
+        self.profile = _is_profiled(bear=self)
+
+        if self.profile and self.debugger:
+            raise ValueError('Cannot run debugger and profiler at the same '
```
jayvdb: start the string on the next line, indented, so you dont need to split it.

coalib/parsing/DefaultArgParser.py
jayvdb: What the ??

It is not unrelated.  --profile and --debug are mutually exclusive.  It must be done in this commit in order for this commit to be atomic.  Otherwise *this* commit has bugs.  UI bugs.  The kind that coala is riddled with because they are deferred to some other commit instead of UI being considered a vital part of every change.

coalib/parsing/DefaultArgParser.py
jayvdb: This is mandatory @Makman2 .  More at https://github.com/coala/coala/pull/5599#discussion_r207759623

coalib/parsing/DefaultArgParser.py
```diff
@@ -251,6 +251,16 @@ def default_arg_parser(formatter_class=None):
         ' additional arguments. If used without arguments, all bears specified'
         ' with --bears will be debugged (even implicit dependency bears).')
 
+    outputs_group.add_argument(
+        '--profile', nargs='?', const=True,
+        help='Enable bear profiling with cProfile. To specify where to dump the'
+        ' profiled files, supply the directory path. If specified directory'
+        ' does not exist it will be created. If the specified path points to an'
+        ' already existing file a ``SystemExit`` is raised. All bears(even'
```
jayvdb: The user reading a help message on their console is never going to see `SystemExit` .  That is internal, and does not belong in the help message.  I already explained this.  `--help` needs to be written from the perspective of a CLI users -- what do *they* experience.  they get logs, and an exit code.

tests/bears/BearTest.py
```diff
@@ -537,6 +560,126 @@ def test_is_debugged(self):
...
+            section.append(Setting('profile', temp_file.name))
+            bear = TestTwoBear(section, self.queue)
+            with self.assertRaisesRegex(SystemExit, '2'):
```
jayvdb: assertRaisesRegex is inappropriate with SystemExit .  check the exit code, which is `e.code` iirc

coalib/bears/Bear.py
```diff
@@ -19,38 +21,75 @@
...
+    """
+    Check whether the bear is in profile mode according to its section-settings.
+
+    :param bear: Bear object.
+    :return:     current working directory if ``profile`` is ``True``, False
+                 if ``profile`` is ``False`` else return directory path
+                 specified in ``profile``.
+    """
+    setting = _setting_is_enabled(bear, key='profile')
+    if setting is True:
```
jayvdb: Stop grouping if together when you have a return.
```py
If not setting:
    return False

If setting is True:
    Return ...

# it must be a Setting
return setting.value
```

RaiVaibhav: There was a reason behind that 
```bash
>>> section.append(Setting('profile', 'abc'))
>>> if not setting:
...     print('abc')
... 
NameError: name 'setting' is not defined
>>> setting = section['profile']
>>> if not setting:
...     print('abc')
... 
  File "/usr/local/lib/python3.6/dist-packages/coala_utils/string_processing/StringConverter.py", line 46, in __bool__
    raise ValueError
ValueError
>>> 
```

RaiVaibhav: To prevent it from grouping I will make the logic to
```python
    if setting is True:
        return getcwd()
    if isinstance(setting, Setting):
        return setting.value
    return False
```

RaiVaibhav: or something like this 
```python
    setting = _setting_is_enabled(bear, key='profile')
    if isinstance(setting, bool):
        return getcwd() if setting else False
    return setting.value
```
**?**

tests/bears/BearTest.py
```diff
@@ -537,6 +560,132 @@ def test_is_debugged(self):
...
+                bear.profile, 'name_TestTwoBear.prof'))
+
+    def test_profiler_with_file_path(self):
+        args = ()
+        kwargs = {}
+        section = Section('name')
+        with NamedTemporaryFile() as temp_file:
+            section.append(Setting('profile', temp_file.name))
+            bear = TestTwoBear(section, self.queue)
+            with self.assertRaises(SystemExit) as e:
```
jayvdb: this is not an `e`.  it is a context manager. use `context` ..

tests/bears/BearTest.py
```diff
@@ -537,6 +560,132 @@ def test_is_debugged(self):
...
+                bear.run_bear_from_section(args, kwargs)
+                self.assertEqual(e.exception.code, 2)
```
jayvdb: because `e.exception...` is confusing

jayvdb:
ack https://github.com/coala/coala/pull/5599/commits/fea6e01bd0b5300b0f46c3528981c1e6558e1867

jayvdb:
@gitmate-bot ff

Issue #5545, opened 2018-06-19
Add settings inspection toolset in Debgguer interface


RaiVaibhav:
Add the code of Settings inspection toolset so that user can access the settings of a Bear in Debug environment.

PR #5610, opened 2018-07-08
Bears.py: Add Debugger Setting toolset

RaiVaibhav:
Add the code of Settings inspection
toolset so that user can access the
settings of a Bear in Debug environment
and add initial test.

Related to https://github.com/coala/coala/issues/5545

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -26,6 +26,13 @@ from .meta import bearclass
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, bear, *args, **kwargs):
+        if not isinstance(bear, Bear):
+            raise ValueError('Positional argument bear is not an instance of '
+                             'Bear class.')
+        super(Debugger, self).__init__(*args, **kwargs)
+        self.bear = bear
+
     def do_quit(self, arg):
         self.clear_all_breaks()
         super().do_continue(arg)
@@ -33,6 +40,18 @@ class Debugger(pdb.Pdb):
 
     do_q = do_quit
     do_exit = do_quit
+
+    def do_settings(self, arg):
+        md = self.bear.get_metadata()
+        section_params_dict = md.create_params_from_section(
+                              self.bear.section)
+        for param in md.non_optional_params:
+            self.message('%s = %r' % (param, section_params_dict[param]))
+        for param in md.optional_params:
+            self.message('%s = %r' % (param, section_params_dict[param] if
+                                      param in section_params_dict else
+                                      md.optional_params[param][2]))
+        return 1
 
 
 def debug_run(func, dbg=None, *args, **kwargs):
@@ -319,7 +338,7 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
                 self.name), str(err))
             return
         if self.debugger:
-            return debug_run(self.run, Debugger(), *args, **kwargs)
+            return debug_run(self.run, Debugger(bear=self), *args, **kwargs)
         else:
             return self.run(*args, **kwargs)
 
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -1,7 +1,7 @@
 import pdb
 from collections import defaultdict
 import datetime
-from io import BytesIO
+from io import BytesIO, StringIO
 import multiprocessing
 import unittest
 from os.path import abspath, exists, isfile, join, getmtime
@@ -56,6 +56,15 @@ class TestBear(Bear):
         self.print('set', 'up', delimiter='=')
         self.err('teardown')
         self.err()
+
+
+class TestOneBear(LocalBear):
+    def __init__(self, section, queue):
+        Bear.__init__(self, section, queue)
+
+    def run(self, x: int, y: str, z: int = 79, w: str = 'kbc'):
+        yield 1
+        yield 2
 
 
 class TypedTestBear(Bear):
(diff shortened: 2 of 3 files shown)


coalib/bears/Bear.py
```diff
@@ -26,6 +26,11 @@
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, *args, **kwargs):
+        self.bear = kwargs.get('bear', None)
+        kwargs.pop('bear', None)
```
Makman2: why not just declaring an optional argument `bear=None`? Then you don't need to pop it off from `kwargs`

coalib/bears/Bear.py
```diff
@@ -34,6 +39,19 @@ def do_quit(self, arg):
...
+        md = self.bear.get_metadata()
+        section_params_dict = md.create_params_from_section(self.bear.section)
+        for param in md.non_optional_params:
+            self.message('%s = %r' % (param, section_params_dict[param]))
+        for param in md.optional_params:
+            if param in section_params_dict:
+                self.message('%s = %r' % (param, section_params_dict[param]))
+            else:
+                self.message('%s = %r' % (param,
+                                          md.optional_params[param][2]))
```
Makman2: that should be
```python
self.message('%s = %r' % (param, section_params_dict[param] if param in section_params_dict else md.optional_params[param][2]))
```

tests/bears/BearTest.py
```diff
@@ -489,6 +498,16 @@ def test_debug_run_with_no_return(self, runcall):
         kwargs = {}
         self.assertIsNone(my_bear.run_bear_from_section(args, kwargs))
 
+    def test_debugger_settings_with_self_onscope(self):
```
Makman2: `with_self_onscope`?

coalib/bears/Bear.py
```diff
@@ -34,6 +38,17 @@ def do_quit(self, arg):
     do_q = do_quit
     do_exit = do_quit
 
+    def do_settings(self, arg):
+        md = self.bear.get_metadata()
```
Makman2: what is the scenario if `bear` is set to `None`?

coalib/bears/Bear.py
```diff
@@ -26,6 +26,10 @@
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, bear=None, *args, **kwargs):
```
Makman2: maybe put `bear=None` after `*args`?

RaiVaibhav:
@gitmate-bot  rebase

coalib/bears/Bear.py
```diff
@@ -34,6 +38,21 @@ def do_quit(self, arg):
...
+                                  self.bear.section)
+            for param in md.non_optional_params:
+                self.message('%s = %r' % (param, section_params_dict[param]))
+            for param in md.optional_params:
+                self.message('%s = %r' % (param, section_params_dict[param] if
+                                          param in section_params_dict else
+                                          md.optional_params[param][2]))
+            return 1
+        else:
+            self.error("Name 'settings' is not defined.")
```
RaiVaibhav: when there bear object is none i.e., `debug_run(self.run, Debugger(bear=None), *args, **kwargs)` then if use tries to use the settings commands it will give a message `"Name 'settings' is not defined."` (self.error is actually `print` function so it will not interrupt the whole program)

coalib/bears/Bear.py
```diff
@@ -319,7 +338,7 @@ def run_bear_from_section(self, args, kwargs):
                 self.name), str(err))
             return
         if self.debugger:
-            return debug_run(self.run, Debugger(), *args, **kwargs)
+            return debug_run(self.run, Debugger(bear=self), *args, **kwargs)
```
RaiVaibhav: Can't changed to `debug_run(self.run, *args, Debugger(bear=self), **kwargs)` for python3.4 during running pytest it will throw SyntaxError `SyntaxError: only named arguments may follow *expression`

coalib/bears/Bear.py
```diff
@@ -26,6 +26,10 @@
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, *args, bear=None, **kwargs):
+        self.bear = bear
+        super(Debugger, self).__init__(*args, **kwargs)
```
Makman2: invoke `super.__init__` always as soon as possible.

Makman2: It's a good habit to not run into weird states of the object, since you don't know what might happen to `self.bear` in the meanwhile. C++ for example enforces you to initialize other supers first (a slightly different syntax though), or Java does the same, which is often a good decision.

coalib/bears/Bear.py
Makman2: If not passing a bear is an erroneous state, raise an exception while instantiating the debugger. This is easiest and will always require you to initialize the debugger properly.

tests/bears/BearTest.py
```diff
@@ -489,6 +498,18 @@ def test_debug_run_with_no_return(self, runcall):
...
 
+    def test_do_settings(self):
+        section = Section('name', None)
+        section.append(Setting('x', '85'))
+        section.append(Setting('y', 'kbc3'))
+        section.append(Setting('z', '75'))
+        my_bear = TestOneBear(section, self.queue)
+        dbg = Debugger(bear=my_bear)
+        arg = ()
+        self.assertEqual(dbg.do_settings(arg), 1)
```
Makman2: without checking that the settings are actually printed this test is not very representative^^

RaiVaibhav: @Makman2  I think thats why we created a ideal test `DebugBearsTest` ?, I will try with mock and make test more effective

coalib/bears/Bear.py
```diff
@@ -26,6 +26,14 @@
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, *args, bear=None, **kwargs):
```
Makman2: since `bear` is actually not optional, you should remove the default value `None`

coalib/bears/Bear.py
```diff
@@ -26,6 +26,14 @@
...
+        if bear is None:
+            raise ValueError('Invalid argument passed. Bear object '
+                             'must passed as a argument for inspecting bear '
+                             'settings in debugging environment.')
```
Makman2: I would actually just say "Bear missing". That is short and easy to understand :P

coalib/bears/Bear.py
```diff
@@ -26,6 +26,14 @@
 
 class Debugger(pdb.Pdb):
 
+    def __init__(self, *args, bear=None, **kwargs):
+        if bear is None:
```
Makman2: Maybe make even a type-check: `if not isinstance(bear, Bear)`

Makman2:
ack 4416403f67737557727aafea6ebb2707b185f4bf

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

PR #5627, opened 2018-07-11
Integrate FileFactory with the core

palash25:
up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/io/FileFactory.py
+++ b/coalib/io/FileFactory.py
@@ -27,7 +27,7 @@ class FileFactory:
     >>> temp.write(bytes('This is a test file.', 'UTF-8'))
     20
     >>> temp.close()
-    >>> ff = FileFactory(temp.name)
+    >>> ff = FileFactory(temp.name, newline=False)
 
     File indices start with zero.
     To retrieve a single line:
@@ -64,13 +64,14 @@ class FileFactory:
     >>> os.remove(ff.name)
     """
 
-    def __init__(self, filename):
+    def __init__(self, filename, newline=True):
         """
         :param filename:
             The filepath.
         """
         self._filename = os.path.abspath(filename)
         self._timestamp = os.path.getmtime(self._filename)
+        self._newline = newline
 
     def get_line(self, line):
         """
@@ -89,7 +90,12 @@ class FileFactory:
         :return:
             A tuple containing the lines of the file.
         """
-        return tuple(self.string.splitlines())
+        lines = self.string.splitlines()
+        if self._newline:
+            return tuple(line if line.endswith('\n') else line + '\n'
+                         for line in lines)
+        else:
+            return tuple(lines)
 
     @cached_property
     def raw(self):
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -6,7 +6,6 @@ import queue
 import subprocess
 
 from coala_utils.string_processing.StringConverter import StringConverter
-from coala_utils.FileUtils import detect_encoding
 
 from coalib.collecting.Collectors import collect_files
 from coalib.misc.Exceptions import log_exception
@@ -30,6 +29,7 @@ from coalib.results.SourceRange import SourceRange
 from coalib.settings.Setting import glob_list, typed_list
 from coalib.parsing.Globbing import fnmatch
 from coalib.io.FileProxy import FileDictGenerator
+from coalib.io.FileFactory import FileFactory
 
 
 ACTIONS = [DoNothingAction,
@@ -268,16 +268,23 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
     :return:                Reads the content of each file into a dictionary
                             with filenames as keys.
     """
-    file_dict = {}
+    file_dict = FileDict()
     for filename in filename_list:
         try:
-            with open(filename, 'r',
-                      encoding=detect_encoding(filename)) as _file:
-                file_dict[filename] = tuple(_file.readlines())
+            file_dict[filename] = FileFactory(filename)
+            # This is a workaround to purposely raise a
+            # ``UnicodeDecodeError`` to move into the
+            # except clause.
+            FileFactory(filename).string
         except UnicodeDecodeError:
             if allow_raw_files:
                 file_dict[filename] = None
                 continue
+            else:
+                # In case raw files are not allowed the
+                # ``FileFactory`` object created for it
+                # needs to be evicted from the dictionary.
+                del file_dict[filename]
             logging.warning("Failed to read file '{}'. It seems to contain "
                             'non-unicode characters. Leaving it out.'
                             .format(filename))
@@ -825,3 +832,31 @@ def execute_section(section,
 
             for runner in processes:
                 runner.join()
+
+
+class FileDict(dict):
+    """
+    Acts as a middleware to provide the bears with the
+    actual file contents instead of the `FileFactory`
+    objects.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+
+    def __getitem__(self, key):
+        val = super().__getitem__(key)
+        if val:
+            return val.lines
+
+    def __delitem__(self, key):
+        super().__delitem__(key)
+
+    def keys(self):
+        return list(super().keys())
+
+    def __len__(self):
+        return super().__len__()
--- a/tests/processes/ProcessingTest.py
+++ b/tests/processes/ProcessingTest.py
@@ -79,6 +79,36 @@ class ProcessingTest(unittest.TestCase):
         self.unreadable_path = os.path.join(os.path.dirname(config_path),
                                             'unreadable')
 
+        factory_test_path = os.path.abspath(os.path.join(
+            os.path.dirname(__file__),
+            'file_factory_test_files'))
+        self.factory_test_file = os.path.join(factory_test_path,
+                                              'factory_test.txt')
+        self.a_bear_test_path = os.path.join(factory_test_path,
+                                             'a_bear_test.txt')
+        self.b_bear_test_path = os.path.join(factory_test_path,
+                                             'b_bear_test.txt')
+        self.c_bear_test_path = os.path.join(factory_test_path,
+                                             'c_bear_test.txt')
+        self.d_bear_test_path = os.path.join(factory_test_path,
+                                             'd_bear_test.txt')
+        self.e_bear_test_path = os.path.join(factory_test_path,
+                                             'e_bear_test.txt')
+        self.n_bear_test_path = os.path.join(factory_test_path,
+                                             'n_bear_test.txt')
+        self.n_bear_test_path_2 = os.path.join(factory_test_path,
+                                               'n_bear_test2.txt')
+        self.x_bear_test_path = os.path.join(factory_test_path,
+                                             'x_bear_test.txt')
+
+        filename_list = [self.factory_test_file, self.a_bear_test_path,
+                         self.b_bear_test_path, self.c_bear_test_path,
+                         self.d_bear_test_path, self.e_bear_test_path,
+                         self.n_bear_test_path, self.n_bear_test_path_2,
+                         self.x_bear_test_path]
+
+        self.file_dict = get_file_dict(filename_list)
+
         self.result_queue = queue.Queue()
         self.queue = queue.Queue()
         self.log_queue = queue.Queue()
@@ -275,13 +305,7 @@ class ProcessingTest(unittest.TestCase):
                  Result.from_values('ABear', 'u', 'f', 5, 1),
                  Result.from_values('ABear', 'u', 'f', 6, 1)]},
             {1: [first_global]},
-            {'f': ['first line  # stop ignoring, invalid ignore range\n',
-                   'second line  # ignore all\n',
-                   'third line\n',
-                   "fourth line  # gnore shouldn't trigger without i!\n",
-                   '# Start ignoring ABear, BBear and CBear\n',
-                   '# Stop ignoring\n',
-                   'seventh']},
+            {'f': self.file_dict[self.factory_test_file]},
             lambda *args: self.queue.put(args[2]),
             section,
             None,
@@ -457,9 +481,7 @@ class ProcessingTest(unittest.TestCase):
         self.assertFalse(check_result_ignore(result, ranges))
 
     def test_yield_ignore_ranges(self):
-        test_file_dict_a = {'f':
-                            ('# Ignore aBear\n',
-                             'a_string = "This string should be ignored"\n')}
+        test_file_dict_a = {'f': self.file_dict[self.a_bear_test_path]}
         test_ignore_range_a = list(yield_ignore_ranges(test_file_dict_a))
         for test_bears, test_source_range in test_ignore_range_a:
             self.assertEqual(test_bears, ['abear'])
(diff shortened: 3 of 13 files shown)


palash25:
These are the failing tests out of which the ranges tests seems to be the toughest one.

```
FAIL tests/coalaCITest.py::coalaCITest::test_find_no_issues
FAIL tests/coalaCITest.py::coalaCITest::test_find_no_issues_debug
FAIL tests/coalaCITest.py::coalaCITest::test_section_ordering
FAIL tests/coalaTest.py::coalaTest::test_coala_with_color
FAIL tests/coalaTest.py::coalaTest::test_coala_without_color
FAIL tests/output/ConsoleInteractionTest.py::ConsoleInteractionTest::test_print_results_multiple_ranges
FAIL tests/processes/ProcessingTest.py::ProcessingTest::test_yield_ignore_ranges
FAIL tests/results/result_actions/ShowPatchActionTest.py::ShowPatchActionTest::test_apply
FAIL tests/testing/LocalBearTestHelperTest.py::LocalBearCheckResultsTest::test_bad_assertComparableObjectsEqual
FAIL tests/testing/LocalBearTestHelperTest.py::LocalBearCheckResultsTest::test_good_assertComparableObjectsEqual
```

coalib/processes/Processing.py
```diff
@@ -267,16 +268,17 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
     :return:                Reads the content of each file into a dictionary
                             with filenames as keys.
     """
-    file_dict = {}
+    file_set = {}
```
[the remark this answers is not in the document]
palash25: We talked about not calling it a file_dict and instead a file_set but that was a long time ago before GSoC I think. I will change it back to a file_dict for now and we can decide on a better name afterwards.

coalib/processes/Processing.py
```diff
@@ -267,16 +268,17 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
...
     for filename in filename_list:
         try:
-            with open(filename, 'r',
-                      encoding=detect_encoding(filename)) as _file:
-                file_dict[filename] = tuple(_file.readlines())
+            file_set[filename] = FileFactory(filename)
+            file_set[filename].string
```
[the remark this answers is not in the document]
palash25: Its to invoke `FileFactory.string` in case of raw files to raise a UnicodeDecode error so that we can move into the except clause that handles the raw files. Its the only workaround we have for the old core. I think you suggested this to me.

I will add a comment explaining it :+1:

coalib/processes/Processing.py
```diff
@@ -498,6 +500,8 @@ def yield_ignore_ranges(file_dict):
         if file is None:
             continue
 
+        file = list(file.lines)
```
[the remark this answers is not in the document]
palash25: you are right that's unnecessary tuple works too :sweat_smile: 

Removed list

tests/test_bears/LineCountTestBear.py
```diff
@@ -13,6 +13,6 @@ def run(self, filename, file):
         """
         yield Result.from_values(
             origin=self,
-            message='This file has {count} lines.'.format(count=len(file)),
+            message='This file has {count} lines.'.format(count=len(file.lines)),
```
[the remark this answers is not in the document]
palash25: So the FileFactories are all collected in `get_file_dict` and then they are passed around in the bears for analysis right. So before this the bears used to receive the file-contents.

Can you explain this to me a little better. What would the bears receive now to run the analysis on whether it is `FileFactory` or `FileFactory.lines`? I don't know all the details of a coala run. 

Is there a part of the codebase where we can make changes so that all the bears receive only the file contents and not the objects so that we don't have to change the bears implementation individually, you know like a middleware between the collection of objects and the actual analysis?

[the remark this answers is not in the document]
Makman2: > Can you explain this to me a little better? What would the bears receive now to run the analysis on whether it is FileFactory or FileFactory.lines? I don't know all the details of a coala run.

Definitely `FileFactory.lines` for now. The old bears expect that currently.

> Is there a part of the codebase where we can make changes so that all the bears receive only the file contents and not the objects so that we don't have to change the bears implementation individually, you know like a middleware between the collection of objects and the actual analysis?

Yeah inside Processing.py. Instead of giving bears the file factory objects, give them `file_factory.lines`. I know we currently loose the lazy-loading capability with that, but that's easiest for now.

If you want to keep the lazy loading capabilities, we could write (like you suggested) some kind of "middleware", meaning we have a proxy dict object that gives access to file contents that works the same way as before for bears, but calls the right functions of `FileFactory` in the background.
Like:
```python
class FileDict:  # mimics dict
    def __getitem__(self, filename):
        return self._my_file_middlewares[filename]
    #...

class FileMiddleware:
    def __getitem__(self, lineindex):
        return self._underlying_file_proxy.lines[lineindex]
    #...
```

coalib/processes/Processing.py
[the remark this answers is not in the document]
Makman2: fine with me :+1:

coalib/processes/Processing.py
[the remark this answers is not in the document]
Makman2: > Its to invoke FileFactory.string in case of raw files to raise a UnicodeDecode error so that we can move into the except clause that handles the raw files. Its the only workaround we have for the old core. I think you suggested this to me.

I know, I just wanted to encourage you to explain it to others reading the code as well that are not into the topic ;)

coalib/processes/Processing.py
```diff
@@ -837,3 +841,34 @@ def execute_section(section,
...
+            content = self.__dict__[key]
+            if content is None:
+                return content
+            else:
+                return content.lines
+        else:
+            raise KeyError
+
+    def __delitem__(self, key):
+        del self.__dict__[key]
```
Makman2: why not using `super`?

coalib/processes/Processing.py
```diff
@@ -837,3 +841,34 @@ def execute_section(section,
 
             for runner in processes:
                 runner.join()
+
+
+class FileDict(dict):
```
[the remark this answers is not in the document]
palash25: Sure thanks for reminding me. Will do it in the next push.

coalib/processes/Processing.py
```diff
@@ -267,16 +267,17 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
...
     """
-    file_dict = {}
+    file_dict = FileDict()
     for filename in filename_list:
         try:
-            with open(filename, 'r',
-                      encoding=detect_encoding(filename)) as _file:
-                file_dict[filename] = tuple(_file.readlines())
+            file_dict[filename] = FileFactory(filename)
+            FileFactory(filename).string
```
shreyans800755: What is this line for ?

palash25: The old core relied on a `UnicodeDecodeError` in case of raw files. This is just a workaround for that. When we invoke `FileFactory.string` in case of raw files it will throw a `UnicodeDecodeError` and move into the except case.

I will add a comment in the next push.

tests/processes/ProcessingTest.py
```diff
@@ -551,7 +560,8 @@ def test_yield_ignore_ranges(self):
             self.assertEqual(test_source_range.end.column, 43)
 
         # This case was a bug.
-        test_file_dict_single_line = {'f': ('# ignore XBEAR',)}
+        test_file_dict_single_line = \
+            {'f': self.file_dict[self.x_bear_test_path]}
```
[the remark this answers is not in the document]
palash25: But then its a tuple wouldn't that change the structure of the object. We want this `{filename: filefactory}` not this `({filename: filefactory})`

tests/processes/file_factory_test_files/factory_test.txt
```diff
@@ -0,0 +1,7 @@
+first line  # stop ignoring, invalid ignore range
+second line  # ignore all
+third line
+fourth line  # gnore shouldn't trigger without i!
```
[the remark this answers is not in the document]
palash25: That's the way it should be its not a typo :wink:

coalib/processes/Processing.py
jayvdb: Not done yet

coalib/processes/Processing.py
```diff
@@ -837,3 +838,33 @@ def execute_section(section,
...
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+
+    def __getitem__(self, key):
+        val = super().__getitem__(key)
+        if val is None:
+            return val
+        else:
+            return val.lines
```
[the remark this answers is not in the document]
jayvdb: @palash25 ?

tests/processes/ProcessingTest.py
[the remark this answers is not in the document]
jayvdb: Adding round brackets does not make it a tuple.

Please remove the `\`.

There are many ways to do this.

coalib/processes/Processing.py
```diff
@@ -267,16 +267,17 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
...
-                      encoding=detect_encoding(filename)) as _file:
-                file_dict[filename] = tuple(_file.readlines())
+            file_dict[filename] = FileFactory(filename)
+            FileFactory(filename).string
         except UnicodeDecodeError:
             if allow_raw_files:
                 file_dict[filename] = None
                 continue
+            else:
+                del file_dict[filename]
```
jayvdb: Deleting the object here makes not sense to me ... the objective of FileFactory was to be able to provide objects which could wrap files which are 1) raw, 2) alternative encodings, 3) etc.

It seems like this is only barely integrated, which is ok I guess at this stage, but you need to add comments where this patch is adding hacks to get the `FileFactory` objects in use, and it seems this `del` seems like it is intentionally reducing the integration of `FileFactory` objects because other parts of core havent been updated yet.

If the 'WIP' was because this wasnt finished yet, unfortunately time has run out, and this patch needs to get merged, so this `del` needs to stay otherwise you'll get more test failures.

coalib/processes/Processing.py
```diff
@@ -267,16 +267,17 @@ def get_file_dict(filename_list, log_printer=None, allow_raw_files=False):
     :return:                Reads the content of each file into a dictionary
                             with filenames as keys.
     """
-    file_dict = {}
+    file_dict = FileDict()
     for filename in filename_list:
         try:
-            with open(filename, 'r',
-                      encoding=detect_encoding(filename)) as _file:
```
jayvdb: I see that FileFactory is not using `detect_encoding`, and is always using `'utf-8'`.  I suspect that is why one of your tests is failing.

Anyway, that is a bug and needs to get fixed.

palash25: Makman had raised this issue in an [earlier review](https://github.com/coala/coala/pull/5627#discussion_r203849101) and said that we can do it later but I will try to use `detect_encoding` in `FileFactory` locally and see if the tests pass.

coalib/processes/Processing.py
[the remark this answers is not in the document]
palash25: Yes it would thanks @sks444 

It worked

coalib/processes/Processing.py
palash25: I needed to evict the file factories when the raw files were not allowed to make the tests pass. I have added the comments explaining the deletion.

coalib/io/FileFactory.py
```diff
@@ -64,13 +64,14 @@ class FileFactory:
     >>> os.remove(ff.name)
     """
 
-    def __init__(self, filename):
+    def __init__(self, filename, newline=True):
```
Makman2: documentation :)

coalib/io/FileFactory.py
```diff
@@ -89,7 +90,12 @@ def lines(self):
         :return:
             A tuple containing the lines of the file.
         """
-        return tuple(self.string.splitlines())
+        lines = self.string.splitlines()
```
Makman2: `splitlines` offers a `keepends` parameter, maybe that's what we should use?

coalib/io/FileFactory.py
```diff
@@ -64,13 +64,14 @@ class FileFactory:
...
 
-    def __init__(self, filename):
+    def __init__(self, filename, newline=True):
         """
         :param filename:
             The filepath.
         """
         self._filename = os.path.abspath(filename)
         self._timestamp = os.path.getmtime(self._filename)
+        self._newline = newline
```
Makman2: Since the bear writer using the file factory can't control this parameter via `__init__` (because the core instantiates them), we should maybe add additional function(s) (maybe `get_lines(options)`?) that allow you to get different kinds of strings and lines?

coalib/processes/Processing.py
Makman2: `FileFactory` aims to be consistent and simple, that was my motivation to fix the encoding. I have thought about different encodings and using `detect_encoding` but wanted to make it possible at the same time to control encodings as well (would e.g. allow to write encoding converter bears). Haven't yet decided finally^^

If it helps for now to make tests pass it's fine with me, but always using `detect_encoding` is a workaround for me^^

coalib/io/FileFactory.py
jayvdb: This was a part of my cep PR iirc

coalib/processes/Processing.py
jayvdb: ok, lets create issues about this and work on them after gsoc.  @palash25 you can mention this in your work report under 'things to be done'

tests/processes/ProcessingTest.py
```diff
@@ -275,13 +305,7 @@ def test_process_queues(self):
...
                  Result.from_values('ABear', 'u', 'f', 6, 1)]},
             {1: [first_global]},
-            {'f': ['first line  # stop ignoring, invalid ignore range\n',
-                   'second line  # ignore all\n',
-                   'third line\n',
-                   "fourth line  # gnore shouldn't trigger without i!\n",
-                   '# Start ignoring ABear, BBear and CBear\n',
-                   '# Stop ignoring\n',
-                   'seventh']},
+            {'f': self.file_dict[self.factory_test_file]},
```
jayvdb: please move all of these changes to existing tests to another PR, or at least a separate commit.

when rewriting code, existing tests should not be modified.  that is how you prove , easily , that you correctly matched the old functionality.

these changes make the reviewers life much more difficult as they need to verify that no functional changes were done to the tests.

And even if a reviewer does that, any time anyone else goes and looks at this change, like when chasing down a bug, they need to verify the exact same thing.  Not nice.

jayvdb: (note, your separate commit or PR can be before or after this PR)

tests/processes/ProcessingTest.py
```diff
@@ -546,7 +556,7 @@ def test_yield_ignore_ranges(self):
         self.assertEqual(source_range.start.line, 1)
         self.assertEqual(source_range.start.column, 1)
         self.assertEqual(source_range.end.line, 1)
-        self.assertEqual(source_range.end.column, 14)
+        self.assertEqual(source_range.end.column, 15)
```
jayvdb: I suspect that this shouldnt be done.  The `# This case was a bug.` above makes me worried.

It seems like the change is due to an added `\n` in the file.

jayvdb: It was https://github.com/coala-analyzer/coala/issues/2563

jayvdb: It doesnt seem that missing EOL@EOF was relevant to that bug.

And a missing trailing `\n` is not a thing in universal newlines mode so it shouldnt be a problem ... so ... accepting this change

tests/processes/ProcessingTest.py
jayvdb: Thanks this was done, and I've confirmed that the commit before the test changes does in fact build : https://travis-ci.org/jayvdb/coala/builds/415063261

Issue #5628, opened 2018-07-12
CI build failure because of httpstat.us SSL certificate error


RaiVaibhav:
https://travis-ci.org/virresh/coala/jobs/402970785#L1642
https://ci.appveyor.com/project/anctartica/coala
https://circleci.com/gh/virresh/coala/6

difficulty/low
importance/high

jayvdb:
Yup.  Lets go with https://httpbin.org at the moment as it seems more reliable.

jayvdb:
httpstat.us failing again/still : https://travis-ci.org/coala/coala/jobs/403314993

jayvdb:
https://github.com/coala/coala/issues/5433 was switching from httpbin to httpstat.us .

We need a more reliable approach :/

jayvdb:
I set up `pytest-httpbin`, and the result was. ... 

```py
requests.exceptions.SSLError: HTTPSConnectionPool(host='127.0.0.1', port=45891): Max retries exceeded with url: /status/418 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1045)')))
```

I change it to https://www.google.com/teapot , which was the original plan behind https://github.com/coala/coala/pull/5435 , and will not surprise anyone that the Google servers did not have an SSL error :P 

IMO we use Google Teapot.

PR #5630, opened 2018-07-12
BearTest: Replace httpstat with Google teapot

RaiVaibhav:
Use httpbin instead of httpbin because
of SSL certification error.

Closes https://github.com/coala/coala/issues/5628

### DIFF
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -515,8 +515,8 @@ class BearDownloadTest(BearTestBase):
         super().setUp()
         self.mock_url = 'https://test.com'
         self.filename = 'test.html'
-        self.teapot_url = 'https://httpstat.us/418'
-        # https://www.google.com/teapot and
+        self.teapot_url = 'https://www.google.com/teapot'
+        # 'https://httpstat.us/418' and
         # http://httpbin.org/status/418 also work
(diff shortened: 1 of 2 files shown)


tests/bears/BearTest.py
```diff
@@ -515,7 +515,7 @@ def setUp(self):
         super().setUp()
         self.mock_url = 'https://test.com'
         self.filename = 'test.html'
-        self.teapot_url = 'https://httpstat.us/418'
+        self.teapot_url = 'https://httpbin.org/status/418'
```
jayvdb: I think we should use the Google Teapot.

See https://github.com/coala/coala/issues/5628#issuecomment-404678858 for background.

RaiVaibhav:
Seems like only travis failing for all three https://travis-ci.org/coala/coala/jobs/403347645

jayvdb:
And failed several times after restarting . We have an issue about those tests. This failure is also becoming quite annoying.

jayvdb:
ack https://github.com/coala/coala/pull/5630/commits/c3489fe06888a7e199de5c763ca42329de064696

jayvdb:
@gitmate-bot ff

jayvdb:
@gitmate-bot ff

Issue #5676, opened 2018-07-28
Fix the API breakage problem because of `debugger` flag in bear base class


RaiVaibhav:
For further information https://github.com/coala/coala/pull/5599/files#r203964627

RaiVaibhav:
currently its block, it will be solved after Profiler interface PR gets merge so all the breakge problem gets solved in one PR 

Maintainer assign this to me.

jayvdb:
I'd prefer that this is done first, and the profile PR doesnt create a breakage unnecessarily causing code churn.

RaiVaibhav:
@jayvdb  The profiler Interface is in my GSoC and time is close so If I first wait for the API PR to get a review and merge then change the code of Profiler Interface PR and then get a further review and then merge the I am already failed because of time as I will also take my time to put all the work and make a document for the final evaluation

Hoping you will understand, as API breakage is also my concern and because of that I opened a PR asap.

RaiVaibhav:
Also, all the previous review will get mixed and right now Profiler PR Github page becomes heavy so I will open same in next PR as sometimes I can't open with my internet connection.

jayvdb:
This is a simple fix, it is implemented already, and can get pushed in.
Updating your other PR after that is not hard.
@Makman2 can hide comments on that PR if it is too large, marking them as 'Outdated'.
https://help.github.com/articles/managing-disruptive-comments/#hiding-a-comment

If your PR is still unusable, then create a new PR and close the old one.

The size of the PR comments is not a reason to merge problematic code.

RaiVaibhav:
>If your PR is still unusable, then create a new PR and close the old one.

No this is not a case :stuck_out_tongue: 

Can you take a look at a comment https://github.com/coala/coala/pull/5677#discussion_r205954170

RaiVaibhav:
Its already fixed

RaiVaibhav:
https://github.com/coala/coala/pull/5677

PR #5677, opened 2018-07-28
Bear: Remove debugger from init

RaiVaibhav:
Add initial Fix the API breakage problem
because of `debugger` flag in bear base
class.

Related to https://github.com/coala/coala/issues/5676

### DIFF
--- a/coalib/bears/Bear.py
+++ b/coalib/bears/Bear.py
@@ -24,6 +24,27 @@ from coalib.settings.ConfigurationGathering import get_config_directory
 from .meta import bearclass
 
 
+def _is_debugged(bear):
+    """
+    Check whether the bear is in debug mode according to its section-settings.
+
+    :param bear: Bear object.
+    :return:     True if ``debug_bears`` is ``True`` or if bear name specified
+                 in ``debug_bears`` setting match with the bear parameter.
+    """
+    if not isinstance(bear, Bear):
+        raise ValueError('Positional argument bear is not an instance of '
+                         'Bear class.')
+    if 'debug_bears' not in bear.section:
+        return False
+    try:
+        return bool(bear.section['debug_bears'])
+    except ValueError:
+        pass
+    return bear.name.lower() in (
+            map(str.lower, bear.section['debug_bears']))
+
+
 class Debugger(pdb.Pdb):
 
     def __init__(self, bear, *args, **kwargs):
@@ -280,8 +301,7 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
     def __init__(self,
                  section: Section,
                  message_queue,
-                 timeout=0,
-                 debugger=False):
+                 timeout=0):
         """
         Constructs a new bear.
 
@@ -301,7 +321,7 @@ class Bear(Printer, LogPrinterMixin, metaclass=bearclass):
         self.section = section
         self.message_queue = message_queue
         self.timeout = timeout
-        self.debugger = debugger
+        self.debugger = _is_debugged(bear=self)
 
         self.setup_dependencies()
         cp = type(self).check_prerequisites()
--- a/tests/bears/BearTest.py
+++ b/tests/bears/BearTest.py
@@ -16,7 +16,7 @@ import requests_mock
 from coalib.bearlib.aspects.collections import AspectList
 from coalib.bearlib.aspects.Metadata import CommitMessage
 from coalib.bearlib.languages.Language import Language, Languages
-from coalib.bears.Bear import Bear, Debugger
+from coalib.bears.Bear import Bear, Debugger, _is_debugged
 from coalib.bears.BEAR_KIND import BEAR_KIND
 from coalib.bears.GlobalBear import GlobalBear
 from coalib.bears.LocalBear import LocalBear
@@ -478,7 +478,8 @@ class BearTest(BearTestBase):
     @patch('pdb.Pdb.do_continue')
     def test_custom_continue(self, do_continue):
         section = Section('name')
-        bear = Bear(section, self.queue, debugger=True)
+        section.append(Setting('debug_bears', 'True'))
+        bear = Bear(section, self.queue)
         args = ()
         self.assertEqual(Debugger(bear).do_quit(args), 1)
         pdb.Pdb.do_continue.assert_called_once_with(args)
@@ -487,7 +488,8 @@ class BearTest(BearTestBase):
     @patch('coalib.bears.Bear.Debugger.runcall', side_effect=((1, 2), 1, 2))
     def test_debug_run_with_return(self, runcall):
         section = Section('name')
-        my_bear = Bear(section, self.queue, debugger=True)
+        section.append(Setting('debug_bears', 'True'))
+        my_bear = Bear(section, self.queue)
         args = ()
         kwargs = {}
         self.assertEqual(my_bear.run_bear_from_section(args, kwargs), [1, 2])
@@ -495,7 +497,8 @@ class BearTest(BearTestBase):
     @patch('coalib.bears.Bear.Debugger.runcall', return_value=None)
     def test_debug_run_with_no_return(self, runcall):
         section = Section('name')
-        my_bear = Bear(section, self.queue, debugger=True)
+        section.append(Setting('debug_bears', 'True'))
+        my_bear = Bear(section, self.queue)
         args = ()
         kwargs = {}
         self.assertIsNone(my_bear.run_bear_from_section(args, kwargs))
@@ -517,6 +520,22 @@ class BearTest(BearTestBase):
         self.assertEqual(output[3], "w = 'kbc'")
         with self.assertRaises(ValueError):
             Debugger(bear=None)
+
+    def test_is_debugged(self):
+        with self.assertRaises(ValueError):
+            _is_debugged(bear=None)
+
+        section = Section('name')
+        uut = Bear(section, self.queue)
+        self.assertEqual(_is_debugged(uut), False)
+        section.append(Setting('debug_bears', 'tRuE'))
+        self.assertEqual(_is_debugged(uut), True)
+        section.append(Setting('debug_bears', '0'))
+        self.assertEqual(_is_debugged(uut), False)
+        section.append(Setting('debug_bears', 'Bear, ABear'))
+        self.assertEqual(_is_debugged(uut), True)
+        section.append(Setting('debug_bears', 'abc, xyz'))
+        self.assertEqual(_is_debugged(uut), False)
 
 
 class BrokenReadHTTPResponse(BytesIO):
--- a/tests/processes/ProcessingTest.py
+++ b/tests/processes/ProcessingTest.py
@@ -566,8 +566,8 @@ class ProcessingTest(unittest.TestCase):
     def test_loaded_bears_with_error_result(self):
         class BearWithMissingPrerequisites(Bear):
 
-            def __init__(self, section, queue, timeout=0.1, debugger=False):
-                Bear.__init__(self, section, queue, timeout, debugger)
+            def __init__(self, *args, **kwargs):
+                Bear.__init__(self, *args, **kwargs)
 
             def run(self):
                 return []
@@ -606,33 +606,28 @@ class ProcessingTest(unittest.TestCase):
     def test_global_instantiation(self):
         class TestOneBear(Bear):
 
-            def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
+            def __init__(self, *args, **kwargs):
                 raise RuntimeError
 
         class TestTwoBear(Bear):
 
             BEAR_DEPS = {TestOneBear}
 
-            def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
+            def __init__(self, *args, **kwargs):
                 raise RuntimeError
 
         class TestThreeBear(Bear):
 
             BEAR_DEPS = {TestTwoBear}
 
-            def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
-                Bear.__init__(self, section, queue, timeout, debugger)
+            def __init__(self, file_dict, section, queue, timeout=0.1):
+                Bear.__init__(self, section, queue, timeout)
 
         global_bear_list = [TestTwoBear, TestThreeBear]
         list1, list2 = instantiate_bears(self.sections['cli'], [],
                                          global_bear_list, {}, self.queue,
                                          console_printer=self.console_printer,
-                                         debug=False,
-                                         debug_bears=['TestTwoBear',
-                                                      'TestThreeBear'])
+                                         debug=False)
 
         self.assertEqual(len(list1), 0)
         self.assertEqual(len(list2), 1)
@@ -640,7 +635,7 @@ class ProcessingTest(unittest.TestCase):
             global_bear_list = [TestOneBear]
             instantiate_bears(self.sections['cli'], [], global_bear_list, {},
                               self.queue, console_printer=self.console_printer,
-                              debug=True, debug_bears=False)
+                              debug=True)
 
 
 class ProcessingTest_GetDefaultActions(unittest.TestCase):
--- a/coalib/bears/GlobalBear.py
+++ b/coalib/bears/GlobalBear.py
@@ -18,8 +18,7 @@ class GlobalBear(Bear):
                  file_dict,
                  section,
                  message_queue,
-                 timeout=0,
-                 debugger=False):
+                 timeout=0):
         """
         Constructs a new GlobalBear.
 
@@ -27,7 +26,7 @@ class GlobalBear(Bear):
 
         See :class:`coalib.bears.Bear` for other parameters.
         """
-        Bear.__init__(self, section, message_queue, timeout, debugger)
+        Bear.__init__(self, section, message_queue, timeout)
         self.file_dict = file_dict
 
     @staticmethod
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -297,8 +297,7 @@ def instantiate_bears(section,
                       file_dict,
                       message_queue,
                       console_printer,
-                      debug=False,
-                      debug_bears=False):
+                      debug=False):
     """
     Instantiates each bear with the arguments it needs.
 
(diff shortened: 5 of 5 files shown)


RaiVaibhav:
[WIP]

tests/processes/ProcessingTest.py
```diff
@@ -566,8 +567,8 @@ def test_yield_ignore_ranges(self):
     def test_loaded_bears_with_error_result(self):
         class BearWithMissingPrerequisites(Bear):
 
-            def __init__(self, section, queue, timeout=0.1, debugger=False):
-                Bear.__init__(self, section, queue, timeout, debugger)
+            def __init__(self, *args, **kwargs):
+                Bear.__init__(self, *args, **kwargs)
```
RaiVaibhav: these changes are already reflected in Profiler interface PR once these changes gets merge there I will rebase (the purpose of reflecting these changes here is preventing the coverage and test failure problem)

tests/processes/ProcessingTest.py
```diff
@@ -606,41 +607,36 @@ def check_prerequisites(cls):
     def test_global_instantiation(self):
         class TestOneBear(Bear):
 
-            def __init__(self, file_dict, section, queue, timeout=0.1,
-                         debugger=False):
+            def __init__(self, *args, **kwargs):
                 raise RuntimeError
 
-        class TestTwoBear(Bear):
+        class TestTwoBear(GlobalBear):
```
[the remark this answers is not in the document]
RaiVaibhav: I just made test more clear check the line `627` (just like a enhancement of test), it will also run fine without changing it to `GlobalBear`

coalib/bears/Bear.py
```diff
@@ -301,7 +302,12 @@ def __init__(self,
         self.section = section
         self.message_queue = message_queue
         self.timeout = timeout
-        self.debugger = debugger
+        self.debugger = (False if 'debug_bears' not in section or (
```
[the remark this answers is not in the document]
RaiVaibhav: @jayvdb I didn't get, you want me to make a different private method which do the same work?

>You will use it again for profile

I don't think so as `debug_bears` accept different argument then `profile`, If I try to mix both it will create a extra code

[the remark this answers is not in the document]
RaiVaibhav: Also IMO creating a helper will not help as both `debug_bears` and `profile` take different argument and accordingly logic change in both case.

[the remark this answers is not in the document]
jayvdb: the helper would take an argument.

If necessary create two helpers.

At the moment these are gibberish logic which I am assuming is similar because it isnt python code which is readable and intuitive.

tests/processes/ProcessingTest.py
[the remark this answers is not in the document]
jayvdb: If the test runs fine without changing it to `GlobalBear`, please do not change the tests.
Create an issue and explain why you think it should be changed.
And it would probably be a newcomer issue.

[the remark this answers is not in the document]
RaiVaibhav: Sorry for the confusion please nvm the above comment, that change was meaningfull actually `GlobalBear` gets one more arg `file_dict` compare to other Bears. so I changed that to `GlobalBear` [here](https://github.com/coala/coala/pull/5677/files#diff-7c16cc0dc5d9e615bbb54239955f7b5eR625)

coalib/bears/Bear.py
```diff
@@ -16,14 +16,31 @@
...
 from coalib.results.TextPosition import ZeroOffsetError
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
 from coalib.settings.ConfigurationGathering import get_config_directory
+from coalib.settings.Setting import typed_list
 
 from .meta import bearclass
 
 
+def _debugger_helper(bear):
```
jayvdb: what does this do?

the function name should describe it.

the docstring should explain it.

coalib/bears/Bear.py
```diff
@@ -16,14 +16,31 @@
...
+    if bear is None:
+        return False
+    elif 'debug_bears' not in bear.section or (
```
jayvdb: help the reader follow the logic.  one step at a time.
```py
if 'debug_bears' not in bear.section:
    return False
if ...:
    ....
```

coalib/bears/Bear.py
```diff
@@ -16,14 +16,31 @@
...
+            bear.section['debug_bears'].value.lower() in FALSE_STRINGS):
+        return False
+    elif bear.name.lower() in (
```
jayvdb: you do not need `elif` if the previous branch did a `return`.

avoiding the `elif` means you can also do some computation and use variables *before* the next `if`, instead of putting so much unreadable stuff into an `if` clause like you have done here

coalib/bears/Bear.py
```diff
@@ -16,14 +16,31 @@
...
+from coalib.settings.Setting import typed_list
 
 from .meta import bearclass
 
 
+def _debugger_helper(bear):
+    if bear is None:
+        return False
+    elif 'debug_bears' not in bear.section or (
+            bear.section['debug_bears'].value.lower() in FALSE_STRINGS):
```
jayvdb: `bool(bear.section['debug_bears'])`

RaiVaibhav: Please check I am following the previous instructions also for the value like `na` or `non` etc this will raise ValueError

jayvdb: Please test what I suggested before saying it will raise an exception .

And see `StringConverter.__bool__` for why it *should* work.
If it doesnt work, it is because there is something else wrong with the `debug_bears` Setting.

RaiVaibhav: Check this https://bpaste.net/show/1faff582d756

RaiVaibhav:
```
>>> section.append(Setting('abc', 'nop'))
>>> bool(section['abc'])
  File "/usr/local/lib/python3.6/dist-packages/coala_utils/string_processing/StringConverter.py", line 46, in __bool__
    raise ValueError
ValueError
>>> 
```

Makman2: Actually, I've seen that this is kinda buggy because we don't use coala's `FALSE_STRINGS`, but `coala_util`'s `FALSE_STRINGS`, which isn't so comprehensive than coala's list. However, @jayvdb is right to use `bool` on `Section`, it works in principle and that's what our section system uses.

coalib/bears/Bear.py
```diff
@@ -16,14 +16,31 @@
...
+        return False
+    elif bear.name.lower() in (
+            map(str.lower, typed_list(str)(bear.section['debug_bears']))) or (
+                typed_list(str)(bear.section['debug_bears'])[0].lower() in
+            TRUE_STRINGS):
+        return True
+    else:
+        return False
```
Makman2: you can return the result of `bear.name.lower() in ...` directly instead of using an if, it's like `if True: return True else: return False` now

RaiVaibhav:
@Makman2  can you take a look, please, I want this to be merge today so that I can update the Profiler Interface code asap.

coalib/bears/Bear.py
jayvdb: @Makman2 , so we need to enhance coala_util's `FALSE_STRINGS` ?  better create an issue quick as those libraries need to get released soon

coalib/bears/Bear.py
```diff
@@ -20,10 +20,29 @@
 from coalib.settings.FunctionMetadata import FunctionMetadata
 from coalib.settings.Section import Section
 from coalib.settings.ConfigurationGathering import get_config_directory
+from coalib.settings.Setting import typed_list
 
 from .meta import bearclass
 
 
+def _debugger_flag_setter(bear):
```
jayvdb: this is not a "setter".  It doesnt set anything.

It determines the value of the `debugger` *setting*.

coalib/bears/Bear.py
RaiVaibhav: Can you review this code :sweat_smile: 
I will create a issue after getting green from Makman

RaiVaibhav: If possible I will solve that :+1:

coalib/bears/Bear.py
```diff
@@ -20,10 +20,29 @@
...
+                 in ``debug_bears`` setting matched with the name of parameter
+                 bear.
+    """
+    if bear is None:
+        return False
+    if 'debug_bears' not in bear.section:
+        return False
+    return (bear.name.lower() in (
+            map(str.lower, typed_list(str)(bear.section['debug_bears']))) or
+            typed_list(bool)(bear.section['debug_bears'])[0])
```
jayvdb: what is `typed_list(bool)(bear.section['debug_bears'])[0]` ?

that is buggy as.  I know what you are trying to do.  You want to hide the ValueError exceptions which could happen.  But it only hides some of them, because that is not what `typed_list(bool)` is designed to do.

You need to let go of that horrible `or` you have in the `return`.  It is forcing you to write more complicated and buggy code.

`bool(bear.section['debug_bears'])` can raise exceptions.  So, you need to catch them.  We even have a MR related to it at https://gitlab.com/coala/coala-utils/merge_requests/34/diffs .

Then after you have finished checking if it is a bool, then do another check if it is a list of bear names.
And do *not* use typed_list(str).  `bear.section['debug_bears']` is already the appropriate type.  Read the documentation about that type for how to use it correctly.

RaiVaibhav: I didn't get this part
>And do not use typed_list(str). bear.section['debug_bears'] is already the appropriate type

Can you tell me how am I going to make this --> 'ABear, Bbear' to a list without using `typed_list`?

RaiVaibhav: >bool(bear.section['debug_bears']) can raise exceptions. So, you need to catch them. 

doen't user know that they have passed wrong value (`ValueError`)?

>do another check if it is a list of bear names.

There are cases and here cases are many like let suppose user passed `ABear, BBear` and If I do something like `bool(bear.section['debug_bears'])` it will give `ValueError`

coalib/bears/Bear.py
RaiVaibhav: >It determines the value of the debugger setting.

No its not this is not the way current code is working.

Does your lower comment of `typed_list` is because of the above confusion?

RaiVaibhav: Its only setting the `debugger` flag to `True` or `False` depending the cases and after that this falg is used to enable debugger.

coalib/bears/Bear.py
RaiVaibhav: After suggesting your changes this will be the code
```python
def _debugger_flag_setter(bear):
    """
    Set the value of `debugger` flag.

    :param bear: Bear object
    :return:     True if ``debug_bears`` is ``True`` or if bear name specified
                 in ``debug_bears`` setting matched with the name of parameter
                 bear.
    """
    if bear is None:
        return False
    if 'debug_bears' not in bear.section:
        return False
    try:
        return bool(bear.section['debug_bears'])
    except ValueError:
        return bear.name.lower() in (
                map(str.lower, typed_list(str)(bear.section['debug_bears'])))
```
And I still don't know why you said to not use `typed_list(str)`

jayvdb: Read `Setting` and `StringConverter`; docs, code, whatever.

Do *not* use `typed_list(str)`!  It is that simple.  Any attempt to use that is not understanding what a `Setting` is, and is obfusicating the code!  It is almost like going `str(bool(None))`.

And for goodness sakes, never do complicated code in a exception handler.  This is bug heaven:
```py
    except ValueError:
        return bear.name.lower() in (
                map(str.lower, typed_list(str)(bear.section['debug_bears'])))
```
Code should tell a story.  Someone else should be able to read it and understand it almost immediately.  If they cant, it needs to be rewritten.

Please do not reply further.  You've been told what not to do, and to lay out your code better.  Submit a patch which does what was requested.  This isnt a debate.

RaiVaibhav: Sorry, does this code is acceptable?
```python
def _debugger_flag_value(bear):
    """
    Value of `debugger` flag.

    :param bear: Bear object
    :return:     True if ``debug_bears`` is ``True`` or if bear name specified
                 in ``debug_bears`` setting matched with the name of parameter
                 bear.
    """
    if bear is None:
        return False
    if 'debug_bears' not in bear.section:
        return False
    try:
        return bool(bear.section['debug_bears'])
    except ValueError:
        pass
    return bear.name.lower() in (
            map(str.lower, list(bear.section['debug_bears'])))

```

jayvdb: im not reviewing code in comments that hasnt had CI check it and coverage fixed.

coalib/bears/Bear.py
```diff
@@ -24,6 +24,26 @@
...
+    if bear is None:
+        return False
+    if 'debug_bears' not in bear.section:
+        return False
+    try:
+        return bool(bear.section['debug_bears'])
+    except ValueError:
+        pass
+    return bear.name.lower() in (
+            map(str.lower, list(bear.section['debug_bears'])))
```
Makman2: Is using `list` necessary? `map` tries to create an iterator as well out of `bear.section[...]` if I'm not mistaken

coalib/bears/Bear.py
```diff
@@ -24,6 +24,26 @@
...
+    Determines the value of ``debugger`` setting.
+
+    :param bear: Bear object.
+    :return:     True if ``debug_bears`` is ``True`` or if bear name specified
+                 in ``debug_bears`` setting match with the bear parameter.
+    """
+    if bear is None:
+        return False
+    if 'debug_bears' not in bear.section:
+        return False
```
Makman2: Can be merged into a single `if`-condition ;)

coalib/bears/Bear.py
Makman2: Hm maybe catch here a `ValueError` as well, and emit a better descriptive message, since bools and string-lists are allowed in that setting. What do you think?
-->
```python
try:
    return bear.name.lower() in ...
except ValueError:
    raise ValueError('could not parse setting "debug_bears" into bool or string-list')
```

coalib/bears/Bear.py
```diff
@@ -24,6 +24,26 @@
 from .meta import bearclass
 
 
+def _debugger_setting_value(bear):
```
Makman2: Would change the name, it's not descriptive enough.
So what does this function do? You want to check whether the bear is in debug mode according to its section-settings. So a name like `_is_debugged` or `_is_debug_mode_enabled` or similar might fit better^^

coalib/bears/Bear.py
```diff
@@ -24,6 +24,26 @@
...
+    """
+    Determines the value of ``debugger`` setting.
+
+    :param bear: Bear object.
+    :return:     True if ``debug_bears`` is ``True`` or if bear name specified
+                 in ``debug_bears`` setting match with the bear parameter.
+    """
+    if bear is None:
+        return False
```
Makman2: Is that actually even valid passing `None` for a bear? We are determining if nothing is being debugged or not? :P

coalib/bears/Bear.py
```diff
@@ -24,6 +24,26 @@
 from .meta import bearclass
 
 
+def _debugger_setting_value(bear):
+    """
+    Determines the value of ``debugger`` setting.
```
Makman2: See above, could get a little update ;)
> So what does this function do? You want to check whether the bear is in debug mode according to its section-settings. So a name like _is_debugged or _is_debug_mode_enabled or similar might fit better^^

coalib/bears/Bear.py
RaiVaibhav: You want me to create something like 
```python
try:
    return (bear.name.lower() in (
            map(str.lower, list(bear.section['debug_bears']))) or bool(bear.section['debug_bears']))
except ValueError:
    raise ValueError('could not parse setting "debug_bears" into bool or string-list')
```
then John already said not to do so, direclty said to only follow the the things.

coalib/bears/Bear.py
RaiVaibhav: Which two condition you want me to merge if above you said `if bear is None ` is not a valid condition (I will raise a value error in replacement)

coalib/bears/Bear.py
RaiVaibhav: you want me to do somehting like this ?
```python
    try:
        return (bear.name.lower() in (
                map(str.lower, list(bear.section['debug_bears']))) or
                    bool(bear.section['debug_bears']))
    except ValueError:
        raise ValueError('could not parse setting "debug_bears" into bool or '
                         'string-list')
```

RaiVaibhav: I have updated till now what I have got :+1:

coalib/bears/Bear.py
RaiVaibhav: I have updated till now what I have got

tests/bears/BearTest.py
```diff
@@ -518,6 +521,24 @@ def test_do_settings(self):
...
+        self.assertEqual(_is_debugged(uut), False)
+        section.append(Setting('debug_bears', 'tRuE'))
+        self.assertEqual(_is_debugged(uut), True)
+        section.append(Setting('debug_bears', '0'))
+        self.assertEqual(_is_debugged(uut), False)
+        section.append(Setting('debug_bears', 'Bear, ABear'))
+        self.assertEqual(_is_debugged(uut), True)
+
+        section.append(Setting('debug_bears', 'abc, xyz'))
+        with self.assertRaises(ValueError):
```
jayvdb: why is 'abc, xyz' a ValueError ?

that looks like a valid list of strings.

RaiVaibhav: Yeah it is valid list of string but it is not valid valid for `debug_bears` setting.

RaiVaibhav: If I understand correctly this is what Makman is saying ?https://github.com/coala/coala/pull/5677#discussion_r206652926

RaiVaibhav: Or I made a mistake IMO, this is what Makman wanted?
```python
    try:
        return bool(bear.section['debug_bears'])
    except ValueError:
        pass
    try:
        return (bear.name.lower() in (
                map(str.lower, bear.section['debug_bears']))
    except ValueError:
        raise ValueError('could not parse setting "debug_bears" into bool or '
                         'string-list')
```

RaiVaibhav: Yeah it will never through a `ValueError`

jayvdb: Of course he doesnt want ValueError's thrown when there isnt a `ValueError`.
You merged the wrong two conditions, making the `bool()` one be caught and re-raised when it shouldnt have been re-raised.

coalib/bears/Bear.py
RaiVaibhav: I made a silly mistake and I will take care of that next time for further update of this reply check [here](https://github.com/coala/coala/pull/5677#discussion_r206724245)

coalib/bears/Bear.py
```diff
@@ -290,6 +310,7 @@ def __init__(self,
         :param message_queue: The queue object for messages. Can be ``None``.
         :param timeout:       The time the bear is allowed to run. To set no
                               time limit, use 0.
+        :param debugger:      If ``True``, a debugger is started.
```
jayvdb: what is this doing here?

tests/processes/ProcessingTest.py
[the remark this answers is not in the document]
jayvdb: I still dont get this.

You didnt change that in c1184ca58 .

So it isnt being 'undone' in this cleanup patch.

Is it 100% necessary?
Or are you doing an unrelated change?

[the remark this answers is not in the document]
jayvdb: My guess is you need to revert to using `Bear.__init__(self, section, queue, timeout, debugger)` like before c1184ca

jayvdb:
ack https://github.com/coala/coala/pull/5677/commits/3cb58246351ecbabd9768151cb6b9fe28c3790fc

jayvdb:
@gitmate-bot ff

PR #5690, opened 2018-08-05
Section Tags Filter

ksdme:
PTAL [cEP-0028](https://github.com/coala/cEPs/pull/171)

### DIFF
--- a/coalib/parsing/filters/decorators.py
+++ b/coalib/parsing/filters/decorators.py
@@ -35,7 +35,7 @@ def typed_filter(type_classes, msg=None):
                     .format(filter=filter.__name__,
                             type_name=tuple(type_classes)))
 
-            return filter(thing, *args, **kargs)
+            return filter(obj, *args, **kargs)
 
         return decorated_filter
 
--- a/coalib/parsing/filters/SectionTagsFilter.py
+++ b/coalib/parsing/filters/SectionTagsFilter.py
@@ -1,0 +1,31 @@
+from coalib.parsing.filters.decorators import typed_filter
+
+
+@typed_filter(('bearclass', 'Bear', 'Section'))
+def section_tags_filter(section_or_bear, args):
+    """
+    Filters the bears or sections by ``tags``.
+
+    :param section_or_bear: A section or bear instance on which filtering
+                            needs to be carried out.
+    :param args:            Set of tags on which it needs to be filtered.
+    :return:                ``True`` if this instance matches the criteria
+                            inside args, ``False`` otherwise.
+    """
+    enabled_tags = list(map(str.lower, args))
+    if len(enabled_tags) == 0:
+        return True
+
+    section = section_or_bear
+    if hasattr(section_or_bear, 'section'):
+        # If it is a bear or bear like object it has
+        # to have an associated section which contains
+        # all settings, including tags.
+        section = section_or_bear.section
+
+    section_tags = section.get('tags', False)
+    if str(section_tags) == 'False':
+        return False
+
+    section_tags = map(str.lower, section_tags)
+    return bool(set(section_tags) & set(enabled_tags))
--- a/coalib/parsing/filters/__init__.py
+++ b/coalib/parsing/filters/__init__.py
@@ -5,9 +5,11 @@ get a list of bears with specific properties.
 from .LanguageFilter import language_filter
 from .CanDetectFilter import can_detect_filter
 from .CanFixFilter import can_fix_filter
+from .SectionTagsFilter import section_tags_filter
 
 
 available_filters = {'language': language_filter,
                      'can_detect': can_detect_filter,
                      'can_fix': can_fix_filter,
+                     'section_tags': section_tags_filter,
                      }
--- a/tests/coalaTest.py
+++ b/tests/coalaTest.py
@@ -206,6 +206,51 @@ class coalaTest(unittest.TestCase):
             self.assertEqual(len(stdout.splitlines()), 2)
             self.assertFalse(stderr)
 
+    def test_execute_with_bad_filters(self, debug=False):
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (lines, filename):
+            retval, stdout, stderr = execute_coala(
+                coala.main, 'coala', '--filter-by', 'language', 'python',
+                '-f', filename, '-b', 'TestBear', '--no-color', '-I',
+                debug=debug)
+
+            self.assertIn(
+                "'language_filter' can only handle ('bearclass',). "
+                'The context of your usage might be wrong.', stdout)
+            # Calling without config, hence 0
+            # else, it'll be 1
+            self.assertEqual(retval, 0)
+
+    def test_execute_with_bad_filters_debug(self):
+        self.test_execute_with_bad_filters(True)
+
+    def test_execute_with_filters(self, debug=False):
+        coala_config = ('[section_one]',
+                        'tags = save',
+                        '[section_two]',
+                        'tags = change',)
+
+        with bear_test_module(), \
+                prepare_file(['#fixme'], None) as (_, filename), \
+                prepare_file(coala_config, None) as (_, configuration):
+            results, retval, _ = run_coala(
+                                    console_printer=ConsolePrinter(),
+                                    log_printer=LogPrinter(),
+                                    arg_list=(
+                                        '-c', configuration,
+                                        '-f', filename,
+                                        '-b', 'TestBear',
+                                        '--filter-by', 'section_tags',
+                                        'save'
+                                    ),
+                                    autoapply=False,
+                                    debug=debug)
+
+            self.assertTrue('section_one' in results)
+
+    def test_execute_with_filters_debug(self):
+        self.test_execute_with_filters(True)
+
     def test_show_capabilities_with_supported_language_debug(self):
         self.test_show_capabilities_with_supported_language(debug=True)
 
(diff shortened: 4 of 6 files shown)


coalib/parsing/DefaultArgParser.py
```diff
@@ -177,6 +177,10 @@ def default_arg_parser(formatter_class=None):
         '-d', '--bear-dirs', type=PathArg, nargs='+', metavar='DIR',
         help='additional directories which may contain bears')
 
+    inputs_group.add_argument(
```
jayvdb: This belongs before --bear-dirs , probably after --files .

Lets wait for more feedback on position

coalib/parsing/FilterHelper.py
```diff
@@ -49,6 +55,37 @@ def apply_filter(filter_name, filter_args, all_bears=None):
...
+    """
+    Returns sections after filtering based on ``filter_args``. It returns
+    all sections if nothing is present in ``filter_args``.
+
+    :param filter_name:
+        Name of the filter.
+    :param filter_args:
+        Arguments of the filter to be passed in.
+        For example:
+        ``['tags', ('java',)]``
```
jayvdb: please use exampes from your cEP, and your own real needs.

this looks like language , not a tag which should be an arbitrary grouping which couldnt be achieved via any other means.

coalib/parsing/FilterHelper.py
```diff
@@ -68,3 +105,53 @@ def apply_filters(filters, bears=None):
...
+def apply_section_filters(filters, sections):
+    """
+    Returns sections after filtering based on ``filters``. It returns
+    intersection of sections if more than one element is present in ``filters``
+    list.
+
+    :param filters:
+        List of args based on ``sections`` has to be filtered.
+        For example:
+        ``[['tags', ('save', 'change')], ['exclude', ('css', '!python')]]``
```
jayvdb: again, dont use tags that are a language .  it is confusing.

tests/settings/ConfigurationGatheringTest.py
```diff
@@ -198,12 +198,16 @@ def test_merge_defaults(self):
                 gather_configuration(lambda *args: True,
                                      self.log_printer,
                                      arg_list=['-S', 'value=1', 'test.value=2',
-                                               '-c', escape(temporary, '\\')] +
+                                               '--tags', 'save', '-c',
```
jayvdb: move the `'-c', ` back to the next line, so you dont modify that line.

it is also confusing to have the `'-c', ` on this line when its arg is on the next line.

coalib/parsing/FilterHelper.py
ksdme: I used them as examples, but I will update it.

coalib/parsing/DefaultArgParser.py
ksdme: Currently,
```
Inputs:
  -b, --bears           names of bears to use
  -f, --files           files that should be checked
  -i, --ignore          files that should be ignored
  --limit-files         filter the `--files` argument's matches further
  -d, --bear-dirs       additional directories which may contain bears
  -t, --tags            tags of sections to execute
```

I think `tags` should definitely be above `bear-dirs`.

coalib/parsing/FilterHelper.py
```diff
@@ -1,11 +1,17 @@
...
+from coalib.parsing.filters import (
+    available_filters, available_section_filters)
+from coalib.parsing.DefaultArgParser import default_arg_parser
 
 
 def is_valid_filter(filter):
     return filter in available_filters
 
 
+def is_valid_section_filter(filter):
```
jayvdb: nope.

lets improve the filters system to support sections.

`def foo_filter(thing, args):`

every filter than checks whether is knows what is in `thing`, e.g. using isinstance , and raises NotImplementedError if it doesnt get it.  Existing filters only know about `Bear`.

You make a small chance to the existing filters to make them compliant with the new API by adding a `raise NotImplementedError` to them.

(Actually most of them understand could easily be improved to understand `Section`, but that is beyond the scope of your change, and each will need to be carefully done wrt the CLI args to ensure we build a good user experience)

coalib/parsing/FilterHelper.py
```diff
@@ -49,6 +55,37 @@ def apply_filter(filter_name, filter_args, all_bears=None):
...
+        List of sections on which filter is to be applied.
+    :return:
+        Filtered sections based on a single section filter.
+    """
+    if not is_valid_section_filter(filter_name):
+        raise InvalidFilterException(filter_name)
+    if not filter_args or len(filter_args) == 0:
+        return all_sections
+
+    filter_function = available_section_filters[filter_name]
```
jayvdb: not needed.  below you would catch NotImplementedError and ignore it.

coalib/parsing/filters/TagSectionFilter.py
```diff
@@ -0,0 +1,21 @@
+def tag_section_filter(section, args):
+    """
+    Filters the sections by ``tags``. ``cli`` is always accepted.
+
+    :param section: Section object.
+    :param args:    Set of tags on which ``section`` is to be filtered.
+    :return:        ``True`` if this section matches the criteria inside args,
+                    ``False`` otherwise.
+    """
+    if section.name == 'cli':
```
jayvdb: I agree with @shreyans800755 that this doesnt belong in the filter.
https://github.com/ksdme/coala/commit/f8fc69c533333d73f891115edd4be14b1d3da458#r29964403

The `cli` section should be kept hidden inside `coalib/settings`, as a special case implementation detail.  It annoys me that we have any UI messages that mention this section -- the user cant see it, so they shouldnt hear about it.

And likewise developers in other parts of coalib shouldnt need to deal with that section.

coalib/coala_main.py
```diff
@@ -159,6 +161,15 @@ def run_coala(console_printer=None,
                 (section_name, sections[section_name])
                 for section_name in targets)
 
+        # Collect all the section filters which are based on args,
```
gaocegege: I think we could remove the last `,`

ksdme:
Updated to use `--filter-by` for both section filters and bear filters. Known issues:
- Requires update to `--filter-by` documentation.

coalib/parsing/FilterHelper.py
```diff
@@ -49,22 +50,83 @@ def apply_filter(filter_name, filter_args, all_bears=None):
...
+        ['section_tags', 'save']]``
     :param bears:
         The bears to filter.
+    :param sections:
+        The sections to filter.
     :return:
-        Filtered bears.
+        Filtered bears or sections.
     """
+    things = bears
```
shreyans800755: Let's call it items rather than things

coalib/parsing/FilterHelper.py
```diff
@@ -49,22 +50,83 @@ def apply_filter(filter_name, filter_args, all_bears=None):
...
+        Parsed CLI args using which the filters are to be collected.
+    :param arg_list:
+        The CLI argument list.
+    :param arg_parser:
+        Instance of ArgParser that is used to parse arg list.
+    :return:
+        List of filters in standard filter format, i.e
+        ``[['filter_name', 'arg1', 'arg2']]``.
+    """
+    if args is None:
```
shreyans800755: Do we have any practical case for this condition ?
If so, is there way we can check this before calling this function ?

shreyans800755: Because this is not the job of this function. This function is just a getter

coalib/parsing/filters/decorators.py
```diff
@@ -0,0 +1,42 @@
...
+            thing_class = thing.__class__
+            all_bases = list(map(lambda klass: klass.__name__,
+                                 getmro(thing_class)))
+
+            for type_class in type_classes:
+                if str(type_class) in all_bases:
+                    break
+            else:
+                raise NotImplementedError(
+                    msg or '\'{filter}\' filter can only handle {type_name}. '
```
jayvdb: it was going so well until you mentioned filter here.

the thing you are wrapping has a name, which you are using here.

This decorator is a developer helper ... to catch when they have done the wrong thing.  it shouldnt give pretty names ... the caller needs to detect this and make it user-friendly.

Why should a `NotImplementedError` say "language_filter filter" .. ?  `language_filter` is better, and the backtrace is even better, so no need to tidy it up.

To help, create a subclass of `NotImplementedError` and stuff into it extra attributes to record what went wrong, such as the callee, expected classes, and the incorrect object which was encountered.   Then whoever is catching it can understand what went wrong and create a user-friendly message.

coalib/parsing/filters/TagSectionFilter.py
```diff
@@ -0,0 +1,31 @@
+from coalib.parsing.filters.decorators import typed_filter
+
+
+@typed_filter(('bearclass', 'Bear', 'Section'))
+def tag_section_filter(section_or_bear, args):
```
jayvdb: `section_tag_filter`

coalib/parsing/FilterHelper.py
ksdme:
```
run_coala

:param arg_parser:        Instance of ArgParser that is used to non-setting arguments.
:param arg_list:              The CLI argument list.
:param args:                   Alternative pre-parsed CLI arguments.
```

This is not just about the tests, since args to `run_coala()` can be optionally passed instead of arg_list, I think we need to merge opts from both, so as to enable the internal usage of this filter. Tests also use can use this by directly calling `run_coala()` with an arg list.

ksdme:
```
run_coala

:param arg_parser:        Instance of ArgParser that is used to non-setting arguments.
:param arg_list:              The CLI argument list.
:param args:                   Alternative pre-parsed CLI arguments.
```

This is not just about the tests, since args to `run_coala()` can be optionally passed instead of arg_list, I think we need to merge opts from both, so as to enable the internal usage of this filter. Tests also use can use this by directly calling `run_coala()` with an arg list.

PR #5699, opened 2018-08-08
Add Support for FileProxies

ksdme:
Adds support for FileProxies and analysis of in-memory files.

Known Shortcomings:
- Cannot yet work with apply-patch actions.
- File Proxies cannot yet work with caching disabled.

### DIFF
--- a/coalib/io/FileProxy.py
+++ b/coalib/io/FileProxy.py
@@ -1,0 +1,358 @@
+import logging
+from os import path
+
+from coala_utils.FileUtils import detect_encoding
+from coala_utils.decorators import (enforce_signature, generate_eq)
+
+
+@generate_eq('filename')
+class FileProxy:
+    """
+    ``FileProxy`` is responsible for providing access to
+    contents of files and also provides methods to update
+    the in memory content register of the said file.
+
+    >>> import logging
+    >>> import tempfile
+    >>> file = tempfile.NamedTemporaryFile(delete=False)
+    >>> file.write(b'bears')
+    5
+    >>> file.close()
+
+    To create a new file proxy instance of a file:
+
+    >>> proxy = FileProxy(file.name, None, 'coala')
+    >>> proxy.contents()
+    'coala'
+
+    >>> proxy.lines()
+    ('coala',)
+
+
+    You can replace the file contents in-memory by using
+    the replace method on a file proxy. Version tracking
+    is a simple way FileProxy provides to handle external
+    incremental updates to the contents.
+
+    >>> proxy.replace('coala-update', 1)
+    True
+    >>> proxy.contents()
+    'coala-update'
+    >>> proxy.version
+    1
+
+    File Proxy instances can also be initialized from files
+    using FileProxy.from_file(). Binary files are also
+    supported using from_file.
+
+    >>> proxy2 = FileProxy.from_file(file.name, None)
+    >>> proxy2.contents()
+    'bears'
+    """
+
+    def __init__(self, filename, workspace=None, contents=''):
+        """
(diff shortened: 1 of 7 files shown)


coalib/misc/Caching.py
```diff
@@ -173,3 +178,137 @@ def get_uncached_files(self, files):
...
+                                'non-unicode characters. Leaving it out.'
+                                .format(filename))
+
+            except (OSError, FileNameError) as exception:
+                log_exception("Failed to read file '{}' because of an unknown "
+                              'error. Leaving it out.'.format(filename),
+                              exception,
+                              log_level=LOG_LEVEL.WARNING)
+
+        logging.debug('Files that will be checked:\n' +
```
jayvdb: this could be moved to the common path in `instantiate_processes` after the dict has been obtained.

coalib/io/FileProxy.py
```diff
@@ -0,0 +1,363 @@
+import logging
+from os import path
+
+from coala_utils.FileUtils import detect_encoding
+from coala_utils.decorators import (enforce_signature, generate_eq,)
+
+
+class FileNameError(LookupError):
+    def __init__(self, error_message):
```
jayvdb: you sure you need this?

it isnt doing anything afaics

coalib/misc/Caching.py
ksdme: I think we need to maintain compatibility with `FileDictFileCache`, since it uses the default implementation of `get_file_dict()` it logs locally as soon as files are collected. Moving this to `instantiate_processes()` will cause logging twice for FileDictFileCache. I think we can have a separate issue for this once this gets merged.

coalib/io/FileProxy.py
ksdme: I wanted to have a distinct error to denote bad file names. I will use `ValueError` instead.

coalib/misc/Caching.py
jayvdb: I mean that `get_file_dict` should also not log this message.

`instantiate_processes` can do that for all implementations.

coalib/io/FileProxy.py
```diff
@@ -0,0 +1,358 @@
+import logging
+from os import path
+
+from coala_utils.FileUtils import detect_encoding
+from coala_utils.decorators import (enforce_signature, generate_eq,)
```
jayvdb: no trailing `,` if followed by `)`

tests/misc/CachingTest.py
```diff
@@ -177,3 +182,102 @@ def test_caching_multi_results(self):
...
+        with bear_test_module(), \
+                prepare_file(['disk-copy\n'], None) as (_, filename):
+
+            memory_data = 'in-memory\n'
+            proxy = FileProxy(filename, None, memory_data)
+            proxymap = FileProxyMap([proxy])
+            self.cache.set_proxymap(proxymap)
+
+            results, exitcode, file_dicts = run_coala(
+                                    console_printer=ConsolePrinter(),
```
jayvdb: strange indentation.

four space indent, from beginning of previous line.

jayvdb:
ack https://github.com/coala/coala/pull/5699/commits/97a1f6c2671d58c3ee7f49ec20e15d583d9ddc4f https://github.com/coala/coala/pull/5699/commits/013b3f80e092c47802fe4e8ad0f8e29ee2fe2f1e https://github.com/coala/coala/pull/5699/commits/59f78c1114f54a4a5b6a8416ebcd6d4d57edb17f

jayvdb:
@gitmate-bot ff

Issue #5683, opened 2018-07-31
deprecate_settings to support splitting a value to two new args


jayvdb:
I would like to split the `checkstyle_configs` arg into two args, as they are two different data-types.

It seems `@deprecate_settings` and `FunctionMetadata` cant support this, as they expect a deprecated arg to only be used once.

Here is a more detailed look at what I tried to do, and failed.

```py

def get_name(value):
    if value in known_checkstyles.keys():
        return value

def get_path(value):
    if value not in known_checkstyles.keys():
        return value

class CheckstyleBear:
    ...

    @deprecate_settings(
        java_style_name=('checkstyle_configs',
                         get_name),
        checkstyle_config=('checkstyle_configs',
                           get_path),
    )
    def create_arguments(
            self, filename, file, config_file,
            checkstyle_config: path = None,
            java_style_name: CheckstyleStyle = 'google',
            ...
```

PR #3204, opened 2016-12-10
Writing_Tests: Correct "Executing Tests" reference

sils:
### Checklist

- [ ] I have rebased properly. Please see [our tutorial on rebasing](http://coala.readthedocs.io/en/latest/Developers/Git_Basics.html#rebasing).
- [ ] I have gone through the [commit guidelines](http://coala.readthedocs.io/en/latest/Developers/Writing_Good_Commits.html) and I've followed them.
- [ ] I have followed the [guidelines for the commit shortlog](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#shortlog).
- [ ] I have followed the [guidelines for the commit body](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#commit-body).
- [ ] I have included the issue URL with the 'Fixes' keyword if the commit fixes a *real bug* and the 'Closes' keyword if it adds a feature or enhancement. For details, check the [issue reference section in our commit guidelines](http://docs.coala.io/en/latest/Developers/Writing_Good_Commits.html#issue-reference).
- [ ] I have [run all the tests](http://coala.readthedocs.io/en/latest/Developers/Executing_Tests.html) and they all pass. If any CI (below) is not green, please fix them. If GitMate issues appear you will have to amend your commits, pushing new commits on top is not sufficient!

### Reviewers

Fixes https://github.com/coala/coala/issues/3199

PR #5701, opened 2018-08-10
deprecate_settings: Support splitting setting

nalinbhardwaj:
Adds support for splitting deprecated arg into multiple new args.
Adds test for it as well.

Closes https://github.com/coala/coala/issues/5683

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/tests/bearlib/DeprecateSettingsTest.py
+++ b/tests/bearlib/DeprecateSettingsTest.py
@@ -1,3 +1,4 @@
+import logging
 import unittest
 
 from coalib.bearlib import deprecate_settings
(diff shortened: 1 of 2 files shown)


tests/bearlib/DeprecateSettingsTest.py
```diff
@@ -10,8 +11,22 @@ def func(new):
...
                          'This docstring will not be lost.')
+
+    def test_multiple_args(self):
+        logger = logging.getLogger()
+
+        with self.assertLogs(logger, 'WARNING') as cm:
+            self.assertEqual(func_2(a=1), 5)
+
+            self.assertRegex(cm.output[0], 'WARNING:root:The setting `a` is '
+                             r'deprecated. Please use `[xy]` instead.')
```
jayvdb: > Please use `[xy]` instead

This needs to be improved.

And we also need a test to cover the other scenario:

> 'The value of `{}` and `{}` are conflicting. `{}` will be used instead.

nalinbhardwaj: > > Please use [xy] instead
>
> This needs to be improved.

Not sure if you/I understand correctly, but this is a regex, so the actual error is "Please use `x` instead" or "Please use `y` instead". The reason this needs a regex is that the order of args seems to depend on the version of the py(py3.4 has the message with y, py3.5 with x).

nalinbhardwaj: > And we also need a test to cover the other scenario:
>
> > 'The value of {} and {} are conflicting. {} will be used instead.

Added.

nalinbhardwaj: No, there's only a single error message. This seems to be a deliberate choice in #3204.

jayvdb:
ack https://github.com/coala/coala/pull/5701/commits/bac641f76bd516823d2d7fcd1a5425fd91c5894f

jayvdb:
reack https://github.com/coala/coala/pull/5701/commits/7122a125a38da442d77903fc0d45a21ce42c2af5

jayvdb:
@gitmate-bot ff

Issue #5739, opened 2018-08-30
Add language for scss


jayvdb:
Like coalib/bearlib/languages/definitions/CSS.py

PR #5748, opened 2018-09-03
SCSS.py: Add SCSS.py file

rishabhgarg25699:
Add a new python file in coalib/bearlib/languages/definitions/SCSS.py

Closes https://github.com/coala/coala/issues/5739

### DIFF
--- a/coalib/bearlib/languages/definitions/SCSS.py
+++ b/coalib/bearlib/languages/definitions/SCSS.py
@@ -1,0 +1,11 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class SCSS:
+    extensions = '.scss',
+    comment_delimiters = '//',
+    multiline_comment_delimiters = {'/*': '*/'}
+    encapsulators = {'(': ')', '[': ']'}
+    versions = {3.1, 3.2, 3.3, 3.4, 3.5, 4.0}
+    interpolation = {'{': '}'}
--- a/tests/TestUtilities.py
+++ b/tests/TestUtilities.py
@@ -57,6 +57,7 @@ LANGUAGE_NAMES = [
     'Python 2.7, 3.3, 3.4, 3.5, 3.6',
     'Ruby',
     'Scala',
+    'SCSS 3.1, 3.2, 3.3, 3.4, 3.5, 4.0',
     'Shell',
     'Swift',
     'Tcl',


coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,48 @@
...
+}
+encapsulators {
+	= {
+		'(': ')', '[': ']';
+	}
+}
+string_delimiter_escape {
+	= {
+		'"': '\\"', "'": "\\'";
+	}
```
shikharvaish28: You need to change the formatting of the whole file. It's not as we expect it to be.

shikharvaish28: This file is not in python at all

shashwatmishra76:
@rishabhgarg25699 You are getting an assertion error under the method 
def test_get_all_languages_with_unknown(self) .It requires you to add 'SCSS 3.1, 3.2, 3.3, 3.4, 3.5, 4.0' to LANGUAGE_NAMES list.

.travis.yml
```diff
@@ -12,15 +12,19 @@ stages:
   - name: unsupported
     if: branch = master AND type = push
 
+.disable_global: &disable_global
```
li-boxuan: Remove those changes which are unrelated to your pr.

coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,11 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class SCSS:
+    extensions = '.scss'
+    comment_delimiters = '//'
```
li-boxuan: should be a tuple

coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,11 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class SCSS:
+    extensions = '.scss'
```
li-boxuan: should be a tuple

.travis.yml
li-boxuan: You didn't do what I said, and you are adding more irrelevant changes :(

I am pretty sure you are not using git command correctly

Fix the problem first. Next time before you push, use `gitk` or `gitg` tools, or use `git diff head~1` to check the changes within your commit.

coalib/bearlib/languages/definitions/SCSS.py
li-boxuan: This problem is still not fixed.

coalib/bearlib/languages/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .Language import Languages
 
 from .definitions.Unknown import Unknown
+from .definitions.Bash import Bash
```
shikharvaish28: Why are you adding ```Bash``` here? I don't think it's relevant here. IMO it's from  d5c0b79  which was earlier into this PR. Please check this.

tests/TestUtilities.py
```diff
@@ -33,6 +33,7 @@
 
 # This list is sorted alphabetically
 LANGUAGE_NAMES = [
+    'Bash',
```
shikharvaish28: Same Here. ```Bash``` is not needed.

coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,11 @@
+from coalib.bearlib.languages.Language import Language
+
+
```
shikharvaish28: Blank space not needed.

tests/TestUtilities.py
```diff
@@ -46,6 +47,7 @@
     'JavaScript Object Notation',
     'JavaServer Pages',
     'Jinja2',
+    'KornShell',
```
shikharvaish28: Not needed.

coalib/bearlib/languages/__init__.py
rishabhgarg25699: It is done by me. When i sync my local repo with remote they are added automatically.
Even I did not know the meaning of bash then when i added this in this file.
And also now you open the coalib/bearlib/languages/__init__.py on github of coala, these all are already there. https://github.com/coala/coala/blob/master/coalib/bearlib/languages/__init__.py  SEE IN THIS LINK BASH IS ALREADY THERE.

rishabhgarg25699: I think that the github is comparing with old coala files.

coalib/bearlib/languages/definitions/SCSS.py
rishabhgarg25699: The gap is there in all the files. So i put this gap

tests/TestUtilities.py
rishabhgarg25699: SEE THIS IS ALREADY PRESENT https://github.com/coala/coala/blob/master/tests/TestUtilities.py
It is comparing wrong  PLEASE SEE https://github.com/coala/coala/pull/5748/commits/01a27859071b51ffcd783eb3b283564b8f130bdb

coalib/bearlib/languages/definitions/SCSS.py
shikharvaish28: Since it's not needed, you will have to remove it.

tests/TestUtilities.py
shikharvaish28: It's because you've used wrong ```git``` commands. 
It looks like that you've messed up your branch, I would suggest you to reset to the old commit and then make the changes once again.

coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,11 @@
...
+@Language
+class SCSS:
+    extensions = ('.scss')
```
li-boxuan: This is wrong. `('.scss')` -> `'.scss',`

coalib/bearlib/languages/definitions/SCSS.py
```diff
@@ -0,0 +1,11 @@
...
+    comment_delimiters = ('//')
```
li-boxuan: Same as above

jayvdb:
@gitmate-bot rebase

jayvdb:
@gitmate-bot ff

jayvdb:
ack https://github.com/coala/coala/pull/5748/commits/d8eccb4e8026c75940f480cbea8f54b17d0307e6

Issue #5330, opened 2018-04-07
Add language definition for Haskell


jayvdb:
https://github.com/coala/coala/tree/master/coalib/bearlib/languages/definitions has no Haskell.
:-(

li-boxuan:
@jayvdb I'm willing to solve this.

juliandorville:
Hi,
I've been looking into contributing to Coala and reviewed the newcomers' guide (so I know to target issues labeled with a newcomer difficulty), and was wondering why this issue was labeled as medium. I think it seems like a pretty simple fix that I could handle, but maybe I'm missing something?
Thanks!

jayvdb:
@juliandorville , a language definition can range from a very basic couple of lines, to being quite detailed like the Python one.

And tests will be required, and those tests would be beyond the typical newcomer.

https://github.com/coala/coala/pull/5428 is a recent example of a language definition modification.

Anyway, an invite to coala org has been sent to you.  Once you accept that, we can assign this issue, and we go from there.  The fact that you are finding a Haskell issue and want to solve it suggests that our newcomer process is not ideal for you.

shashwatmishra76:
@jayvdb I want to try my hands with this one.I guess this issue requires me to add tests for Haskell.Can I be assigned to this?

jayvdb:
Yes some tests would be good.  I've assigned it to you @rishabhgarg25699

Issue #1690, opened 2016-02-24
handling languages inside other languages


AbdealiJK:
If I have a php file, it is inevitable that I'd probably have chunks of html inside a few of them. We probably want some method of finding specific languages inside other languages ?
Another example is awk inside bash
JSON inside Javascript, etc

jack17529:
@AbdealiJK  can I please know what is required to deal with this issue and how to proceed ? So that I can make sure to assign me this.

meetmangukiya:
@jack17529 this is one of the gsoc projects, so definitely this is something that can't be done soonish and also cannot be assigned

jack17529:
@meetmangukiya  then which issues to solve to get selected for GSOC 2017 as a super strong candidate.

lalitmee:
@AbdealiJK I think this is an exciting project to work on. Since I have faced many problems in my editors when I do work like you said `HTML` in `PHP` and `HTML` in `JavaScript` or `JavaScript` in `HTML`. It would be great if we can add this functionality to **coala** for linting.

PR #5759, opened 2018-09-16
Haskell.py: Add Haskell.py file

rishabhgarg25699:
Add a new python file in coalib/bearlib/languages/definitions/Haskell.py

Closes https://github.com/coala/coala/issues/5330

### DIFF
--- a/coalib/bearlib/languages/definitions/Haskell.py
+++ b/coalib/bearlib/languages/definitions/Haskell.py
@@ -1,0 +1,18 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs',
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010
+    special_chars = list('!#$%&;*:+.,/<=>?@\\^_|-~')
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
+        'case', 'of', 'infixl', 'foreign',
+        'data', 'data family', 'default', 'type family',
+        'deriving', 'instance', 'do', 'forall',
+        'hiding', 'if', 'then', 'else', 'import',
+        'infixr', 'let', 'in', 'mdo', 'module',
+        'proc', 'qualified', 'rec', 'type', 'infix',
+        'where', 'newtype', 'show']
--- a/tests/TestUtilities.py
+++ b/tests/TestUtilities.py
@@ -41,6 +41,7 @@ LANGUAGE_NAMES = [
     'Extensible Markup Language 1.0',
     'Fortran',
     'Golang',
+    'Haskell 1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010',
     'Hypertext Markup Language 2.0, 3.2, 4.0, 4.1, 5, 5.1',
     'Java',
     'JavaScript',


coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,18 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs'
```
li-boxuan: Should be a tuple. I remember I told you once in your previous pr, c.f. https://github.com/coala/coala/pull/5748#discussion_r215634169, but you are making the same mistake again 😅

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,18 @@
...
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
+    special_chars = '!#$%&*+./<=>?@\\^|-~'
+    comment_delimiters = '-- |',
```
li-boxuan: What is `-- |`? The comment delimiter in Haskell should be `--`.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,18 @@
...
+    multiline_comment_delimiters = {'-|': '-'}
+    grouping_expressions = {'$'}
```
li-boxuan: I've seen several attributes here like `grouping_expressions`, `string_accessing_n-th_character`, but not seen in any other language definition file or Language.py. Could you explain a bit about what they are for?

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,16 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = ('.hs')
```
li-boxuan: Again, you made the same mistake as https://github.com/coala/coala/pull/5748#discussion_r216180172

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,16 @@
...
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
+    special_chars = '!#$%&*+./<=>?@\\^|-~'
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
+        'case', 'class', 'data', 'default', 'deriving', 'do', 'else',
+        'forall', 'if', 'import', 'in', 'infix', 'infixl', 'infixr',
+        'let', 'module', 'newtype', 'of', 'qualified', 'then', 'type',
+        'where', '_', 'foreign', 'ccall', 'as', 'safe', 'instance', 'unsafe']
+    variable_declaration = {'::'}
```
li-boxuan: What is this field for? Btw `::` is not a variable declaration but type declaration.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,16 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = ('.hs')
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
+    special_chars = '!#$%&*+./<=>?@\\^|-~'
```
li-boxuan: use `list()`. see other language definition files

coalib/bearlib/languages/definitions/Haskell.py
rishabhgarg25699: First you say make the tuple of this. When I make tuple then you say previous one is right .

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,16 @@
...
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
+    special_chars = list('!#$%&*+./<=>?@\\^|-~')
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
+        'case', 'class', 'data', 'default', 'deriving', 'do', 'else',
+        'forall', 'if', 'import', 'in', 'infix', 'infixl', 'infixr',
+        'let', 'module', 'newtype', 'of', 'qualified', 'then', 'type',
+        'where', '_', 'foreign', 'ccall', 'as', 'safe', 'instance', 'unsafe']
+    type_declaration = {'::'}
```
li-boxuan: Sorry if https://github.com/coala/coala/pull/5759#discussion_r217915557 is confusing. What I mean is this field should better be removed.

tests/TestUtilities.py
```diff
@@ -42,6 +42,7 @@
     'Fortran',
     'Golang',
     'Hypertext Markup Language 2.0, 3.2, 4.0, 4.1, 5, 5.1',
+    'Haskell 1.0, 1.1, 1.2, 1.3, 1.4',
```
Man-Jain: I think these are all in alphabetical order. So Haskell should come above HTML.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@Language
+class Haskell:
+    extensions = '.hs',
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
+    special_chars = list('!#$%&*+./<=>?@\\^|-~')
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
```
wisn: Please refer to this [HaskellWiki](https://wiki.haskell.org/Keywords) for a complete list of Haskell keywords. There are several missing keywords. Also, why does underscore `_` listed as keyword but the other chars are not?

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,15 @@
...
+        'case', 'class', 'data', 'default', 'deriving', 'do', 'else',
+        'forall', 'if', 'import', 'in', 'infix', 'infixl', 'infixr',
+        'let', 'module', 'newtype', 'of', 'qualified', 'then', 'type',
+        'where', '_', 'foreign', 'ccall', 'as', 'safe', 'instance', 'unsafe']
```
wisn: I never saw any `ccall` in Haskell before. Where did you find this?

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,15 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs',
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4
```
wisn: Haskell versioning is quite unique. The major version `1.x` is before Haskell 98 released.

I would list them as `1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010`. However, we need to clarify this.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,15 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs',
```
wisn: Another extension for Haskell is `.lhs` which is Literate Haskell. I'm not sure. Should we add it as well?

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,20 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs', '.lhs',
```
jayvdb: you do analysis before blindly adding something because of review comments.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,20 @@
...
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010
+    special_chars = list('!#$%&*+./<=>?@\\^|-~')
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
+        '!', '-', '--', '-<', '-<<', '->', '::', ';',
+        '<-', ',', '=', '=>', '>', '?', '#', '*', '@', '[|, |]',
+        '_', '`', '{, }', '{-, -}', '|', '~', 'case', 'of',
```
jayvdb: many of these belong in `encapsulators`, which you have omitted from this language definition.

And as mentioned by @shashwatmishra76 , we do not need to repeat each `special_chars` here as keywords.

coalib/bearlib/languages/definitions/Haskell.py
jayvdb: Literate Haskell is Haskell language nested 'within' another language (https://github.com/coala/coala/issues/1690); i.e. languages where they use a enclosing language of some sort to demarcate code, and the outer language is usually some textual content in another language.  https://wiki.haskell.org/Literate_programming has some more examples.

Literate Haskell needs to be a different language definition, describing those demarcations.

coalib/bearlib/languages/definitions/Haskell.py
wisn: > you do analysis before blindly adding something because of review comments.

I clearly give a question mark in the previous review tho.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,19 @@
...
+
+@Language
+class Haskell:
+    extensions = '.hs',
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010
+    special_chars = list('!#$%&*+./<=>?@\\^|-~')
+    comment_delimiters = '--',
+    multiline_comment_delimiters = {'{-': '-}'}
+    keywords = [
+        '-<', '-<<', '::', ';', '<-', ',', '_', '{, }',
```
li-boxuan: `'{, }'` should be `'{',  '}'` ...?

rishabhgarg25699: I think it is right because opening and closing bracket is one keyword.If we use opening bracket then it is compulsory to use close bracket. They can not used separately.

li-boxuan: Your point makes sense, but this: 

`{, }`

is definitely not a keyword. Maybe wait for other reviewers' voice.

wisn: There is no such `{, }` keyword in Haskell. It should be `'{', '}'` because it is the same as [multiline_comment_delimiters](https://wiki.haskell.org/Keywords#.7B-.2C_-.7D). You write it `'{-', '-}'` instead of `'{-, -}'`, right?

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,19 @@
...
+        'case', 'of', 'class', 'infixl', 'foreign',
+        'data', 'data family', 'data instance', 'default',
+        'deriving', 'riving instance', 'do', 'forall',
```
wisn: `deriving` and `instance` already exists. I think you don't need to write it again as `deriving instance`. It is also a typo by the way.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,19 @@
...
+        'hiding', 'if', 'then', 'else', 'import', 'infix',
+        'infixr', 'instance', 'let', 'in', 'mdo', 'module',
+        'proc', 'qualified', 'rec', 'type', 'type family',
+        'type instance', 'where', 'newtype']
```
wisn: Since `type` and `instance` already exists, I don't think you need to write it again as `type instance`. Also, where is `show` keyword?

coalib/bearlib/languages/definitions/Haskell.py
wisn: I think a single char like `,` and `_` could be moved to `special_chars` list.

rishabhgarg25699: I will removed this because it came in  multiline_comment_delimiters.

coalib/bearlib/languages/definitions/Haskell.py
```diff
@@ -0,0 +1,18 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class Haskell:
+    extensions = '.hs',
+    versions = 1.0, 1.1, 1.2, 1.3, 1.4, 98, 2010
+    special_chars = list('::!#$%&;*+.,/<=>?@\\^_|-~')
```
wisn: Doesn't `special_chars` means it is a list of char so `::` is equal to `:`?
@jayvdb please help :joy:

coalib/bearlib/languages/definitions/Haskell.py
wisn: > I will removed this because it came in multiline_comment_delimiters.

Actually that is not what I means. However, it is fine.

rishabhgarg25699:
@wisn Please check now it is correct.

coalib/bearlib/languages/definitions/Haskell.py
Makman2: Bump I still don't see all keywords listed at @wisn's URL here in the definition (for eaxmple `data instance`)

rishabhgarg25699:
@Makman2 @wisn suggested earlier that 
deriving and instance already exists.Don't need to write it again as deriving instance. It is also a typo by the way.And 
Since type and instance already exists, I don't think you need to write it again as type instance.
See previous to previous commits of wisn

coalib/bearlib/languages/definitions/Haskell.py
wisn: @Makman2 `data` keyword and `instance` keyword already exists. I don't think we need to write it again as `data instance`. As for `data family`, since `family` is not a keyword by itself, I think we need to write it.

Makman2:
ack 0e876dcad6b56261e3c2c942311d35ab16019e6e

jayvdb:
@gitmate-bot rebase

jayvdb:
ack https://github.com/coala/coala/pull/5759/commits/d45b756d90a3814d84d102456f88df02b964f6e0

jayvdb:
@gitmate-bot ff

Issue #5345, opened 2018-04-08
Shouldn't be getting Executing section cli... when [cli] is empty


ishanSrt:

atb00ker:
Hi, i am trying to figure out this issue.
Please let me know if i have got it correctly:
`Executing section cli` should be printed on console only if the cli arguments are passed, otherwise that section should not be created -- i.e. the `parse_cli()` in `coala/coalib/parsing/CliParsing.py` should `append_to_sections` and `return sections` if there are cli arguments else simply `return None`

ishanSrt:
@atb00ker 
>-- example while running coala --do something --some option: Executing section cli should be printed; while running the same with .coafile; it should not be printed.

yes
I guess what happens currently is settings without a section or in the default section too (Please test this, this was a bug a while ago, maybe its solved now) were renamed and placed inside `[cli]` while parsing.

susg:
Terminal displays "executing section cli" even when there is no section named cli. So we need to stop that from happening. Is this what you mean?

jayvdb:
Yup, assigned to you @susg ;)

susg:
hi, what should be done when there is "default" section. Since it gets updated as "cli", so should "executing section cli" be printed then ?

PR #5770, opened 2018-09-25
coala_main.py: Prevent display of [cli] when empty

susg:
This conditional statement ensures that
`executing section cli` is not printed when
there is no section named `cli`.

Fixes https://github.com/coala/coala/issues/5345

up your stuff for merge!

### Checklist

- [X] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [X] I ran coala over my code locally. (*All commits have to pass
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/coala_main.py
+++ b/coalib/coala_main.py
@@ -183,7 +183,8 @@ def run_coala(console_printer=None,
                 section['default_actions'] = '*: ShowPatchAction'
                 section['show_result_on_top'] = 'yeah'
 
-            print_section_beginning(section)
+            if not (section.name == 'cli' and len(section.contents) == 0):
+                print_section_beginning(section)
             section_result = execute_section(
                 section=section,
                 global_bear_list=global_bears[section_name],


coalib/coala_main.py
```diff
@@ -183,7 +183,8 @@ def run_coala(console_printer=None,
                 section['default_actions'] = '*: ShowPatchAction'
                 section['show_result_on_top'] = 'yeah'
 
-            print_section_beginning(section)
+            if not (section.name == 'cli' and len(section.contents) == 0):
```
Naveenaidu: The above approach wouldn't work.
The reason being `section.contents` is never empty. Even if your `.coafile` is something like below where your [cli] section is empty. 
```
[cli]

[python]
files = admin.py
bears = SpaceConsistencyBear, QuotesBear
language = Python
preferred_quotation = '

default_actions = **: ApplyPatchAction
```
When i tried to debug it, `section.contents` always returns a `Ordered dict` of length 3 when the contents in the [ cli ] section is empty. I tried to print the value of `section.contents` when the [ cli ] section is empty. It's something like:- 
```
OrderedDict([('comment0', <Setting object(key='comment0', value='', origin='/home/theprophet/git/thepursuit/oth/.coafile', from_cli=False, to_append=False) at 0x7f5ed01aa1d0>)])

```
From the above result I can positively say that section always maintains some value in its contents which it uses for the proper working of coala. 
That's the reason, you checking if `len(section.contents) == 0` will not work. Since this doesn't work your `if` condition fails and the message `Executing Section CLI` will always prompt up.

I guess you need to change your approach.

P.S:- Please test your changes locally before creating a PR. This will save yours and the reviewers time :)

susg: > OrderedDict([('comment0', <Setting object(key='comment0', value='', origin='/home/theprophet/git/thepursuit/oth/.coafile', from_cli=False, to_append=False) at 0x7f5ed01aa1d0>)])

Yeah this true when `debug=True`, but when `debug=False` output is `OrderedDict()`.

jayvdb:
@gitmate-bot rebase

jayvdb:
@susg , you need to add a test case for this specific case.

susg:
hi @jayvdb , inorder to write test for the above changes, we need to run coala without cli args and also the coafile should not have sections named as `default`. I tried using `execute_coala` and provided path for system_coafile, but it also takes the coafile present in coala folder which has `default` in s    ections. Hence the if statement is always true.

PR #5811, opened 2018-10-13
Setting.py: Add SourcePosition to Setting.origin

akshatkarani:
This adds SourcePosition to Setting.origin which helps
in identifying the starting line number of a setting in 
.coafile.

Related to https://github.com/coala/coala/issues/5044

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/parsing/ConfParser.py
+++ b/coalib/parsing/ConfParser.py
@@ -8,6 +8,7 @@ from coalib.parsing.LineParser import LineParser
 from coalib.settings.Section import Section
 from coalib.settings.Setting import Setting
 from coalib.bearlib import deprecate_settings
+from coalib.results.SourcePosition import SourcePosition
 
 
 class ConfParser:
@@ -90,6 +91,7 @@ class ConfParser:
         current_section = self.get_section(current_section_name)
         current_keys = []
         no_section = True
+        line_number = 0
 
         for line in lines:
             (section_name,
@@ -98,6 +100,7 @@ class ConfParser:
              append,
              comment) = self.line_parser._parse(line)
 
+            line_number += 1
             if comment != '':
                 self.__add_comment(current_section, comment, origin)
 
@@ -135,7 +138,8 @@ class ConfParser:
                     current_section.add_or_create_setting(
                         Setting(key,
                                 value,
-                                origin,
+                                SourcePosition(
+                                    str(origin), line=line_number),
                                 to_append=append,
                                 # Start ignoring PEP8Bear, PycodestyleBear*
                                 # they fail to resolve this
@@ -149,7 +153,8 @@ class ConfParser:
                         True).add_or_create_setting(
                             Setting(key,
                                     value,
-                                    origin,
+                                    SourcePosition(
+                                        str(origin), line=line_number),
                                     to_append=append,
                                     # Start ignoring PEP8Bear, PycodestyleBear*
                                     # they fail to resolve this
--- a/coalib/results/SourcePosition.py
+++ b/coalib/results/SourcePosition.py
@@ -25,6 +25,7 @@ class SourcePosition(TextPosition):
         """
         TextPosition.__init__(self, line, column)
 
+        self.filename = file
         self._file = abspath(file)
 
     @property
@@ -36,3 +37,11 @@ class SourcePosition(TextPosition):
         if use_relpath:
             _dict['file'] = relpath(_dict['file'])
         return _dict
+
+    def __str__(self):
+        source_position = self.filename
+        if self.line is not None:
+            source_position += ':' + str(self.line)
+        if self.column is not None:
+            source_position += ':' + str(self.column)
+        return source_position
--- a/coalib/settings/ConfigurationGathering.py
+++ b/coalib/settings/ConfigurationGathering.py
@@ -367,8 +367,7 @@ def get_config_directory(section):
 
     However if its origin is already a directory this will be preserved:
 
-    >>> files = section['files']
-    >>> files.origin = os.path.abspath('/tmp/dir/')
+    >>> files = Setting('files', '**', origin=os.path.abspath('/tmp/dir/'))
     >>> section.append(files)
     >>> os.makedirs(section['files'].origin, exist_ok=True)
     >>> get_config_directory(section) == section['files'].origin
--- a/coalib/settings/Setting.py
+++ b/coalib/settings/Setting.py
@@ -8,6 +8,7 @@ from coala_utils.decorators import (
 from coala_utils.string_processing.StringConverter import StringConverter
 from coalib.bearlib.languages.Language import Language, UnknownLanguageError
 from coalib.parsing.Globbing import glob_escape
+from coalib.results.SourcePosition import SourcePosition
 
 
 def path(obj, *args, **kwargs):
@@ -155,7 +156,7 @@ class Setting(StringConverter):
     def __init__(self,
                  key,
                  value,
-                 origin: str = '',
+                 origin: (str, SourcePosition) = '',
                  strip_whitespaces: bool = True,
                  list_delimiters: Iterable = (',', ';'),
                  from_cli: bool = False,
@@ -196,7 +197,7 @@ class Setting(StringConverter):
 
         self.from_cli = from_cli
         self.key = key
-        self.origin = str(origin)
+        self._origin = origin
 
     def __path__(self, origin=None, glob_escape_origin=False):
         """
@@ -296,3 +297,22 @@ class Setting(StringConverter):
                              'incomplete. Please access the value of the '
                              'setting in a section to get the complete value.')
         return self._value
+
+    @property
+    def origin(self):
+        """
+        Returns the filename.
+        """
+        if isinstance(self._origin, SourcePosition):
+            return self._origin.filename
+        else:
+            return self._origin
+
+    @property
+    def line_number(self):
+        if isinstance(self._origin, SourcePosition):
+            return self._origin.line
+        else:
+            raise TypeError("Instantiated with str 'origin' "
+                            'which does not have line numbers. '
+                            'Use SourcePosition for line numbers.')
--- a/tests/parsing/ConfParserTest.py
+++ b/tests/parsing/ConfParserTest.py
@@ -98,6 +98,11 @@ class ConfParserTest(unittest.TestCase):
         self.assertRegex(self.cm.output[0],
                          'A setting does not have a section.')
 
+        # Check if line number is correctly set when
+        # no section is given
+        line_num = val.contents['setting'].line_number
+        self.assertEqual(line_num, 1)
+
     def test_parse_foo_section(self):
         foo_should = OrderedDict([
             ('a_default', 'val'),
@@ -146,6 +151,13 @@ class ConfParserTest(unittest.TestCase):
 
         self.assertEqual(val['comment1'].key, 'comment1')
 
+        # Check starting line number of
+        # settings in makefiles section.
+        line_num = val.contents['another'].line_number
+        self.assertEqual(line_num, 12)
+        line_num = val.contents['append'].line_number
+        self.assertEqual(line_num, 20)
+
     def test_parse_empty_elem_strip_section(self):
         empty_elem_strip_should = OrderedDict([
             ('a', 'a, b, c'),
@@ -166,6 +178,22 @@ class ConfParserTest(unittest.TestCase):
         for k in val:
             is_dict[k] = str(val[k])
         self.assertEqual(is_dict, empty_elem_strip_should)
+
+        # Check starting line number of
+        # settings in empty_elem_strip section.
+        line_num = val.contents['b'].line_number
+        self.assertEqual(line_num, 24)
+
+    def test_line_number_name_section(self):
+        # Pop off the default, foo, makefiles and empty_elem_strip sections.
+        self.sections.popitem(last=False)
+        self.sections.popitem(last=False)
+        self.sections.popitem(last=False)
+        self.sections.popitem(last=False)
+
+        key, val = self.sections.popitem(last=False)
+        line_num = val.contents['key1'].line_number
+        self.assertEqual(line_num, 30)
 
     def test_remove_empty_iter_elements(self):
         # Test with empty-elem stripping.
--- a/tests/results/SourcePositionTest.py
+++ b/tests/results/SourcePositionTest.py
@@ -25,12 +25,17 @@ class SourcePositionTest(unittest.TestCase):
             repr(uut),
             "<SourcePosition object\\(file='.*filename', line=1, "
             'column=None\\) at 0x[0-9a-fA-F]+>')
+        self.assertEqual(str(uut), 'filename:1')
 
         uut = SourcePosition('None', None)
         self.assertRegex(
             repr(uut),
             "<SourcePosition object\\(file='.*None', line=None, column=None\\) "
             'at 0x[0-9a-fA-F]+>')
+        self.assertEqual(str(uut), 'None')
+
+        uut = SourcePosition('filename', 3, 2)
+        self.assertEqual(str(uut), 'filename:3:2')
 
     def test_json(self):
         with prepare_file([''], None) as (_, filename):
--- a/tests/settings/SettingTest.py
+++ b/tests/settings/SettingTest.py
@@ -11,6 +11,7 @@ from coalib.settings.Setting import (
 )
 from coalib.parsing.DefaultArgParser import PathArg
 from coalib.parsing.Globbing import glob_escape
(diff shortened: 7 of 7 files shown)


coalib/settings/Setting.py
```diff
@@ -160,7 +160,8 @@ def __init__(self,
                  list_delimiters: Iterable = (',', ';'),
                  from_cli: bool = False,
                  remove_empty_iter_elements: bool = True,
-                 to_append: bool = False):
+                 to_append: bool = False,
+                 line_number: int = 0):
```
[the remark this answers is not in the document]
akshatkarani: Why do we need to convert to `Setting.origin` to string? Also we can get the filename from `origin.file`

coalib/parsing/ConfParser.py
```diff
@@ -135,7 +138,8 @@ def __parse_lines(self, lines, origin):
                     current_section.add_or_create_setting(
                         Setting(key,
                                 value,
-                                origin,
+                                SourcePosition(str(origin)),
+                                line_number=line_number,
```
jayvdb: `line_number` goes into the `SourcePosition`

coalib/settings/Setting.py
```diff
@@ -216,6 +219,7 @@ def __path__(self, origin=None, glob_escape_origin=False):
         :raises ValueError:        If no origin is specified in the setting
                                    nor the given origin parameter.
         """
+        origin = origin.file
```
jayvdb: this doesnt belong here.  it wont be backwards compatible here.

See below, which has `if hasattr(self, 'origin') ...`

To be backwards compatible you will need to add `if isinstance(origin, SourcePosition) ..`

coalib/collecting/Collectors.py
```diff
@@ -126,11 +126,13 @@ def list_glob_results(values=None):
...
 
 def collect_files(file_paths, log_printer=None, ignored_file_paths=None,
-                  limit_file_paths=None, section_name=''):
+                  limit_file_paths=None, section_name='',
+                  line_num=[]):
     """
     Evaluate globs in file paths and return all matching files
 
     :param file_paths:         File path or list of such that can include globs
+    :param line_num:           List of line numbers which correspond to files
```
jayvdb: remove this parameter, and all changes to coalib/processing and coalib/collecting , etc.

we do not want to be passing around line numbers.   We pass around `SourcePosition` objects, which contain the filename and the line number together as a single object.

Your PR should not try to implement the better error messages everywhere.  Focus on `Setting.origin` supporting `SourcePosition` in a backwards compatible way.

Create unit tests which demonstrate the the Setting.origin are being set with the correct values.

coalib/settings/Setting.py
```diff
@@ -155,12 +156,13 @@ class Setting(StringConverter):
     def __init__(self,
                  key,
                  value,
-                 origin: str = '',
+                 origin=SourcePosition(''),
```
jayvdb: Dont remove type information.  And it must be backwards compatible, which means it still must handle `str`.

> origin: (str, SourcePosition) = ''

coalib/settings/Setting.py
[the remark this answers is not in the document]
jayvdb: Remove this parameter.

We do not pass around line numbers.  We pass around `SourcePosition` objects.

coalib/parsing/ConfParser.py
```diff
@@ -149,7 +153,8 @@ def __parse_lines(self, lines, origin):
                         True).add_or_create_setting(
                             Setting(key,
                                     value,
-                                    origin,
+                                    SourcePosition(str(origin)),
+                                    line_number=line_number,
```
jayvdb: line_number belongs inside the `SourcePosition`

coalib/settings/FunctionMetadata.py
```diff
@@ -140,11 +140,13 @@ def return_arg(x):
...
             annotation = return_arg
 
+        lines = section._get_lines(param)
         try:
             return annotation(section[param])
         except (TypeError, ValueError):
             raise ValueError('Unable to convert parameter {!r} into type '
-                             '{}.'.format(param, annotation))
+                             '{}.'.format(param, annotation) +
+                             '\n.coafile lines: ' + str(lines))
```
jayvdb: This is the one usage of the line numbers which you might like to keep in the patch, as it is in `coalib/settings`, but I suggest that you save it in a backup branch, and remove it until you have good tests cases all passing and getting positive reviews of your test cases.

coalib/processes/Processing.py
```diff
@@ -376,6 +376,7 @@ def instantiate_processes(section,
     filename_list = collect_files(
         glob_list(section.get('files', '')),
         None,
+        line_num=section._get_lines('files'),
```
jayvdb: needs to be removed, because you are accessing a semi-private method of an object not in the same component.  i.e. `coalib/processes` shouldnt access `._foo` of an object from `coalib/section` .  (however coalib/section & coalib/settings are almost the same component, so the rule can be bent between those two components, if necessary)

jayvdb:
looking better ;)

The gitmate errors here are all things you need to fix.

And then you will need to add tests to reach 100% coverage.

p.s. there is a `coala -ci` autopep8 error about unrelated code, so your CI will be red until that is fixed.

coalib/results/SourcePosition.py
```diff
@@ -36,3 +37,6 @@ def __json__(self, use_relpath=False):
         if use_relpath:
             _dict['file'] = relpath(_dict['file'])
         return _dict
+
+    def __str__(self):
+        return self.filename
```
jayvdb: this should show the filename, line number if it exists, and the column number if it also exists

separated by colons.

coalib/parsing/ConfParser.py
```diff
@@ -82,14 +83,15 @@ def __add_comment(self, section, comment, origin):
         section.append(Setting(
             key,
             comment,
-            origin,
+            SourcePosition(str(origin)),
```
jayvdb: a SourcePosition is only useful if we are adding a line number.

Either add a line number here, or restore this to be only `origin`

coalib/settings/Section.py
```diff
@@ -41,7 +42,8 @@ def append_to_sections(sections,
         sections[section_name.lower()] = Section(section_name)
 
     sections[section_name.lower()].append(Setting(
-        key, str(value), origin, from_cli=from_cli, to_append=to_append))
+        key, str(value), SourcePosition(origin),
```
jayvdb: I'm only 90%-ish sure about this ... so feel free to tell me why I am wrong here...

this shouldnt be necessary, again because it doesnt add a line number.

If tests are breaking because of this, my guess is that means something isnt working properly when Setting.origin is not a `SourcePosition`

coalib/settings/ConfigurationGathering.py
```diff
@@ -403,7 +403,7 @@ def get_config_directory(section):
     if 'project_dir' in section:
         return path(section.get('project_dir'))
 
-    config = os.path.abspath(section.get('files', '').origin)
+    config = os.path.abspath(str(section.get('files', '').origin))
```
jayvdb: when `__str__` is changed to also give the line and col, this is going to break.

See comment about @property for `SourcePosition.origin`

coalib/settings/Setting.py
```diff
@@ -196,7 +197,7 @@ def __init__(self,
 
         self.from_cli = from_cli
         self.key = key
-        self.origin = str(origin)
+        self.origin = origin
```
jayvdb: for backwards compatibility, we need `self.origin` to always return the filename.

So here, `self._origin = origin`

And then add

```py
@property
def origin(self):
    if isinstance(self._origin, SourcePosition):
        return self._origin.filename
    else:
        return self._origin
```

But please give it a good docstring to explain what it does.

coalib/settings/Setting.py
```diff
@@ -223,6 +224,9 @@ def __path__(self, origin=None, glob_escape_origin=False):
         if hasattr(self, 'origin') and self.origin != '':
```
jayvdb: when we use a @property, `hasattr(self, 'origin')` wont work.

It was already broken code, because `__init__` sets `self.origin` , so this `hasattr` would never be `False`.

Anyways, you should be able to simply remove it, and then you also dont need the `isinstance(origin, SourcePosition)` below because `self.origin` always returns a string, the filename, which is what we want.

coalib/settings/Setting.py
akshatkarani: @jayvdb Although unrelated to this PR I don't quite understand why we have `self.key = key` here and then below also have
```
@property
def key(self):
```

coalib/settings/Setting.py
akshatkarani: If I remove `hasattr(self, 'origin')` then some tests are breaking. It is because `origin` is string in some cases when it is called by `__glob__`.
So i guess we should not remove it.

coalib/settings/Section.py
akshatkarani: This does seem to be unnecessary. I will remove it.

coalib/settings/Setting.py
jayvdb: ok, add a comment about it in the code here.

coalib/results/SourcePosition.py
akshatkarani: I guess the CLI section is parsed by `CliParsing.py` and there are no changes made there, so no line number.

akshatkarani:
CLI section is parsed in `CliParsing.py` and there are no changes there, so no line number there. And as you said no line number is needed.

akshatkarani:
Had to add `origin.setter` as in some tests origin was being set and they were failing.
Changed `SourcePositionTest.py` as now string conversion happens differently.
I did add some asserts for line number but didn't for CLI section as I had some doubt.
Should I add a new CLI section in `example_file` and add tests below or something else.

coalib/settings/Setting.py
li-boxuan: That looks like a bug to me.

coalib/settings/Setting.py
```diff
@@ -296,3 +297,26 @@ def value(self):
...
+        else:
+            return self._origin
+
+    @origin.setter
+    def origin(self, origin):
+        self._origin = origin
+
+    @property
+    def line_number(self):
+        if isinstance(self._origin, SourcePosition):
```
li-boxuan: Can be just me, but I personally think `self.origin` is better here. Assume you add some validation in `origin.setter`, but if some bad guy sets `self._origin` explicitly, they can bypass the validation because you are using `self._origin` here.

coalib/settings/Setting.py
```diff
@@ -296,3 +297,26 @@ def value(self):
                              'incomplete. Please access the value of the '
                              'setting in a section to get the complete value.')
         return self._value
+
+    @property
+    def origin(self):
+        """
+        Ensures that Setting.origin always returns a string,
```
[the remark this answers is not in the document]
akshatkarani: @li-boxuan  But i will call it as `Setting.origin` only. What would be good docstring otherwise?

coalib/settings/Setting.py
akshatkarani: @li-boxuan  Even i think it would be better to do it your way. @jayvdb  What do you think?

coalib/settings/Setting.py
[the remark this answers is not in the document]
jayvdb: @akshatkarani you shouldnt mention `Setting.origin` here, as it is the docstring of `Setting.origin`.  and instead of `string...`, just say `filename`

coalib/settings/Setting.py
jayvdb: ya, agreed,

```py
source_position = self.origin
if isinstance(...)
    ...
```

coalib/settings/Setting.py
```diff
@@ -296,3 +297,26 @@ def value(self):
...
+    @origin.setter
+    def origin(self, origin):
+        self._origin = origin
+
+    @property
+    def line_number(self):
+        if isinstance(self._origin, SourcePosition):
+            return self._origin.line
+        else:
+            raise TypeError("Given 'str' object does not have "
```
jayvdb: "Given 'str' object" doesnt mean anything to the person reading it.

> Instantiated with str 'origin' which does not have line numbers.  Use SourcePosition for line numbers.

coalib/settings/Setting.py
akshatkarani: Can't use `self.origin` as getter is set to return `str`

li-boxuan: Oh yep, my fault. Nvm then.

coalib/settings/Setting.py
```diff
@@ -296,3 +297,26 @@ def value(self):
...
+    def origin(self):
+        """
+        Returns the filename.
+        """
+        if isinstance(self._origin, SourcePosition):
+            return self._origin.filename
+        else:
+            return self._origin
+
+    @origin.setter
```
jayvdb: no, we need to remove this.

it is extremely bad practise for `a.value = a.value` to have any effect.

think about it.

if the property getter always returns a string filename, then if the caller uses the setter with a `SourcePosition`, the object will be lost.

```py
sp = SourcePosition('foo.txt',...)
s = Setting(origin=sp,...)
print(type(s.origin))
filename = s.origin  # a string for backwards compatibility
s.origin = filename
print(type(s.origin))
```

coalib/results/SourcePosition.py
```diff
@@ -36,3 +37,8 @@ def __json__(self, use_relpath=False):
         if use_relpath:
             _dict['file'] = relpath(_dict['file'])
         return _dict
+
+    def __str__(self):
+        return ('file=' + self.filename +
```
jayvdb: getting very close ... so now ... wtf is this string.

why did you chose this string instead of the one I suggested?

do not just add what I suggested instead.

Please defend your choice.  Explain it.  Maybe as a code comment, or here as a comment in GitHub.

Once this is merged, this is the string representation of a SourcePosition object --- any change to it after it is merged is a BIG BAD BREAKING CHANGE.

So we need to be sure this is the best , most stable , most appropriate, way to convert a SourcePosition into a string.

This is [serious business](https://www.urbandictionary.com/define.php?term=serious%20business)! :P

akshatkarani: What you suggested was to separate it by colons but it think it is more natural to separate it by semi-colon. 
Also about only printing line and column if it exists, even if it doesn't exist printing it will do no harm.
I will remove `file=` but `line=` and `column=` should be kept to know which is what.

li-boxuan: Btw, why not comma :P

akshatkarani: Even comma is fine.

jayvdb: `filename:lineno:colno` is dense, but is commonly understood

and it has no i18n issues.  words in a `__str__` are bad news for i18n, as a str(..) must be the same for every user, so they cant be internationalised, and they force English onto all users, or force parsers creating by programmers to use English keywords to parse the output of `str(..)` before adding i18n in order to show something localised for a non-English UI.

coalib/settings/Setting.py
jayvdb: yup. unfortunately , we didnt think about this when we suggested to add a setter.

jayvdb: this hasnt been done.

akshatkarani: If setter is not there then on running pytest, a test is failing.
In docstring of `get_config_directory` in `ConfigurationGathering.py` there is a line (line number 371) which sets the `origin` and if setter is not there then error is shown `AttributeError: can't set attribute`.

jayvdb: That doctest needs to be fixed.  It was showing what was possible, by creating a ficticious scenario.
It was cheating.  It needs to be fixed.

coalib/settings/ConfigurationGathering.py
```diff
@@ -367,8 +367,7 @@ def get_config_directory(section):
 
     However if its origin is already a directory this will be preserved:
 
-    >>> files = section['files']
-    >>> files.origin = os.path.abspath('/tmp/dir/')
+    >>> files = Setting("files", "**", origin=os.path.abspath('/tmp/dir/'))
```
jayvdb: single quotes :P

akshatkarani: done :smile:

jayvdb:
ack https://github.com/coala/coala/pull/5811/commits/6af1b6049e3728d12adef1951b89c6e0279448c8

jayvdb:
@gitmate-bot ff

Issue #5822, opened 2018-10-16
Add language definition for g4 files


virresh:
This g4 file refers to an ANTLRv4 grammar file.
See examples of g4 files and more information about them at: https://github.com/antlr/antlr4/blob/master/doc/getting-started.md

Dhiraj240:
is it about adding generated lexer and parser of c,c++ etc. languages from grammar repo?

virresh:
Nope.
See https://github.com/coala/coala/tree/master/coalib/bearlib/languages/definitions for examples

PR #5828, opened 2018-10-18
antlr.py: Add g4 file

Dhiraj240:
This adds the language definitions
for g4 files in coala.

Closes https://github.com/coala/coala/issues/5822

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/bearlib/languages/definitions/antlr.py
+++ b/coalib/bearlib/languages/definitions/antlr.py
@@ -1,0 +1,24 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class antlr:
+    aliases = 'antlr3', 'antlr4'
+    versions = 3, 4
+
+    extensions = '.g', '.g4'
+    comment_delimiters = '//',
+    multiline_comment_delimiters = {'/*': '*/'}
+    string_delimiters = {"'": "'"}
+    multiline_string_delimiters = {}
+    indent_types = {'{': '}'}
+    encapsulators = {'(': ')', '[': ']'}
+    keywords = [
+        'import', 'fragment', 'lexer',
+        'parser', 'grammar', 'returns',
+        'locals', 'throws', 'catch',
+        'finally', 'mode', 'options',
+        'tokens',
+    ]
+    special_chars = list(r'/**+-*/.;\,()[]{}\=<>|&^~?%!')
+    string_delimiter_escape = {"'": "\\'"}
--- a/coalib/bearlib/languages/__init__.py
+++ b/coalib/bearlib/languages/__init__.py
@@ -7,6 +7,7 @@ from .Language import Language
 from .Language import Languages
 
 from .definitions.Unknown import Unknown
+from .definitions.antlr import antlr
 from .definitions.Bash import Bash
 from .definitions.C import C
 from .definitions.CPP import CPP
--- a/tests/TestUtilities.py
+++ b/tests/TestUtilities.py
@@ -33,6 +33,7 @@ C_BEARS_COUNT = 2
 
 # This list is sorted alphabetically
 LANGUAGE_NAMES = [
+    'antlr 3, 4',
     'Bash',
     'C',
     'C#',


Dhiraj240:
@ishanSrt I already followed it that a language definition has been added for g4 i.e grammar files.

coalib/bearlib/languages/definitions/g4.py
```diff
@@ -0,0 +1,21 @@
+from coalib.bearlib.languages.Language import Language
+
+
+@Language
+class g4:
+    aliases = 'antlr4', 'grun'
```
virresh: I think `grun` is an alias for their test benching code, probably doesn't fit in here.

coalib/bearlib/languages/definitions/g4.py
```diff
@@ -0,0 +1,21 @@
...
+    indent_types = {'{': '}'}
+    encapsulators = {'(': ')', '[': ']'}
+    keywords = [
+        'import', 'fragment', 'lexer',
+        'parser', 'grammar', 'returns',
+        'locals', 'throws', 'catch',
+        'finally', 'mode', 'options',
+        'tokens']
+    special_chars = list(r'/**+-*/.;\,()[]{}\=<>|&^~?%!')
+    string_delimiter_escape = {"'": "\\'"}
```
virresh: Possibly you meant to escape both the start and end tags for a string delimiter ?
Also this is a new usage, I've not seen it before. Only seen the braces being escaped on a need basis.
Would be great if you could put a link which may expand on this usage as a comment in this PR for future reference :sweat_smile:

coalib/bearlib/languages/definitions/g4.py
Dhiraj240: Yeah but don't you think that ```grun ``` is also a part of ```antlr-4.7.1-complete.jar``` i.e both of these aliases ```antr4``` and ```grun``` is part of antlr.jar because antr4 testrig is used for testing grammar.

coalib/bearlib/languages/definitions/g4.py
Dhiraj240: Yes i meant to escape both the start and end tags for a string delimiter.I followed [this](https://github.com/antlr/antlr4/blob/master/doc/lexicon.md) where both ```char``` and ```string``` use single quote.Also at the very first line they have written that they follow syntax of C so should i include this link at the PR body as taken references from so and so resource.

virresh: So for escaping both start and end, should'nt it be `{"\\'": "\\'"} ?

coalib/bearlib/languages/definitions/g4.py
virresh: Yeah, so the point of aliases is not to focus on the features of ANTLR's jar, the point of aliases in this file is to focus on language specific features of the grammar files, of which grun is not a part of

Also please not that `grun` is not any officially endorsed thing, it's just a bash shell alias (similar to `ll` which is given to `org.antlr.v4.gui.TestRig` for convenience sake, it could very well be any arbitrary name which we want.

ishanSrt:
🤨 don't see any file named **definitions** @Dhiraj240

Dhiraj240:
@ishanSrt okay, i thought that if we change a file then only we write such thing.I will change it for sure and will remember that t has to be written while an addition too.
Thanks.

coalib/bearlib/languages/definitions/g4.py
Dhiraj240: Alright, got it.

coalib/bearlib/languages/definitions/g4.py
Dhiraj240: yes, even same case is applied to ```antlr4``` to which is also an alias for bash shell.i will do the changes also i will change language name as ```antlr``` and put on versions.Then it will be ```aliases = 'antlr4', 'antlr3' ``` just check gitter chat.

coalib/bearlib/languages/definitions/g4.py
Dhiraj240: But check [this](https://github.com/coala/coala/blob/master/coalib/bearlib/languages/definitions/C.py#L22) for other languages too.

virresh: Oh
Github escaped my comment
I'd meant to write:
`{"\\'": "\\'"}`

Since a single backslash is equivalent to no backslash
And I'm not quite sure about the one present in c definition, not sure what's the purpose of that, but seems like its an error.

coalib/bearlib/languages/definitions/antlr.py
```diff
@@ -0,0 +1,23 @@
...
+    indent_types = {'{': '}'}
+    encapsulators = {'(': ')', '[': ']'}
+    keywords = [
+        'import', 'fragment', 'lexer',
+        'parser', 'grammar', 'returns',
+        'locals', 'throws', 'catch',
+        'finally', 'mode', 'options',
+        'tokens']
+    special_chars = list(r'/**+-*/.;\,()[]{}\=<>|&^~?%!')
+    string_delimiter_escape = {"\\'": "\\'"}
```
jayvdb: this isnt valid.

See the other ones which have a `string_delimiter_escape` , the key must be the same as the key in the string_delimiters

coalib/bearlib/languages/definitions/antlr.py
```diff
@@ -0,0 +1,23 @@
...
+    string_delimiters = {"'": "'"}
+    multiline_string_delimiters = {}
+    indent_types = {'{': '}'}
+    encapsulators = {'(': ')', '[': ']'}
+    keywords = [
+        'import', 'fragment', 'lexer',
+        'parser', 'grammar', 'returns',
+        'locals', 'throws', 'catch',
+        'finally', 'mode', 'options',
+        'tokens']
```
jayvdb: trailing comma, with closing bracket on the next line, dedented.

jayvdb:
ack https://github.com/coala/coala/pull/5828/commits/8a7714a65ef463516a483d9f2ccd8607ff63a9cf

jayvdb:
@gitmate-bot ff

PR #5892, opened 2018-11-12
coalib/: Use trollius for 3.4.x support

jayvdb:
This is now needed for https://github.com/coala/coala-bears/pull/1386


jayvdb:
trollius isnt updated, so it isnt usable.

https://github.com/coala/coala/pull/5894 instead

PR #5893, opened 2018-11-13
Moban repo support

jayvdb:
**For short term contributors:** we understand that getting your commits well
defined like we require is a hard task and takes some learning. If you
look to help without wanting to contribute long term there's no need
for you to learn this. Just drop us a message and we'll take care of brushing
up your stuff for merge!

### Checklist

- [ ] I read the [commit guidelines](http://coala.io/commit) and I've followed
      them.
- [ ] I ran coala over my code locally. (*All commits have to pass
      individually.* It is not sufficient to have "fixup commits" on your PR,
      our bot will still report the issues for the previous commit.) You will
      likely receive a lot of bot comments and build failures if coala does not
      pass on every single commit!

After you submit your pull request, **DO NOT click the 'Update Branch' button.**
When asked for a rebase, consult [coala.io/rebase](https://coala.io/rebase)
instead.

Please consider helping us by reviewing other peoples pull requests as well:

- pick up any PR at <https://coala.io/review>
- review it (check <https://coala.io/reviewing> for more info)
- if you are sure that it needs work, use `corobo mark wip <URL>` to get it out
  of the review queue.

The more you review, the more your score will grow at coala.io and we will
review your PRs faster!

PR #5894, opened 2018-11-13
Python 3.4.3 support

jayvdb:
Sits on top of https://github.com/coala/coala/pull/5893


.travis.yml
```diff
@@ -3,6 +3,7 @@ language: python
 python:
   - 3.4.4
   - 3.5
+  - 3.4.2
```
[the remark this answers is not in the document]
jayvdb: it would slow down all builds, unnecessarily.

We could add it to master builds.

Or I think we can also only run jobs on the nightly `cron`, which would reduce delays if multiple PRs are getting merged.

.moban.yaml
```diff
@@ -65,3 +67,6 @@ targets:
   - netlify.toml: docs/netlify.toml
   - .misc/appveyor.yml: ci/appveyor.yml.jj2
   - .misc/run_with_env.cmd: run_with_env.cmd
+copy:
+  # homeassistant license MIT
+  - coalib/misc/Asyncio.py: async.py
```
li-boxuan: What is the benefit of using `moban` to copy `async.py` from https://github.com/NAStools/homeassistant/blob/master/homeassistant/util/async.py? That project seems to be out of maintenance.

Drawbacks I can see:
1. licence is not stated in `async.py`, but in this file.
2. `async.py` needs to be excluded in `.coafile`.
3. `async.py` has other functions, e.g. `fire_coroutine_threadsafe`, which we don't use.
4. We cannot easily test it.

jayvdb:
@gitmate-bot rebase

.moban.yaml
jayvdb: The reason to do it this way is so that external code is 'managed'.  We can improve the way it is managed by improving moban.

Your point '3' is the only significant problem, and that can be solved by moban having patch support, so we can exclude the other code e.g. https://github.com/moremoban/moban/issues/75 . A header can also be added using a patch.

I am not worried about the licensing being missing from the file, as anyone modifying this file will likely be informed via code review, or will be reverted when someone looks at the moban job.  We should distribute the MIT license, and any other license of code we're including, but by no means is this "substantial portions" according to the MIT license so we're not breaking the license by not including it.  And by having a control file which mentions where code came from, and thus were the license text can be found, we're not hiding the fact the code is copied.

The fact it is unmaintained is a problem for the future, as PRs upstream will not get merged.  Again, a .patch could solve that, but i would expect that this type of code would be moved to a new `coala-compat` package, or something like that, or a sub-module of coala-utils.

Same with the .coafile problems.  A patch can fix it.

The drawbacks of not doing it:

- another piece of copied code will fade into the coala codebase and its history effectively lost
- coala doesnt have a process for allowing externally written code into the repo.
- moban doesnt have a user (coala) pushing for it to handle more complicated scenarios than just 'copy'.
- coala doesnt start the process of identifying and fixing any existing copied code.  This one is MIT, but we will likely find some which isnt so freely licensed.  e.g. we have CC-BY-SA-licensed 'code' in some repos, and a major issue over in coala-quickstart
- etc.

Being in `.moban.yaml` is an annoyance.  That is intentional ;-)

jayvdb:
@gitmate-bot rebase

Issue #5931, opened 2019-01-05
Make the code coverage reach to 100%


he1l0world:
https://codecov.io/gh/coala/coala/src/master/coalib/processes/Processing.py#L65
https://codecov.io/gh/coala/coala/src/master/coalib/parsing/Globbing.py#L172

Maybe we can imporve code coverage to 100% by changing "# pragma posix: no cover" to "# pragma: no cover"

he1l0world:
I saw the code coverage has reached 100% but somehow it failed in other parts. Can anyone guide me? Thank you.

li-boxuan:
https://codecov.io/gh/coala/coala/compare/3eaae8875055d5340b707adaa53c289dd6b96725...3d80d91367cf416c020afa336efa1735fe967bc9/changes coverage reaches 100% after fixing Appveyor build failure https://github.com/coala/coala/pull/5937. Thx for reporting anyway :)

PR #5936, opened 2019-01-06
Improve the code coverage

he1l0world:
Processing.py: Change code coverage
Globbing.py: Change code coverage
Change # pragma posix: no cover --> # pragma: no cover
Fixes https://github.com/coala/coala/issues/5931

### DIFF
--- a/coalib/parsing/Globbing.py
+++ b/coalib/parsing/Globbing.py
@@ -169,7 +169,7 @@ def translate(pattern):
                 regex += '.*'
             # On Windows, '*' matches everything but the filesystem
             # separators '/' and '\'.
-            elif platform.system() == 'Windows':  # pragma posix: no cover
+            elif platform.system() == 'Windows':  # pragma: no cover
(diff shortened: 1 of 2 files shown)


he1l0world:
I saw the code coverage has reached 100% but somehow failed. Is anyone can guide me? Thanks

coalib/parsing/Globbing.py
```diff
@@ -169,7 +169,7 @@ def translate(pattern):
                 regex += '.*'
             # On Windows, '*' matches everything but the filesystem
             # separators '/' and '\'.
-            elif platform.system() == 'Windows':  # pragma posix: no cover
+            elif platform.system() == 'Windows':  # pragma: no cover
```
li-boxuan: This does not look like the right way to solve the problem. By changing `# pragma posix: no cover` to `# pragma: no cover` you are hiding the existing problem. Instead, we should find why code coverage was less than 100%.

Looks like it was because Appveyor build was failing, leading to this line not being tested.

PR #5994, opened 2019-02-15
Add LineLengthTestHelper

### LINKED ISSUE #5877: Create a line length test helper
A line length test helper will take a chunk of source , and a regex, and verify the bear accepts source under the line length and rejects lines over the line length.

The max_line_length setting can be automatically determined by the lines in the source provided by the caller.  The bear should pass this test.

The regex is then used to increase the size of the source line lengths by 1 character, in accordance with the source language rules.

The helper confirms that the source was increased by 1 character only, and re-runs the bear, and expects it to fail.

bkhanale:
This helper can be used to test the `max_line_length`
setting of the bear.

Closes https://github.com/coala/coala/issues/5877

up your stuff for merge!

### Checklist

      them.
      pass on every single commit!

instead.

  of the review queue.

review your PRs faster!

### DIFF
--- a/coalib/testing/LineLengthTestHelper.py
+++ b/coalib/testing/LineLengthTestHelper.py
@@ -1,0 +1,78 @@
+import logging
+import queue
+
+from coalib.settings.Section import Section
+from coalib.settings.Setting import Setting
+from coalib.testing.BearTestHelper import generate_skip_decorator
+from coalib.testing.LocalBearTestHelper import LocalBearTestHelper
+
+
+def verify_line_length(bear,
+                       source_file,
+                       filename=None,
+                       settings={},
+                       aspects=None,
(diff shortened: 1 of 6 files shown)


bkhanale:
I implemented a slightly different approach to the problem. The original solution was to find the max_line_length setting from the source, increase the length of the max_line by 1 through a regex provided. But here there were two problems, one being the regex itself. It is hard to provide the regex for a language syntax (specially when the there are multiple max_lines) and the second being it is also difficult to generate the string from a regex and increase the line length by 1.

The alternative solution takes care of both of the above problems and also solves the issue. The solution is that we can get the max_line_length from the source and then *decrease* the max_line_length setting and run the bear, which will fail eventually.

@jayvdb Let me know if this is good enough :-)

PR #6005, opened 2019-03-30
ConsoleInteraction: Remove nothing_done()

roshnet:


li-boxuan:
You provided an invalid link in your [opening comment](https://github.com/coala/coala/pull/6005#issue-265958963) :(

roshnet:
@li-boxuan Sorry about the mistake.

tests/output/ConsoleInteractionTest.py
```diff
@@ -634,7 +634,8 @@ def test_print_section_beginning(self):
 
     def test_nothing_done(self):
```
KVGarg: Since the function doesn't exist anymore as well as the parameter. So, IMO this function is of no need now. Also, in issue description its mentioned that we don't need this test

PR #6029, opened 2019-06-19
Result.py: Add actions attribute

akshatkarani:
This adds a new attribute `actions` to the Result class. It is a list of action objects which are specific to the origin to that Result.
This also makes changes in `ConsoleInteraction` module and `Processing` module to incorporate these additions and support bears to define their own actions.

### DIFF
--- a/coalib/processes/Processing.py
+++ b/coalib/processes/Processing.py
@@ -32,13 +32,13 @@ from coalib.io.FileProxy import FileDictGenerator
 from coalib.io.File import File
 
 
-ACTIONS = [DoNothingAction,
-           ApplyPatchAction,
-           PrintDebugMessageAction,
-           ShowPatchAction,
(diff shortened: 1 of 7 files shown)


coalib/processes/Processing.py
```diff
@@ -145,30 +150,34 @@ def autoapply_actions(results,
             else:
                 not_processed_results.append(result)
                 continue
+        if action not in bear_actions or action in result.actions:
```
akshatkarani: This condition ensures that if action is in `bear_actions` then it must be in `result.actions`.

abhishalya:
Ping @Makman2

abhishalya:
Can you change the PR title = commit shortlog.

Its a bit confusing everytime I see it :P

Makman2:
From my side nothing more to complain :) can be merged.

Issue #6034, opened 2019-06-22
ConsoleInteraction.py: `color_letter` function is buggy


akshatkarani:
`color_letter` function in ConsoleInteraction module does not work as expected when there is an upper case letter after the upper case letter in parenthesis.
e.g. If input line is `Edit (C)ommit Message` then output is `Edit (M)ommit Message`

akshatkarani:
The problem with the current function is that is loops over all the characters and the last uppercase letter in the line gets highlighted. But it could be the case that the uppercase letter in parenthesis is not the last uppercase letter.

priyansh19:
Just wanna know that is this issue still open for contribution?

akshatkarani:
@priyansh19 No this issue is reserved for GSoC

PR #6039, opened 2019-06-30
ConsoleInteraction: Modify color_letter function

akshatkarani:
This modifies color_letter function to run as correctly
when there is a uppercase letter after "the" uppercase
letter in parenthesis and also if action has a associated warning then
it gets highlighted.

Fixes https://github.com/coala/coala/issues/6034

### DIFF
--- a/coalib/output/ConsoleInteraction.py
+++ b/coalib/output/ConsoleInteraction.py
@@ -97,22 +97,24 @@ DIFF_EXCERPT_MAX_SIZE = 4
 
 
 def color_letter(console_printer, line):
-    x = -1
-    y = -1
-    letter = ''
-    for i, l in enumerate(line, 0):
-        if line[i] == '(':
-            x = i
-        if line[i] == ')':
-            y = i
-        if l.isupper() and x != -1:
-            letter = l
+    x = line.find('(')
+    if x == -1:
+        letter = ''
+        y = x + 1
+    else:
+        letter = line[x + 1]
+        y = x + 2
+    warn = line.rfind('[')
+    if warn == 0:
+        warn = len(line)
     first_part = line[:x+1]
-    second_part = line[y:]
+    second_part = line[y:warn]
+    warning_part = line[warn:]
 
     console_printer.print(first_part, end='')
     console_printer.print(letter, color='blue', end='')
-    console_printer.print(second_part)
+    console_printer.print(second_part, end='')
+    console_printer.print(warning_part, color='blue')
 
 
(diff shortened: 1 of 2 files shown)


akshatkarani:
If action has warning the it looks something like
Screenshot from 2019-06-30 17-41-46

akshatkarani:
@abhishalya Can you review?

abhishalya:
Haven't tested it, but the changes look good to me :+1:

coalib/output/ConsoleInteraction.py
```diff
@@ -97,22 +97,24 @@ def highlight_text(no_color, text, style, lexer=TextLexer()):
...
-            letter = l
+    x = line.find('(')
+    if x == -1:
+        letter = ''
+        y = x + 1
+    else:
+        letter = line[x + 1]
+        y = x + 2
+    warn = line.rfind('[')
+    if warn == 0:
```
li-boxuan: Shouldn't it be `-1` if not found?

tests/output/ConsoleInteractionTest.py
```diff
@@ -189,6 +190,22 @@ def tearDown(self):
...
+
+        line2 = '[  ] *0. Apply (P)atch'
+        with retrieve_stdout() as stdout:
+            color_letter(self.console_printer, line2)
+            self.assertEqual(line2 + '\n', stdout.getvalue())
+
+        line3 = '[  ] 3. Apply (P)atch [Note: This will do something]'
+        with retrieve_stdout() as stdout:
+            color_letter(self.console_printer, line3)
+            self.assertEqual(line3 + '\n', stdout.getvalue())
```
li-boxuan: I am not familiar with this part, but is it able to verify the color of output?

coalib/output/ConsoleInteraction.py
akshatkarani: No, notice `[` is always the first character when you are asked for an action.

tests/output/ConsoleInteractionTest.py
akshatkarani: No, it just checks the parsing of the line.

akshatkarani: I will look if there is a way to verify color as well

akshatkarani: I tried using `termcolor` for generating colored string but there is always a difference in some bits.

Makman2:
:+1: can be merged.

abhishalya:
@jayvdb Can we have a final review on this..

Issue #6053, opened 2019-08-04
pytest fails to run: TypeError: attrib() got an unexpected keyword argument 'converter'


jayvdb:
```
$ py.test
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.3/bin/py.test", line 7, in <module>
    from pytest import main
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pytest.py", line 10, in <module>
    from _pytest.fixtures import fixture, yield_fixture
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/_pytest/fixtures.py", line 15, in <module>
    from _pytest import nodes
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/_pytest/nodes.py", line 11, in <module>
    from _pytest.mark.structures import NodeKeywords, MarkInfo
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/_pytest/mark/__init__.py", line 4, in <module>
    from .structures import (
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/_pytest/mark/structures.py", line 306, in <module>
    class MarkInfo(object):
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/_pytest/mark/structures.py", line 309, in MarkInfo
    _marks = attr.ib(converter=list)
TypeError: attrib() got an unexpected keyword argument 'converter'
```

jayvdb:
Prior to that

> Requirement already satisfied: attrs in /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages (from packaging>=16.8->-r requirements.txt (line 7))

A bit annoying we cant see the version installed, but attrs-19.1.0 is the latest, but pytest only needs "attrs>=17.4.0"
https://github.com/pytest-dev/pytest/blame/d4351ac5a287eb96521f09618bbbe1c73f4e6884/setup.py

KVGarg:
Might be `pytest` was not installed completely at that point of time, because I just pushed my forked master branch to travis and the `pytest` is running as it should.
https://travis-ci.org/KVGarg/coala/jobs/568352135
Just added 2 changes, first to install python `v3.6.3` as if not added it would have installed `v3.6.7` and other was to forcefully install `setuptools==21`

jayvdb:
> Might be pytest was not installed completely at that point of time

[...]

Issue #4708, opened 2017-09-10
travis-ci installing incorrect version of setuptools


stellargo:
The following error in travis-ci has been spotted in two recent pull requests https://github.com/coala/coala/pull/4650 and https://github.com/coala/coala/pull/4646 where even after specifying that setuptools 17 has to be installed, it goes on to install setuptools-35.0.2.
I tried this on my local environment which did not give any issue. 
```
2.08s$ pip install pip==9 setuptools==17
Collecting pip==9
  Using cached pip-9.0.0-py2.py3-none-any.whl
Collecting setuptools==17
  Using cached setuptools-17.0-py2.py3-none-any.whl
Installing collected packages: pip, setuptools
  Found existing installation: pip 9.0.1
    Uninstalling pip-9.0.1:
      Successfully uninstalled pip-9.0.1
  Found existing installation: setuptools 36.3.0
    Uninstalling setuptools-36.3.0:
      Successfully uninstalled setuptools-36.3.0
Successfully installed pip-9.0.0 setuptools-35.0.2
0.45s$ python .misc/check_setuptools.py
Checking setuptools==17.0
Failed! pip list reports wrong setuptools:
```

gabru-md:
It happens because of the python environment using different paths.

```sh
$ which pip
$ sudo which pip
```
will give you two different environment paths.

atultherajput:
I am getting same error for this PR https://github.com/coala/coala/pull/4709

jayvdb:
https://github.com/coala/coala/blob/master/.travis.yml#L1 - travis is not using sudo.
The problem is somewhere else.

shreyans800755:
@jayvdb I had this issue in a PR. https://github.com/coala/coala/pull/4726/commits/8f747b3ccb6c8acf201a1cc21ffb541c5190c5d0 commit resolved it. However, `-v -v -v` might not be required.
After some digging, I figured out that, python 2.7 and 3.5 are installed by default. In rest of the versions, `setuptools` was installed multiple time (36.3.0 and 36.0.1 as far as I can remember). But while installing version 17.0, it was uninstalling only one of the versions. So, there was already another version available, due to which version 17 was not being installed.
I still couldn't figure out the root cause. But this commit solves the issue at least in my case.

PR #6054, opened 2019-08-04
.travis.yml: Build on xenial using template

jayvdb:
https://github.com/coala/coala/issues/6053 is only
occurring in the Travis trusty environment.
Upgrading to xenial bypasses that problem, but
requires a few workarounds, especially replacing the
Python 3.4.2 job with Python '3.4' which causes
Travis to provide Python 3.4.8.

Add gravis to template to switch jdk versions on Xenial.

Uninstall setuptools multiple times on Xenial for the same
reason it was required once on Trusty.  Also remove
unnecessary pipenv which causes pip to complain about
incompatibilities.

Disable new template for JinjaBear due to its inability
to handle indentation.

.misc/check_moban.sh is no longer needed as the
Travis template provides a moban job, and the latest
moban handles branches.

Related to https://github.com/coala/coala/issues/4708
Related to https://github.com/coala/coala/issues/6053

Implements https://gitlab.com/coala/mobans/merge_requests/135 , which needs to be approved first.


.moban.dt/core-travis.yml.jj2
```diff
@@ -0,0 +1,151 @@
+{% extends 'ci/travis.yml.jj2' %}
+
+{% block custom_python_versions %}
+python:
+{% if travis_dist not in ['precise', 'trusty'] %}
```
jayvdb: Note that if we find a better solution for https://github.com/coala/coala/issues/6053 on trusty, the old Python 3.4.2 minimum dep will be restored, or maybe before then the need for py34 has evaporated.

jayvdb: Template moved to https://gitlab.com/coala/mobans/merge_requests/135 , as coala-quickstart and others need the same fixes

.coafile
```diff
@@ -134,6 +134,8 @@ ignore = node_modules/
 [jinja2]
 bears = Jinja2Bear
 files = **.jj2
+check_end_labels = False
+ignore = .moban.dt/core-travis.yml.jj2
```
KVGarg: is this required? As you have moved this file

jayvdb: Good catch.  That should be removed.

.travis.yml
```diff
@@ -14,19 +16,27 @@ stages:
...
 
-.check_moban: &check_moban
+.moban: &moban
   <<: *disable_global
   python: 3.6
   stage: moban
-  install: pip install moban
-  script: .misc/check_moban.sh
-  if: branch != master OR type = pull_request
+  install: pip install moban>=0.0.4
```
KVGarg: Just a thought, In [test-requirements.txt](https://github.com/coala/coala/blob/master/test-requirements.txt#L15) we're installing another version of `moban` for python version > 3.0 and here we're installing a different version. Shouldn't they be same?

PR #6057, opened 2019-08-12
Support for bears to suggest multiple patches

akshatkarani:
This PR is on top of https://github.com/coala/coala/pull/6029.

Added a new attribute in Result class, `alternate_diffs`. It is a list of dictionaries. For each alternate patch, a corresponding diff should be created and added to `alternate_diffs`.
Then in `ConsoleInteraction` module `get_alternate_patch_actions` function adds mulitple instances of `AlternatePatchAction` to the list of action.
`AlternatePatchAction` when applied updates the `result.diffs` and applies `ShowPatchAction`.


abhishalya (approved):
LGTM :tada:

PR #6082, opened 2019-10-10
link fixedfor dive into python

MaskedCarrot:
fixed the link for dive into python in writting_Linter_Bears.rst

### DIFF
--- a/docs/Developers/Writing_Linter_Bears.rst
+++ b/docs/Developers/Writing_Linter_Bears.rst
@@ -146,7 +146,7 @@ For the exact list of named groups ``@linter`` recognizes, see the `API
 documentation <https://api.coala.io/en/latest/>`__.
 
 Please refer to `Python3 re module <https://docs.python.org/3/library/re.html>`_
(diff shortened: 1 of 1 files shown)


docs/Developers/Writing_Linter_Bears.rst
```diff
@@ -146,7 +146,7 @@ For the exact list of named groups ``@linter`` recognizes, see the `API
 documentation <https://api.coala.io/en/latest/>`__.
 
 Please refer to `Python3 re module <https://docs.python.org/3/library/re.html>`_
-and `Dive into python <http://www.diveintopython3.net/regular-expressions.html>`_
+and `Dive into python <http://diveintopython3.problemsolving.io>`_
```
utkarsh2102, line 149: I'd rather link it with https://diveintopython3.net/regular-expressions.html than what you're proposing, I believe. So in a way, what's there does make sense and this change is not needed in the first place :)

MaskedCarrot:
closing this as there is no change needed.

PR #6134, opened 2021-03-12
fixes issue #6133

### LINKED ISSUE #6133: test pipeline of travis.ci broken
Pipeline fails while installing dependencies.

anshalshukla:
I've updated the configuration to fix the test pipeline. 
The fixes made the test pipeline run successful on my repository.
fixes #6133

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,7 @@ env:
     - PATH="$PATH:$TRAVIS_BUILD_DIR/node_modules/.bin"
     - BEARS_ZIP_URL=https://codeload.github.com/coala/coala-bears/zip
     - GRAVIS="https://raw.githubusercontent.com/DanySK/Gravis-CI/master/"
-    - JDK="adopt@1.8.0-212"
+    - JDK="graalvm-ce-java8@21.0.0"
 
 cache:
   pip: true
@@ -104,6 +104,8 @@ before_install:
   # Apart from builds with python 2.7 and 3.5, it installs setuptools
   # twice. So, it is required to uninstall one version manually.
   - pip uninstall setuptools --yes
+  - pip uninstall setuptools --yes || true
+  - pip uninstall setuptools --yes || true
   - pip uninstall setuptools --yes || true
   - pip uninstall setuptools --yes || true
   - pip uninstall pipenv --yes || true
@@ -134,8 +136,8 @@ script:
   - python setup.py bdist_wheel
   - pip install ./dist/*.whl
   - curl -fsSL -o coala-bears.zip $BEARS_ZIP_URL/master
-  - pip install coala-bears.zip[alldeps]
   - coala --non-interactive
+  - pip install coala-bears.zip[alldeps]
   - python setup.py docs
   - .ci/check_man.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ before_install:
   # Use pip 9, so that setuptools can be downgraded.
   # https://github.com/coala/coala/issues/4708
   # Apart from builds with python 2.7 and 3.5, it installs setuptools
-  # twice. So, it is required to uninstall one version manually.
+  # multiple times. So, it is required to uninstall manually.
   - pip uninstall setuptools --yes
   - pip uninstall setuptools --yes || true
   - pip uninstall setuptools --yes || true
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 appdirs~=1.4
 cached-property~=1.4.3
 cli-helpers~=1.0.2
-git+https://gitlab.com/coala/coala-utils#egg=coala-utils-0.7.0
+coala-utils==0.7.0
 colorlog>=2.7,<4.0
-git+https://gitlab.com/jayvdb/package_manager#egg=dependency-management-0.5.0.dev0
+dependency-management
 packaging>=16.8
 Pygments~=2.1
 PyPrint~=0.2.6
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -14,19 +14,19 @@ codecov~=2.0.5
 freezegun~=0.3.9
 moban~=0.5.0 ; python_version > '3.0'
 pytest~=3.6.4
-pytest-cov~=2.4
+pytest-cov>=2.4.0,<2.6
 pytest-env~=0.6.0
 pytest-error-for-skips~=1.0
 pytest-instafail~=0.3.0
 pytest-mock~=1.1
 pytest-profiling~=1.3.0
 pytest-reorder~=0.1.0
-git+https://github.com/jayvdb/pytest-reqs@coala#egg=pytest-reqs
+pytest-reqs==0.1
 pytest-timeout~=1.3.0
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -77,7 +77,7 @@ env:
     - PATH="$PATH:$TRAVIS_BUILD_DIR/node_modules/.bin"
     - BEARS_ZIP_URL=https://codeload.github.com/coala/coala-bears/zip
     - GRAVIS="https://raw.githubusercontent.com/DanySK/Gravis-CI/master/"
-    - JDK="adopt@1.8.0-212"
+    - JDK="graalvm-ce-java8@21.0.0"
```
abhishalya, line 80: I don't know where this comes from but you should be able to use any build of `adopt@1.8.0-xxx`. Check the travis docs to see which builds are available.

.travis.yml
```diff
@@ -102,10 +102,12 @@ before_install:
   # Use pip 9, so that setuptools can be downgraded.
   # https://github.com/coala/coala/issues/4708
   # Apart from builds with python 2.7 and 3.5, it installs setuptools
-  # twice. So, it is required to uninstall one version manually.
+  # multiple times. So, it is required to uninstall manually.
   - pip uninstall setuptools --yes
   - pip uninstall setuptools --yes || true
   - pip uninstall setuptools --yes || true
+  - pip uninstall setuptools --yes || true
+  - pip uninstall setuptools --yes || true
```
abhishalya, line 110: This is absurd, do you know why this is required?

.travis.yml
```diff
@@ -134,8 +136,8 @@ script:
   - python setup.py bdist_wheel
   - pip install ./dist/*.whl
   - curl -fsSL -o coala-bears.zip $BEARS_ZIP_URL/master
-  - pip install coala-bears.zip[alldeps]
   - coala --non-interactive
+  - pip install coala-bears.zip[alldeps]
```
abhishalya, line 140: No this is wrong, how can you run coala without installing coala-bears?

requirements.txt
```diff
@@ -1,9 +1,9 @@
 appdirs~=1.4
 cached-property~=1.4.3
 cli-helpers~=1.0.2
-git+https://gitlab.com/coala/coala-utils#egg=coala-utils-0.7.0
+coala-utils==0.7.0
```
abhishalya, line 4: No, we will have to use git source for this.

requirements.txt
```diff
@@ -1,9 +1,9 @@
...
 colorlog>=2.7,<4.0
-git+https://gitlab.com/jayvdb/package_manager#egg=dependency-management-0.5.0.dev0
+dependency-management
```
abhishalya, line 6: Similar as before, use the git source.

test-requirements.txt
```diff
@@ -14,19 +14,19 @@ codecov~=2.0.5
...
 pytest-profiling~=1.3.0
 pytest-reorder~=0.1.0
-git+https://github.com/jayvdb/pytest-reqs@coala#egg=pytest-reqs
+pytest-reqs==0.1
 pytest-timeout~=1.3.0
 pytest-travis-fold~=1.3.0
 requests-mock~=1.2
 ipdb~=0.11
-pip<10
+pip
```
abhishalya, line 29: No, this won't work. There is a hard check on pip version, coala won't work with anything above pip 9. There is a hard check on coala setup.py too to check this.

abhishalya (changes requested):
Thanks for contriuting, I have added some comments.

Also once you see a green build, you'll also have to do similar changes in coala-mobans which is where we keep the templates for some of the files which eventually generates the actual files. It is required that the generated files and the files here are exactly the same. Although this is something you should do after you see a green build here.

abhishalya:
Also as a general rule, always work on non-master branches :)

.travis.yml
anshalshukla: pip install language-check need jdk8 and above to run on python 3.6 and above. Any specific reason to use adopt@1.8.0-xxx? I have taken this version from Travis docs

.travis.yml
anshalshukla: More than 2 versions of setuptools are getting installed which needs to manually uninstalled.

requirements.txt
anshalshukla: I'm not sure if that works, there is a issue regarding this by @jayvdb https://github.com/di/pytest-reqs/issues/25

anshalshukla: pytest-reqs fails to recognize it as an requirement.

requirements.txt
anshalshukla: pytest-reqs fails to recognise it as an requirement.

test-requirements.txt
anshalshukla: Umm, but it is working fine for me. As far as I understand, there is no such hard check for pip, but pip 9 was being used so that setup-tools can be downgraded; there is a hard check on that in coala/.misc/check_setuptools.py

requirements.txt
abhishalya: You can always adjust the pytest-reqs version.

.travis.yml
abhishalya: This should not happen, so there's something wrong in the build then

.travis.yml
abhishalya: We would want to use openjdk and we wouldn't want to use private sources if this is one of them.

test-requirements.txt
abhishalya: Yes, there is. See https://github.com/coala/coala/blob/master/setup.py#L441

requirements.txt
anshalshukla: Well, I tried doing that, but it leads to some weird error not recognising pluggy and other 2-3 dependencies.

.travis.yml
anshalshukla: Isn't it just for dependencies? I tried running with this before but it downgrades the PyYAML package to 3.xx but coala requires PyYAML package > 5.xx .
Editing dependencies is not that easy in this case as it is deeply rooted.

abhishalya:
@anshalshukla Let me migrate repository from travis org to com. Might take some time until I get everything reported here properly. It would help me in having small build queue.

test-requirements.txt
anshalshukla: Is there any specific reason for that because I can see that even in logs of the previous run [link](https://travis-ci.org/github/coala/coala/jobs/682807502) that pip-18.1(line252) was used.

abhishalya: Yes, a lot of reasons. I couldn't possibly list them down here. This is what we will be changing in the python update project.

.travis.yml
anshalshukla: Yeah, so graalvm is an open-source project and uses OpenJDK runtime under the hood. I found it somewhere in the supported JVM list of Travis.ci

test-requirements.txt
anshalshukla: Okay, while exploring bears repo I understood your concerns, will make the necessary changes.

anshalshukla:
I'll soon be creating a new PR to fix this issue.

Issue #6150, opened 2021-03-16
Uniform string formatting


anshalshukla:
Some files in the test module use f string format, whereas others use placeholder syntax. This should be changed and made uniform across all files.

#difficulty/newcomer

Shania99:
I am new to open source and would like to take this up.

abhishalya:
@Shania99 Sure, go ahead :)

codehobbyist06:
Sir, I am a newcomer and would like to work on this issue. Can you pls tell if it is already assigned?

Shania99:
@codehobbyist06 Hey! I've been working on this actually.

codehobbyist06:
Ok.

alphaX86:
Hi @anshalshukla so what kind of formatting you need uniformly in all test module files?

alphaX86:
IK the issue is already WIP but I'm asking this to be clear and understandable

shobhi1310:
If someone is working on it, it should be marked as assigned to avoid confusion.

alphaX86:
Hmm @shobhi1310 I agree

anshalshukla:
@alphaX86 I had f-strings on my mind. Since we already use python3.6 using f-strings would be nice in my opinion.

alphaX86:
Hmm alright, and could you tell what files I should exactly look on? Coz tests folder is huge and IDK where to exactly look and change on

Shania99:
@anshalshukla I am almost done with changing the formatting but I used the old school method of string formatting and not f-strings. Would that be okay?

anshalshukla:
You can maybe use grep to search for placeholders and then make the changes. But this issue has already been taken by someone recently so you can look for issues that are unassigned or have been inactive for a while.

anshalshukla:
@Shania99 I'm not particularly sure about which one is preferred. But it should be uniform across test module and should have support in future versions of python as well since coala is planning for an upgrade.
@abhishalya can take a stand on what is preferred.

abhishalya:
f-strings are the way to go from here on :)

Liad-n:
Are there any PRs for this issue yet?

echo-outlier:
Can I take this up? I can fix this and make sure that f string format is used in place of placeholder syntax uniformly across all files.

Anupam-USP:
Hey! Is this issue still open? Or can I take it?

PR #6158, opened 2021-04-03
Use f-string formatting to format strings

jkklapp:
Closes https://github.com/coala/coala/issues/6150

PR #6164, opened 2021-04-14
Fix for #6150

echo-outlier:
You can review my code. I have just used f string format in 3 files. This is a fix for issue #6150 which is yet open. So kindly merge it. If there is any problem you can inform me. I have change to the f string formatting in only 3 of your files but I can fix this in all your files where there is a need of f string formatting.

### DIFF
--- a/coalib/misc/BuildManPage.py
+++ b/coalib/misc/BuildManPage.py
@@ -70,7 +70,7 @@ class BuildManPage(Command):
         _license = dist.get_license()
         appname = self._parser.prog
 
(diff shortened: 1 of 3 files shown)


abhishalya (changes requested):
Please follow commit guidelines properly.

utkarsh2102 (changes requested):
Plus, I'd rather have the 2 commits squashed before re-requesting a review :)
And also, a good PR header please, if that's not much to ask!

abhishalya:
Duplicate of https://github.com/coala/coala/pull/6158

